PR #49, opened 2014-08-30
version output will not print extra newline

boneskull:
This is fussy of me, but the version output may be easier to programmatically consume if `process.stdout.write()` was used instead of `console.log()`, which will throw an extra newline at the end.

### DIFF
--- a/index.js
+++ b/index.js
@@ -457,7 +457,7 @@ function Argv (processArgs, cwd) {
                 process.exit(0);
             }
             else if (key === versionOpt) {
-                console.log(version);
+                process.stdout.write(version);
                 process.exit(0);
             }
         });
--- a/test/usage.js
+++ b/test/usage.js
@@ -714,10 +714,12 @@ describe('usage', function () {
         process._exit = process.exit;
         process._env = process.env;
         process._argv = process.argv;
+        process.stdout._write = process.stdout.write;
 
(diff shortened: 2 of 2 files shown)


nylen:
@bcoe will you reconsider this?

bcoe:
@nylen I share some of the same concerns about adding too much terminal specific logic to the existing yargs code-base -- I'm not to excited to add too many more dials/knobs to index.js.

Having said this, I'd like to start taking on some refactoring of yargs, that would make it easier to isolate terminal-specific logic. My thought is this:

`minimist.js could be renamed parser.js`
'Index.js' contains logic that I think could be split into a few files logically;

* logic for outputting text.
* logic for building up usage information.
* additional parsing logic which might be better contained in parser.js.

I'd love to start thinking about some of these refactors, before adding much functionality... Also it would be good to get coverage close to 100% before any major refactors.

Long story short, open another ticket for this feature request. And we should also open another ticket discussing the work we'd like to do on yargs. I want to respect @chevex's wishes to keep yargs agnostic of the terminal, but I think with some work we can have the best of both worlds (I'd love a JSON outputter, as an example)

nylen:
All fair points, thanks for the thoughtful response.  I've opened #81 and should be able to help with refactoring.

PR #57, opened 2014-10-13
Mention .option in README

eush77:


bcoe:
@eush77 I don't know how I feel about adding a top-level section-header for an alias, what do you think of instead adding a note at the end of the section, something like:

can be aliases as `option`.

eush77:
@bcoe I just wanted to be consistent with how aliases are documented now.
Take a look at https://github.com/chevex/yargs#demandkey-msg--boolean.

bcoe:
@eush77 +1 for consistency :+1:

PR #70, opened 2014-11-23
fix #69: Allow explicit `default: undefined` for flags

### LINKED ISSUE #69: Allow undefined in boolean field parsing
I want to get the user's preference on update or not, so I added a `--update` flag to parsing options:
```
var argv = yargs.Options('--update', {
  boolean: true
}).argv;
```

In this case, when user set `--update` in the arguments, `argv.update` will be set, and when user miss this flag, `false` will be put there.  But a more sensible interpretation is that user didn't give a specific preference, neither did my parsing option, in which case `undefined` seems more favorable. And, in this context, whether to update should be determined by some mechanism like last update time.

I propose that, allow return `undefined` when `default: undefined` is specifically set.

ashi009:
It’s possible to know if the user not set certain flags, after set `default: undefined` for that field.

_This change is very unlikely to break anything, and it behaves the same when not set `default` or set it `true/false`._

#69


bcoe:
@ashi009 would you mind adding a unit test around this new feature. All the existing unit-tests pass, but it would help prevent any regressions, and help document the feature, if you could toss a test in.

You should probably rebase against master, before adding your test :+1:

ashi009:
Yes will do that.

bcoe:
@ashi009, I've merged your pull request here:

https://github.com/chevex/yargs/pull/87

and it's published in `npm install yargs@2.3.0`, let me know if this does the trick.

ashi009:
Thank you!

Issue #12, opened 2014-03-11
Default value for config arguments


dogoku:
Given the following setup

    yargs.config("c").describe("c", "Config file path").default("c","config.json")

It seems that even though the default value gets read (eventually), the config file is never read, since `setArg` is not called for a default value.

soluml:
Still experiencing this issue as well!

mathroc:
@bcoe is anything preventing you to do a PR based on your "12-fix" branch ? I've been using it and it's working great :+1:

bcoe:
@mathroc, @dogoku mind giving this branch a try:

https://github.com/chevex/yargs/pull/79

If it solves your problems I'd like to merge it and release version 2.0 of yargs.

dogoku:
I'll give it a go later today

mathroc:
@bcoe I tried to install `bcoe/yargs#test-fix-config-changes` but it seems the branch is not present anymore in your repository. am I missing something ?

bcoe:
@mathroc sorry for the confusion, I'm actually committing directly to the `yargs` repo now, and have a branch  called `bcoe/test-fix-config-changes`.

FYI, I just bounced the behavior we've been bike-shedding off my coworker @zeke at npm, he agrees with CLI arguments smashing config arguments.

zeke:
smash

mathroc:
@bcoe oh I didn't notice, it's better but there is still some case not working I think, I'll explain it on the PR

bcoe:
This problem should now be fixed on master, via:

https://github.com/chevex/yargs/pull/79

mathroc:
it's working great! thank you!

PR #79, opened 2015-02-05
additional changes to how JSON configuration is loaded, based on discussion in #12

bcoe:
In this branch I've changed the behavior of how we load config.js files, based on a discussion in #12 

If you have a config.json, with the value `foo=bar`, and run the command `./bin/foo --foo=blarg`,
rather than yargs interpreting this as `['blarg', 'bar']`, which does not seem like the expected behavior, `blarg` will override `bar`.

This will require a major version bump. @mathroc, mind trying this branch out?

Note: I was sick of my console output getting broken by yargs when running tests, and hunted down this bug/added test-coverage/leak detection.


mathroc:
I've made a simple program trying different cases and i can see a few issues. 
you can see the files here : https://gist.github.com/mathroc/0acfaeceff852d1b848c

when executing `node test.js` i have this output : 

 > arg_no_default____no_config____no_cli___ undefined
 > arg_default_true__no_config____no_cli___ true
 > arg_default_false_no_config____no_cli___ false
 > arg_no_default____config_false_no_cli___ false
 > arg_default_true__config_false_no_cli___ false
 > arg_default_false_config_false_no_cli___ false
 > arg_no_default____config_true__no_cli___ true
 > arg_default_true__config_true__no_cli___ true
 > arg_default_false_config_true__no_cli___ true
 > **arg_no_default____no_config____cli_true_ undefined**
 > arg_default_true__no_config____cli_true_ true
 > **arg_default_false_no_config____cli_true_ false**
 > arg_no_default____config_true__cli_true_ true
 > arg_default_true__config_true__cli_true_ true
 > arg_default_false_config_true__cli_true_ true
 > **arg_no_default____config_false_cli_true_ false**
 > **arg_default_true__config_false_cli_true_ false**
 > **arg_default_false_config_false_cli_true_ false**
 > **arg_no_default____no_config____cli_false undefined**
 > **arg_default_true__no_config____cli_false true**
 > arg_default_false_no_config____cli_false false
 > **arg_no_default____config_true__cli_false true**
 > **arg_default_true__config_true__cli_false true**
 > **arg_default_false_config_true__cli_false true**
 > arg_no_default____config_false_cli_false false
 > arg_default_true__config_false_cli_false false
 > arg_default_false_config_false_cli_false false

the bolded lines are not correct IMHO

bcoe:
@mathroc several of the errors were caused by a misunderstanding in the `yargs` DSL, to negate a key the syntax is `--no-foo`, rather than `--no_foo`. Having said this, your script was awesome and helped identify a bug with config loading that I have put a test around.

When running your script, with the following command:

```bash
node test.js --arg_no_default____no_config____cli_true_ --arg_default_true__no_config____cli_true_ --arg_default_false_no_config____cli_true_ --arg_no_default____config_true__cli_true_ --arg_default_true__config_true__cli_true_ --arg_default_false_config_true__cli_true_ --arg_no_default____config_false_cli_true_ --arg_default_true__config_false_cli_true_ --arg_default_false_config_false_cli_true_ --no-arg_no_default____no_config____cli_false --no-arg_default_true__no_config____cli_false --no-arg_default_false_no_config____cli_false --no-arg_no_default____config_true__cli_false --no-arg_default_true__config_true__cli_false --no-arg_default_false_config_true__cli_false --no-arg_no_default____config_false_cli_false --no-arg_default_true__config_false_cli_false --no-arg_default_false_config_false_cli_false | diff expected_output.txt -
```

output matches expected now :+1:, I'm going to go ahead and merge this., let me know if you find any more issues.

jackyzy823:
here's  a situation of boolean option in config file.
```javascript
require('yargs').options({
    f: {
      alias: 'force',
      type: 'boolean',
      requiresArg: false
    }}).config('config');
```
and usage is
```
node xxx.js --config xxx.json
```
 in ```xxx.json``` there's ``` { "f": true }``` ,and ``` -f``` is not mentioned in ```CLI```. it's expected to be ```true``` for ```f(force)``` ,however the result of this is ```false```.

I supppose that those boolean options not mentioned in ```CLI``` should be according to config file.

PS. string options works correctly.

PR #85, opened 2015-02-08
test coverage now at 100%

bcoe:
In prep for a 3.x release of yargs (with the stated goal of making yargs easier to extend) I've:

* renamed `minimist` -> `parser`, since our parser has now diverged significantly from `minimist`.
* added the `wordwrap` dependency, so that we don't have to maintain a fork.
* renamed the testing files, to better match the logical divisions I'd like to make in `index.js`: `validator`, `argv`, `usage`, `parser`.
* I've gotten test-coverage to 100%, to make refactoring less frightening.
* I documented a couple bonus features floating around `yargs`, that I discovered via unit-testing.

This should put us into a good position to entertain adding some new features to `yargs`, without worrying about breaking anything. cc: @nylen


nylen:
:+1: I will look into further steps in my copious free time :)

PR #86, opened 2015-02-08
clarify opt parameters

ngoldman:
Added all possible `opt` parameters to the README since that wasn't totally clear from the `.option()`/`.options()` section of the doc. Also added an `.options()` example.


bcoe:
@ngoldman :+1:, great clarification.

PR #93, opened 2015-02-19
make the version feature more useful

bcoe:
`.version()` is cool, but you usually want to load the version # programmatically rather than having it hard-coded. This change allows you to optionally pass a function to `.version()`, which can be (as an example) used to load the version # from the package.json.


chevex:
Love this feature.

Issue #60, opened 2014-10-26
.default should support functions


kilianc:
```js
.default('id', getRandomId, 'self assigned')
```

otherwise it prints the function body, I had to hack the `toString` method and move the code before `yarg` which is ugly.

```js
function getRandomId() {
  var rand = crypto.randomBytes(30).toString('base64')
  var time = Date.now().toString()
  return crypto.createHash('sha256').update(rand + time).digest('hex')
}

getRandomId.toString = function () {
  return 'self assigned'
}
```

bcoe:
@kilianc does #95 do the trick?

PR #95, opened 2015-02-20
uses function as default value resulted in a broken usage message

bcoe:
Using a function as a default value resulted in a broken usage message, see #60.

Providing an unnamed function for a default will now result in these usage instructions:

`[default: (generated-value)]`

You can also now provide a named function:

```js
var argv = require('./')
  .help('h')
  .default('random', function randomValue() {
    return Math.random() * 256;
  }).argv;
```

which would result in the usage instruction:

`[default: (random-value)]`

what do you think @kilianc?

### DIFF
--- a/index.js
+++ b/index.js
@@ -50,6 +50,7 @@ function Argv (processArgs, cwd) {
             string: [],
             alias: {},
             default: {},
+            defaultDescription: {},
             requiresArg: [],
             count: [],
             normalize: [],
@@ -94,13 +95,14 @@ function Argv (processArgs, cwd) {
         return self;
     };
 
-    self.default = function (key, value) {
+    self.default = function (key, value, defaultDescription) {
         if (typeof key === 'object') {
             Object.keys(key).forEach(function (k) {
                 self.default(k, key[k]);
             });
         }
         else {
+            options.defaultDescription[key] = defaultDescription;
             options.default[key] = value;
         }
         return self;
--- a/README.md
+++ b/README.md
@@ -371,6 +371,16 @@ Set `argv[key]` to `value` if no option was specified on `process.argv`.
 
 Optionally `.default()` can take an object that maps keys to default values.
 
+But wait, there's more! the default value can be a `function` which returns
+a value. The name of the function will be used in the usage string:
+
+```js
+var argv = require('yargs')
+  .default('random', function randomValue() {
+    return Math.random() * 256;
+  }).argv;
+```
+
 .demand(key, [msg | boolean])
 -----------------------------
 .require(key, [msg | boolean])
--- a/README.md
+++ b/README.md
@@ -364,7 +364,7 @@ Optionally `.alias()` can take an object that maps keys to aliases.
 Each key of this object should be the canonical version of the option, and each
 value should be a string or an array of strings.
 
-.default(key, value)
+.default(key, value, [description])
 --------------------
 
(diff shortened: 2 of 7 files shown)


kilianc:
I would have preferred to explicitly set the name, because if it's a library we can't control the function name

nylen:
I like this as the default behavior.  You could add an optional name argument, but you can always set it explicitly by wrapping in another function:

```js
argv.default('random', function randomValue() {
    return lib.someOtherFunction();
});
```

I didn't know about `S().dasherize` when I wrote [`toCamelCase`](https://github.com/bcoe/yargs/blob/v3.1.0/lib/parser.js#L35-L39), you could probably also replace that with [`camelize`](https://github.com/jprichardson/string.js#--camelize) as a future change.

kilianc:
I honestly don't get how and why the name of my function is related to the help screen.

bcoe:
@kilianc @nylen, I could go either way:

On one hand: I like that using the function name eliminates the need for an additional argument, and it makes it easy to standardize how you output the `default:` string, i.e., it's the dasherized function name with `(` around it.

Having said this, perhaps the approach is to clever, and accepting a third parameter is better.

bcoe:
@kilianc, for what it's worth I also like that it encourages you to use a descriptive function name :)

Why don't I add an optional third parameter which will take precedence, and let you set the `defaultName` yourself?

kilianc:
I think that function name works but I still want control over it if I am doing something particular

bcoe:
@kilianc @nylen, okay submitted for your review.

index.js
```diff
@@ -94,13 +95,14 @@ function Argv (processArgs, cwd) {
         return self;
     };
 
-    self.default = function (key, value) {
+    self.default = function (key, value, defaultDescription) {
         if (typeof key === 'object') {
             Object.keys(key).forEach(function (k) {
                 self.default(k, key[k]);
```
bcoe: with the concept of defaultDescription, one thing I don't like is that you can't set a defaultDescription if you pass in an object for `key` ... I continue to be so so about the concept of a `defaultDescription`, is wrapping the method in a named function too weird?

kilianc:
I don't know about implementation specifics, but what I know is that I need to be able to write what I want in my `--help`. I would honestly just override function `name` or `toString`. Creating an wrapper function with a custom name in order to pass a name, it's a clear symptom that we are doing it wrong, but again. Just humble opinion here.

bcoe:
@kilianc point taken, mind trying out the current implementation; I think it should do the trick for you.

kilianc:
I will checkout the branch tomorrow thanks @bcoe !

kilianc:
@bcoe what about object declaration like:

```
  .option('id', {
    alias: ['n', 'name'],
    type: 'string',
    default: uuid.v4,
    describe: 'the log stream id'
  })
```

how do I name my default 'uuid' in this case?

kilianc:
@bcoe poke

brewster1134:
so after playing with this a bit, i think the implementation needs updated a bit...

* when showing help/usage, the default function should be called and the result should be used as the default value, instead of the name of the function
* if an option is passed in from the cli, the default function should _not_ be called

the current implementation, to me, only seems valuable if the option(s) passed in from cli, are used as arguments to call the function with, instead of just using it instead of whatever the function returns

perhaps i am missing something... can the original use case for adding this feature be explained?

PR #109, opened 2015-03-01
fix: count should not consume the next argument

schnittstabil:
Actually, `count` drops the next argument:
```js
require('yargs').count('verbose').parse([ '--verbose', 'I AM LOST!', 'foo' ])
{ _: [ 'foo' ], verbose: 1, '$0': 'node' }
```
That's because `count` and `boolean` are handled somewhat differently.

Maybe this PR needs some refactoring as I've only doubled the `boolean` checks.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -90,7 +90,8 @@ module.exports = function (args, opts) {
                 var next = args[i + 1];
 
                 if (next !== undefined && !next.match(/^-/)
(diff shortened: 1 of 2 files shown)


bcoe:
@schnittstabil great catch, I'm happy with this patch I think adding an extra `checkAllAliases` is pretty clean. this patch is released in `yargs@3.4.5`.

PR #118, opened 2015-03-08
added support for bash completion to yargs

bcoe:
yargs now supports bash completion for commands and options.

1. have your application call `yargs.showCompletionScript()`, to print out a script that can be installed in a user's `.bashrc`.
2. set the `.completion('completion')` flag to enable bash completion.

```js
var yargs = require('yargs')
   .completion('completion'),
   argv = yargs.argv;
```

1. when the user executes yargs with the `completion` command, the bash completion script will be outputted., this should be concatenated to `.bash_profile` or `.bashrc`.
2. that's all there is to it, yargs will magically generate completions for arguments and options.

see #3


README.md
```diff
@@ -702,6 +702,58 @@ yargs.showHelp();
...
+.completion(cmd, [description], [fn]);
+-------------
+
+Enable bash-completion shortcuts for commands and options.
+
+`cmd`: when present in `argv._`, will result in the `.bashrc` completion script
+being outputted. To enable bash completions, concat the generated script to your
+`.bashrc`, or `.bash_profile`.
+
+`descrption`: provide a description in your usage instructions for the command
```
isaacs: `s/descrption/description/` probably?

bcoe: thanks! for the copy edit :+1:

Issue #122, opened 2015-03-10
demand() does not play well with commands


raine:
```
Usage: tool do-stuff <arg>
```

If I have a command `tool do-stuff <arg>` that demands at least one argument (`arg`), I'll have to use `demand(2)`. The default error message will be weird in this case:

```
Not enough non-option arguments: got 1, need at least 2
```

Of course you can override the message with your own, but I think I'd rather not have a message at all in this case, because the help will be self-explanatory in that it's expecting non-optional arguments (`<arg>`).

Using `demand()` with empty string as msg will not work, it's the same as not passing it at all. So maybe we could have a way to use `demand()` without a message, as that at alone will print the help which is an message in itself.

What do you think?

Cheers.

raine:
To clarify, even allowing an empty string would in my opinion not be sufficient, because it prints an unnecessary empty line. Unless of course empty string meant no line at all.

raine:
What about something like this:

https://github.com/raine/yargs/commit/f57430ce02491ec75c26a25cc59b92daeaa533d7

Fail with help without showing a specific error if `msg` is explicitly set as `null`?

```js
.demand(1, null)
```

bcoe:
@raine published your changes in `yargs@3.5.4`, with one slight modification `''` will also result in no output.

raine:
Makes sense. Thanks.

Sent from my iPhone

--
Raine

PR #123, opened 2015-03-10
validation/demand: fail without message if msg is null

raine:
#122


bcoe:
:+1: looks good to me, I'll get this merged tonight after I rock climb.

PR #127, opened 2015-03-14
add support for config specified in .js files

pirxpilot:
Let me know what you think. I need this for code42day/postcss-cli#5 - but it did turn out a bit more complicated than what I imagined.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -276,7 +276,7 @@ module.exports = function (args, opts) {
             var configPath = argv[configKey] || configLookup[configKey];
             if (configPath) {
               try {
-                  var config = JSON.parse(fs.readFileSync(configPath, 'utf8'));
+                  var config = require(path.resolve(process.cwd(), configPath));
                   Object.keys(config).forEach(function (key) {
                       // setting arguments via CLI takes precedence over
                       // values within the config file.
--- a/test/fixtures/settings.js
+++ b/test/fixtures/settings.js
@@ -1,0 +1,6 @@
+module.exports = {
+    calculate: function (a) {
+        return a + 55;
+    },
+    herp: 'derp'
+};
--- a/test/parser.js
+++ b/test/parser.js
@@ -297,6 +297,17 @@ describe('parser tests', function () {
             argv.should.have.property('foo').and.deep.equal('bar');
         });
 
+        it('should load options and values from a JS file when config has .js extention', function () {
+            var jsPath = path.resolve(__dirname, './fixtures/settings.js');
+            var argv = yargs([ '--settings', jsPath, '--foo', 'bar' ])
+                .config('settings')
+                .argv;
+
+            argv.should.have.property('herp', 'derp');
+            argv.should.have.property('foo', 'bar');
+            argv.should.have.property('calculate').and.be.a('function');
+        });
+
         it('should raise an appropriate error if JSON file is not found', function(done) {
           var argv = yargs([ '--settings', 'fake.json', '--foo', 'bar' ])
             .alias('z', 'zoom')


bcoe:
Thanks for the patch! I wanted to reply on the postcss thread, but got wrapped up in the work week sorry.

I have some personal biases towards JavaScript-based configuration files. My experience has been that it encourages too much logic to be included in the configuration file itself. This can result in a lot of refactoring when it comes time to pull configuration into a system like `etcd`, in a production environment.

Having said this, I could have my arm twisted into swapping out `JSON.parse` for `require` -- however, the `noEval` functionality frightens me a bit, I think of configuration as a static thing whereas if I understand the purpose of `noEval` it's to expose helper functions in `yargs` that generate different values each time they are executed?

Would love to hear your counter-points, there's rarely one right or wrong answer with regards to any of these things :)

@nylen, @chevex  have any opinions with regards to this?

pirxpilot:
Thanks for checking out the code.

> I have some personal biases towards JavaScript-based configuration files.

I totally agree (as long as it's a bias *against* them). I avoid that whenever I can. That said maybe yargs should not be as opinionated as you or I are. The sad fact of life is that there modules out there - postcss plugins among them - which do use (maybe even legitimately need using) JS config. Also gulp/grunt with their JS format encourages functions as options. If yargs does not support it, it just means that folks cannot do with CLI what they can do with other interfaces (and that makes them use grunt/gulp, which in turn encourages what we were trying to avoid: even more JS in config...)

> I could have my arm twisted into swapping out `JSON.parse` for `require` 

I am somewhat ambivalent on that. It is certainly shorter ;-). Then again, if we were ever to add yet another format we would probably need to select the parser based on a file extension anyway. In any case I put it in a separate commit on top of the others. If you were to take the PR but not this change it's easy to do.

> the `noEval` functionality frightens me a bit

You and me both ;-) The only things that makes me feel better about it is yargs unit test coverage. I am *almost* certain that introducing `noEval` did not break anything. Also - and I probably did not spend enough time looking at the code to claim that - I somehow feel that *evaluating* functions during setting keys is a special case and *not evaluating* is a normal behavior but clearly I don't have better ideas to express it in the code.

You're right BTW it is about helper functions - but it's not like the fact that they return different values is important here. That fact that they will executed in a different context and with parameters yargs cannot possible know about is the key.

bcoe:
@pirxpilot one option would be, like you suggest, making it so in both cases we don't evaluate the function.

Digging through the parser, the reason that functions are historically executed immediately, is for handling counting functionality:

```value = function(orig) { return orig !== undefined ? orig + 1 : 0; };```

I'd be comfortable making this a special case, and making it so both functions loaded from configuration files, and functions set as defaults are not executed -- I think this will require a major bump though.

What are your thoughts?

pirxpilot:
I am probably the last person you should ask since such change would clearly make *my* life easier ;-)
All I can say it is that if you do do that, I'll gladly rebase and rework this patch to remove `noEval` tricks.

pirxpilot:
Also - and feel free to point out if this is a complete nonsense - wouldn't it be better if we execute functions before they end up in the `defaults` map. If you think this is something might be worth trying let me know. I'll see if I can come up with a patch.

bcoe:
@pirxpilot what I'm trying to figure out is the distinction between loading config variables from a file, and setting default configuration variables.

Why does it make sense to execute functions that are passed in as default values, vs. not evaluating functions when they are loaded in from a JavaScript configuration file -- I see a configuration file as being similar to a set of default values.

Sorry to drag my feet on getting this merged, it's been a really busy few weeks at npm -- and I want to make sure we get this right :+1:

pirxpilot:
Yup. No problem. The worst case scenario for me is that I'll just have to use yargs clone for postcss-cli, which is a pain but not the end of the world. And I want to get it right as well.

I guess the source of the confusion is that until now it was impossible to pass a function as a value of an option. And because of that yargs code could safely assume that whenever it sees the function in such context it can evaluate it and store the result instead. As soon as we allow for `.js` config files functions become first class citizens. And in most cases they cannot and should not be evaluated by yargs because they are in no way yargs parameters (they are parameters of whatever module is using yargs to parse command line). This is different from functions passed as defaults: they are just syntactic sugar - an easier way of passing arguments to yargs itself. And they can and should be evaluated whenever it is convenient for yargs.

So the price of accepting '.js' files as config is that the code has to give up on its - `whenever I see a function I evaluate it and store the result` assumption.

I actually think it can be achieved without changing yargs interface. When I have a moment I'll code it and send a separate patch.

pirxpilot:
As promised: a simplified version. It now relies on #130 (and tests will be failing until #130 is merged). The upside is code change becomes trivial.

bcoe:
@pirxpilot thanks for the contribution, I'm really happy how this pull turned out.

your changes are now published in `yargs@3.6.0`.

PR #129, opened 2015-03-18
Correct `.showHelp` docs

robatron:
Corrects the `.showHelp` docs by changing the parameter from a function to a `console` function, to match what the code is doing:

https://github.com/bcoe/yargs/blob/master/lib/usage.js#L209


bcoe:
thanks for the contribution @robatron, you rock.

PR #130, opened 2015-03-19
adjust the way yargs evaluates functions passed as defaults

pirxpilot:
This paves the way for cleaner implementation of #127 (as in: no need for `noEval` flag)

### DIFF
--- a/index.js
+++ b/index.js
@@ -126,6 +126,10 @@ function Argv (processArgs, cwd) {
             });
         }
         else {
+            if (typeof value === 'function') {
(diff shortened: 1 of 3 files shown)


bcoe:
@pirxpilot awesome, thanks for the patch. I happen to be rolling out a new version of npm this week, so am swamped -- but I'm hoping to hack on yargs this weekend.

pirxpilot:
No worries. Since I need it anyway, I'll go ahead and prepare an alternative version of #127 that does not use `noEval` flag.

bcoe:
really dig this refactor :+1:

Issue #134, opened 2015-04-03
Describe doesn't work with alias + defaults


megawac:
Minor bug I think, tested in v3.6

**cli.js**
```js
var argv = require('yargs')
  .demand(1)

  .default('output', null)
  .alias('o', 'output')
  .nargs('o', 1)
  .describe('output', 'file to generate')

  .argv;
```

```sh
$ ./cli.js
```
```
Options:
  -o, --output                                                   [default: null]

Not enough non-option arguments: got 0, need at least 1
```

bcoe:
@megawac could you give this branch a try:

https://github.com/bcoe/yargs/pull/135

and let me know if it solves your problem?

Issue #128, opened 2015-03-17
showHelp() shows incorrect option list


zaro:
I just migrated several scripts from optimist to yargs and I noticed that yargs omits some options when generating help. Here is an example :

```javascript
var opts= require('yargs')
        .usage(
                "Usage: $0 [options]"
        )
        .describe('createdAfter','CreatedAfter ListOrders parameter')
        .alias('a','createdAfter')
        .describe('createdBefore','CreatedBefore ListOrders parameter')
        .alias('b','createdBefore')
        .describe('lastUpdatedAfter','LastUpdatedAfter ListOrders parameter')
        .alias('u','lastUpdatedAfter')
        .describe('lastUpdatedBefore','LastUpdatedBefore ListOrders parameter')
        .alias('v','lastUpdatedBefore')
        .describe('incremental','Perform incremental update based on last synccessfull sync')
        .alias('i','incremental')
        .default('i',false)
        .boolean('i')
        .describe('force','Force recheck')
        .alias('f','force')
        .default('f',false)
        .boolean('f')
        .describe('description','Sync transaction description')
        .alias('d','description')
        .default('d',"#{ require('os').hostname() }:#{ process.title } - #{ process.pid }")
        .string('d');
var argv = opts.argv;

opts.showHelp();
```

This outputs :
```
Usage: example.js [options]

Options:
  -i, --incremental                                             [default: false]
  -f, --force                                                   [default: false]
  -d, --description  
[default: "#{ require('os').hostname() }:#{ process.title } - #{ process.pid }"]
```
As you can see the first 4 options are missing.

bcoe:
@zaro good find. I tend to use the `option` method which allows you to create an alias and description at the same time -- I will put a test case around this use case and get you a patch tomorrow hopefully.

bcoe:
@zaro FYI, it appears to be something silly with the usage logic; when building up a list of parameters to print, it runs into trouble if you define aliases as `alias -> described-key` rather than `described-key -> alias`, a quick fix is this:

```javascript
var opts= require('yargs')
        .usage(
                "Usage: $0 [options]"
        )
        .describe('createdAfter','CreatedAfter ListOrders parameter')
        .alias('createdAfter', 'a')
        .describe('createdBefore','CreatedBefore ListOrders parameter')
        .alias('createdBefore', 'b')
        .describe('lastUpdatedAfter','LastUpdatedAfter ListOrders parameter')
        .alias('lastUpdatedAfter', 'u')
[...]

PR #135, opened 2015-04-04
the chaining API now cares less about key ordering

bcoe:
if you swapped the order of the `alias` and the `primary` key, when using the chaining API, options such as: `describe`, `default`, `string`, `require`, would not be shown in the usage string.

## In this pull:

* we now make an attempt to apply usage information to both the primary and the alias key.
* coverage has been fixed by using the patched-blanket library.

see: #134, #128


megawac:
:+1: fixes #134

PR #139, opened 2015-04-05
also detect iojs bin

bcoe:
see: https://github.com/bcoe/yargs/issues/138


rstacruz:
Awesome!

btw, even if you specify `#!/usr/bin/env node`, node is merely a symlink to `iojs`. `process.argv[0]` will always be `iojs` :)

PR #146, opened 2015-04-12
Fix $0 when 'node' or 'iojs' in script path

nylen:
This was causing `argv.$0` to be `''` for me since a lot of my code files live at `/somedir/node/projectname/file.js`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,7 +29,6 @@ function Argv (processArgs, cwd) {
                 ? b : x
         })
         .join(' ').trim();
-    ;
 
     if (process.env._ != undefined && process.argv[1] == process.env._) {
         self.$0 = process.env._.replace(
--- a/index.js
+++ b/index.js
@@ -23,7 +23,7 @@ function Argv (processArgs, cwd) {
         .map(function (x) {
             // ignore the node bin, specify this in your
             // bin file with #!/usr/bin/env node
-            if (~x.indexOf('node') || ~x.indexOf('iojs')) return;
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -23,7 +23,7 @@ function Argv (processArgs, cwd) {
         .map(function (x) {
             // ignore the node bin, specify this in your
             // bin file with #!/usr/bin/env node
-            if (~x.indexOf('node') || ~x.indexOf('iojs')) return;
+            if (/^(node|iojs)\b/.test(x)) return;
```
bcoe: I'm pretty sure there are cases where the bin can be:

`/usr/bin/node`, or `/user/bin/iojs`, perhaps we could modify the regex to something like:

`/(node|iojs)$/`?

nylen: Done.  I also made this check only apply to `process.argv[0]` - it's possible, though pretty unlikely, that someone could name their binary `/path/to/project/scripts/node`.

bcoe:
@nylen will get merged ASAP, just finishing up pulling in `standard` and cleaning up our coding style, I'll merge this after.

bcoe:
@nylen I've published your changes in `yargs@next`, give it a try and let me know if it does the trick for you:

```npm install yargs@next```

I did a large cosmetic overhaul, pulling in the standard style guide (https://github.com/feross/standard) let me know if you see any weirdness.

nylen:
:+1: wfm :)

PR #151, opened 2015-04-21
refactor yargs' table layout logic to use new helper library

bcoe:
### What Have You Done!

I found the logic in yargs for laying out multi-column tables to be some of the most confusing in the codebase. It had grown fairly organically, and we were doing backflips to use the same code for: `commands`, `examples`, and `options`.

Motivated by this, I've created the library `cliui`:

https://www.npmjs.com/package/cliui

Which provides an abstraction for multi-column UIs.

### Testing Help Please

This is a fairly major change, I've tested the usage output with several of my own yargs applications, but I'd love a second set of eyes.

Mind helping with testing? @annonymouse, @nylen, @pirxpilot, @raine

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,4 +4,4 @@ node_js:
   - "0.11"
   - "0.12"
   - "iojs"
-after_script: "NODE_ENV=test YOURPACKAGE_COVERAGE=1 ./node_modules/.bin/mocha --require blanket --reporter mocha-lcov-reporter | ./node_modules/coveralls/bin/coveralls.js"
+after_script: "NODE_ENV=test YOURPACKAGE_COVERAGE=1 ./node_modules/.bin/mocha --require patched-blanket --reporter mocha-lcov-reporter | ./node_modules/coveralls/bin/coveralls.js"
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -112,7 +112,7 @@ module.exports = function (yargs) {
     // the usage string.
     if (usage) {
       var u = usage.replace(/\$0/g, yargs.$0)
-      ui.div({text: u, padding: [0, 0, 1, 0]})
+      ui.div(u + '\n')
     }
 
     // your application's commands, i.e., non-option
@@ -203,7 +203,7 @@ module.exports = function (yargs) {
     // the usage string.
     if (epilog) {
       var e = epilog.replace(/\$0/g, yargs.$0)
-      ui.div({text: e, padding: [0, 0, 1, 0]})
+      ui.div(e + '\n')
     }
 
     return ui.toString()
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -226,6 +226,10 @@ module.exports = function (yargs) {
       width = Math.max(v[0].length, width)
     })
 
+    // if we've enabled 'wrap' we should limit
+    // the max-width of the left-column.
+    if (wrap) width = Math.min(width, parseInt(wrap * 0.5, 10))
+
     return width
   }
 
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -107,34 +107,27 @@ module.exports = function (yargs) {
       ui = cliui({
         width: wrap,
         wrap: wrap ? true : false
-      }),
-      width
+      })
 
     // the usage string.
     if (usage) {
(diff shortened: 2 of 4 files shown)


pirxpilot:
I'll test with postcss-cli and report back.

pirxpilot:
These are the changes.

**Before** - yargs 3.6.0

````
Usage: bin/postcss -use|-p plugin [--config|-c config.json] --output|-o output.css input.css

Options:
  -c, --config   JSON file with plugin configuration             
  -u, --use      postcss plugin name (can be used multiple times)  [required]
  -o, --output   Output file                                     
  -d, --dir      Output directory                                
  --safe         Enable postcss safe mode.                       
  -v, --version  Show version number                             
  -h, --help     Show help                                       

Examples:
  postcss --use autoprefixer -c options.json -o screen.css screen.css                   Use autoprefixer as a postcss plugin            
  postcss --use autoprefixer --autoprefixer.browsers "> 5%" -o screen.css screen.css    Pass plugin parameters in plugin.option notation
  postcss -u postcss-cachify -u autoprefixer -d build *.css                             Use multiple plugins and multiple input files   
````

**After** - yargs from cliui branch

````
Usage: bin/postcss -use|-p plugin [--config|-c config.json] --output|-o output.css input.css

Options:
  -c, --config  JSON file with plugin configuration
  -u, --use  postcss plugin name (can be used multiple times)
                 [required]
  -o, --output  Output file
  -d, --dir  Output directory
  --safe  Enable postcss safe mode.
                 [boolean]
  -v, --version  Show version number
  -h, --help  Show help

Examples:
  postcss --use autoprefixer -c options.json -o screen.css screen.css  Use autoprefixer as a postcss plugin
  postcss --use autoprefixer --autoprefixer.browsers "> 5%" -o screen.css screen.css  Pass plugin parameters in plugin.option notation
  postcss -u postcss-cachify -u autoprefixer -d build *.css  Use multiple plugins and multiple input files
````

Looks like in *after* version options are not aligned, `[required]` mysteriously ends up in new line, and examples are not aligned either.
I might be doing something stupid in [postcss-cli](https://github.com/code42day/postcss-cli) code. For example I am [calling `wrap()` with no args](https://github.com/code42day/postcss-cli/blob/master/index.js#L32). And things do look better if I remove that line. Will gladly fix and retest if you have any suggestions.

bcoe:
@pirxpilot I did some refactoring of `cliui` that moved yargs closer to your original output.

It's worth noting that `.wrap()` will actually disable wrapping, because it's equivalent to `.wrap(undefined)`.

You should test and make sure things look closer to the original format now, but you might find you get a nicer ui if you don't call `.wrap()`.

pirxpilot:
Nice. With disabled wrapping it looks like it did before `cliui`.
Enabling wrapping (that is when I test with `wrap()` call removed) makes it look [interesting](https://gist.github.com/pirxpilot/82f9714b6a176a28286b) 
In any case: I am happy.

bcoe:
@pirxpilot argh, I'll make it so the right-hand-column has a reasonable maximum width., so that we don't end up with the behavior outlined in that gist...

I'd love to get you running without `.wrap()`., I will have another pull shortly.

bcoe:
@pirxpilot a max-width is now set on the left-hand column if `wrap` is enabled, give that a shot.

pirxpilot:
Nice! :shipit: Looks like that now:
````
Usage: bin/postcss -use|-p plugin [--config|-c config.json] --output|-o output.
css input.css

Options:
  -c, --config   JSON file with plugin configuration
  -u, --use      postcss plugin name (can be used multiple times)     [required]
  -o, --output   Output file
  -d, --dir      Output directory
  --safe         Enable postcss safe mode.                             [boolean]
  -v, --version  Show version number
  -h, --help     Show help

Examples:
  postcss --use autoprefixer -c options.    Use autoprefixer as a postcss plugin
  json -o screen.css screen.css
  postcss --use autoprefixer --             Pass plugin parameters in plugin.
  autoprefixer.browsers "> 5%" -o screen.   option notation
  css screen.css
  postcss -u postcss-cachify -u             Use multiple plugins and multiple
  autoprefixer -d build *.css               input files
````

PR #159, opened 2015-05-16
clarify .config() docs

linclark:
The language used to explain how `.config()` works is ambiguous. It could be read as saying that `.config("config.json")` would work. I'm not sure that this language is much better, so feel free to change if you think of something better.


bcoe:
@linclark awesome, thanks for the update. It looks like the `standard` coding style has been updated slightly, I can fix these failing tests.

linclark:
Rebased with `various-silly-fixes`.

Issue #162, opened 2015-05-26
issues with "array" parameters


getify:
I'm having trouble getting "array" type parameters to work. Perhaps I'm missing how to do it, and this can just be an issue about improving the docs. Here's what I'm trying to do:

```
./foobar --greetings=hello world friends --otherparam=...
```

I want `"hello"`, `"world"`, and `"friends"` as three entries in the array for the argument key `greetings`. Of course, I actually want to be able to have any number of entries from one to a hundred or whatever.

Here's how I've tried to set that up:

```js
ARGS = yargs
    .usage(..)
    // ..
    .array("greetings")
    .default("greetings",[],"empty list")
    // ..
```

I even tried adding in an `nargs("greetings",100)`, but that didn't seem to have any effect. It parses "greetings" as an array, but only puts the first value in, leaving the other two param values in `ARGS._` (and thus being incorrectly handled by other parts of my parsing logic).

What am I missing, or is this is a known issue?

bcoe:
@getify thank you for the bug report, arrays were originally implemented with this arg format in mind:

`$0 --greetings greeting1 --greetings greeting2`

Although I can see the format you're outlining being more convenient, it also sounds like we could use more test cases around `nargs`... the problem is, `nargs` will eat argument no matter what which doesn't work great for variable length arrays. Perhaps if you specify an array type, we should eat arguments until the next flag argument?

getify:
> arrays were originally implemented with this arg format in mind:
> 
> `$0 --greetings greeting1 --greetings greeting2`

Yeah the problem with that is that it's not really obvious from the help text being generated that a user has to re-specify the same param name over and over again. When the help text says "array" or "list" or something like that, I think pretty much all users are going to assume space (or maybe comma) delimited lists of values after the param name. I can't remember that I've ever used a CLI program where I had to specify a list in the way you're describing.

> it also sounds like we could use more test cases around nargs... the problem is, nargs will eat argument no matter what which doesn't work great for variable length arrays.

If I do `--greetings=val1 val2 val3`, then `nargs` seems to have no effect and only grabs `val1`. If I do `--greetings val1 val2 val3`, it works. That's very non-intuitive as well. Should/could it do the same with or without the `=` syntax to avoid that pitfall (not just for me but for users of my program)?

> Perhaps if you specify an array type, we should eat arguments until the next flag argument?

[...]

PR #164, opened 2015-05-27
more intuitive parsing of array arguments see #162

bcoe:
This pull request adds more intuitive parsing of array arguments, based on a discussion with @getify (see #162). Here's how things work now:

`--foo a b c` where `foo` is an array, will now be parsed as `['a', 'b', 'c']`.

This pull request includes fixes for both `array` and `nargs`, adding support for `--foo=a b c`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,4 @@ node_modules/
 *.swp
 test.js
 coverage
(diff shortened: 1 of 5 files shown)


getify:
<3

PR #165, opened 2015-05-27
expose the method yargs.terminalWidth()

### LINKED ISSUE #155: Wrap to window size
First
https://github.com/bcoe/yargs/blob/master/lib/usage.js#L297-300

mentions 100 in the comment and uses 80 in the code :)
What I wonder if there is a way to say ' wrap to window width'. RIght now, I'd need to call:
.wrap(require('window-size').wsize)

bcoe:
The terminal width is now exposed by `yargs.terminalWidth()`, this can be used along with `wrap()` to maximize the width of usage instructions.

fixes #155

### DIFF
--- a/README.md
+++ b/README.md
@@ -618,7 +618,7 @@ Optionally `.nargs()` can take an object of `key`/`narg` pairs.
 .config(key)
 ------------
 
-Tells the parser that if the option specified by `key` is passed in, it 
+Tells the parser that if the option specified by `key` is passed in, it
 should be interpreted as a path to a JSON config file. The file is loaded
 and parsed, and its properties are set as arguments.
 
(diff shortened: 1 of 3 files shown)


nelson6e65:
How can I use this in Typescript ESM?

Getting error:
```ts
import yargs, { terminalWidth } from 'yargs';

// SyntaxError: The requested module 'yargs' does not provide an export named 'terminalWidth'
```


```ts
import yargs, { CommandModule } from 'yargs';

yargs(hideBin(process.argv))
  .wrap(yargs.terminalWidth())

// TypeError: yargs.terminalWidth is not a function
```

nelson6e65:
NVM, I solved by using:

```ts
const myYargs = yargs(hideBin(process.argv));

myYargs.wrap(myYargs.terminalWidth())
```

PR #168, opened 2015-05-29
force a build

bcoe:


bcoe:
Guess I'll just have to get used to the fact that Travis doesn't work any more.

PR #178, opened 2015-06-17
Avoid help, version, validation, and the completion command while handling the completion option

### LINKED ISSUE #177: Option completion doesn't work with strict arg parsing
If `strict()` is called, completion of options generates errors.  Here's an example `test.js` script:

```js
#!/usr/bin/env node
require('yargs').completion('completion').help('help').strict().argv;
```

I can enable completion with this:
```bash
$ eval "$( ./test.js completion )"
```

Then if I try to complete the `--help` option after typing `--h[TAB]`, things fail:
```
$ ./test.js --hCommands:
  completion  generate bash completion script

Options:
  --help  Show help  [boolean]

Unknown argument: h
```

This makes sense, given that this is what is being called:
```
./test.js --get-yargs-completions ./test.js --h
```

### LINKED ISSUE #179: Infinite completion loop
Given this `test.js` script:
```
#!/usr/bin/env node
var argv = require('./index')
  .completion('completion')
  .argv;

console.log('argv: %j', argv);
```

Enabling completion:

```bash
$ eval "$( ./test.js completion )"
```

And typing `./test.js [TAB][TAB]`, it looks like you get stuck in an endless completion loop with this shown:
```
./test.js completion 
```

You need to send `SIGINT`to get back to sanity (or just watch the endless `-bash: fork: Resource temporarily unavailable`).

tschaub:
While handling the completion option, `parseArgs` is called with a "partial" set of arguments.  With strict parsing, these partial arguments may cause the task to exit with an error.  This screws up completion.  See #177 for an example script.

The changes here add a `parseOnly` argument to the `parse` method.  While the completion values are being determined, `yargs.parse(args, true)` is called to avoid validation.  This allows people to complete options whose partial values may not be valid options.

The test helper needed to be updated so that logging doesn't continue after `process.exit()` has been called.  While processing the `--get-yargs-completions` option in the tests, the call to `process.exit()` doesn't stop execution, and errors were being logged about `get-yargs-completions` being an unknown argument.  To make an assertion about there being no logged errors in the tests, the helper stops errors from being logged after `process.exit()` has been called.

Fixes #177.
Fixes #179.

### DIFF
--- a/index.js
+++ b/index.js
@@ -229,8 +229,8 @@ function Argv (processArgs, cwd) {
     return self
   }
 
-  self.parse = function (args) {
-    return parseArgs(args)
+  self.parse = function (args, skipValidation) {
+    return parseArgs(args, skipValidation)
   }
 
   self.option = self.options = function (key, opt) {
@@ -399,7 +399,7 @@ function Argv (processArgs, cwd) {
     enumerable: true
   })
 
-  function parseArgs (args) {
+  function parseArgs (args, skipValidation) {
     var parsed = Parser(args, options)
     var argv = parsed.argv
     var aliases = parsed.aliases
@@ -452,6 +452,10 @@ function Argv (processArgs, cwd) {
         return
       }
     })
+
+    if (skipValidation) {
+      return argv
+    }
 
     validation.nonOptionCount(argv)
     validation.missingArgumentValue(argv)
--- a/index.js
+++ b/index.js
@@ -408,6 +408,26 @@ function Argv (processArgs, cwd) {
 
     self.parsed = parsed
 
+    if (parseOnly) {
+      return argv
+    }
+
+    // process completion option before all others
+    if (completionOpt in argv) {
+      // we allow for asynchronous completions,
+      // e.g., loading in a list of commands from an API.
+      completion.getCompletion(function (completions) {
+        ;(completions || []).forEach(function (completion) {
+          console.log(completion)
+        })
+
+        if (exitProcess) {
+          process.exit(0)
+        }
+      })
+      return argv
+    }
+
     // generate a completion script for adding to ~/.bashrc.
     if (completionCommand && ~argv._.indexOf(completionCommand)) {
       self.showCompletionScript()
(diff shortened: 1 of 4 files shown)


tschaub:
It looks like there is more work needed to properly handle completion of an option for a command.

The simple case now works:
```
foo --h[TAB]
```

But this still fails:
```
foo bar --h[TAB]
```

tschaub:
In addition to the issue above, option completion is also not working as expected with aliases.  Given the following `test.js` script:
```js
#!/usr/bin/env node
var argv = require('./index')
  .completion('completion')
  .help('help')
  .alias('help', 'h')
  .strict()
  .argv;

console.log('argv: %j', argv);
```

Setting up completion:
```bash
$ eval "$( ./test.js completion )"
```

And then typing `./test.js --h[TAB]` results in this:
```
$ ./test.js --help, 
```

(note the trailing comma)

tschaub:
In 856db9c1f5f94cd537041e6d277335f3df1bfc6c, the completion option is always handled before the help or version options.  Previously, the order options were handled depended on the order of `Object.getKeys(argv)`.  So when an option starting with an `h` or a `v` was being completed, the help or version option handling might have come first (which is not what you want when you're passing the completion option).

It would be cleaner to separate the `parse` method into parse, validate, and process methods (where process would handle things like help and version).  But I was hoping to minimize the changes in an effort to get in a fix for option completion.

tschaub:
In 2c64dd0a9181da59997a9d29046311d7bc5fe4f1, the completion option is handled before the completion command.  Without this change, `./test.js completion [TAB]` results in an infinite loop (see #179).  While I know this is becoming harder to review, fixing this issue in this branch is a more straightforward change than fixing it in master.

bcoe:
@tschaub whoops, I was on a long flight today and accidentally patched many of these same issues:

https://github.com/bcoe/yargs/pull/181

I think I also solved: #179 on this branch, mind taking a look?

Sorry I stepped on your toes, would you be interested in tackling:

https://github.com/bcoe/yargs/issues/180

Your work is very much appreciated :100:

bcoe:
We were 100% on the same page with regards to these fixes, I'm only opting for #181 since it has a patch for the infinite problem.

tschaub:
No problem.  The changes in #181 look good.  I'll test out the branch now.

PR #181, opened 2015-06-18
patches for completion, brought about by #177

bcoe:
patches for completion brought about by #177:

* there were problems with completing the completion argument itself, related to it attempting to print out the completion script.
* there were problems completing `--help`, caused by the help output firing.
* there were problems related to strict argument parsing, because `--get-yargs-completions` was not whitelisted.

### DIFF
--- a/index.js
+++ b/index.js
@@ -356,7 +356,7 @@ function Argv (processArgs, cwd) {
     }
 
     // register the completion command.
-    completionCommand = cmd
+    completionCommand = cmd || 'completion'
     completionOpt = completion.completionKey
     self.command(completionCommand, desc || 'generate bash completion script')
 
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -426,18 +435,10 @@ function Argv (processArgs, cwd) {
       }
     }
 
+    // we must run completions first, a user might
+    // want to complete the --help or --version option.
     Object.keys(argv).forEach(function (key) {
```
tschaub: I'm curious about this style of coding (`Object.keys(argv).forEach(func))` when you're only checking for the presence of a single key).  I would find this more straightforward:

```js
if (completionOpt in argv) {
  // do whatever
}
```

Are you intentionally guarding against `completionOpt` being in the `Object` prototype?

bcoe: no, it's because I cargo-culted the code from inside the loop that did check for multiple cases. I like your snipped better, and will update the pull.

PR #183, opened 2015-06-18
Avoid completing already provided commands

tschaub:
This makes it so a command that is already in the provided arguments will not be one of the suggested completions.  Since yargs doesn't represent commands more than once in the parsed `argv`, I assume this is non-controversial.

However, this does mean that if a program supports both `foo` and `foobar` as commands, typing `program foo[TAB]` would complete to `foobar`.  I could add some additional logic so that both `foo` and `foobar` were suggested completions in this case.  @bcoe would you be interested in support for this?

### DIFF
--- a/lib/completion.js
+++ b/lib/completion.js
@@ -31,7 +31,9 @@ module.exports = function (yargs, usage) {
 
     if (!current.match(/^-/)) {
       usage.getCommands().forEach(function (command) {
-        completions.push(command[0])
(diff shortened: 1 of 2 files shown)


bcoe:
I'm fine with this, I think it's a nice optimization.

I'll get this merged as well when I cut a release later tonight.

tschaub:
Thanks @bcoe.  I look forward to making use of the completion improvements.

PR #185, opened 2015-06-22
Enable completion of options for commands

### LINKED ISSUE #180: Option completion doesn't work for commands
Given this `test.js` script:
```js
#!/usr/bin/env node
var argv = require('./index')
  .completion('completion')
  .command('bar', 'bar command', function(yargs) {
    var subArgv = yargs
      .option('foo', {
        describe: 'foo option'
      })
      .argv;

    console.log('subArgv: %j', subArgv);
  })
  .argv;

console.log('argv: %j', argv);
```

Enabling completion:
```bash
$ eval "$( ./test.js completion )"
```

And typing `./test.js bar --f` doesn't result in the `foo` option being completed.

tschaub:
This allows options for commands to be completed.  It requires that a command is configured using `completion`.  E.g.

```js
#!/usr/bin/env node
var argv = require('yargs')
  .completion()
  .command('foo', 'foo command', function(yargs) {
    var sub = yargs
      .completion()
      .option('bar', 'bar option')
      .argv;
  })
  .argv;
```

It does not allow for completion of options for nested commands.  This would be possible, but I'm pretty sure it would involve modification of `process.argv` while processing the `--get-yargs-completions` options.  Not too bad, but this is a pretty simple change and may handle most use cases.

Fixes #180.

### DIFF
--- a/index.js
+++ b/index.js
@@ -433,16 +433,6 @@ function Argv (processArgs, cwd) {
       }
     }
 
-    // if there's a handler associated with a
-    // command defer processing to it.
-    var handlerKeys = Object.keys(self.getCommandHandlers())
-    for (var i = 0, command; (command = handlerKeys[i]) !== undefined; i++) {
-      if (~argv._.indexOf(command)) {
-        self.getCommandHandlers()[command](self.reset())
-        return self.argv
-      }
-    }
-
     // we must run completions first, a user might
     // want to complete the --help or --version option.
     if (completionOpt in argv) {
@@ -458,6 +448,16 @@ function Argv (processArgs, cwd) {
         }
       })
       return
+    }
+
+    // if there's a handler associated with a
+    // command defer processing to it.
+    var handlerKeys = Object.keys(self.getCommandHandlers())
+    for (var i = 0, command; (command = handlerKeys[i]) !== undefined; i++) {
+      if (~argv._.indexOf(command)) {
+        self.getCommandHandlers()[command](self.reset())
+        return self.argv
+      }
     }
 
     Object.keys(argv).forEach(function (key) {
--- a/lib/completion.js
+++ b/lib/completion.js
@@ -29,6 +29,13 @@ module.exports = function (yargs, usage) {
       }
     }
 
+    var handlers = yargs.getCommandHandlers()
+    for (var i = 0, ii = previous.length; i < ii; ++i) {
+      if (handlers[previous[i]]) {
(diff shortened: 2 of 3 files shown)


bcoe:
@tschaub this is awesome, thanks for taking this on. A couple things I noticed during initial testing:

* we should probably make it so the `completion` command itself only completes when it's at the top level.
* I managed to get this strange output when I completed a `--` argument: 

```shell
Benjamins-MacBook-Pro:renv benjamincoe$ ./bin/renv.js config --for-bash: compgen: --: invalid option
compgen: usage: compgen [-abcdefgjksuv] [-o option] [-A action] [-G globpat] [-W wordlist] [-P prefix] [-S suffix] [-X filterpat] [-F function] [-C command] [word]
```

This might actually be related to #186, I can work evenings this week (after 6PM PDT) to get this out the door. I feel like it's 99% of the way there.

tschaub:
> I managed to get this strange output when I completed a -- argument

Good catch.  That is indeed from #186.  See 0765cbcefb921577b4d3b2fd81eb20aa32da8b4b for a fix.

> we should probably make it so the completion command itself only completes when it's at the top level

Agreed it's a bit weird, but it is still a valid command (and the output is legit).  I was also wondering about hiding `completion` altogether (not even displaying it in the usage) - maybe handling it with another built in option like `--get-yargs-completion-script`.  This could allow people to enable completion without needing to call `completion` in the config.  But that's more involved.

I'll see what it looks like to avoid suggesting `completion` if there is already one command provided.

tschaub:
One alternative would be to support "hidden" commands.  Hidden commands wouldn't show up in usage and wouldn't be suggested for completion.  Then `yargs.completion()` would create a hidden command.  Not sure if others actually want to see `completion` in the help output though.

tschaub:
Here's a branch that adds support for "hidden" commands and hides the `completion` command: https://github.com/bcoe/yargs/compare/bcoe:master...tschaub:hidden

Hidden commands don't show up in usage.  Only commands in usage are eligible for completion.

Discuss.

bcoe:
@tschaub give #187 a spin. By making it so we always look for `get-yargs-completions`, we eliminate the need to call `completion()` on the inner yargs.

This solves the issue of `completion` being suggested as a completion for inner commands, without needing to introduce the concept of hidden options.

tschaub:
#187 looks like a good solution.  Thanks @bcoe.

PR #186, opened 2015-06-22
Default to filename completion if no other matches are found

tschaub:
If calling a script with `--get-yargs-completions` doesn't result in any matches for the current word, we can fall back to the default filename completion.

See suggestion in https://github.com/bcoe/yargs/issues/180#issuecomment-113799086.

### DIFF
--- a/completion.sh.hbs
+++ b/completion.sh.hbs
@@ -16,6 +16,12 @@ _yargs_completions()
     type_list=`{{app_path}} --get-yargs-completions $args`
 
     COMPREPLY=( $(compgen -W "${type_list}" -- ${cur_word}) )
+
+    # if no match was found, fall back to filename completion
+    if [ ${#COMPREPLY[@]} -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


bcoe:
@tschaub I like this a lot, let the bash do what it's good at :+1:

tschaub:
0765cbcefb921577b4d3b2fd81eb20aa32da8b4b guards against bad output when trying to complete an option that doesn't match any configured options.  Let me know if you'd like these squashed.

completion.sh.hbs
```diff
@@ -16,6 +16,12 @@ _yargs_completions()
     type_list=`{{app_path}} --get-yargs-completions $args`
 
     COMPREPLY=( $(compgen -W "${type_list}" -- ${cur_word}) )
+
+    # if no match was found, fall back to filename completion
+    if [ ${#COMPREPLY[@]} -eq 0 ]; then
+      COMPREPLY=( $(compgen -f -- "${cur_word}" ) )
```
bcoe: awesome :+1:

PR #187, opened 2015-06-23
a couple minor tweaks to the command-specific completion logic in #185

bcoe:
I've made a couple minor tweaks to the command-specific-completion logic in #185 

we now always look for the `get-yargs-completions` option, so `yargs.completion()` only needs to be called when registering a top-level command to generate a completion script.

### DIFF
--- a/index.js
+++ b/index.js
@@ -66,7 +66,6 @@ function Argv (processArgs, cwd) {
     strict = false
     helpOpt = null
     versionOpt = null
-    completionOpt = null
     commandHandlers = {}
     self.parsed = false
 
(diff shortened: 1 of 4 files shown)


bcoe:
@tschaub mind giving this branch a spin with your most up-to-date bash completion script? I think this solves my complaint about `completion` being recommended multiple times.

If things are looking good, we can cut a release candidate tomorrow `yargs@next`.

tschaub:
Nice solution @bcoe.  This is working well for me.

I've got a follow up question regarding hidden commands, but I'll carry on with that in a fresh issue.

bcoe:
@tschaub, cool I'll get a release out tonight.

PR #190, opened 2015-06-24
Hide commands with false for description

### LINKED ISSUE #188: Support for hidden commands
There are cases where it would be nice to hide some commands from the help/usage output.  An example would be setup or post-install work that is not part of the typical user workflow (e.g. `completion`).  The help output becomes less useful when it is diluted with these one-time or setup type commands.

This could be supported by not adding usage for commands with `false` for a description.  E.g.
```js
yargs.command('post-install', false, handler);
```

If someone wanted the `completion` command to not show up in the help doc, this could be made to work:
```js
yargs.completion('completion', false);
```

Only commands that are registered with usage are eligible for completion, so "hidden" commands would not show up as completion candidates.  I think this is consistent with being hidden.

Prior art:
 * [commander's `noHelp`](https://github.com/tj/commander.js/pull/383)
 * [nomnom's `hidden`](https://github.com/harthur/nomnom#hidden)
 * [argparse's `argparse.SUPPRESS`](https://docs.python.org/dev/library/argparse.html#help)

tschaub:
This makes it so calling `yargs.command('some-command', false)` will create a hidden command.  Hidden commands are not shown in the help output and are not eligible for completion.

Fixes #188.

### DIFF
--- a/README.md
+++ b/README.md
@@ -509,8 +509,10 @@ present script similar to how `$0` works in bash or perl.
 
 Document the commands exposed by your application.
 
-use `desc` to provide a description for each command your application accepts (the
-values stored in `argv._`).
+Use `desc` to provide a description for each command your application accepts (the
+values stored in `argv._`).  Set `desc` to `false` to create a hidden command.
+Hidden commands don't show up in the help output and aren't available for
+completion.
 
 Optionally, you can provide a handler `fn` which will be executed when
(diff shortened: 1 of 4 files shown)


bcoe:
@tschaub this is great. I've been having a busy week, but will try to get a release out by early in the weekend.

tschaub:
No rush.  Keep up the good work!

PR #203, opened 2015-07-08
Only check for commands at current nesting level

### LINKED ISSUE #202: Nested commands are treated as a flat structure
Given the following command hierarchy and `argv`, `create` will be invoked instead of triggering `check` on the `profile` handler:

```
$0
    create
    profile
        ls
        rm
        ... (no 'create' at this level)
```

ARGV: `['profile', 'create']`.

This is because yargs does not store the nested nature of commands and instead maintains them in a flat hash. Although `profile create` is an invalid command, because both `profile` and `create` each exist, the incorrect command will be invoked.

See: https://github.com/bcoe/yargs/blob/master/index.js#L434

There doesn't seem to be an easy solution to this problem since yargs doesn't maintain any state at the initialization stage as to the relationship between commands. The only case in which this becomes a problem is when a command is defined at a level above the 'category' in which the sub-command is invoked.

What do you think?

ggoodman:
Fixes #202 by tracking the current level of command nesting instead of searching the entire `argv._` for matching commands (even if these may have previously been matched).

### DIFF
--- a/index.js
+++ b/index.js
@@ -432,11 +432,11 @@ function Argv (processArgs, cwd) {
     // command defer processing to it.
     var commandName = argv._[self.commandIndex]
     var handler = self.getCommandHandlers()[commandName]
-    
+
     if (handler) {
-        self.commandIndex++;
-        handler(self.reset())
-        return self.argv;
+      self.commandIndex++
+      handler(self.reset())
+      return self.argv
     }
 
     // generate a completion script for adding to ~/.bashrc.
--- a/index.js
+++ b/index.js
@@ -417,6 +417,7 @@ function Argv (processArgs, cwd) {
     argv.$0 = self.$0
 
     self.parsed = parsed
+    self.commandIndex = self.commandIndex || 0
 
     // while building up the argv object, there
     // are two passes through the parser. If completion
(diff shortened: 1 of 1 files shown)


ggoodman:
As far as I can tell all tests pass on this. The issue appears to be at the Travis level.

bcoe:
Hey @ggoodman thanks for the pull, the build is failing because a few standard style nits -- standard is run as part of the build process:

```
standard: Use JavaScript Standard Style (https://github.com/feross/standard)
  /home/travis/build/bcoe/yargs/index.js:435:4: Trailing spaces not allowed.
  /home/travis/build/bcoe/yargs/index.js:437:6: Expected indentation of 6 characters.
  /home/travis/build/bcoe/yargs/index.js:437:28: Extra semicolon.
  /home/travis/build/bcoe/yargs/index.js:438:6: Expected indentation of 6 characters.
  /home/travis/build/bcoe/yargs/index.js:439:6: Expected indentation of 6 characters.
  /home/travis/build/bcoe/yargs/index.js:439:25: Extra semicolon.
```

I'd also like to add some tests for this behavior, I'm probably going to hack on yargs a bit this weekend. If you could provide some sample code which would have failed prior to this patch, I'd even be happy to work on adding some tests.

ggoodman:
```js
Yargs = require('yargs');

Yargs.command('top', 'Top-level command that exists', function (yargs) {
    var argv = yargs.usage('Call me, maybe.')
        .help('help')
        .argv;
});

Yargs.command('parent', 'Top-level command category', function (yargs) {
    var argv = yargs.command('child', 'The apple didn\'t fall far from the tree', function (yargs) {
        var argv = yargs.usage('Just try me.')
            .help('help')
            .argv;
    })
    .argv;
});

Yargs.parse(process.argv);
```

Trigger `top` unexpectedly (`parent` completely ignored):

```bash
./cli.js parent top
```

I didn't actually try running this code, but this should give you an idea of what triggers unexpected behaviour.

bcoe:
see #206.

PR #204, opened 2015-07-08
ESLint

ChristianMurphy:
Replaces Standard with ESLint.
ESLint is set to use Standard ESLint configuration.
Then adds additional validations.

ref https://github.com/feross/eslint-config-standard

### DIFF
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -160,11 +160,11 @@ module.exports = function (yargs) {
         var desc = descriptions[key] || ''
         var type = null
 
-        if (!options.boolean.indexOf(key)) type = '[boolean]'
-        if (!options.count.indexOf(key)) type = '[count]'
-        if (!options.string.indexOf(key)) type = '[string]'
-        if (!options.normalize.indexOf(key)) type = '[string]'
-        if (!options.array.indexOf(key)) type = '[array]'
+        if (~options.boolean.indexOf(key)) type = '[boolean]'
+        if (~options.count.indexOf(key)) type = '[count]'
+        if (~options.string.indexOf(key)) type = '[string]'
+        if (~options.normalize.indexOf(key)) type = '[string]'
+        if (~options.array.indexOf(key)) type = '[array]'
 
         var extra = [
             type,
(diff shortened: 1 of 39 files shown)


lib/usage.js
```diff
@@ -158,11 +160,11 @@ module.exports = function (yargs) {
...
-        if (~options.boolean.indexOf(key)) type = '[boolean]'
-        if (~options.count.indexOf(key)) type = '[count]'
-        if (~options.string.indexOf(key)) type = '[string]'
-        if (~options.normalize.indexOf(key)) type = '[string]'
-        if (~options.array.indexOf(key)) type = '[array]'
+        if (!options.boolean.indexOf(key)) type = '[boolean]'
+        if (!options.count.indexOf(key)) type = '[count]'
+        if (!options.string.indexOf(key)) type = '[string]'
+        if (!options.normalize.indexOf(key)) type = '[string]'
+        if (!options.array.indexOf(key)) type = '[array]'
```
isaacs: Wrong.

lib/usage.js
```diff
@@ -247,11 +249,11 @@ module.exports = function (yargs) {
...
-        if (~options.boolean.indexOf(alias)) yargs.boolean(key)
-        if (~options.count.indexOf(alias)) yargs.count(key)
-        if (~options.string.indexOf(alias)) yargs.string(key)
-        if (~options.normalize.indexOf(alias)) yargs.normalize(key)
-        if (~options.array.indexOf(alias)) yargs.array(key)
+        if (!options.boolean.indexOf(alias)) yargs.boolean(key)
+        if (!options.count.indexOf(alias)) yargs.count(key)
+        if (!options.string.indexOf(alias)) yargs.string(key)
+        if (!options.normalize.indexOf(alias)) yargs.normalize(key)
+        if (!options.array.indexOf(alias)) yargs.array(key)
```
isaacs: Wrong.  Bitwise NOT !== boolean NOT.  `~` is only falsey if the value is `-1` (ie, the thing is not found).  `!` is falsey if the value is `-1` or > `0` (ie, if the thing is either not found, or found after the first position.)

lib/usage.js
ChristianMurphy: restored original bitwise not

lib/usage.js
ChristianMurphy: restored original bitwise not

bcoe:
Hey @ChristianMurphy I really appreciate the hard work you put into this pull request, but a big part of why I love @feross' standard is that it leaves no `. eslintrc` to argue about.

Rather than accepting this pull request, I encourage you to start a conversation regarding some of these stylistic additions on the `standard` repo, I will happily update yargs' style to match any additions to `standard`.

ChristianMurphy:
Thanks @bcoe and @isaacs for the prompt feedback. :smile: 

@bcoe would you be interested in any of the changes made here being merged, but not being enforced?

The most notable changes being
- strict mode
- replacing `var === undefined` with `typeof var === 'underfined'`
- replacing `throw Error()` with `throw new Error()`
- applying standard style on examples

PR #206, opened 2015-07-13
slight tweaks to @ggoodman's work in #203

bcoe:
I've made a couple slight tweaks to the pull request in #203 (also see #202).

First off, thanks for reporting and tackling this @ggoodman, this makes commands significantly more powerful. I've made a couple small changes:

* I added some unit-tests, including one I found for the edge-case where the node-bin is in `_.argv`.
* I modified your approach slightly:
  * the core of the problem was that commands were getting checked in the order they were defined.
  * the solution was to instead iterate over `argv._` in order.

Could you give this branch a try and let me know if it does the trick for you.

### DIFF
--- a/index.js
+++ b/index.js
@@ -417,7 +417,6 @@ function Argv (processArgs, cwd) {
     argv.$0 = self.$0
 
     self.parsed = parsed
-    self.commandIndex = self.commandIndex || 0
 
     // while building up the argv object, there
     // are two passes through the parser. If completion
@@ -430,14 +429,7 @@ function Argv (processArgs, cwd) {
 
     // if there's a handler associated with a
     // command defer processing to it.
-    var commandName = argv._[self.commandIndex]
-    var handler = self.getCommandHandlers()[commandName]
-
-    if (handler) {
-      self.commandIndex++
-      handler(self.reset())
-      return self.argv
-    }
+    if (handleCommand(argv)) return self.argv
 
     // generate a completion script for adding to ~/.bashrc.
     if (completionCommand && ~argv._.indexOf(completionCommand) && !argv[completion.completionKey]) {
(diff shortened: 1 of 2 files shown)


ggoodman:
@bcoe I've tried your tweaks in this branch but note that this doesn't seem to fully resolve the issue. Given a 'category' like `profile` with commands like `init` and `ls` (but no `create`) as well as a top-level 'command' called `create`, I note the following behaviour:

`cli profile create` --> `Unable to read the file `$(pwd)/create` (It actually prints a path, but what is important is that the path is relative to the working directory and not the cli tool). Expected behaviour would have been to fall back to my top-level `check()` handler that would have printed `no such command profile create` under my proposed changes.

See [auth0/nested-yargs](https://github.com/auth0/nested-yargs) for the library I created to structure this 'category' and 'command' hierarchy.

bcoe:
Could you paste a failing code example again? The problem with the counter
based solution is that it's too strict. As an example your original sample
code failed because of the node bin in process.argv.

I'm happy to put a test around any other edge cases you find, and fix them.

bcoe:
@ggoodman bump, it's been a busy week at npm; but if you could gist me a couple code examples to put tests around again it would be amazing :)

bcoe:
@ggoodman I've written this code example, but I can't create the behavior you're describing on this branch:

```js
var argv = require('yargs')
  .command('profile', 'profile commands', function (yargs) {
    console.log('run profile')

    yargs
      .command('ls', 'ls command', function (yargs) {
        console.log('run ls')
      })
      .command('init')
      .argv
  })
  .command('create', 'create some junk', function (yargs) {
    console.log('run create')
  })
  .argv
```

`./test.js create` prints `run create` like I'd expect.
`./test.js profile ls` prints `run profile run ls` as I'd expect.
`./test.js profile ls create`, prints `run profile run ls`.
`./test.js profile create ls`, prints `run profile run ls` like I'd expect.

I think you've really got some good ideas about how to hammer out yargs' command handling, but I think I'm slightly confused at this point.

bcoe:
@ggoodman closing this pull request for the time being, happy to revisit if we can get a few more test cases around the behavior -- as I said, love your ideas surrounding how to improve command handling.

PR #207, opened 2015-07-21
Use shields from shields.io

SimenB:
For consistency.

image

vs

image

Low quality paste, but the difference is mainly that travis is now svg instead of png, so looks (much) better on hi-res screens

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,10 +5,10 @@ Yargs be a node.js library fer hearties tryin' ter parse optstrings.
 
 With yargs, ye be havin' a map that leads straight to yer treasure! Treasure of course, being a simple option hash.
 
-[![Build Status](https://travis-ci.org/bcoe/yargs.png)](https://travis-ci.org/bcoe/yargs)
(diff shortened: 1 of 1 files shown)


bcoe:
I like this in theory, but I'm finding that I'm getting timeouts and 503s from img.shields.io when I try to access these new URLs:

https://img.shields.io/travis/bcoe/yargs.svg

SimenB:
That link works for me? Maybe they were down just as you tested it?

On Thu, 23 Jul 2015 07:20 Benjamin E. Coe <notifications@github.com> wrote:

> I like this in theory, but I'm finding that I'm getting timeouts and 503s
> from img.shields.io when I try to access these new URLs:
>
> https://img.shields.io/travis/bcoe/yargs.svg
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bcoe/yargs/pull/207#issuecomment-123971399>.
>
-- 

--
Simen Bekkhus

bcoe:
Things are working now, bad timing I think badges.io was down -- I'm happy to pull this in though :+1:

Issue #92, opened 2015-02-17
yargs' documentation needs your help to make it better


billev2k:
The pirate talk is cute, but ... ultimately detracts from the usability of the documentation, and thus, the product. It's particularly difficult for anyone for whom pirate english is not their first language. Which is most of us.

Please create documentation that's designed to communicate how to use the code.

mikermcneil:
> difficult for anyone for whom pirate english is not their first language

@billev2k  haha awesome :)

> Please create documentation that's designed to communicate how to use the code.

Not sure I agree that the pirate speak `ultimately detracts from the usability of the documentation`, but you're totally right that the README could do a better job explaining the usage of this module.  @chevex: If it would be helpful, @irlnathan and I are willing to spend a bit of time enhancing the readme (and just to clarify, "enhancing" !== "removing pirate speak" :)  )

bcoe:
@billev2k tis a mighty fine point, and I tend to be agreen' with ya.

Personally I love the pirate theme, it's fun. But, I think we could celebrate pirate roots of yargs in the first few paragraphs, while tightening up the bulk of the documentation. Personally I would like to see:

* a pirate-themed intro.
* a few well documented real-world examples of the yargs DSL in action (sans pirate talk).
* thorough, standardized, documentation of all the API methods. (sans pirate talk).

What do you think?

acstll:
I just read the entire README and I don't mind the pirate talk but I was like "please just show me the API".

So +1 for this:

> - a pirate-themed intro.
> - a few well documented real-world examples of the yargs DSL in action (sans pirate talk).
> - thorough, standardized, documentation of all the API methods. (sans pirate talk).

bcoe:
@acstll, @billev2k, @mikermcneil I agree with everyone in this thread. I would love some help trimming down the README a bit. patches more than welcome.

ejosafat:
I agree with the idea of pirate lingo being cute, but when english is not your native language, well, it gets too difficult to understand the documentation when you first have to translate from pirate to english, you spend so much effort doing it!!

nexdrew:
@bcoe What do you think about reorganizing the README content into a format similar to [the standard-readme draft](https://github.com/zcei/standard-readme/issues/1)?

Would basically be something like:

```
(Heading Block)
# yargs
![badges]
Piratey description

(Install Block)
# Install
How to install

(Usage/Example Block)
# Examples
Minimal examples?
Show most common usages

(API Block)
# API
## method signature
Description
[...]

PR #208, opened 2015-07-22
Minor README improvements

nexdrew:
I was reading over the docs recently and I noticed a couple problems, so I thought I'd attempt a bit of cleanup. Namely:

- Fix a few spelling and grammatical mistakes (e.g. "retrived" -> "retrieved", "then" -> "than", etc)
- Attempt to make content more consistent
- Update examples with actual output from yargs v3.15.0

Hope this helps. :wink:

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ var util = require('util');
 var argv = require('yargs').argv;
 
 if (argv.s) {
-    util.print(argv.fr ? 'Le perroquet dit: ' : 'The parrot says: ');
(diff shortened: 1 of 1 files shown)


bcoe:
@nexdrew thank you so much \o/

We could use a ton of help with the README.md, thanks for taking a first stab at the problem: see #92

nexdrew:
Thanks, Ben, I'll take a look. Overall, I think the documentation is pretty good, but I may have a couple ideas about how to better organize its content, roughly following the high-level outline you mention in #92. I'll move any further discussion there.

Issue #199, opened 2015-07-06
Please allow usage strings to be customized and/or formatted.


waynebloss:
Hi and thanks for your work. One feature I would like to request is to allow the developer to supply the usage strings instead of using the hard-coded "Commands:" and "Options:".

Alternatively, it would be nice to just be able to configure yargs to add one newline after each of those strings so that the usage output isn't formatted so tightly.

bcoe:
@waynebloss I like this idea, I think it would be good to make every string customizable so that we can make strides towards internationalizing yargs. I'll put some thought into what the API would look like for this on the weekend; I'd like to make strings customizable without complicating the API significantly -- how would you feel about setting all of interface's strings in a single configuration step?

waynebloss:
@bcoe Yarg! A single configuration step sounds good. I think all of the [i18n](npmjs.com/package/i18n) modules I've seen for Node store their strings in a single dictionary/key-value style object which is usually "flat" by default, but can also be nested.

Here's a flat example:

```javascript
yargs.swabTheDeckMatey({
  commandsHeader: 'Commands:',
  optionsHeader: 'Options:',
  otherStuff: 'Value'
})...
```

And the nested one:

```javascript
yargs.swabTheDeckMatey({
  header: {
    commands: 'Commands:',
    options: 'Options:'
  },
  otherStuff: 'Value'
})...
```

The i18n npm package is pretty handy and it will do both - not saying to use it because you probably don't need all that functionality, but it might be good to take a look at.

waynebloss:
Here's the gist of the flat/nested conversation that is probably better illustrated - https://www.npmjs.com/package/i18n#object-notation

bcoe:
@waynebloss awesome, would love to hack on this on the weekend. Keep this thread updated with any thoughts you have.

bcoe:
@waynebloss could you take a look at https://github.com/bcoe/yargs/pull/210 and try out the branch, I finally got a chance to tackle this issue.

waynebloss:
@bcoe Hello, I will set a reminder to do that near the end of the week or maybe Saturday.

bcoe:
@waynebloss I've released this functionality on the `next` tag of yargs on npm:

`npm install yargs@next`

nexdrew:
[...]

PR #210, opened 2015-07-28
adds i18n support and allows for usage strings to be customized

bcoe:
Based on a discussion with @waynebloss I've internationalized yargs, this kills two birds with one stone:

* people have asked to be able to customize usage strings in yargs (see #199), this allows this do be done in a generic way.
* this also opens up yargs to being translated into multiple languages; I think this is really cool.

## The i18n Library Used

@waynebloss pointed me at [i18n](https://www.npmjs.com/package/i18n), I thought this was really cool and I liked the API, but there were some issues with it:

* it had more dependencies than I'd like to pull into yargs (debug, mustache).
* it seems as though i18n leaks into the global scope: http://ejohn.org/blog/a-strategy-for-i18n-and-node/

Using i18n as a model, I put together a little library called [y18n](https://www.npmjs.com/package/y18n):

* it has almost the same API as i18n.
* leaks no global variables.
* has no dependencies.

## Help Wanted

This is a fairly large change, and pulls in a newly minted library, I would greatly appreciate any help I could get QAing it: @waynebloss, @nexdrew, @nylen


lib/validation.js
```diff
@@ -45,11 +47,12 @@ module.exports = function (yargs, usage) {
...
       })
 
-      if (missingRequiredArgs.length === 1) {
-        usage.fail('Missing argument value: ' + missingRequiredArgs[0])
-      } else if (missingRequiredArgs.length > 1) {
-        var message = 'Missing argument values: ' + missingRequiredArgs.join(', ')
-        usage.fail(message)
+      if (missingRequiredArgs.length > 0) {
+        usage.fail(__n(
+          'Missing argument value: %s', 'Missing argument values: %s',
```
nexdrew: Minor nit: I like how you put the singular and plural strings on their own lines elsewhere (e.g. lines 84-85, lines 116-117) since it helps make the differences much more obvious IMO (even based on alignment alone). I would repeat that pattern here.

bcoe: :+1:

test/validation.js
```diff
@@ -139,6 +139,7 @@ describe('validation tests', function () {
         .choices('system', ['linux', 'mac', 'windows'])
         .choices('arch', ['x86', 'x64', 'arm'])
         .fail(function (msg) {
+          console.log(msg)
```
nexdrew: I'm assuming this was a sanity check. Did you intend to leave it in here?

bcoe: @nexdrew good catch; I'll nuke that.

nexdrew:
Not sure if y18n is shooting for feature parity with [i18n](https://www.npmjs.com/package/i18n) or [i18n-2](https://www.npmjs.com/package/i18n-2) (excluding integration with express or templating libraries), but it currently doesn't handle the following "nested" usage of `__n` from their docs:

```js
var singular = i18n.__n('There is one monkey in the %%s', 'There are %d monkeys in the %%s', 1, 'tree');
var plural = i18n.__n('There is one monkey in the %%s', 'There are %d monkeys in the %%s', 3, 'tree');
console.log(singular); //There is one monkey in the tree
console.log(plural); //There are 3 monkeys in the tree
```

I assume the equivalent with y18n would be this:

```js
var singular = y18n.__n('There is one monkey in the %s', 'There are %s monkeys in the %s', 1, 1, 'tree');
var plural = y18n.__n('There is one monkey in the %s', 'There are %s monkeys in the %s', 3, 3, 'tree');
console.log(singular); //There is one monkey in the 1 tree => WHOOPS
console.log(plural); //There are 3 monkeys in the tree
```

The only way this seems to work with y18n (or with `util.format()` in general) is to know not to pass the first placeholder argument for the singular scenario:

```js
var singular = y18n.__n('There is one monkey in the %s', 'There are %d monkeys in the %s', 1, 'tree');
console.log(singular); //There is one monkey in the tree => BINGO
```

Again, not sure this is really an issue to care about, but it *is* a difference.

lib/usage.js
```diff
@@ -151,23 +152,23 @@ module.exports = function (yargs) {
...
+        if (~options.string.indexOf(key)) type = '[' + __('string') + ']'
+        if (~options.normalize.indexOf(key)) type = '[' + __('string') + ']'
+        if (~options.array.indexOf(key)) type = '[' + __('array') + ']'
 
         var extra = [
             type,
-            demanded[key] ? '[required]' : null,
-            options.choices && options.choices[key] ? '[choices: ' +
+            demanded[key] ? '[' + __('required') + ']' : null,
+            options.choices && options.choices[key] ? '[' + __('choices') + ': ' +
```
nexdrew: Most usage and validation text containing colons includes the colon as part of the localized string value (e.g. `__('Commands:')`, `__('Options:')`, `__('Examples:')`, etc), which seems appropriate. The only exceptions are here (`__('choices') + ': '`) and line 293 (`__('default') + ': '`). Perhaps the colon should be included with these as well?

README.md
```diff
@@ -875,6 +875,83 @@ if (command === 'hello') {
 }
 ```
 
+.locale(locale)
+---------------
+
+Set a locale other than the default `en` locale:
+
+```js
+var argv = require('./')
```
nexdrew: This should probably be `require('yargs')` instead of `require('./')`. :smiley_cat:

nexdrew:
Just to clarify the intent:

1. yargs will now come with a static set of supported languages for the usage and validation text it auto-generates for consumers (additional languages will be provided by future PRs)
2. Consumers can explicitly select one of yargs' supported languages to use, either statically or based on their own CLI option, via `locale()`
3. Consumers can selectively override yargs' text for a given language using `updateStrings()`
4. yargs **DOES NOT** attempt to provide a standard way for CLIs to support multiple languages
 - Either something similar to `help('h')` like `lang('l')` which interprets `-l en` or `-l=fr` as a locale option
 - Or something like `multiLang(true)` which enables interpretation of options like `--en` or `--fr` to set a locale
5. yargs **DOES NOT** attempt to provide consumers with a general means of supporting multiple languages for their own text (meaning text not generated by yargs code)
6. yargs **DOES NOT** attempt to set a default locale based on the end user's system (e.g. via a `LANG` env var or similar)
7. Neither y18n nor yargs attempts to provide locale-specific value formatting or parsing (e.g. numbers, dates, currency, etc) - "locale" really only means "language" at this point

If this is all correct, then I'd say the PR is pretty much spot on.

README.md
```diff
@@ -875,6 +875,83 @@ if (command === 'hello') {
...
+.updateLocale(obj)
+------------------
+.updateStrings(obj)
+------------------
+
+Override the default strings used by yargs with the key/value
+pairs provided in `obj`:
+
+```js
+var argv = require('./')
```
nexdrew: Just noticed this is same as line 884, `require('yargs')` would be better. :repeat:

bcoe:
@nexdrew thanks for the great code review. Addressing your points in order:

1. yes, I'd love to add more languages in future pull requests -- not quite sure what the best approach is for this yet? I was thinking perhaps:
  * open an issue on yargs and flag a few past contributors.
  * tweet about it and try to get a few people interested.
2. yes I was thinking users would explicitly select a language. I was looking at @sindresorhus' awesome https://www.npmjs.com/package/os-locale, but the problem is the lookup is asynchronous which breaks yargs' synchronous core API -- I was thinking we could perhaps point developers at this library, but it might be overkill to pull it in ourselves?
3. I felt the `updateStrings` option allows us to kill two birds with one stone, giving people more ability to format yargs' output: https://github.com/bcoe/yargs/issues/199
  * I felt like this feature might be a bit too niche in isolation, but when introducing it in the context of i18n it's kind of neat.
4. yes, I think that this responsibility stays in the hands of the consuming library; although perhaps we can point them in the direction of `y18n`?
5. likewise, I think this is outside the realm of what we ever want yargs to handle.
6.see my comment with regards to `os-locale`.
7. correct, although this seems like a natural extension down the line?

@nexdrew my overarching question. Does this add value? I like that it addresses a feature people have asked for, while also giving us the opportunity to internationalize ... I could see it helping yargs to get adoption in non-english-speaking countries, what are your thoughts?

bcoe:
@nexdrew I patched y18n to address the API issue you brought up:

https://github.com/bcoe/y18n/pull/3

nexdrew:
:+1: on y18n v3.0.0. I retested and it looks good.

Back to your question:

> Does this add value?

I think it certainly does. Using this as a vehicle to support i18n *and* user customization is a cool feature and a win-win.

I'm just slightly concerned about users' expectations when they read "yargs now supports i18n" in docs, changelog, or announcement. As a user myself, I would expect at least one of the following:

1. Some automatic user locale detection and/or language-switching that my CLI now has out-of-the-box (or that I can opt-in to)
2. Some standard (or opinionated) way that allows my CLI's end users to specify a locale/language to use (something like `help('h')` that I can easily enable)
3. Some way to give yargs different localized versions of *my* text that it outputs for me

Maybe I'm getting ahead of myself. You're probably right that this is out of scope for an initial offering. I mean your enhancements certainly haven't blocked us from doing any of the above; rather, they provide a good foundation to incrementally build upon. But I at least want to bring these up now, if only to consider them for future enhancements.

Otherwise, I think this is some excellent work. I haven't really seen any standard way of tackling i18n for CLIs or any other libs that provide this for CLI apps, so I think this definitely puts/keeps yargs on the forefront. Kudos! :chocolate_bar:

bcoe:
@nexdrew I'm hearing a vote to ship, iterate, and start digging into auto locale detection once we have some translations? FYI, what's your npm user?

nexdrew:
Yes, nicely put.

I am [abg](https://www.npmjs.com/~abg) on npm. I'd be honored. :smile:

bcoe:
@nexdrew done! I'm really grateful for your hard work and thoughtful code reviews.

PR #220, opened 2015-08-03
Added ES locale

zkat:
Here you go <3

### DIFF
--- a/locales/es.json
+++ b/locales/es.json
@@ -1,0 +1,36 @@
+{
+  "Commands:": "Comandos:",
+  "Options:": "Opciones:",
+  "Examples:": "Ejemplos:",
+  "boolean": "boolean",
+  "count": "cuenta",
+  "string": "cadena de caracteres",
+  "array": "tabla",
+  "required": "requisito",
+  "default:": "defecto:",
+  "choices:": "selección:",
+  "generated-value": "valor-generado",
+  "Not enough non-option arguments: got %s, need at least %s": "Hacen falta argumentos no-opcionales: Número recibido %s, necesita por lo menos %s",
+  "Too many non-option arguments: got %s, maximum of %s": "Demasiados argumentos no-opcionales: Número recibido %s, máximo es %s",
+  "Missing argument value: %s": {
+    "one": "Falta argumento: %s",
+    "other": "Faltan argumentos: %s"
+  },
+  "Missing required argument: %s": {
+    "one": "Falta argumento requerido: %s",
+    "other": "Faltan argumentos requeridos: %s"
+  },
+  "Unknown argument: %s": {
+    "one": "Argumento desconocido: %s",
+    "other": "Argumentos desconocidos: %s"
+  },
+  "Invalid values:": "Valores inválidos:",
+  "Argument: %s, Given: %s, Choices: %s": "Argumento: %s, Recibido: %s, Selección: %s",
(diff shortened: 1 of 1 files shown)


nexdrew:
This is great. Thank you so much @zkat!

One more quick thing, I noticed recently we missed a couple strings in the original locale implementation. Would you mind adding the following to `es.json` in anticipation of an upcoming fix, please?

```json
"Show help": "<Your translation here>",
"Show version number": "<Your translation here>"
```

nexdrew:
Found another one:

```json

"Path to JSON config file": "<Translation needed>"
```

Might want to wait until I finish reviewing all missing i18n lookups. :pensive:

nexdrew:
Blimey! I think that's all. Sorry for the hiccup!

```json
"Path to JSON config file": "<Translation needed>",
"Show help": "<Your translation here>",
"Show version number": "<Your translation here>"
```

@zkat If you wouldn't mind adding these to `es.json`, I'd really appreciate it!

zkat:
ping me again when you're done or tell @bcoe to poke me. :) I'll gladly
update things!

On Mon, Aug 3, 2015, 12:05 Andrew Goode <notifications@github.com> wrote:

> Found another one:
>
> "Path to JSON config file": "<Translation needed>"
>
> Might want to wait until I finish reviewing all missing i18n lookups. [image:
> :pensive:]
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/bcoe/yargs/pull/220#issuecomment-127372150>.
>
-- 

kat

bcoe:
@zkat \o/ you're awesome. @nexdrew, perhaps you can sync up with @LoicMahieu too and get some additional French translation?

nexdrew:
@bcoe Sure. I was thinking I would attempt to add the French translation myself (via Google Translate, ick I know), and then see if @LoicMahieu would review the PR for us.

Considering these additional strings, I have an interesting problem for the fix and would like your opinion. The problem is that we need to defer the y18n lookup until after the locale has been set, so that we get consistent behavior b/w these 2 scenarios:

```js
// locale defined BEFORE help option
var argv = require('yargs').locale('fr').help('h').argv
// locale defined AFTER help option
var argv = require('yargs').help('h').locale('fr').argv
```

Here are the options for implementation that I see:

1. Defer the default description lookup until parsing (`parseArgs()` in index.js) to guarantee desired locale and description/custom strings have already been set
2. Keep default description lookup in option methods (e.g. `addHelpOpt()`) and then attempt to manage/reset state if `locale()` or `updateLocale()` methods are subsequently called

Since the 2nd seems kludgy, I am leaning towards the 1st.

Thoughts?

nexdrew:
Actually, if we don't call `describe()` when the option is added, then we run into problems maintaining "order of insertion" for the usage/help content.

Instead of managing state to keep track of what needs a y18n default description and the order in which it was inserted, I'm thinking it might be better to embed the state into the description itself as a special prefix (something like `'__yargsString__:Show help'`) and then check for the prefix and do the y18n lookup when building the usage content in `usage.help()`.

I know that sounds hacky, but it might simplify some things. I'm gonna give it a shot.

nexdrew:
Attempt to address missing y18n lookups with PR #221.

@zkat Feel free to add the 3 strings mentioned above to `es.json` at your convenience. Thanks!

zkat:
@nexdrew @bcoe all done. Any others you'd like?

nexdrew:
That's all for now, thanks @zkat!

PR #221, opened 2015-08-03
Add missing locale strings to y18n lookup

nexdrew:
:heavy_plus_sign: Further testing revealed the following strings used in `index.js` were missing from y18n lookup:

```json
"Path to JSON config file": "Path to JSON config file",
"Show help": "Show help",
"Show version number": "Show version number"
```

This PR fixes that.

Note that [the first commit](https://github.com/bcoe/yargs/commit/3a523eb086c4c877f2e837b3ee9f02523e6fe0fa) takes the naive approach, which results in using the `en` version for `'Show help'` in this scenario:

```js
var argv = require('yargs').help('h').locale('pirate').argv
```

[The second commit](https://github.com/bcoe/yargs/commit/01e91c7f9ad51d67f235540567ed8362d97760c6) addresses this by deferring the y18n lookup until the whole usage content is generated, instead of doing it when the option is defined.

Also note that this PR **does not** attempt to address the problem of updating locale strings *before* the desired locale is set (if it even *is* a problem):

```js
var argv = require('yargs')
  .help('h')
  .updateStrings({
    'Show help': 'Use this here help desc instead' // applies to default locale en
  })
  .locale('pirate') // changes locale and thus discards previously updated strings
  .argv
```

See related discussion in PR #220.

### DIFF
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -91,6 +91,11 @@ module.exports = function (yargs, y18n) {
     wrap = cols
   }
 
+  var deferY18nLookupPrefix = '__yargsString__:'
+  self.deferY18nLookup = function (str) {
+    return deferY18nLookupPrefix + str
+  }
+
   self.help = function () {
     normalizeAliases()
 
@@ -158,6 +163,8 @@ module.exports = function (yargs, y18n) {
         var kswitch = switches[key]
         var desc = descriptions[key] || ''
         var type = null
+
+        if (~desc.lastIndexOf(deferY18nLookupPrefix)) desc = __(desc.substring(deferY18nLookupPrefix.length))
 
         if (~options.boolean.indexOf(key)) type = '[' + __('boolean') + ']'
         if (~options.count.indexOf(key)) type = '[' + __('count') + ']'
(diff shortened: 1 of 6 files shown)


lib/usage.js
```diff
@@ -91,6 +91,11 @@ module.exports = function (yargs, y18n) {
     wrap = cols
   }
 
+  var deferY18nLookupPrefix = '__yargsString__:'
```
bcoe: This is a bit weird, but like you say, also fairly elegant! :shipit:

nexdrew:
@bcoe Just so I'm clear, :shipit: = ok to merge?

Do you want me to attempt a release once the `es.json` and `fr.json` PRs are merged, or is that something I should leave to you? I assume the general steps are something like:

1. update CHANGELOG.md in master
2. npm version
3. git push && git push --tags
4. npm publish --tag next
5. test yargs@next
6. npm dist-tag add yargs@x.y.z latest

bcoe:
:shipit: = ship it.

You've got the steps correct for the release, but let's wait for @kat to update the Spanish translation before we release. I talked to her today she's on it :)

You can push the next release once we get things merged :+1: -- I think I'll have time for some yargs hacking this weekend, perhaps we can close a few more issues.

PR #222, opened 2015-08-04
Add missing locale strings to FR locale

nexdrew:
:fr: Some strings were found to be previously missing from y18n lookup; this adds them to `fr.json` with an attempted French translation:

```json
"Path to JSON config file": "Chemin du fichier de configuration JSON",
"Show help": "Affiche de l'aide",
"Show version number": "Affiche le numéro de version"
```

See related discussion in PRs #220 and #221.

@LoicMahieu If you have time, I would greatly appreciate a quick review of these French translations. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -676,7 +676,7 @@ Choose yer command:
 
 Options for me hearties!
   --option, -o  'tis a mighty fine option               [requi-yar-ed]
-  --help        Show help                                    [boolean]
+  --help        Parlay this here code of conduct             [boolean]
 
 Ex. marks the spot:
   test.js run foo  shiver me timbers, here's an example for ye
@@ -687,8 +687,8 @@ Ye be havin' to set the followin' argument land lubber: option
 Locales currently supported:
 
 * **en:** American English.
-* **fr:** French.
 * **es:** Spanish.
+* **fr:** French.
 * **pirate:** American Pirate.
 
 To submit a new translation for yargs:
@@ -933,6 +933,9 @@ My Commands -->
 Options:
   --help  Show help  [boolean]
 ```
+
+If you explicitly specify a `locale()`, you should do so *before* calling
+`updateStrings()`.
 
 .usage(message, [opts])
 ---------------------
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -235,7 +235,7 @@ module.exports = function (args, opts, y18n) {
   function eatNargs (i, key, args) {
     var toEat = checkAllAliases(key, opts.narg)
 
-    if (args.length - (i + 1) < toEat) error = Error(__('not enough arguments following: %s', key))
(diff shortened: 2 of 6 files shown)


LoicMahieu:
Looks perfect :rocket:
Just a remark not related specially to french language. It's about "JSON" or "json", we use the both "syntax" for describing the same thing. It would be good to harmonize ?
Personally I prefer the upper case.
/cc @bcoe

nexdrew:
Thanks @LoicMahieu!

I agree there are some inconsistencies in the text that yargs outputs. Not only the treatment of "JSON" vs "json" (which should probably be upper-case, as you suggest), but also a couple error messages do not start with a capital letter.

I'll see what I can do to clean this up. Thanks again!

nexdrew:
With [commit 57af0ab](https://github.com/bcoe/yargs/commit/57af0abea631eadb6c936858730f6e8fc957505f), I made the following changes to address the inconsistencies in current locale strings:

- Change "json" -> "JSON" for relevant y18n keys and values for all locales
- Make sure all error messages start with a capital letter, including y18n keys and values for all locales

All tests so far look good. I will make sure the necessary changes are propagated to PR #224 for the Portuguese translation as well.

nexdrew:
@bcoe [My last change](https://github.com/bcoe/yargs/pull/222#issuecomment-128071530) will "break" user code that attempts to customize the following 2 error message keys via `updateLocale()`:

- `"not enough arguments following: %s"`
- `"invalid json config file: %s"`

"Break" meaning their code will no longer override the default messages, since the y18n keys have changed.

It seems we would either:

1. Stick with this change and bump major version when it is merged/released
2. Revert the key change and live with the inconsistency between keys and values

What is your preference?

bcoe:
@nexdrew given how new this feature is, I think I'm comfortable making this a `minor` bump:

1. I doubt many consuming libraries have customized these strings.
2. I would argue slight copy edit could be considered a patch release -- I do think, as you've outlined, we should avoid it since it can be a hassle for upstream libraries.

Now that we have translations in for French and Spanish, feel free to cut a release :+1:

Let's hold off on releasing Portuguese, @seldo would like to announce this translation as part of his talk at Brazil JS.

In other exciting news @sindresorhus is looking into creating a sync version of os-locale, so that we can automatically detect people's locale :dancer:

nexdrew:
@bcoe Sounds good. I'll merge this as is and cut a `3.18.0` release today.

nexdrew:
`yargs@next` now points to `3.18.0` :grimacing:

nexdrew:
Ran a bunch of tests without problem. `3.18.0` is now `yargs@latest`. :sunglasses:

bcoe:
@nexdrew \o/

PR #223, opened 2015-08-04
Make booleans work in config files

sgentle:
Hi there,

Booleans don't seem to be working in config files. I added a test and a potential fix, but I'm not sure if it's the best way to proceed. Happy for feedback!

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -328,7 +328,7 @@ module.exports = function (args, opts, y18n) {
           Object.keys(config).forEach(function (key) {
             // setting arguments via CLI takes precedence over
             // values within the config file.
-            if (argv[key] === undefined) {
+            if (argv[key] === undefined || (flags.bools[key] && !argv[key])) {
               delete argv[key]
               setArg(key, config[key])
             }
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -12,7 +12,7 @@ module.exports = function (args, opts, y18n) {
 
   var __ = y18n.__
   var error = null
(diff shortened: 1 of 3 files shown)


sgentle:
I realised my original fix doesn't work when you want to override a boolean in the config from the cli.

nexdrew:
Thanks @sgentle! I will try to take a look at this today. I'd like to get PR 222 settled before merging others, but hopefully that will be soon.

nexdrew:
Looks pretty good to me.

The only potentially undesirable side-effect I have noticed is that the `help()` option and `version()` option can now be given a truthy value (e.g. `true`, `1`, `{}`, `[]`) from the config file. As you can imagine, this results in displaying the help content or version number when only specifying the config argument in command line, which is a little odd.

We obviously don't have this problem with the current version of yargs, but that's only because it ignores *all* boolean arguments defined in a config file, as you've found.

I think this is an acceptable oddity, but I'd rather defer to @bcoe's judgment.

@bcoe Thoughts? Seems to be a case of "you get what you ask for" to me. Otherwise, we'd have to update the parser to have knowledge of the `help()` and `version()` options.

bcoe:
I'll review and merge for early tomorrow. Love the fix, and nothing bad jumps out at me, but I'm always especially careful in code that touches the parser.

bcoe:
see #226

bcoe:
@sgentle I've published this to the next branch of yargs:

`npm install yargs@next`

Give this a shot, and let me know if it does the trick for you.

PR #226, opened 2015-08-08
a couple slight tweaks to @sgentle's configuration fixes

bcoe:
@sgentle made a couple slight tweaks to #223 to make it even more awesome:

* defaulted now takes into account aliases, this way you can have the `help` key set in your config file, and if you have `.alias('h', 'help')` yargs will set both the `h` and `help` key based on your JSON config.
* I made it so that you can specify `config` in the options block, because why not?

```js
yargs.option('config', {
  config: true
})
```

reviewers: @sgentle, @nexdrew.

### DIFF
--- a/index.js
+++ b/index.js
@@ -276,6 +276,8 @@ function Argv (processArgs, cwd) {
 
       if (demand) {
         self.demand(key, demand)
+      } if ('config' in opt) {
+        self.config(key)
       } if ('default' in opt) {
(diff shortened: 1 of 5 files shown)


sgentle:
Great. Thanks for the merge!

nexdrew:
Got this branch up to date with master and resolved conflict in `test/parser.js`. Should be good to go now. :+1:

PR #228, opened 2015-08-10
Update README.md

etiktin:
I added a comment about working in environments where the argv[1] is not the script name (e.g. Electron), to the `.argv` documentation.


bcoe:
Awesome, thank you.

PR #231, opened 2015-08-16
detect a user's locale using os-locale

bcoe:
This pull moves us to phase 2.0 of locales in yargs, we now automatically detect the operating system's locale using @sindresorhus' [os-locale](https://github.com/sindresorhus/os-locale) module.

```sh
LC_ALL=pt_BR ./test.js --help
```

outputs:

```sh
Opções:
  -h, --help  Mostra ajuda                                             [boolean]
  --foo       what up fool                                          [padrão: 99]
  --config    i be config yo                             [padrão: "./cool.json"]
  -x                                                                [padrão: 10]
```

now we just need to work on getting some more translations.

reviewers: @nexdrew, @nylen

PR #235, opened 2015-08-17
Added German translation

maxrimue:
This adds German translation as said in the title.

Note: `array`, `string` and `boolean` are untouched because translating them to German rather causes confusion than enlightenment as I experienced it, I think that more people know the English words instead of the German words (if you still want me to translate them, tell me).

`count` is untouched because I'm not certain if I am aware of the context it's used in as I'm not having much experience with this module, could you provide an example? I'll fix it then.

### DIFF
--- a/locales/de.json
+++ b/locales/de.json
@@ -3,7 +3,7 @@
   "Options:": "Optionen:",
   "Examples:": "Beispiele:",
   "boolean": "boolean",
-  "count": "count",
+  "count": "Zähler",
   "string": "string",
   "array": "array",
   "required": "erforderlich",
--- a/locales/de.json
+++ b/locales/de.json
@@ -1,0 +1,36 @@
+{
+  "Commands:": "Kommandos:",
+  "Options:": "Optionen:",
+  "Examples:": "Beispiele:",
+  "boolean": "boolean",
+  "count": "count",
+  "string": "string",
+  "array": "array",
+  "required": "erforderlich",
+  "default:": "Standard:",
+  "choices:": "Möglichkeiten:",
+  "generated-value": "Generierter-Wert",
+  "Not enough non-option arguments: got %s, need at least %s": "Nicht genügend Argumente ohne Optionen: %s vorhanden, mindestens %s benötigt",
+  "Too many non-option arguments: got %s, maximum of %s": "Zu viele Argumente ohne Optionen: %s vorhanden, maximal %s erlaubt",
+  "Missing argument value: %s": {
+    "one": "Fehlender Argumentwert: %s",
+    "other": "Fehlende Argumentwerte: %s"
+  },
+  "Missing required argument: %s": {
+    "one": "Fehlendes Argument: %s",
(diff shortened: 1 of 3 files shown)


nexdrew:
@maxrimue Wow, thanks for the translation, and thanks for fixing the incompatibility with latest `standard`!

Here's an example of ["count"](https://www.npmjs.com/package/yargs#yargs-even-counts-your-booleans) in help content (which is, admittedly, not documented very well):

```js
var argv = require('../yargs')
  .locale('de')
  .option('c', {
    count: true,
    describe: 'Wert auf der Anzahl von "c" Argumente gegeben basiert'
  })
  .help('h')
  .wrap(null)
  .argv
console.dir(argv)
```

Run as `node test-de.js -h`:

```
Optionen:
  -c  Wert auf der Anzahl von "c" Argumente gegeben basiert  [count]
  -h  Hilfe anzeigen  [boolean]
```

Run as `node test-de.js -ccc -c`:

```
{ _: [], h: false, c: 4, '$0': 'test-de.js' }
```

So it represents a boolean flag with a value of the flag count instead of just `true`. Does that help determine what you need for the translation?

locales/de.json
```diff
@@ -0,0 +1,36 @@
...
+  "Options:": "Optionen:",
+  "Examples:": "Beispiele:",
+  "boolean": "boolean",
+  "count": "count",
+  "string": "string",
+  "array": "array",
+  "required": "erforderlich",
+  "default:": "Standard:",
+  "choices:": "Möglichkeiten:",
+  "generated-value": "Generierter-Wert",
```
nexdrew: I notice these translated strings (`Standard:`, `Möglichkeiten:`, `Generierter-Wert`) start with a capital letter, even though they're typically not capitalized in other languages. Is this intentional?

maxrimue: @nexdrew Yes, nouns in German are capitalised in general, not only at the beginning of sentences

maxrimue:
@nexdrew What you show there is some usage of the word in written code, I thought that the translation was for output generated by the module though?

maxrimue:
@nexdrew Ready to be merged

nexdrew:
:+1: Awesome, thanks @maxrimue!

Merge notes:
- If we merge PR #231 first (which more explicitly makes the `standard@5.1.0` jump), then we'll need to resolve the conflicting/identical fix here.
- We'll need to remember to add **de** locale to list in README.md, which I assume will be first in the list if we stick to alphabetical-by-code listing

maxrimue:
@nexdrew Okay, how about I make the changes for the README, you merge this pull request, and when you merge #231, there won't be any merge conflicts as we changed the same?

nexdrew:
@maxrimue That will work.

maxrimue:
@nexdrew That's it then

bcoe:
This is great @maxrimue, thanks for the contribution.

maxrimue:
No problem, glad to help!

Issue #140, opened 2015-04-06
Allow more settings for .option()


mscdex:
Currently AFAIK there is no way to use the following with a `.option()` object:

```javascript
var args = require('yargs')
  .option('config', {
    // ...
    config: true
  })
  .option('port', {
    // ...

    // notice no actual default set, this logic is done externally
    // as it depends on the values of other options
    defaultDescription: '80 for HTTP and 443 for HTTPS'
  })
  .argv;
```

So the workarounds for these two missing options are:

```javascript
var args = require('yargs')
  .option('config', {
    // ...
  })
  .config('config')
  .option('port', {
    // ...
  })
  // using `null` is necessary as `undefined` does *not* work
  .default('port', null, '80 for HTTP and 443 for HTTPS')
  .argv;
```

mscdex:
Also I'm not sure how to be able to use "dot notation" when using `option()`?

Using `.option('foo.bar', ...)` just results in a flat object with keys like `"foo.bar": undefined` instead of just `foo: undefined` or `foo: {}` if no foo.* arguments were passed or `foo: { bar: 'asdf' }` if `foo.bar` was passed.

bcoe:
@mscdex happy to add a few more fields to option notation, and to figure out why `dot` notation is giving you trouble. Would you mind opening this as a couple issues though:

* an issue for the `dot-notation` bug.
* an issue for the additional parameters you'd like to see added to `option`

Thanks!

bcoe:
@mscdex #226 adds `config=true` to the options array, I don't quite understand the problem you're having with default descriptions:

```js
yargs.option('port', {
  describe: ''80 for HTTP and 443 for HTTPS',
  default: someVariableILoaded
})
```

As for setting defaults, etc, for dot-notation, let's open a separate ticket for this.

nexdrew:
@bcoe I think @mscdex wants to define a description for the default value (not a description for the option itself) via the `option()` object. I think the only way to do this currently is by calling `default()` and either passing a 3rd argument or passing a named function for the 2nd argument.

Seems easy enough to support a `defaultDescription` property on the `option()` object.

nexdrew:
@mscdex Added `defaultDescription` in `yargs@3.21.0`, which also allows you to explicitly define `default` as `undefined` in the `option()` object:

```js
var argv = require('yargs')
  .option('port', {
    defaultDescription: '80 for HTTP and 443 for HTTPS',
    default: undefined // works this way, but not if property omitted
[...]

Issue #194, opened 2015-07-01
using functions for default doesn't allow to change defaultDescription


kilianc:
This is super ugly ATM because not only it uses `Function.name` which is readonly and forces the developer to use a wrapper function to rename it, but also you can't use parenthesis in the name.

Example:

```js
const moniker_ = require('moniker').choose

let argv = yargs
  .option('id', {
    alias: 'name',
    type: 'string',
    default: function moniker() { return moniker_() } ,
    describe: 'The log stream id'
  })
```

Nasty names_ and wrapper function, when all I wanted is to print `moniker()` there. I used to override `Function.toString()` which is ugly but at least I was able to achieve the desired result.

bcoe:
@kilianc, you are able to change the default description if you use the `.default()` method rather than the `.option()` shorthand:

```js
let argv = yargs
  .option('id', {
    alias: 'name',
    type: 'string',
    describe: 'The log stream id'
  })
  .default('id', require('moniker').choose, 'moniker')
```

Any thoughts, with regards to better syntax -- I like the `.option()` shorthand (and tend to use it myself), but I want to be careful not to overcomplicate the feature ... does it seem reasonable to expect a user to drop down to a method like `.default()` if they need to set complex configuration settings?

kilianc:
probably using a new key `defaultName` if it's a function fixes everything

nexdrew:
See similar request for `defaultDescription` in #140.

nexdrew:
@kilianc `yargs@3.21.0` should take care of this via a `defaultDescription` property defined in the object passed to `option()`:

```js
var argv = require('yargs')
  .option('id', {
    alias: 'name',
    type: 'string',
    default: require('moniker').choose,
    defaultDescription: 'moniker',
    describe: 'The log stream id'
  })
  .argv
```

Please give it a test run and let us know if it satisfies your use case. Thanks!

nexdrew:
Closing based on the addition of `defaultDescription`. If this doesn't satisfy your use-case, please open another issue. Thanks!

PR #237, opened 2015-08-20
Add defaultDescription to option()

nexdrew:
This should fix issues #140 (excluding dot notation problems) and #194.

With this change, you can now do the following:

##### Example from 140
```js
var argv = require('yargs')
  .option('port', {
    defaultDescription: '80 for HTTP and 443 for HTTPS',
    default: null || undefined
  })
  .argv
```

##### Example from 194

```js
var argv = require('yargs')
  .option('id', {
    alias: 'name',
    type: 'string',
    default: require('moniker').choose,
    defaultDescription: 'moniker',
    describe: 'The log stream id'
  })
  .argv
```

And yargs help content will include `[default: 80 for HTTP and 443 for HTTPS]` and `[default: moniker]`, respectively.

Note that I slightly refactored the `default()` method to use this order of precedence for `defaultDescription`:

1. given `defaultDescription` argument
2. predefined `options.defaultDescription[key]` (via `option()`)
3. function description
4. (default value)

Not sure if I went overboard with the unit tests, but I think they all represent different valid scenarios. In fact, I probably could've written twice as many but tried to keep it minimal.

### DIFF
--- a/index.js
+++ b/index.js
@@ -153,11 +153,11 @@ function Argv (processArgs, cwd) {
         self.default(k, key[k])
       })
     } else {
+      if (defaultDescription) options.defaultDescription[key] = defaultDescription
       if (typeof value === 'function') {
-        defaultDescription = usage.functionDescription(value, defaultDescription)
+        if (!options.defaultDescription[key]) options.defaultDescription[key] = usage.functionDescription(value)
         value = value.call()
       }
-      options.defaultDescription[key] = defaultDescription
       options.default[key] = value
     }
     return self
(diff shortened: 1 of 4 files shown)


bcoe:
@nexdrew this seems great to me, no objections to erring on the safe side with regards to tests.

 Perhaps close #140, and ask @mscdex to open a separate issue describing the dot notation problems; also we can probably follow up with #194 and ask if this appropriately addresses the issue?

I'm off the Vancouver to climb all weekend with my wedding party, will be back early next week.

nexdrew:
Cool, don't worry, I'll take care of it. Have a good time!

Issue #225, opened 2015-08-05
multiple commands


jmls:
so, I'm pretty sure that this can be done, but for some reason I am struggling

what I want to have is

```javascript
myapp create project [options]
myapp create module[options]
myapp list [options]
```
etc

so, if the user enters just "myapp" I want to show

```javascript
usage: myapp <command>
commands:
  create : create a new [project|module]
  list : list items in project
```
if the user enters "myapp create"

```javascript
usage: myapp create <item> [options]
item:
  project : create a new project
  module : create a new module
```

at this point I can do the defaults / required for [options] for each particular command

what am I missing ?

nexdrew:
Here's an adaptation from [the `command()` docs](https://www.npmjs.com/package/yargs#command-cmd-desc-fn) that should help get you started:

```js
var yargs = require('yargs')
var argv = yargs
  .usage('usage: $0 <command>')
  .command('create', 'create a new [project|module]', function (yargs) {
    argv = yargs
      .usage('usage: $0 create <item> [options]')
      .command('project', 'create a new project', function (yargs) {
        console.log('creating project :)')
      })
      .command('module', 'create a new module', function (yargs) {
        console.log('creating module :)')
      })
      .help('help')
      .updateStrings({
        'Commands:': 'item:'
      })
      .wrap(null)
      .argv
    checkCommands(yargs, argv, 2)
  })
  .command('list', 'list items in project', function (yargs) {
    console.log('listing projects :)')
  })
  .help('help')
  .wrap(null)
  .argv

checkCommands(yargs, argv, 1)

function checkCommands (yargs, argv, numRequired) {
  if (argv._.length < numRequired) {
    yargs.showHelp()
  } else {
    // check for unknown command
  }
}
```

---

```
$ myapp 
usage: myapp <command>

Commands:
  create  create a new [project|module]
  list    list items in project

Options:
  --help  Show help  [boolean]
```

```
$ myapp create
usage: myapp create <item> [options]

item:
  project  create a new project
  module   create a new module

Options:
  --help  Show help  [boolean]
```

```
$ myapp create project
creating project :)
```

```
$ myapp create module
creating module :)
```

```
$ myapp list
listing projects :)
```

Good luck, and thanks for using yargs!

bcoe:
@nexdrew also see #202, I'd like to find a way to make commands a bit more intuitive -- without complicating the API too much (convention vs. configuration).

nexdrew:
[...]

Issue #163, opened 2015-05-26
using optional handler for .command() causes inconsistent / default behavior


burl:
When using the optional 3rd argument to .command(), yargs behaves as if it has no knowledge of other declared argument types:

```javascript
#!/usr/bin/env node
var argv = require('yargs')
    .boolean('v')
    .command('foo', 'do foo')                 // knows about -v
    .command('bar', 'do bar', function() {})  // as if -v was not declared
    .demand(1)
    .argv;
console.dir([argv.v, argv._])
```

Actual Results:

```shell
localhost$ ./test-yargs.js -v foo
[ true, [ 'foo' ] ]
localhost$ ./test-yargs.js -v bar
[ 'bar', [] ]
```

Expected Results:
```shell
localhost$ ./test-yargs.js -v foo
[ true, [ 'foo' ] ]
localhost$ ./test-yargs.js -v bar
[ true, [ 'bar' ] ]
```

bcoe:
Hey @burl, when you registry a handler for `bar`, it resets the argument parsing as it's passed into the closure. This is so you can document sub-commands. To get the behavior you're looking for, you could do something like this:

```js
#!/usr/bin/env node
var argv = require('yargs')
    .boolean('v')
    .command('foo', 'do foo')                 // knows about -v
    .command('bar', 'do bar', function(yargs) {
       yargs.boolean('v')
           // document options specific to the command `bar`.
    })  // as if -v was not declared
    .demand(1)
    .argv;
console.dir([argv.v, argv._])
```

Any thoughts about how we could better document this. The problem that we're specifically trying to solve, is being able to document sub-commands -- without making documenting the top-level options too confusing.

burl:
Aah, I see now.  I was not expecting the reset state.  Perhaps just calling that out in the explanation for .command() is good enough.

burl:
I was attempting to parse a command line for a git-style interface with the concept of "global switches" in addition to commands and command options/switches.

ie:  command [global-switches...] <subcommand> [other-switches-and-options....]

Git for instance lets one do this:

   git --no-pager log -q

But, not:

   git -q --no-pager log

because '-q' is not a valid global option.

(unfortunately, this is also valid and unambiguous to git:   git -p log -p)

So this is sort of like two contexts, divided by the position of the subcommand. The left context is "global" and the right context is "command".  Argument permutation may only happen within each context -- not across contexts, and the "global" context needs to become "strict" by default.  The option parser needs to know all the valid subcommands as well.

[...]

PR #239, opened 2015-08-27
pass argv as a second argument to command handlers

bcoe:
until we can sit down and start overhauling how the command functionality works (I think this will be what drives us to `yargs@4.x`), I think this might be a good stop-gap to make commands more useful.

If someone wants to use some top-level options in their command handlers, they can now do this by using the outer `argv` which is passed as the second argument to the command handler.

I found myself wanting this for a command line app I'm working on today.

What do you think of this @nexdrew, is it a worthwhile tweak?

### DIFF
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -38,6 +38,17 @@ describe('yargs dsl tests', function () {
     Object.keys(argv).should.include('cool')
   })
 
+  it('populates argv with placeholder keys when passed into command handler', function (done) {
+    yargs(['blerg'])
+      .option('cool', {})
+      .command('blerg', 'handle blerg things', function (yargs, argv) {
+        Object.keys(argv).should.include('cool')
+        return done()
+      })
+      .exitProcess(false) // defaults to true.
+      .argv
+  })
+
   it('accepts an object for implies', function () {
     var r = checkOutput(function () {
       return yargs(['--x=33'])
@@ -235,7 +246,12 @@ describe('yargs dsl tests', function () {
   describe('command', function () {
     it('allows a function to be associated with a command', function (done) {
       yargs(['blerg'])
-        .command('blerg', 'handle blerg things', function () {
+        .command('blerg', 'handle blerg things', function (yargs, argv) {
+          // a fresh yargs instance for perorming command-specific parsing,
+          // and an argv instance containing the parsing performed thus far
+          // should be passed to the command handler.
+          (typeof yargs.option).should.equal('function')
+          argv._[0].should.equal('blerg')
           return done()
         })
         .exitProcess(false) // defaults to true.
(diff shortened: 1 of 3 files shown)


nexdrew:
I definitely like this and even thought about doing it myself a couple times. Initiating review...

test/yargs.js
```diff
@@ -219,7 +230,12 @@ describe('yargs dsl tests', function () {
   describe('command', function () {
     it('allows a function to be associated with a command', function (done) {
       yargs(['blerg'])
-        .command('blerg', 'handle blerg things', function () {
+        .command('blerg', 'handle blerg things', function (yargs, argv) {
+          // a fresh yargs instance for perorming command-specific parsing,
```
nexdrew: s/perorming/performing/

Superficial nit, sorry!

bcoe:
@nexdrew I might hold off on merging this branch until the weekend; I'd love to sync up on IRC or slack and talk about what the command API should look like.

I'd like to keep the simple spirit of yargs, while making it a bit more intuitive for handling nested commands and inner argvs.

test/yargs.js
```diff
@@ -219,7 +230,12 @@ describe('yargs dsl tests', function () {
...
+          // and an argv instance containing the parsing performed thus far
+          // should be passed to the command handler.
+          (typeof yargs.option).should.equal('function')
```
nexdrew: Should probably assert that `yargs` has been reset somehow. Maybe something like:

```js
yargs.getCommandHandlers().should.deep.equal({}) // 'blerg' command should not exist after reset
```

test/yargs.js
```diff
@@ -219,7 +230,12 @@ describe('yargs dsl tests', function () {
   describe('command', function () {
     it('allows a function to be associated with a command', function (done) {
       yargs(['blerg'])
-        .command('blerg', 'handle blerg things', function () {
+        .command('blerg', 'handle blerg things', function (yargs, argv) {
+          // a fresh yargs instance for performing command-specific parsing,
+          // and an argv instance containing the parsing performed thus far
+          // should be passed to the command handler.
+          (typeof yargs.option).should.equal('function')
```
nexdrew: Think my last comment got lost in the ether. Here it is again:

Should probably assert that `yargs` has been reset somehow. Maybe something like:

```js
yargs.getCommandHandlers().should.deep.equal({}) // 'blerg' command should not exist after reset
```

nexdrew:
@bcoe I totally agree with your sentiments. I've been thinking about what the next version of command handling/API might look like on and off for a while now. I like the simplicity of yargs, but I think we can definitely make defining/handling subcommands and their options/context much more intuitive. I think some combination of my (naive) proposal and the ideas you and @danielvy put forth in issues #225 and #163 should give us a decent starting point. At least seems we should make it easier to define commands along with their options, handler, and subcommands.

I'd love to talk through some of this. Unfortunately, I will be traveling for my grandmother's funeral over the next 2 days (Friday and Saturday), but I should be available after that.

BTW, I think this PR looks good now. :+1:

bcoe:
closing this in favor of overhauling how commands work for yargs@4.x

PR #240, opened 2015-08-28
hotfix for os-locale on some windows platforms

bcoe:
If os-locale throws, we will now default to en when detecting locale.

CC: @sindresorhus, @kevinsawicki, @MatthewNewland

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,7 +19,7 @@ function Argv (processArgs, cwd) {
   var validation = null
   var y18n = Y18n({
     directory: path.resolve(__dirname, './locales'),
-    locale: osLocale.sync(),
+    locale: guessLocale(),
     updateFiles: false
   })
 
@@ -529,6 +529,14 @@ function Argv (processArgs, cwd) {
     setPlaceholderKeys(argv)
 
     return argv
+  }
+
+  function guessLocale () {
+    try {
+      return osLocale.sync()
+    } catch (err) {
+      return 'en'
+    }
   }
 
   function setPlaceholderKeys (argv) {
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -531,6 +531,14 @@ function Argv (processArgs, cwd) {
     return argv
   }
 
+  function guessLocale () {
+    try {
+      return osLocale.sync()
```
kevinsawicki: This is still a bit worrisome for Atom's case where app startup time is critical and this call exec's processes synchronously.

bcoe: two options I see:

1. is there an environment variable I can detect and flip this feature off specifically for Atom.
2. what if I add an option `.detectLocale(false)`?

Shall I release this in the interim to fix the exception?

sindresorhus:
Fixed in `os-locale` 1.2.1. Weird, I thought `wmic` were available everywhere.

I guess this PR still makes sense though, just in case.

index.js
kevinsawicki: > Shall I release this in the interim to fix the exception?

I think Atom will just pin to version `3.19.0` for now and cut a new release.

Just a heads up, earlier this week we moved `apm` to use `3.19.0` as well because of issues with long paths on Windows introduced by the `os-locale` addition. I know this is a Windows issues, but I just wanted to mention it in case you came across it on that repo.

> what if I add an option .detectLocale(false)?

:+1: That seems good

sindresorhus: It only spawns when neither of the following environment variables are defined:

```
env.LC_ALL || env.LANGUAGE || env.LANG || env.LC_MESSAGES
```

Might be a good idea to add an option to `os-locale` to prevent spawning `{spawn: false}`?

bcoe:
@kevinsawicki @sindresorhus I'm going to get this hot fix out the door so that we're not breaking anything on Windows, we can discuss in this thread:

https://github.com/bcoe/yargs/issues/241

PR #247, opened 2015-09-03
Upgrade standard to 5.2.x

nexdrew:
Required fixing the following nits:

- yargs/index.js:35:8: '?' should be placed at the end of the line.
- yargs/index.js:584:8: '?' should be placed at the end of the line.
- yargs/index.js:585:8: ':' should be placed at the end of the line.
- yargs/test/integration.js:41:4: Block must not be padded by blank lines.
- yargs/test/usage.js:350:6: Block must not be padded by blank lines.
- yargs/test/usage.js:366:6: Block must not be padded by blank lines.
- yargs/test/usage.js:395:6: Block must not be padded by blank lines.

These fixes would be required on next build anyway, according to current `devDependencies` semver range.

PR #248, opened 2015-09-03
Reinstate os-locale with option to prohibit spawning

nexdrew:
`os-locale@1.3.1` allows us to nuke the previous stop-gap solution. Now we can tell `os-locale` not to spawn any subprocesses, so we don't introduce unnecessary latency or, even worse, problems with installing `spawn-sync`, for downstream dependents like Atom. :sunglasses:

If the build fails due to `standard` problems, anything not handled by PR #247 will be subsequently fixed here.


nexdrew:
@bcoe When you get a chance, this PR needs review. Thanks!

bcoe:
awesome \o/ will make the time this evening.

bcoe:
Great work :sunglasses:, :shipit:

nexdrew:
Published as `yargs@3.24.0`

PR #250, opened 2015-09-03
Fix order of locale environment variable

jvdm:
During detection of the language to select translations the code is
[relying][2] on the same environment variables [libc][1]'s use, but
you are not considering them in the same order `glibc`
(and `gettext(3)`) do.

Specifically, you want to get the value of ``LC_MESSAGES`` locale
category (do not mistake here with the environment variable with the
same name).  To do that, the manual says:

    If locale is an empty string, "", each part of the locale that
    should be modified is set according to the environment variables.
    The details are implementation-dependent.  For glibc, first
    (regardless of category), the environment variable LC_ALL is
    inspected, next the environment variable with the same name as the
    category (see the table above), and finally the environment
    variable LANG.  The first existing environment variable is used.

Furthermore, ``LANGUAGE`` is a GNU extension to be used as a fallback
for all those.

Following the above, this patch fix the proper order of environment
variable selection to consider `LC_MESSAGES` before `LANG`, and put
`LANGUAGE` as the last option.

 [1]: http://manpages.ubuntu.com/manpages/hardy/man3/setlocale.3posix.html
 [2]: https://github.com/bcoe/yargs/blob/master/lib/locale.js#L9

### DIFF
--- a/lib/locale.js
+++ b/lib/locale.js
@@ -6,7 +6,7 @@ module.exports = function () {
 
 function getEnvLocale () {
   var env = process.env
-  var ret = env.LC_ALL || env.LANGUAGE || env.LANG || env.LC_MESSAGES
+  var ret = env.LC_ALL || env.LC_MESSAGES || env.LANG || env.LANGUAGE
   return getLocale(ret)
 }
 


nexdrew:
@jvdm Thanks for the info! We are in the middle of replacing `lib/locale.js` with [`os-locale`](https://github.com/sindresorhus/os-locale). Would you mind submitting the same change there instead?

jvdm:
Looking into how `gettext` [find the correct locale](https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html#The-LANGUAGE-variable)
it seems that `LANGUAGE` is used to set the preferred locale to use if
`LC_MESSAGES` category is set to something different from `C`.

My patch is not correct if you want to follow `gettext` convention, so please
verify and reject as needed.

> @jvdm Thanks for the info! We are in the middle of replacing lib/locale.js
> with os-locale. Would you mind submitting the same change there instead?

I'll take a look at it @nexdrew. Thanks.

bcoe:
@jvdm this is awesome, but let's move the discussion to os-locale since we are going to be pulling that in as a dependency again :+1:

PR #253, opened 2015-09-12
make the yargs test suite run on Windows

bcoe:
* fixed various issues with `/` vs. `\` paths.
* upgraded nyc to a version that has `spawn-wrap` and `foreground-child` that are Windows ready.


nexdrew:
:+1: LGTM

PR #254, opened 2015-09-13
Added Japanese translation

oti:


bcoe:
@oti this is awesome, thanks for the translation. The implication failed warning is caused by this:

```js
var yargs  = require('./')
  .implies('a', 'b')
  .argv
```

```shell
Benjamins-MacBook-Pro:yargs benjamincoe$ ./test.js -a

Implications failed:
  a -> b
```

I'm going to land this translation, and if needed you can provide a patch that fixes any wording if need be.

PR #258, opened 2015-09-18
Add Node v4.x to CI builds

nexdrew:
Also opted to drop Node v0.11, to keep number of builds the same (4 per platform) while still (theoretically) covering the same range (v0.10 - v4).

To make AppVeyor config consistent with Travis, also changed `3.0` to `3` to use latest iojs version (e.g. 3.3.1) instead of latest 3.0 version (which is always 3.0.0).

Note that, at least for Travis, we could use `stable` instead of `4` (to anticipate version 5+), but I decided to stick with prior convention.

Also note: I expect the build to fail (at the moment) for Node v0.10 until [issue 14 of `standard-engine`](https://github.com/Flet/standard-engine/issues/14) is fixed. Looks like we could revert our `standard` dependency back to version `5.0.0` to workaround this, but then we'd have to remember to update it once the `eslint` git dependency of `standard-engine` is fixed/removed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,6 @@ script: npm test
 after_success: npm run coverage
 env:
   - NODE_VERSION="0.10"
-  - NODE_VERSION="0.11"
   - NODE_VERSION="0.12"
   - NODE_VERSION="iojs"
+  - NODE_VERSION="4"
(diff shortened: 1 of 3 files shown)


nexdrew:
In order to kick off another build and verify the [`standard-engine` fix for Node v0.10](https://github.com/Flet/standard-engine/pull/15), I went ahead and updated dependency declarations to latest versions.

nexdrew:
Could possibly wait to merge this until upstream dependencies `os-locale` and `y18n` publish new versions.

bcoe:
Awesome work, let's wait for the upstream dependencies, might as well.

Issue #260, opened 2015-09-21
Split up required and optional args in help output


ammeep:
:wave: @bcoe  Love this project. While using this on my own project I found that I had a handful of optional arguments which produced a very busy help output.

With a small change I was able to split up the arguments which are required from those which are optional. 
<img width="685" alt="screen shot 2015-09-21 at 4 22 39 pm" src="https://cloud.githubusercontent.com/assets/1024672/10002349/af3a01f0-60fa-11e5-886c-b935ea12fb60.png">

I'll open a pull request which closes this issue out also, but I wanted to give you a heads up first.

nexdrew:
@ammeep I don't want to speak for Ben, but I do want to say thanks!

I think this is a good idea, but we might want to consider implementing [option](https://github.com/bcoe/yargs/issues/152#issuecomment-125863571) [groups](https://github.com/bcoe/yargs/issues/152#issuecomment-126380983) as a more flexible way to achieve this. Then CLI authors could organize options in usage content any way they'd like.

Another thing to consider is i18n. If option grouping could be based on any arbitrary string, then it would be nice if there was an "official" way to support different languages/locales for the consumer-defined strings (i.e. some kind of built-in support for translating "Required options:" into the OS locale/language that yargs is going to use for its own strings). CLI authors can use [y18n](https://www.npmjs.com/package/y18n) on their own, but there's currently no way to couple/sync that with the y18n instance/files that yargs uses.

So I think a more robust solution would include (a) arbitrary option grouping that (b) accounted for i18n in some way. Thoughts?

bcoe:
Hey @ammeep, sorry for the slow response it's been a crazy busy few weeks for me. I love the idea of providing more flexibility for grouping together options! I'm with @nexdrew, however, that it would be slick if we could provide a generic API for handling the grouping together of options.

@andrew and I have been discussing the work we'd like to take on for `yargs@v4.x`:

* I'd like to overhaul the command API a little bit, I think it could be made more intuitive.
* Andrew already had a few thoughts about making argument layout more flexible.
* I'd like to triage the existing tickets, there might be some other potentially breaking changes worth making if we're planning on a major bump.

Long story short, I'd love to start a conversation about some API changes that are worth making for the next major version, and I think we can provide the functionality you're looking for in the process. I sent you an invite to the yargs Slack, I'll be hanging out and hacking a bit this weekend.

bcoe:
[...]

PR #261, opened 2015-09-21
[WIP]Required vs options

ammeep:
This will close out #260 by splitting up the required vs optional output in the usage output.

screen shot 2015-09-21 at 4 22 39 pm

This change introduces two new variables ``requiredKeys`` and ``optionalKeys``. It moves useage printing into its own function, and calls it for each type of key. 

### Todo
- [x] Implement fix
- [ ] Fix unit tests
- [ ] Add any additional tests

### DIFF
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -164,7 +164,7 @@ module.exports = function (yargs, y18n) {
       return acc
     }, {})
 
-    var printOptions = function(keys){
+    var printOptions = function (keys) {
       keys.forEach(function (key) {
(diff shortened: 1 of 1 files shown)


bcoe:
@ammeep  closing this in favor of https://github.com/bcoe/yargs/issues/274, which I think is a great generic solution -- let's us support your behavior and other category names :)

I encourage you and @nexdrew to sync up on this feature, so that we make sure everyone's needs are addressed.

ammeep:
Sounds good :+1:

PR #266, opened 2015-09-29
Add more cases to comprehensive parser test.  Mainly for -- .

geophree:
Another argument parsing library had a problem with `--` after `--`, here I add extra data to test that and various other edge cases in your comprehensive parser test.

### DIFF
--- a/test/parser.js
+++ b/test/parser.js
@@ -117,7 +117,7 @@ describe('parser tests', function () {
       '-h', 'awesome', '--multi=quux',
       '--key', 'value',
       '-b', '--bool', '--no-meep', '--multi=baz',
-      '--', '--not-a-flag', 'eek'
+      '--', '--not-a-flag', '-', '-h', '-multi', '--', 'eek'
     ])
     parse.should.have.property('c', true)
     parse.should.have.property('a', true)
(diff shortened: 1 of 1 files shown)


nexdrew:
@geophree Thanks!

So you just want to expand the "short-circuit option parsing" test data to make sure we don't regress on handling `--` after `--` in the future?

Not sure this provides a whole lot of value, but I'll buy it for a dollar. LGTM :+1: Thanks!

geophree:
Yep, that's exactly it.  Your test suite is pretty comprehensive, I figured I'd push it a little closer to "completely" comprehensive.  As I mentioned, another command parser (who shall remain nameless) had a problem with this case.  I submitted a PR to fix their implementation & test as well.  Thanks for making such an awesome arg parsing library!

nexdrew:
Ahh, gotcha. Nice work!

Most of the credit for `yargs` goes to @bcoe, @substack, and @chevex. Indeed, they have built an awesome library. I'm just along for the ride. :wink:

Thanks, again, for contributing back!

bcoe:
@geophree thanks for the contribution and kind words \o/ :tada: 

There's nothing better in OSS then having someone add test cases for you.

PR #269, opened 2015-09-30
Add failing test to show that errors are outputted twice when exitPro…

tepez:
When exitProcess is false, errors are printed twice to stderr.
This is a, currently failing, spec that shows this.
Will open an issues next and update.

### DIFF
--- a/test/usage.js
+++ b/test/usage.js
@@ -285,6 +285,32 @@ describe('usage tests', function () {
     )
   })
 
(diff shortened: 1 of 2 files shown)


nexdrew:
PR #273 also includes commit [db263e4](https://github.com/tepez/yargs/commit/db263e4986ee18db0a94177d1866982ec4cfc339), so if it is merged, this PR will/should be closed as well.

PR #271, opened 2015-09-30
Don't validate arguments if help or version is used and exitProcess=false

tepez:
Currently if `exitProcess=false` and the help or version options are given the arguments will be validated anyhow.

I'd argue that this is never the desired behavior. For example, if there are required arguments then if the users enters `node . --help` he would get errors for every required arguments missing.

This PR fixes this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -511,32 +511,40 @@ function Argv (processArgs, cwd) {
       return
     }
 
+    var helpOrVersion = false
+
     Object.keys(argv).forEach(function (key) {
       if (key === helpOpt && argv[key]) {
         self.showHelp('log')
+        helpOrVersion = true
         if (exitProcess) {
           process.exit(0)
         }
       } else if (key === versionOpt && argv[key]) {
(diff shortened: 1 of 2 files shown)


nexdrew:
Awesome! LGTM at first blush. :+1:

nexdrew:
@bcoe See any problems with this PR? It looks good to me. I'd like to land this and publish a new release.

bcoe:
@nexdrew this seems great to me, if you sign off I sign off :+1:

nexdrew:
@bcoe Merged. What do you think, publish 3.27.0 now or wait until the `wrap-ansi` -> `cliui` enhancements are released?

bcoe:
@nexdrew no need to block on wrap-ansi, might take a bit longer to make all those ducks line up.

nexdrew:
@bcoe Ok, I'll go ahead and publish 3.27.0. Thanks!

Issue #270, opened 2015-09-30
Errors are printed twice if exitProcess is false


tepez:
When exitProcess is false, errors are outputted twice.
See failing spec in: https://github.com/bcoe/yargs/pull/269

An error is first outputted [here](https://github.com/bcoe/yargs/blob/master/lib/usage.js#L38).
It's then thrown a few lines below.

It will then be caught and outputted again [here](https://github.com/bcoe/yargs/blob/9a290d699650bb66dd28582ba61befecbf1a0551/index.js#L451).

I think a simple solution would be to create something like a `ParseError` type. When we catch an error of  this type up the stack - we don't print it again.

Will you accept a PR?

nexdrew:
@tepez Thanks! Looks like `.exitProcess(false)` is not very popular and could use some love. :)

Just to be clear, the help content (if `showHelpOnFail` is `true`) and error message are output via `usage.fail()`, which is called twice (if `getExitProcess()` returns falsy) - **first** in a validation method (like `customChecks()` or `requiredArguments()`) and **second** in the catch block of the `argv` getter method.

I think your proposed solution is a good one, but I'm going to investigate other options for sake of due diligence. I'll report back with any findings/ideas.

tepez:
Thanks @nexdrew. I'll wait.

There is another issue with `exitProcess(false)` that we could easily solve with a new `ParseError` type. Currently, when we catch an error that was thrown somewhere in the CLI main function, we have no way to know if it was thrown by yargs or somewhere else.

If it was thrown by yargs - it was already printed so there is no need to print it again.
If it was thrown somewhere else - we do need to print it.

nexdrew:
I do agree we should fix this problem, but note that a possible workaround would be to use your own failure function via [`.fail(fn)`](https://www.npmjs.com/package/yargs#fail-fn), which will bypass failure message output and error throwing.

tepez:
Great idea @nexdrew !
This simple function solved all duplications for me:
```coffee
yargs.fail((msg) ->
  throw new Error(msg)
)
```

nexdrew:
@tepez Great, glad I could help!

As far as potential solutions to this problem goes, here's what I have so far:

1. Add private variable to `lib/usage.js` and use it within the `fail()` method to only output help content and failure message once. (This is rather trivial to implement and does not affect any current unit tests, but I'm not sure if it has other side effects.)
[...]

PR #273, opened 2015-10-01
Output failure message only once

nexdrew:
Fixes issue #270 (concerning when `.exitProcess(false)` is used) by implementing the simplest solution mentioned [in comment](https://github.com/bcoe/yargs/issues/270#issuecomment-144551525) there.

Includes and satisfies test specs from PR #269.

### DIFF
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -28,17 +28,22 @@ module.exports = function (yargs, y18n) {
     return self
   }
 
+  var outputFailure = true
   self.fail = function (msg) {
     if (fails.length) {
       fails.forEach(function (f) {
         f(msg)
       })
     } else {
-      if (showHelpOnFail) yargs.showHelp('error')
-      if (msg) console.error(msg)
-      if (failMessage) {
-        if (msg) console.error('')
-        console.error(failMessage)
+      // don't output failure message more than once
+      if (outputFailure) {
+        outputFailure = false
+        if (showHelpOnFail) yargs.showHelp('error')
+        if (msg) console.error(msg)
+        if (failMessage) {
+          if (msg) console.error('')
+          console.error(failMessage)
+        }
       }
       if (yargs.getExitProcess()) {
         process.exit(1)
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -28,7 +28,7 @@ module.exports = function (yargs, y18n) {
     return self
   }
 
-  var outputFailure = true
+  var failureOutput = false
   self.fail = function (msg) {
     if (fails.length) {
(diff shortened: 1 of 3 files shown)


nexdrew:
@bcoe Got time for a quick review? :grin: For this and [PR 271](https://github.com/bcoe/yargs/pull/271).

bcoe:
@nexdrew yes, consider the code review enqueued -- sorry that I've been so AWOL lately., can't wait to put some time into OSS.

nexdrew:
@bcoe No worries. :palm_tree:

lib/usage.js
```diff
@@ -28,17 +28,22 @@ module.exports = function (yargs, y18n) {
     return self
   }
 
+  var outputFailure = true
```
bcoe: I'd be tempted to start the variable in a false state, just a personal preference when I'm using a variable as a flag. perhaps: `failureOutput` = `false`, and set it to `true` once you spit out an error -- just a minor nit feel free to ignore.

test/yargs.js
```diff
@@ -147,18 +147,34 @@ describe('yargs dsl tests', function () {
...
+              .demand('cat')
+              .showHelpOnFail(false)
+              .exitProcess(false)
+              .argv
+          } catch (err) {
+            // ignore the error, we only test the output here
+          }
+        })
+        expect(r.logs).to.deep.equal([])
+        expect(r.errors).to.deep.equal(['Missing required argument: cat'])
```
bcoe: these tests look great, did you confirm that they fail before adding your switch logic?

bcoe:
this is great :shipit:

lib/usage.js
nexdrew: I thought of that, but then I noticed `showHelpOnFail` above defaults to `true`, so I figured it depended on the use-case. I also thought `if (outputFailure)` was easier to read/follow than `if (!something) `. But I'm happy to change it if you'd like.

test/yargs.js
nexdrew: Yes. These tests came from @tepez (thanks!) and actually [failed in Travis](https://travis-ci.org/bcoe/yargs/builds/82895740) with PR #269 before my fix, so you don't have to take my word for it. :smile:

lib/usage.js
nexdrew: Changed with [commit e61d536](https://github.com/bcoe/yargs/commit/e61d53637299dff8a0667b5aea813c564aa3e13e).

nexdrew:
I will merge once the build passes with the last commit.

PR #277, opened 2015-10-13
upgrade to version of cliui that handles ansi-colors

bcoe:
I've updated the layout logic in cliui to take into account the ansi escape codes:

https://github.com/bcoe/cliui/pull/11
https://github.com/bcoe/cliui/pull/12

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
   },
   "devDependencies": {
     "chai": "^3.3.0",
+    "chalk": "^1.1.1",
     "coveralls": "^2.11.4",
     "hashish": "0.0.4",
     "mocha": "^2.3.3",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   ],
   "dependencies": {
     "camelcase": "^1.2.1",
-    "cliui": "^3.0.0",
+    "cliui": "^3.0.2",
     "decamelize": "^1.0.0",
     "os-locale": "^1.4.0",
     "window-size": "^0.1.2",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   ],
   "dependencies": {
     "camelcase": "^1.2.1",
-    "cliui": "^2.1.0",
+    "cliui": "^3.0.0",
     "decamelize": "^1.0.0",
     "os-locale": "^1.4.0",
     "window-size": "^0.1.2",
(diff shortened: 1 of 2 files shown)


nexdrew:
Found a small problem concerning option alignment when ANSI colors are used in a description for an option. Commit 9963cbd adds 2 tests to demonstrate this with a minimal example:

1. Alignment without ANSI colors
2. Alignment with ANSI colors

As is, the 1st test passes but the 2nd test fails. Not sure what the root problem is, will need to investigate later.

nexdrew:
FYI the test comes from an example I put together in [issue 251](https://github.com/bcoe/yargs/issues/251#issuecomment-138958452). Here's what the problem looks like:

screen shot 2015-10-13 at 12 54 14 am

phated:
This is my before and after runs from gulp:

### Before
screen shot 2015-10-13 at 3 17 23 pm

### After
screen shot 2015-10-13 at 3 16 56 pm

As you can see, the wrapping colors now work correctly.  I'm not sure what the result of the types column should look like though.

Thanks for all the work on this @bcoe!

bcoe:
@phated, @nexdrew sorry for being slow to update it's been a busy few weeks (I also apologize for how much of a bikeshed :bike: :house: this is at this point, want to make sure we solve this issue for good).

I've made some changes to how widths are calculated for ansi escape codes, @phated your right-hand-column should align `[boolean]`, `[count]`, etc, appropriately now, @nexdrew could you confirm that I've fixed the issues you were seeing?

phated:
Looks good!

Screenshot of latest changes:
screen shot 2015-10-16 at 10 53 42 am

Thanks @bcoe for all the work you have been putting into this!

nexdrew:
`yargs@3.28.0` introduces some slight problems for yargonaut (when combining colors with figlet fonts), but that's yargonaut's problem. Everything else looks good. :+1:

bcoe:
@nexdrew your sure this is a yargonaut bug, and not something we can fix in cliui or yargs (while we're tearing these libraries apart any ways).

nexdrew:
@bcoe Well, I think cliui might be unnecessarily trimming leading whitespace when ANSI colors are used. See [yargonaut issue 7](https://github.com/nexdrew/yargonaut/issues/7) for example.

Was planning to look into it and offer a fix, if necessary. Unfortunately I am AFK at the moment and won't have a chance to look til later tonight or tomorrow.

PR #280, opened 2015-10-18
Find the config file in the nearest parent folder

cbas:
Travels up the folder hierarchy until a configuration file is found.
Same behaviour as npm, babel, jscs, jshint, etc.

### DIFF
--- a/index.js
+++ b/index.js
@@ -457,7 +457,7 @@ function Argv (processArgs, cwd) {
   })
 
   function parseArgs (args) {
-    var parsed = Parser(args, options, y18n)
+    var parsed = Parser(args, options, y18n, cwd)
     var argv = parsed.argv
     var aliases = parsed.aliases
 
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -2,12 +2,13 @@
 // from minimist: https://www.npmjs.com/package/minimist
 var camelCase = require('camelcase')
 var path = require('path')
+var nearestSync = require('nearest-file-path').nearestSync
 
 function increment (orig) {
   return orig !== undefined ? orig + 1 : 0
 }
 
-module.exports = function (args, opts, y18n) {
+module.exports = function (args, opts, y18n, cwd) {
   if (!opts) opts = {}
 
   var __ = y18n.__
@@ -325,7 +326,8 @@ module.exports = function (args, opts, y18n) {
       var configPath = argv[configKey] || configLookup[configKey]
       if (configPath) {
         try {
-          var config = require(path.resolve(process.cwd(), configPath))
+          var nearestPath = nearestSync(configPath, cwd)
+          var config = require(nearestPath)
 
           Object.keys(config).forEach(function (key) {
             // setting arguments via CLI takes precedence over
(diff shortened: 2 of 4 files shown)


cbas:
Hmm what gives with AppVeyor? Runs fine on Travis. Is this a known issue?

cbas:
Q: Should this behaviour be opt-in to remain backwards compatible? Could add an argument like `.config(key, [description], [bubble])` for example where `bubble` is a Boolean and defaults to false.

bcoe:
The AppVeyor tests are definitely passing, so this appears to be surfacing a legitimate difference between the tests on Linux vs. the tests on Windows.

I appreciate the pull request, but I'm not quite sure that I'd like to add this behavior. By convention, it's common to have a `.foorc` crawl the directory structure until it hits a matching file -- for a JSON configuration file I think this behavior is not expected.

When I'm deploying to production I might have a `config.json` in the project directory, it would be unexpected behavior to have the application pluck a `config.json` from a few levels up the directory hierarchy. Does anyone else want to chime in on this, @nexdrew? just my gut feeling.

nexdrew:
I've been giving this some thought, and I'm not quite sure. It might make sense if you want to *always* use a config file, and specify a default file name to look for, so that CLI users don't have to specify a config option at all (unless overriding the default), but I'm not sure it makes sense when a config argument is explicitly given by the CLI user.

I tend to agree with @bcoe on this but could be persuaded otherwise. If we *do* add the functionality, I would prefer it to be opt-in instead of enabled by default.

@cbas Could you possibly run through a couple use-cases for us?

cbas:
@bcoe @nexdrew I appreciate the consideration and feedback, thank you.

1. I will need to take a look at the Windows issue. I thought it was odd to be failing in a test following the one I added, as that should be unrelated. Maybe the tweak to `cwd` somehow broke things (just a guess). :worried:

1. The use case is based on an observation that several popular tools currently use JSON-formatted config files combined with the directory-traversing lookup behaviour. The files are still called `rc`-files but contain, in fact, JSON. Examples of tools with directory traversing lookup behaviour would be:
  - `npm` with `package.json`
    e.g. `npm test` works the same from any child directory
  - `babel` with `.babelrc` (actually JSON)
  - `jscs` with `.jscs` (actually JSON)
  - `jschint` with `.jshintrc` (actually JSON)

nexdrew:
Don't mean to be blunt, but why not use something like [rc](https://www.npmjs.com/package/rc) instead? Just convenience?

cbas:
:flushed: TIL :sob:

nexdrew:
@cbas No worries! It's not a bad idea, but I would rather offer a more robust solution, and I think other modules might fit that bill better than yargs. Just my 2¢. :wink:

cbas:
Thanks again for the feedback. Building this was a nice learning project. It gave me a much better understanding of yargs. Hope to contribute something suitable in the future! :+1:

bcoe:
@cbas was about to say, I hope this doesn't discourage you from contributing \o/ One thing worth discussing in the future, is does `yargs@4.x` itself just use `rc`?

nexdrew:
Definitely worth consideration, I think. Props to @cbas for bringing this up!

PR #281, opened 2015-10-18
fix dot notation parsing issues brought up in #279

### LINKED ISSUE #279: Array instead of plain value when default value and alias provided
Consider the following example:

``` javascript
var argv = require('yargs')
    .option('foo.bar', {
        default: 100,
        alias: 'f'
    })
    .argv;

console.warn(argv.foo.bar);
```

```bash
$ node app.js --foo.bar 200
[ 200, 100 ]
```
For me expected result in this case would be that `argv.foo.bar` is equal to `200`

bcoe:
fixes #279 addresses several parsing issues brought up by @mir3z in #279:

* an alias referencing an inner key was not respected (weird things happened like the inner object being turned into an array).
* an outer `foo.bar` key was being set by the `setPlaceholderKeys` method.
* an exception could be created by setting an object key equal to a non-object value.

I'd love an extra set of eyes on these changes, since it's always a bit frightening to muck with the guts of the yargs parser: CC: @mir3z, @nexdrew

### DIFF
--- a/index.js
+++ b/index.js
@@ -565,6 +565,9 @@ function Argv (processArgs, cwd) {
 
   function setPlaceholderKeys (argv) {
     Object.keys(options.key).forEach(function (key) {
+      // don't set placeholder keys for dot
+      // notation options 'foo.bar'.
+      if (key.match(/.*\..*/)) return
       if (typeof argv[key] === 'undefined') argv[key] = undefined
     })
   }
--- a/index.js
+++ b/index.js
@@ -567,7 +567,7 @@ function Argv (processArgs, cwd) {
     Object.keys(options.key).forEach(function (key) {
       // don't set placeholder keys for dot
       // notation options 'foo.bar'.
-      if (key.match(/.*\..*/)) return
+      if (~key.indexOf('.')) return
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -565,6 +565,9 @@ function Argv (processArgs, cwd) {
 
   function setPlaceholderKeys (argv) {
     Object.keys(options.key).forEach(function (key) {
+      // don't set placeholder keys for dot
+      // notation options 'foo.bar'.
+      if (key.match(/.*\..*/)) return
```
nexdrew: Showing my ignorance, but is it necessary to use RegExp here? Could we not get away with `indexOf('.')` instead? Not sure [which](http://jsperf.com/test-vs-indexof-fast) [performs](http://jsperf.com/regexp-vs-indexof) [better](http://jsperf.com/indexof-vs-looping-charat), but I'd prefer the simplest possible solution.

bcoe: I think you're right that we could probably simply opt for:

```js
~key.indexOf('.')
```

I was thinking about cases like `.foo` or `foo.`, but those are strange option names.

PR #282, opened 2015-10-19
completion: support promises

### LINKED ISSUE #278: completion promise support
I think it would be great if the `completion` function could accept an promise as well. As it is now, if you want to do something async you need to add an extra `done` parameter.

It could be implemented like this:

```javascript
const ret = completer(...)

if (typeof ret.then === 'function') {
  ret.then(function (value) {
    // use value
  })
} else {
  // use ret
}
```

It would allow it to be used like this with babel:

```javascript
yargs
  .completion('completion', async function (current, argv) {
    const listPromise = findListCompletions(current)
    const installPromise = findInstallCompletions(current)

    const [list, install] = await [listPromise, installPromise]

    return [...list, ...install]
  })
```

Or with plain promises:

```javascript
yargs
  .completion('completion', function (current, argv) {
    const listPromise = findListCompletions(current)
    const installPromise = findInstallCompletions(current)

    return Promise.all([listPromise, installPromise]).then(function (res) {
      return res[0].concat(res[1])
    })
  })
```

LinusU:
This fixes #278 

I had to rework the helper to support async tests since promises defer things to the next tick, and since I have to use `nextTick` as to not swallow any fatal errors and just never report them.

A promise polyfill is included as a **dev-**dependency and is used during the tests.


bcoe:
@LinusU this looks solid, I'll look a bit closer tonight when I'm not at work.

test/helpers/utils.js
```diff
@@ -20,19 +21,43 @@ exports.checkOutput = function (f, argv) {
...
+
+  if (typeof cb === 'function') {
+    process.exit = function () {
+      exit = true
+      cb(null, done())
+    }
+    process.emit = function (ev, value) {
+      if (ev === 'uncaughtException') {
+        done()
+        cb(value)
```
nexdrew: Anything wrong with changing this to `cb(value, done())` instead?

lib/completion.js
```diff
@@ -19,8 +19,19 @@ module.exports = function (yargs, usage) {
     // to completion().
     if (completionFunction) {
       if (completionFunction.length < 3) {
+        var result = completionFunction(current, argv)
+
+        // promise based completion function.
+        if (typeof result.then === 'function') {
+          return result.then(function (list) {
+            process.nextTick(function () { done(list) })
```
nexdrew: I'm perfectly fine with this, but (just out of curiosity) is the use of `process.nextTick()` necessary here? Does it help guard against `done(list)` throwing an error?

LinusU: Yes, it makes sure that if `done(list)` would throw an error, it wouldn't get swallowed by the promise and never reported.

test/helpers/utils.js
LinusU: Not really, didn't want it to also return the result thought. I don't think it's best practice to include both error and result, and with the move towards promises it won't be possible anymore.

nexdrew:
LGTM :+1: Thanks, @LinusU!

LinusU:
No problems, happy to help!

bcoe:
@LinusU your work is published to yargs' next release candidate, give it a shot:

`npm install yargs@next`

Assuming we don't run into any issues, I'll promote the release tomorrow.

LinusU:
Super! Thanks \o/

Issue #274, opened 2015-10-04
Support grouping of options in help content


nexdrew:
This is a feature requested by at least a few different folks. See some discussion and ideas in [issue 152 here](https://github.com/bcoe/yargs/issues/152#issuecomment-125863571) and in [issue 260](https://github.com/bcoe/yargs/issues/260).

Personally, I like the API methods that accept the group name/key first, e.g. `.group(groupName, options)`, because I think it's cleaner and more intuitive, but note that this would be a deviation from most other API methods that accept the option key first. I think the deviation is justifiable though. Any options defined without a group would default to the existing yargs-defined pseudo group `'Options:'`, for backwards compatibility.

Another thing to consider is i18n support for consumer-defined strings, which should probably be another separate enhancement, but I think it would be nice to introduce the concept here. Perhaps the group name could be treated as a y18n key that yargs automatically resolves from consumer-defined locale files - similar to yargs-defined keys but resolved from (a `locales` directory in) the parent module. That way, consumers could get the benefit of yargs locale detection and y18n lookup without any additional API/code changes (besides adding their own locale files). This idea could be extended to option and command descriptions as well. Something to think about.

Cellule:
+1
I have been using my own set of groups for sub commands to share common options and how to use them. 
It works pretty well, but the help message is getting out of hand.

bcoe:
@nexdrw I like where you're going with this a lot. A couple comments:

* I think the default behavior should be an API identical to the existing one (you already suggest this). the great thing is that we get the additional grouping functionality, without any breaking changes -- groups such as `Options` simply alias the new named group functionality _love it!_.
* I like your idea that we'd be clever about sourcing keys from `i18n`; going forward, if certain group names become common, we could add these translations to the translation files we ship.

bcoe:
@timmckenzie, @ammeep, @nexdrew yargs now has the new `.group()` command which allows you to split your arguments across multiple categories give it a spin:

`npm i yargs@3.30.0`

Issue #152, opened 2015-04-22
API to set the order of help menu items


megawac:
Hi I was wondering if you would consider adding an API to set the order of cli help options. This is useful when 2 commands are related to one another (such as an implication)

Another feature that would branch from this would be adding new line separators or groups of commands.

I can provide some examples later if you'd like

timmckenzie:
My application has about 75 command line arguments.  It would be great to be able group them.

bcoe:
@timmckenzie @megawac I like the idea of grouping CLI arguments better, but I'd like to avoid adding too much more logic for composing the parser.

Would the combination of the following make sense:

1. start alphabetically sorting command line arguments.
2. provide a better API for splitting complex CLI apps into multiple sub-commands, see this pull request:
  * https://github.com/bcoe/yargs/pull/157

mnquintana:
@bcoe Any plans to support setting the order of commands? I'd like to be able to [move Atom from optimist to yargs](https://github.com/atom/apm/pull/346#issuecomment-97490887), but we don't really want to change the order of options in the process.

bcoe:
@mnquintana @kevinsawicki, I would love to get apm using yargs. yargs fixes various bugs with optimists' layout engine and parser, and I think is a worthwhile upgrade.

Is argument ordering the main blocker, here are my thoughts:

* I don't like the idea of an API for ordering, I think it complicates things more than they need to be.
* what if arguments were guaranteed to maintain the same order as they are set in, so:

```js
yargs
 .option('foo', {})
 .option('bar', {})
 .option('snuh', {})
```

would output,

```shell
$0
--foo     blah
--bar     blah
--snuh   blah
```

I figure special arguments like `version`, and `help`, would always be appended to the end of the list perhaps?

* as an alternative option, should we just display arguments in alphabetical order?

kevinsawicki:
> Is argument ordering the main blocker

The main blocker to me was the observed layout differences in https://github.com/atom/apm/pull/346#issuecomment-97490887

nexdrew:
Not sure if ordering is still a consideration, but yargs now preserves declaration/insertion order, which is really nice and should satisfy most use-cases. :white_check_mark:

However, grouping has not been addressed yet and would be a really nice feature. Perhaps add a `group` method and `option()` key that maps arguments to a display string?

e.g.

```js
var argv = require('yargs')
  .demand('air').group('air', 'Must have:')
  .demand('food').group('food', 'Must have:')
[...]

PR #289, opened 2015-11-07
allow options to be placed in custom groups

bcoe:
Allow options to be placed in groups under headings other than the default `Options:` heading, like so:

```js
var yargs = require('yargs')(['--help'])
  .help('help')
  .group('batman', 'Heroes:')
  .describe('batman', "world's greatest detective")
  .argv
```

***
    Options:

    Heroes:
      --batman  world's greatest detective

# Some Notes

This is a first pass at tackling #274, which proposes a solution that should help address several aggregate issues: #152, #260, #261 

I was originally trying out @nexdrew's suggestion in #274 of tacking parameters in the order `group, options` as I started writing tests, this felt too weird to me, as it deviated from other options -- it actually made me think we should clean up any other options that deviate in `yargs@4.x`, specifically `.version()`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -315,7 +315,7 @@ function Argv (processArgs, cwd) {
     return options
   }
 
-  var groups = {'Options:': []}
+  var groups = {}
   self.group = function (opts, groupName) {
     groups[groupName] = (groups[groupName] || []).concat(opts)
     return self
--- a/index.js
+++ b/index.js
@@ -317,9 +317,7 @@ function Argv (processArgs, cwd) {
 
   var groups = {'Options:': []}
   self.group = function (opts, groupName) {
-    if (!Array.isArray(opts)) opts = [opts]
-    if (!groups[groupName]) groups[groupName] = []
-    Array.prototype.push.apply(groups[groupName], opts)
+    groups[groupName] = (groups[groupName] || []).concat(opts)
     return self
   }
   self.getGroups = function () {
--- a/index.js
+++ b/index.js
@@ -282,6 +282,8 @@ function Argv (processArgs, cwd) {
         self.nargs(key, opt.nargs)
       } if ('choices' in opt) {
         self.choices(key, opt.choices)
+      } if ('group' in opt) {
+        self.group(key, opt.group)
       } if (opt.boolean || opt.type === 'boolean') {
         self.boolean(key)
         if (opt.alias) self.boolean(opt.alias)
@@ -311,6 +313,17 @@ function Argv (processArgs, cwd) {
   }
   self.getOptions = function () {
     return options
+  }
+
+  var groups = {'Options:': []}
+  self.group = function (opts, groupName) {
+    if (!Array.isArray(opts)) opts = [opts]
+    if (!groups[groupName]) groups[groupName] = []
+    Array.prototype.push.apply(groups[groupName], opts)
+    return self
+  }
+  self.getGroups = function () {
+    return groups
   }
 
   self.wrap = function (cols) {
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -101,6 +101,7 @@ module.exports = function (yargs, y18n) {
     return deferY18nLookupPrefix + str
   }
 
+  var defaultGroup = 'Options:'
   self.help = function () {
     normalizeAliases()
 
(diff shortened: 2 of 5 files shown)


bcoe:
Reviewers: @ammeep, @nexdrew

index.js
```diff
@@ -313,6 +315,17 @@ function Argv (processArgs, cwd) {
     return options
   }
 
+  var groups = {'Options:': []}
+  self.group = function (opts, groupName) {
+    if (!Array.isArray(opts)) opts = [opts]
+    if (!groups[groupName]) groups[groupName] = []
+    Array.prototype.push.apply(groups[groupName], opts)
```
nexdrew: Not that it helps readability or efficiency any, but these 3 lines could be reduced to:

```js
groups[groupName] = (groups[groupName] || []).concat([].concat(opts))
```

Feel free to ignore this. :grinning:

index.js
```diff
@@ -313,6 +315,17 @@ function Argv (processArgs, cwd) {
     return options
   }
 
+  var groups = {'Options:': []}
```
nexdrew: Any worry about duplicating/defining the `'Options:'` string in both index.js and lib/usage.js?

lib/usage.js
```diff
@@ -151,20 +153,38 @@ module.exports = function (yargs, y18n) {
...
 
-      return acc
-    }, {})
+      // actually generate the switches string --foo, -f, --bar.
+      var switches = normalizedKeys.reduce(function (acc, key) {
+        acc[key] = [ key ].concat(options.alias[key] || [])
+        .map(function (sw) {
+          return (sw.length > 1 ? '--' : '-') + sw
+        })
+        .join(', ')
```
nexdrew: `standard` doesn't seem to care, but can we indent these 4 lines (since they continue the value end of the `acc[key]` assignment)?

nexdrew:
Looks like the same option can possibly end up in multiple groups, like so:

```js
var argv = require('../yargs')
  .group('batman', 'Heroes:')
  .group('batman', 'Villains:')
  .help('help').wrap(null)
  .argv
```

```
Options:
  --help  Show help  [boolean]

Heroes:
  --batman

Villains:
  --batman
```

Is this intentional? I mean, some might think it a bug while others might think it a feature. :smile_cat:

nexdrew:
Also, it looks like custom groups are displayed in the order of declaration (which is good), but there doesn't seem to be a way to change where the default `'Options:'` group is displayed (it's always displayed *before* custom groups).

Could we either (a) allow some explicit way to define the desired order of groups, or (2) just display the default group last instead of first?

nexdrew:
For the (b) option above - displaying default group last instead of first - see [commit dc350f2](https://github.com/nexdrew/yargs/commit/dc350f2bfb17de2aaa49c9313d57296295a235af) against the same branch on my fork.

index.js
nexdrew: I also addressed this in [commit dc350f2](https://github.com/nexdrew/yargs/commit/dc350f2bfb17de2aaa49c9313d57296295a235af).

nexdrew:
For an ugly fix of cleaning duplicated keys from multiple groups, see [commit 5d7eaa8](https://github.com/nexdrew/yargs/commit/5d7eaa83a46fcce30486fc2296a0a4e896419a08) from the branch on my fork.

index.js
bcoe: I like this change; for whatever reason I was on a `prototype.apply` binge on the :airplane: ... probably tired.

bcoe: https://jsperf.com/array-prototype-push-apply-vs-concat/13 note that using `apply` is significantly slower, despite the extra assignment used by `concat` interesting.

index.js
bcoe: I like your solution in https://github.com/nexdrew/yargs/commit/dc350f2bfb17de2aaa49c9313d57296295a235af, will cherry pick over.

bcoe:
> Looks like the same option can possibly end up in multiple groups, like so:

@nexdrew this had crossed my mind, and what I came to was that it's a reasonable feature -- I can imagine a world where someone is using multiple headers to group arguments, and an argument might reasonably fall into two groups -- what do you think? I don't feel super strongly.

Other than this, I think I've addressed your other concerns so far -- how are things working for you?

nexdrew:
I'm ok with leaving the multiple groups functionality as it is.

Only a couple minor tweaks left that I'd like to make, I'll go ahead and add another commit to this branch. After that, feel free to review and merge. :+1:

nexdrew:
I'm done. I vote :shipit:. Nice work!

bcoe:
@ammeep, @Cellule, @timmckenzie, @megawac, this pull introduces grouping functionality to yargs. 

Mind giving it a spin and letting me know if it would solve the problems you brought up in #152, #274, and #260?

nexdrew:
:+1: Awesome!

PR #285, opened 2015-10-27
Added the locale pt-BR (Portuguese Brazil).

JulioGold:
Just added the locale pt-BR (Portuguese Brazil).

Obrigado

PR #290, opened 2015-11-07
rename pt-BR -> pt_BR per y18n and os-locale expectations

nexdrew:
I obviously missed this during review of #285, but region-specific locale files should contain an underscore b/w lang and region instead of a dash, to match the POSIX values returned from `os-locale` and `lcid`.

So add a test, rename the file, and update README. :v:


bcoe:
:shipit:

nexdrew:
Rats. I guess the "fetch PR branch, rebase with amended commit message, and merge --ff-only to master" didn't allow GitHub to see this PR as merged (since the original commit was amended, I guess).

Looks like the best merge options are to either (a) cherry-pick with "Closes #ID PR: " commit message or (b) just use GitHub's merge button.

Anyway, closing this since the changes were merged with commit 80a753c.

PR #293, opened 2015-11-08
Polish locale translation

kamilogorek:


bcoe:
@mmalecki perhaps you could take a quick look at this too?

mmalecki:
Sorry @bcoe, I meant to check all those translations in context (an example application). All this sounds good on its own tho!

PR #297, opened 2015-11-14
Fix cjk width problem on usage table

disjukr:
it solves https://github.com/bcoe/cliui/issues/14


bcoe:
@disjukr great work \o/ I've restarted the build on appveyor, looks like a problem with appveyor not your pull.

Issue #305, opened 2015-11-30
Read option values from env vars


nexdrew:
In the spirit of [the 3rd factor of a 12 factor app](http://12factor.net/config) and to better support services that traditionally pass env var values as CLI options anyway, let's consider supporting the ability to read env vars for values of configured options (and cut out the middleman which usually takes form in a shell script).

## Potential Approaches

1. The Automagical Approach

    With the flip of a configuration switch, `yargs` could automatically take option names, convert them to their env var equivalent, and use those env vars that are defined as values for options that have not been specified on the command line.

    For example, an option like `--foo-bar` could be translated to env vars `FOO_BAR` or `FOOBAR`, with the following results:

    ```
    $ FOO_BAR=x program --foo-bar y
    { _: [], 'foo-bar': 'y', fooBar: 'y', '$0': program' }
    ```

    ```
    $ FOO_BAR=x program
    { _: [], 'foo-bar': 'x', fooBar: 'x', '$0': program' }
    ```

    ```
    $ program
    { _: [], '$0': program', 'foo-bar': undefined }
    ```

    The API could be either a single configuration switch that enables env var lookups for all options (e.g. `yargs.envVars(true)` or could be configured per option (e.g. `yargs.envVar('foo-bar', true)`) or could possibly be enabled for only required options or only non-required options.

    To avoid name clashes with common env vars like `PATH` or `HOME`, the API should support a configurable prefix, e.g. `yargs.envVars('PREFIX_')`.

2. The Explicit Approach

    `yargs` could simply allow an equivalent env var to be explicitly specified per option. If the option is not given on the command line, then `yargs` would then look for the configured env var and use its value if defined.

    For example, an option `--foo-bar` could be explicitly mapped to a `FOO_BAR` env var (or even `SOMETHING_TOTALLY_DIFFERENT`) and the results from above would be the same.

    The API could be as simple as `yargs.envVar('foo-bar', 'FOO_BAR')` or `yargs.option('foo-bar', { envVar: 'FOO_BAR' })`.

3. Some Hybrid of 1 and 2

    The Automagic Approach is much less verbose, making it easy for consumers to build support for env vars without many code changes and without manually specifying a bunch of env var names that will probably be very similar to option names anyway. However, it has the problem of potential name conflicts with env vars not meant for the `yargs` program, requiring a prefix solution or similar. It also has the problem of not allowing overrides on a per-option use-case.

[...]

PR #308, opened 2015-12-02
Apply env vars to argv via prefix-based API

nexdrew:
First stab at env var support, implementing [@bcoe's suggestions](https://github.com/bcoe/yargs/issues/305#issuecomment-160851479) as discussed in #305.

This will allow env vars matching a given prefix to be parsed and applied to `argv` via simple configuration method `.env('PREFIX')`. Prefix can be omitted to apply _all_ env vars to `argv`.

Aliases and type configuration are supported when args are defined by env vars.

New order of precedence for program argument values:

1. CLI flags
2. Config file (if configured)
3. Env var (if configured)
4. Default values (if configured)

There is also a special case to support specification of a config file via env var.

For more details, consult the additional README content or the unit tests.

Things not addressed by this PR:

- Indication of env var support in usage/help text
- Ability to explicitly map a configured option to an arbitrary env var name
- Support for displaying env var value over configured default value in usage/help text

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -220,7 +220,14 @@ module.exports = function (args, opts, y18n) {
     }
   }
 
+  // order of precedence:
+  // 1. command line arg
+  // 2. value from config file
+  // 3. value from env var
+  // 4. configured default value
+  applyEnvVars(opts, argv, true) // special case: check env vars that point to config file
   setConfig(argv)
+  applyEnvVars(opts, argv, false)
   applyDefaultsAndAliases(argv, aliases, defaults)
 
   Object.keys(flags.counts).forEach(function (key) {
@@ -346,6 +353,20 @@ module.exports = function (args, opts, y18n) {
           })
         } catch (ex) {
           if (argv[configKey]) error = Error(__('Invalid JSON config file: %s', configPath))
+        }
+      }
+    })
+  }
+
+  function applyEnvVars (opts, argv, configOnly) {
+    if (typeof opts.envPrefix === 'undefined') return
+
+    var prefix = typeof opts.envPrefix === 'string' ? opts.envPrefix : ''
+    Object.keys(process.env).forEach(function (envVar) {
+      if (prefix === '' || envVar.lastIndexOf(prefix, 0) === 0) {
+        var key = camelCase(envVar.substring(prefix.length))
+        if (((configOnly && flags.configs[key]) || !configOnly) && (!(key in argv) || flags.defaulted[key])) {
+          setArg(key, process.env[envVar])
         }
       }
     })
--- a/README.md
+++ b/README.md
@@ -614,6 +614,63 @@ Optionally `.describe()` can take an object that maps keys to descriptions.
 
 Should yargs attempt to detect the os' locale? Defaults to `true`.
 
+.env([prefix])
+--------------
+
+Tell yargs to parse environment variables matching the given prefix and apply
+them to argv as though they were command line arguments.
+
+If this method is called with no argument or with an empty string or with `true`,
+then all env vars will be applied to argv.
+
+Program arguments are defined in this order of precedence:
+
+1. Command line args
+2. Config file
+3. Env var
+4. Configured defaults
+
+```js
+var argv = require('yargs')
+  .env('MY_PROGRAM')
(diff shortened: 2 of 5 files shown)


lib/parser.js
```diff
@@ -220,7 +220,14 @@ module.exports = function (args, opts, y18n) {
     }
   }
 
+  // order of precedence:
+  // 1. command line arg
```
bcoe: I think I'd be tempted to have the arguments passed as CLI arguments to take precedence over environment variables.

bcoe:
This is looking great @nexdrew, wonderful unit tests :100: 

My only comment is I think my expectation would be that, if I had:

`YARGS_FOO=banana ./bin/app.js --foo=apple`

I'd expect the value of `foo` to be `apple`.

lib/parser.js
nexdrew: They do. CLI args first, then cfg file, then env var, then default. So CLI args trump all. :sunglasses:

bcoe: aha! I was reading the list in the wrong order ... all is well. this rocks.

nexdrew:
Totally agree, that's the way it works. See the `should prefer command line value over env var` parser test.

bcoe:
:shipit:

zeke:
:ok_hand:

bcoe:
@zeke let us know if you try out this feature, I haven't played with it yet in production.

jaridmargolin:
Stumbled on this thread while seeing what I considered unexpected behavior in yargs. Curious on the decision here:

```
New order of precedence for program argument values:

CLI flags
Config file (if configured)
Env var (if configured)
Default values (if configured)
```

Wouldn't one expect env vars to to take precedence over config files? See:

https://github.com/spf13/viper#why-viper
https://github.com/lorenwest/node-config/wiki/Environment-Variables#custom-environment-variables

nexdrew:
@jaridmargolin That's an interesting point, I can certainly understand the argument, especially along the lines of 12 factor apps.

My original rationale was based on the assumption that a config file must be specified on the command line as an arg, making it essentially an extension of specifying multiple args at once. Also, I imagine most uses of config files would be to support project-specific files like `.babelrc` or `.eslintrc`, which should typically override global env vars.

That being said, I can understand when it would make sense to override a default config file (not given on command line) with env vars, like test db credentials/settings.

Perhaps we could make the precedence configurable?

bcoe:
@nexdrew @jaridmargolin another argument for config file taking precedence over environment, is this:

```yaml
FOO=${MY_ENVIRONMENT_VARIABLE}
BAR=banana
```

If we eventually allow for config files, like the `.npmrc` format, these files could defer to an environment variable; the reverse is harder.

jaridmargolin:
@nexdrew @bcoe thanks for the quick responses.

> Also, I imagine most uses of config files would be to support project-specific files like .babelrc or .eslintrc, which should typically override global env vars.

While I understand the use cases described above, I think the argument could just as easily be in reverse. Imagine I am working with a testing framework (this may or may not be my exact use case). The config is setup to launch chrome by default. I however notice firefox specific bugs in my project. If I set an environment variable to specify that I would like use firefox for my current session, I would expect it to override all values excluding command line arguments.

> My original rationale was based on the assumption that a config file must be specified on the command line as an arg, making it essentially an extension of specifying multiple args at once.

Again, I understood the rationale here, but I also believe consistency is key. When utilizing a command line tool, I think their should/is a standard and understood precedence of how arguments are handled. And I think as any of us develop tools, we do so in the hopes of creating intuitive interfaces that avoid "ahah" moments.

I think git is a great example of a project that exposes a root level config, a project level config, and respects env vars set on the shell session. Order of precedence:

1. CLI flags
2. Env var
3. Local Config file (./.git/config)
4. Global Config file (~/.gitconfig)
5. Default values

See: https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables#Committing

jaridmargolin:
@nexdrew @bcoe - Is it worth opening a new issue to discuss?

nexdrew:
@jaridmargolin I think perhaps it is. Even if we don't change the default order of precedence, we could at least look into an API that allows you to change it.

PR #309, opened 2015-12-03
update dependencies

bcoe:


nexdrew:
LGTM assuming tests pass :+1:

Would be nice to enable Greenkeeper for yargs. I would have done so already, but I think you need owner permissions to do so.

PR #313, opened 2015-12-08
Fixed: nested subcommand order

stelcheck:
https://github.com/bcoe/yargs/issues/225#issuecomment-162884778

### DIFF
--- a/index.js
+++ b/index.js
@@ -40,7 +40,8 @@ function Argv (processArgs, cwd) {
     )
   }
 
-  var options
+  self.level = 1;
+
   self.resetOptions = self.reset = function () {
     // put yargs back into its initial
     // state, this is useful for creating a
@@ -518,8 +519,10 @@ function Argv (processArgs, cwd) {
     // command defer processing to it.
     var handlerKeys = Object.keys(self.getCommandHandlers())
     for (var i = 0, command; (command = handlerKeys[i]) !== undefined; i++) {
-      if (~argv._.indexOf(command)) {
+      if (argv._.indexOf(command) === self.level - 1) {
+        self.level += 1;
         runCommand(command, self, argv)
+        self.level -= 1;
         return self.argv
       }
     }


bcoe:
@stelcheck thanks for this contribution, there's an ongoing discussion about what `yargs@4.x` looks like:

https://github.com/bcoe/yargs/issues/310

I've been swamped for the past week and haven't had a chance to chime in with feedback, but let's plan to get fixes for command handling released as part of `4.x` \o/

index.js
```diff
@@ -40,7 +40,8 @@ function Argv (processArgs, cwd) {
     )
   }
 
-  var options
```
nexdrew: Was this a mistake? Removing `options` pretty much breaks everything. :)

index.js
```diff
@@ -40,7 +40,8 @@ function Argv (processArgs, cwd) {
...
+  self.level = 1;
```
nexdrew: Not really any need to make `level` public. `var level = 1` would do.

index.js
stelcheck: Might be, let me have a look.

index.js
```diff
@@ -518,8 +519,10 @@ function Argv (processArgs, cwd) {
     // command defer processing to it.
     var handlerKeys = Object.keys(self.getCommandHandlers())
     for (var i = 0, command; (command = handlerKeys[i]) !== undefined; i++) {
-      if (~argv._.indexOf(command)) {
+      if (argv._.indexOf(command) === self.level - 1) {
+        self.level += 1;
         runCommand(command, self, argv)
+        self.level -= 1;
```
nexdrew: No semicolons, please. We use [standard style](https://www.npmjs.com/package/standard).

index.js
stelcheck: I need the level to be modified incrementally, don't think it would be possible with a public here. It needs to follow on the object instance.

nexdrew:
Please add some tests.

bcoe:
Hey @stelcheck, I actually started commands using this implementation -- it was my gut that this would be how folks would like commands to work (similar to `git`). Folks found the ordering a bit confusing, which was how we ended up with the concept of top-level commands, with the `reset()` method.

I would like to figure out a better way to handle nested commands, could you join the discussion in https://github.com/bcoe/yargs/issues/310

One idea I had, was perhaps something like this:

```js
yargs.command(['branch', 'ls'], function () {})
```

This would be fairly explicit which is nice.

bcoe:
see: https://github.com/bcoe/yargs/pull/368

bcoe:
@stelcheck commands have been overhauled in yargs 4.x, give it a spin:

https://github.com/bcoe/yargs/blob/4.x/README.md#commandcmd-desc-builder-handler

`npm i yargs@next`

stelcheck:
Wonderful! I'll give it a shot for climber then.

PR #316, opened 2015-12-13
Rename sigletonify to singletonify

josephfrazier:


bcoe:
@josephfrazier great catch :+1:

Issue #312, opened 2015-12-05
Remove dynamic requirement in parser.js


scriptjs:
Dynamic requirements are poor practice that prevent bundlers from being able to package the library. The following function needs to be rewritten in parser.js

```
function setConfig (argv) {
    var configLookup = {}

    // expand defaults/aliases, in-case any happen to reference
    // the config.json file.
    applyDefaultsAndAliases(configLookup, aliases, defaults)

    Object.keys(flags.configs).forEach(function (configKey) {
      var configPath = argv[configKey] || configLookup[configKey]
      if (configPath) {
        try {
          var config = require(path.resolve(process.cwd(), configPath))

          Object.keys(config).forEach(function (key) {
            // setting arguments via CLI takes precedence over
            // values within the config file.
            if (argv[key] === undefined || (flags.defaulted[key])) {
              delete argv[key]
              setArg(key, config[key])
            }
          })
        } catch (ex) {
          if (argv[configKey]) error = Error(__('Invalid JSON config file: %s', configPath))
        }
      }
    })
  }
```

nexdrew:
It's using `require` to read a JSON config file (not a library/module) that is specific to the end user's machine. See #304. We will eventually remove the use of `require` here to support other config file formats. Thanks.

scriptjs:
@nexdrew I understand why there may have been the lure to use it this way. Really, this ought to be a fs.read, not a dynamic require. The problem with CommonJS in the first place is that require is a function. es6 solves this by making everything declarative and static with its import statements. The use of require like is simply just not what you want to do. Is game off in almost all circumstances for any bundler.

bcoe:
@scriptjs if I understand this edge-case, it would be if someone bundles yargs using something like `CommonJS`, and then in the context of a browser loads a config file?

scriptjs:
@bcoe. Its actually not an edge case. Most folks today are already writing in es6 or in the process of transitioning. If anyone attempts to bundle yargs for node, they will encounter this issue. It will not bundle without forking the code and eliminating the issue. On that basis, not sure why this issue was closed.

[...]

PR #321, opened 2015-12-24
allow a custom parser to be passed to config()

### LINKED ISSUE #304: suggestion: support arbitrary filename for configuration file
Right now it seems only `.json` files can be used for config files, because they are `require()`'ed. Would be nice to support any filename, such as `.foorc`. Very minor but it took me a bit to figure out why this wasn't working. Thanks.

bcoe:
Allow a custom parsing function to be passed to config.

Fixes #304 

Reviewers: @sebastienbarre, @indieisaconcept, @nexdrew

### DIFF
--- a/README.md
+++ b/README.md
@@ -925,6 +925,7 @@ Valid `opt` keys include:
 - `boolean`: boolean, interpret option as a boolean flag, see [`boolean()`](#boolean)
 - `choices`: value or array of values, limit valid option arguments to a predefined set, see [`choices()`](#choices)
 - `config`: boolean, interpret option as a path to a JSON config file, see [`config()`](#config)
+- `configParser`: function, provide a custom config parsing function, see [`config()`](#config)
 - `count`: boolean, interpret option as a count of boolean flags, see [`count()`](#count)
 - `default`: value, set a default value for the option, see [`default()`](#default)
 - `defaultDescription`: string, use this description for the default value in help content, see [`default()`](#default)
--- a/README.md
+++ b/README.md
@@ -549,17 +549,19 @@ var argv = require('yargs')
   .argv;
 ```
 
-<a name="config"></a>.config(key, [parseFn], [description])
+<a name="config"></a>.config(key, [description], [parseFn])
 ------------
 
 Tells the parser that if the option specified by `key` is passed in, it
 should be interpreted as a path to a JSON config file. The file is loaded
-and parsed, and its properties are set as arguments. If present, the
-`description` parameter customizes the description of the config (`key`) option
+and parsed, and its properties are set as arguments.
+
+An optional `description` can be provided to customize of the config (`key`) option
 in the usage string.
 
-An optional `parseFn` argument can be passed to provide a custom config
-parser:
+An optional `parseFn` can be used to provide a custom parser. The parsing
+function must be synchronous, and should return an object containing
+key value pairs or an error.
 
 ```js
 var argv = require('yargs')
--- a/README.md
+++ b/README.md
@@ -556,7 +556,7 @@ Tells the parser that if the option specified by `key` is passed in, it
 should be interpreted as a path to a JSON config file. The file is loaded
 and parsed, and its properties are set as arguments.
 
-An optional `description` can be provided to customize of the config (`key`) option
+An optional `description` can be provided to customize the config (`key`) option
 in the usage string.
 
 An optional `parseFn` can be used to provide a custom parser. The parsing
--- a/README.md
+++ b/README.md
@@ -549,7 +549,7 @@ var argv = require('yargs')
   .argv;
 ```
 
-<a name="config"></a>.config(key, [description])
+<a name="config"></a>.config(key, [parseFn], [description])
 ------------
 
 Tells the parser that if the option specified by `key` is passed in, it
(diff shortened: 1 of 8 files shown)


bcoe:
CC: @maxrimue, would also be great to get your review.

maxrimue:
LGTM, you could consider using `for` loops instead of `.forEach` since that is [faster](https://jsperf.com/for-vs-foreach/37) and would also allow you to get rid of the semicolons there, but perhaps sacrificing performance for readability isn't a bad idea, too

lib/parser.js
```diff
@@ -341,7 +341,14 @@ module.exports = function (args, opts, y18n) {
...
       if (configPath) {
         try {
-          var config = require(path.resolve(process.cwd(), configPath))
+          var config = null
+          var resolvedConfigPath = path.resolve(process.cwd(), configPath)
+
+          if (typeof flags.configs[configKey] === 'function') {
+            config = flags.configs[configKey](resolvedConfigPath)
+          } else {
+            config = require(resolvedConfigPath)
```
nexdrew: Per #312, should we aspire to remove the dynamic `require()` here? Maybe use [`jsonfile`](https://www.npmjs.com/package/jsonfile) instead or check file extension first?

Or would you rather wait til 4.x to change this? I guess changing it could certainly cause backwards compatibility problems.

lib/parser.js
```diff
@@ -352,7 +359,7 @@ module.exports = function (args, opts, y18n) {
             }
           })
         } catch (ex) {
-          if (argv[configKey]) error = Error(__('Invalid JSON config file: %s', configPath))
+          if (argv[configKey]) error = Error(__('Invalid config file (%s): %s', ex.message, configPath))
```
nexdrew: This change will need to be reflected in all locale files.

README.md
```diff
@@ -558,6 +558,17 @@ and parsed, and its properties are set as arguments. If present, the
 `description` parameter customizes the description of the config (`key`) option
 in the usage string.
 
+An optional `parseFn` argument can be passed to provide a custom config
+parser:
```
nexdrew: Should we perhaps mention that the function should be synchronous?

Or maybe just describe the expected API/contract of the parser function? E.g.

> The parser function should accept a file path argument and synchronously return an object whose keys will be applied to `argv`.

lib/parser.js
bcoe: Good point -- I'm tempted to keep the default error; unless someone returns an error object -- and then we'll output that error (folks using the closure can perform their own locale work).

lib/parser.js
bcoe: I recall that someone switched it to require so that they could have `.js` config files -- I think we should hold off on this change, in `v4` perhaps we direct people to provide a custom require hook that uses `require()`.

bcoe:
@maxrimue quite valid points, but I'm tempted to stick with the `forEach` in this case mainly because it looks a bit cleaner with the array coercion:

```js
Array.isArray(key) ? key : [key]
```

The above starts looking pretty hideous embedded in a `for` statement.

bcoe:
@nexdrew ready for final review.

README.md
```diff
@@ -549,15 +549,28 @@ var argv = require('yargs')
...
-and parsed, and its properties are set as arguments. If present, the
-`description` parameter customizes the description of the config (`key`) option
+and parsed, and its properties are set as arguments.
+
+An optional `description` can be provided to customize the config (`key`) option
 in the usage string.
 
+An optional `parseFn` can be used to provide a custom parser. The parsing
+function must be synchronous, and should return an object containing
+key value pairs or an error.
```
nexdrew: Should an error be returned or thrown? Sorry to be nit-picky, but it's not quite clear to me as a reader.

lib/parser.js
```diff
@@ -341,7 +341,15 @@ module.exports = function (args, opts, y18n) {
       var configPath = argv[configKey] || configLookup[configKey]
       if (configPath) {
         try {
-          var config = require(path.resolve(process.cwd(), configPath))
+          var config = null
+          var resolvedConfigPath = path.resolve(process.cwd(), configPath)
+
+          if (typeof flags.configs[configKey] === 'function') {
+            config = flags.configs[configKey](resolvedConfigPath)
+            if (config instanceof Error) error = config
```
nexdrew: If `config` is an `Error`, then the enumerable properties from the error will be applied to argv in this function below, which, I'm assuming, is not intended.

Perhaps throw the error here instead, or maybe just return after setting `error`?

README.md
nexdrew: I see from the code and tests that, indeed, an `Error` can be returned. Perhaps we should have a sentence describing the options for error handling/propagation here?

lib/parser.js
bcoe: returning seems reasonable.

nexdrew:
My last nit is that it would be nice to use a custom config parser function with the `option` object shorthand, something like:

```js
function parseIni (configPath) {
  return require('ini').parse(fs.readFileSync(configPath, 'utf-8'))
}

var argv = require('yargs')
  .option('i', {
    alias: 'ini',
    describe: 'The ini config file',
    config: true,
    configParser: parseIni
  })
  .help('h').alias('h', 'help')
  .argv

console.log(argv)
```

Otherwise LGTM :+1:

bcoe:
@nexdrew rather than document whether an error should be thrown or returned, both now work -- better that we eliminate room for confusion.

You can now provide a config parser in the option short hand.

nexdrew:
Awesome! I vote :shipit:

PR #325, opened 2016-01-02
Make tests pass on Windows

isaacs:
Note: depends on https://github.com/bcoe/nyc/pull/123

Probably want to coordinate some releases.  Spawn-wrap and foreground-child are both sitting on a `next` dist-tag at the moment, and tap has a gnarly set of stuff on https://github.com/isaacs/node-tap/pull/204

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   },
   "scripts": {
     "pretest": "standard",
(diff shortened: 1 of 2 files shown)


isaacs:
Probably travis is going to fail because the changes to nyc haven't been landed :)

bcoe:
closing in favor of #326, which has `nyc@next`.

Issue #327, opened 2016-01-03
Allow args string in .parse()


kellyselden:
Use case is your program wraps other programs:

```sh
my-app --an-option test --a-wrapped-programs-options="--test x --another=s"
```

The closest config I got was something like:

```js
var yargs = require('yargs');

var argv = yargs.argv;

var aWrappedProgramsOptions = argv['a-wrapped-programs-options'];

aWrappedProgramsOptions = yargs.reset()
    .parse(aWrappedProgramsOptions);
```

But parse only takes an array. I thought I could do something like `aWrappedProgramsOptions.split(' ')`, but that doesn't account for `=` delimiters. If I go further, I start implementing some of the logic in yargs already.

Would it make sense to allow strings in `.parse()`?

bcoe:
@kellyselden you should be able to do something like this:

```js
#!/usr/bin/env node

var opts = require('yargs')('--test x --another=s'.split(' '))
    .argv

console.log(opts)
```
yargs let's you pass an array representing `argv` as an argument to `yargs`.

Does this do the trick for you?

kellyselden:
That works. I didn't realize yargs would handle `=` for you, and you wouldn't have to split on it. Is this use case worth a README update? I can update it if you want.

bcoe:
@kellyselden happy to consider a README update, if you can find an elegant way to outline this tidbit of information :+1:

bcoe:
@kellyselden give: https://github.com/bcoe/yargs/pull/331 a shot and let me know what you think.

kellyselden:
LGTM

bcoe:
@kellyselden this feature is now released on the `next` tag of yargs. Give this a shot:

`npm i yargs@next`

Let me know what you think :+1:

kellyselden:
@bcoe Works great. Was able to remove this line https://github.com/kellyselden/eslint-mocha/pull/17/files#diff-dec258563532f674fb4d25d7d9952de6L22

Thank you.

PR #328, opened 2016-01-03
add comment for splitting a string for parsing

kellyselden:
re #327

An explanation of how to go from an args string to array for the `.parse()` function, while not worrying about the `=`.

This would just be a time saver, as someone could also search closed issues and find mine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,6 +364,12 @@ or use `.parse()` to do the same thing:
 require('yargs').parse([ '-x', '1', '-y', '2' ])
 ````
 
+If you have a random string of args you want to parse, you only have to split on the space as yargs handles the `=` for you:
+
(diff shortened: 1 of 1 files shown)


bcoe:
@kellyselden I was thinking about it, this approach doesn't solve the use-case of parsing a string that looks like this:

`--x="hello world" -y 99`, as you'll split the string like so: `['"hello', 'world"']` -- I'm tempted to instead add this functionality:

```js
yargs.parse('--x="hello world" -y 99')
```

Which would handle tokenizing an argument string appropriately.

bcoe:
@kellyselden I'm going to re-open #327 in favor of allowing `.parse()` to intelligently parse a string.

Issue #329, opened 2016-01-03
fix inconsistency with version and default API


bcoe:
the `.version()` method takes the `key` as the second argument, rather than the first -- let's fix this when we move to the `4.x` API.

CC: @maxrimue

maxrimue:
Reopened as `.default()`'s API inconsistency has never been fixed

bcoe:
at this point folks are used to this API surface.

PR #330, opened 2016-01-03
Fix .version() API inconsistency

maxrimue:
This pull request changes the version() API to solve the inconsistency as raised by @bcoe in #329

### DIFF
--- a/README.md
+++ b/README.md
@@ -1120,12 +1120,13 @@ present script similar to how `$0` works in bash or perl.
 
 `opts` is optional and acts like calling `.options(opts)`.
 
-.version(version, [option], [description])
+.version([option], [description], version)
 ----------------------------------------
 
 Add an option (e.g. `--version`) that displays the version number (given by the
-`version` parameter) and exits the process. If present, the `description`
-parameter customizes the description of the version option in the usage string.
+`version` parameter) and exits the process. If no option is provided, it will default to `--version`.
+If present, the `description` parameter customizes the description of the version
+option in the usage string.
 
 You can provide a `function` for version, rather than a string.
 This is useful if you want to use the version from your package.json:
--- a/index.js
+++ b/index.js
@@ -378,11 +378,22 @@ function Argv (processArgs, cwd) {
   }
 
   var versionOpt = null
-  self.version = function (ver, opt, msg) {
-    versionOpt = opt || 'version'
-    usage.version(ver)
+  self.version = function (opt, msg, ver) {
+    if (arguments.length === 1) {
+      ver = opt
+      versionOpt = 'version'
+      msg = usage.deferY18nLookup('Show version number')
+    } else if (arguments.length === 2) {
+      versionOpt = opt
+      ver = msg
+      msg = usage.deferY18nLookup('Show version number')
+    } else {
+      versionOpt = opt
+    }
+
+    usage.version(ver || undefined)
     self.boolean(versionOpt)
-    self.describe(versionOpt, msg || usage.deferY18nLookup('Show version number'))
+    self.describe(versionOpt, msg)
     return self
   }
 
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -370,7 +370,7 @@ module.exports = function (yargs, y18n) {
 
   // logic for displaying application version.
   var version = null
-  self.version = function (ver, opt, msg) {
+  self.version = function (ver) {
     version = ver
   }
 
--- a/test/usage.js
+++ b/test/usage.js
@@ -862,7 +862,7 @@ describe('usage tests', function () {
     it('should display version', function () {
       var r = checkUsage(function () {
         return yargs(['--version'])
-        .version('1.0.1', 'version', 'Show version number')
+        .version('version', 'Show version number', '1.0.1')
         .wrap(null)
         .argv
       })
@@ -874,12 +874,22 @@ describe('usage tests', function () {
       r.logs[0].should.eql('1.0.1')
     })
 
+    it('accepts version option as first argument, and version number as second argument', function () {
+      var r = checkUsage(function () {
+        return yargs(['--version'])
+        .version('version', '1.0.0')
+        .wrap(null)
+        .argv
+      })
+      r.logs[0].should.eql('1.0.0')
+    })
+
     it('should allow a function to be provided, rather than a number', function () {
       var r = checkUsage(function () {
         return yargs(['--version'])
-        .version(function () {
+        .version('version', function () {
           return require('./fixtures/config').version
-        }, 'version')
+        })
         .wrap(null)
         .argv
       })
@@ -902,7 +912,7 @@ describe('usage tests', function () {
       it('should not validate arguments (required argument)', function () {
         var r = checkUsage(function () {
           return yargs(['--version'])
-            .version('1.0.1', 'version', 'Show version number')
+            .version('version', 'Show version number', '1.0.1')
             .demand('some-opt')
             .wrap(null)
             .exitProcess(false)
@@ -923,7 +933,7 @@ describe('usage tests', function () {
         var r = checkUsage(function () {
           return yargs(['--version', '--some-opt'])
             .nargs('some-opt', 3)
-            .version('1.0.1', 'version', 'Show version number')
+            .version('version', 'Show version number', '1.0.1')
             .wrap(null)
             .exitProcess(false)
             .argv


index.js
```diff
@@ -378,11 +378,13 @@ function Argv (processArgs, cwd) {
   }
 
   var versionOpt = null
-  self.version = function (ver, opt, msg) {
-    versionOpt = opt || 'version'
-    usage.version(ver)
+  self.version = function (opt, msg, ver) {
+    var version = ver || msg || opt || undefined
```
bcoe: I think we can simply this logic a bit, and avoid the ternary operator. I might be tempted to do something like this:

1. check the # of arguments, if there's only **1** assume that it represents the version #.
2. if there are **2** assume that it's `opt`, `version`.
3. if there are **3** assume `opt`, `description`, `version`.



```js
self.version = function (opt, desc, version) {
  if (arguments.length === 1) {
    version = opt
    opt = 'version'
  } else if (arguments.length === 2) {
    version = desc
    desc = null
  }

  desc = desc || usage.deferY18nLookup('Show version number')
  usage.version(version)
  self.boolean(opt)
  self.sescribe(opt, desc)
}
```

bcoe:
@maxrimue this is looking great, one thing I noticed ... this means that:

```
.default(key, value, [description])
```

Is inconsistent with the rest of the API too ... description should probably be the second value right?

bcoe:
testing GitHub integration, CC: @nexdrew, @Irina perhaps you could give @maxrimue some feedback too.

maxrimue:
@bcoe Yes, I think so too, `description` should be the second value as some other functions also accept descriptions as their second parameters

index.js
maxrimue: @bcoe I absolutely agree with this idea, to be honest, the only reason why I kept it this short was because the code before seemed rather valuing shortness either, I would like to make it more readable as well

bcoe: @maxrimue :+1: this is what code review is all about, nothing to feel bad about (mind you, I definitely have a guilt complex about some of the terrible code I've added to yargs :stuck_out_tongue: ).

maxrimue: @bcoe Well it does its job doesn't it? :D I was thinking about that we could maybe create a general global function for this parameter thing though, it would make sense if there are some other functions that need to check first what parameters are provided etc., so we can write code like this:
```javascript
self.version = function (opt, desc, version) {
  var args = getParameters([opt, desc, version], ['opt', 'desc'], ['version']);
  usage.version(args.version)
  self.boolean(args.opt)
  self.describe(args.opt, args.desc)
}
```

The function call passes the parameters, tells which parameters are optional and which required, and then returns an object with the same names or so, was just a random idea though.

lrlna:
I think I need a bit more time with the codebase before I make reasonable code comments. Give me a bit more time ^.^

Did have a question about the reason for keeping `description` as second argument. I've normally seen args "snowball" their way down; e.g. :
- if *1* arg -- opt
- if *2* arg -- opt, version
- if *3* arg -- opt, version, desc 
- etc etc etc 

Do we just have other pieces of code taking `description` as 2nd ?

maxrimue:
@lrlna Yes, this pull request is just for providing consistency. I think the reason for why we don't have that snowball effect here is that different functions would require different orders, causing inconsistencies, which could cause confusion among people using this module

lrlna:
@maxrimue cool -- thanks for clarifying!

index.js
nexdrew: How do y'all feel about defaulting the `version` value to the parent package.json value (if found)? It has always bugged me that I have to code this manually each time, and I always end up doing the same thing. Am I just missing something?

maxrimue: @nexdrew Well, you could create a function that finds out that value itself and use it for `.version()`, but you're right there, there could be a function that does exactly that as a default value for the version

nexdrew: I always just want to customize the option (usually `-v` with `--version` alias) and it's annoying to `require(../package.json).version` every time. (Though we probably shouldn't use `require()` internally for this, to avoid the dynamic require problem.)

maxrimue:
@bcoe I made it more readable now, what do you think? Is the `else` path maybe too much?

README.md
```diff
@@ -1120,12 +1120,13 @@ present script similar to how `$0` works in bash or perl.
...
 `opts` is optional and acts like calling `.options(opts)`.
 
-.version(version, [option], [description])
+.version([option], [description], version)
 ----------------------------------------
 
 Add an option (e.g. `--version`) that displays the version number (given by the
-`version` parameter) and exits the process. If present, the `description`
-parameter customizes the description of the version option in the usage string.
+`version` parameter) and exits the process, if not provided, it will default to `--version`.
```
nexdrew: I would make the last part its own sentence:

```
`version` parameter) and exits the process. If not provided, it will default to `--version`.
```

nexdrew:
I like this.

But, just to get this out off my chest, it feels wrong to make this breaking change for such a trivial thing in the name of consistency - without a way to deprecate or account for the old usage. I mean, we _could_ just add another method, though that would leave the API inconsistency intact.

Ok, now that that's out of the way, carry on. :+1:

nexdrew:
As this represents the first breaking change to the API, I'd like to suggest that we add a new section to the CHANGELOG that documents known breaking changes b/w 3.x and 4.x, with notes on how to upgrade.

Maybe something like this:

# 4.x Breaking Changes

- PR #330: `.version()` method now expects the `version` argument to be last instead of first.

    Example: change `.version('1.0.0', 'v', 'Print app version')` to `.version('v', 'Print app version', '1.0.0')`

lib/usage.js
```diff
@@ -370,8 +370,8 @@ module.exports = function (yargs, y18n) {
 
   // logic for displaying application version.
   var version = null
-  self.version = function (ver, opt, msg) {
-    version = ver
+  self.version = function (opt, msg, ver) {
+    version = ver || msg || opt || undefined
```
bcoe: I'm pretty sure you can get away with just making this:

```js
self.version = function (ver) {
  version = ver
}
```

index.js
```diff
@@ -378,11 +378,27 @@ function Argv (processArgs, cwd) {
...
+
+    if (arguments.length === 1) {
+      version = opt
+      versionOpt = 'version'
+      versionMsg = usage.deferY18nLookup('Show version number')
+    } else if (arguments.length === 2) {
+      versionOpt = opt
+      version = msg
+      versionMsg = usage.deferY18nLookup('Show version number')
+    } else {
```
bcoe: minor nit, but if rather than creating the new variables:

```js
var version = null
var versionMsg = null
```

We simply override the variables being passed in:

```
self.version = function (opt, msg, ver) {
  if (arguments.length === 1) {
   version = opt
 } else if (arguments.length === 2) {
   version = msg
 }
 versionOpt = opt || 'version'

usage.version(ver)
self.boolean(versionOpt)
self.describe(versionOpt, msg || usage.deferY18nLookup('Show version number'))
```

We can get things a little bit more concise \o/

bcoe:
@maxrimue this is looking great, just a couple more comments -- we should also make sure to open up a ticket to track @nexdrew's suggestion of defaulting to the `version` in package.json. This might be a great next ticket for you to tackle @maxrimue :smile:

bcoe:
@nexdrew I agree, but how do you feel about landing a bunch of our proposed breaking changes on the `4.x` branch -- move forward in one slightly painful jump.

index.js
maxrimue: @bcoe I'd like to get the assignment of `versionOpt` into the if/else path too, since `opt || 'version'` wouldn't work if only the `opt` parameter is provided:
```javascript
self.version('4.0.0');

self.version = function (opt, msg, ver) {
  if (arguments.length === 1) {
   version = opt
 } else if (arguments.length === 2) {
   version = msg
 }
 versionOpt = opt || 'version'
}

// version = '4.0.0'
// versionOpt = '4.0.0'
```

nexdrew:
@bcoe Yes, I totally agree we should be merging these changes to the `4.x` branch. I just want to make sure we document the known breaking changes so, when we start to publish 4.x versions, our users have a clear migration path. E.g. "here's what we changed" and "here's what you need to do to upgrade".

index.js
maxrimue: My idea would be, that if we add a default value for the third parameter, `version`, we can make the third parameter optional as well, so we wouldn't need this big logic there any more, thoughts?

nexdrew: @maxrimue Thanks for clarifying that.

I'm a fan of the snowball effect here, but I'm afraid it might make specifying a version number other than the default (which could be a higher priority use-case) more cumbersome than necessary. I guess it's not too bad if we can translate a `null` description as "use the default description", e.g. `version('version', null, '4.0.0')`.

If we're concerned with the multi-parameter API, perhaps we should just accept an object?

```js
version({ option: 'version', version: '4.0.0', describe: 'Current version of this module' })
```

That would avoid the ambiguity of the parameters and keep everything on a level playing ground. It would also allow us to easily support other parameters, e.g. `alias`.

We could even support a "string or object" parameter, assuming we can make a decision on which use-case gets priority for the string (I would argue it should be the option flag, similar to the `.help()` method).

Don't want to throw a wrench into the gears of this PR, but what do you think?

maxrimue: @nexdrew I actually was thinking about accepting an object as well. I think, accepting strings via the snowball effect (with support for `null`) <i>or</i> an object would be the easiest to code, yet the easiest to use as an API, what do you think?

nexdrew: That would definitely be the most flexible API. Just not sure if setting that API precedent will paint us into a corner later on.

You've got my vote though. Would love to hear opinions of others on this.

bcoe: Personally I like the first option better, I'm going to tend to just specify the option and the version #, and rarely change the description. If I ever want to change the description I'm fine with just specifying three values.

Rather than accepting an object, which I think starts to over complicate things for such a simple option., I like @nexdrew's idea that we'd default to the version # in the nearest package.json (@sindresorhus has a module for finding this file in a consistent way). 

Once we have this 99% of the time I feel people will write:

```js
.version()
// or
.version('v')
// or
.version('v', '99.9')
```

I don't think this would avoid needing to set an object as a parameter or a null description.

maxrimue: @bcoe So you think we should make the version parameter optional as well?

nexdrew: I totally agree with the use-cases @bcoe listed above. And I would _love_ to be able to just use `.version()` with no args. :heart_eyes:

maxrimue: @bcoe @nexdrew What do you think? Should I integrate the default value for the version thing in this pull request so we can simplify things? If I don't, that would mean that we merge logic that is just there for covering the version parameter to be non-optional, once we put in the default value for the version, we wouldn't need that code anymore anyway.

nexdrew: If it saves work, then I say go for it.

nexdrew:
Not sure if we have a mantra or guiding principle for 4.x, but in my eyes it should be something like:

**Make the existing features easier, less verbose, more consistent. And add some new cool features.**

Quite possible that's too obvious or simplistic, but having a "statement of objective" might help us measure the value of ideas.

E.g. if this is something I can do in 3.x, is it now "easier" (to do or understand), "less verbose" (requires fewer keystrokes), or "more consistent" (works the same way as other parts of the API) in 4.x? If the answer is yes to 2/3 or greater, then adopt. If 1/3 or less, then drop it.

Granted, the answers will not always be a clear "yes" or "no", and they may heavily depend upon the use-cases, but a warped yard stick is better than no yard stick, right?

index.js
bcoe: @nexdrew @maxrimue in this pull let's concentrate on getting this logic working with appropriate tests:

```js
.version('version', '99.9')
.version('version', 'my version description', '99.9')
.version('version', 'my version description', function () {return '99.9'})
```

Then, in a follow up pull request we can use:

https://www.npmjs.com/package/find-up

To lookup the closed `package.json`, and add the following functionality:

```js
.version()
.version('version')
```

I think this is most consistent with the API elsewhere, and would simplify the version feature _a lot_.

test/usage.js
```diff
@@ -877,9 +877,9 @@ describe('usage tests', function () {
     it('should allow a function to be provided, rather than a number', function () {
```
bcoe: I think it might be worth adding one more test, for:

```js
.version('version', '1.0.0');
```

`accepts version option as first argument, and version number as second argument`.

We'll eventually add additional tests for:

```js
.version()

// and

.version('version')
```

bcoe:
Great work @maxrimue.

PR #331, opened 2016-01-03
allow yargs.parse() and yargs(foo).argv to accept a raw argument string

### LINKED ISSUE #327: Allow args string in .parse()

bcoe:
@kellyselden made the great suggestion that yargs be able to handle a raw argument string, prior to it being parsed by the OS. Like so:

```js
var a1 = yargs.parse('-x=\'marks "the" spot\' --foo "break \'dance\'"')
```

One great advantage to this is that it makes yargs easier to embed in unorthodox environments, .e.g., a REPL in the web-browser.

fixes #327

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Upcoming
 
+- [#331](https://github.com/bcoe/yargs/pull/331) Handle parsing a raw argument string (@kellyselden)
 - [#325](https://github.com/bcoe/yargs/pull/325) Tweaks to make tests pass again on Windows (@isaacs)
 - [#321](https://github.com/bcoe/yargs/pull/321) Custom config parsing function (@bcoe)
 
--- a/README.md
+++ b/README.md
@@ -946,6 +946,8 @@ Valid `opt` keys include:
 
 Parse `args` instead of `process.argv`. Returns the `argv` object.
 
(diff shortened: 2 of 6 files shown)


kellyselden:
After looking at your tests, it looks like this will work quite well. Nice quick work!

nexdrew:
@bcoe Sorry about this, I started to take a look a couple different times and never finished my review.

For some reason I thought the tests already used a string of args, so thought we could reuse logic there, but that's not the case - they use an array of strings.

So overall this LGTM. Nice work!

bcoe:
@nexdrew I was thinking that we could actually update the tests to use a string instead, why don't I do that good use of this new feature.

PR #337, opened 2016-01-09
Default value for version

maxrimue:
This PR adds a default value for the `version` parameter in `.version()`, making the parameter optional. In addition with #330 (this PR includes its commit), following ways of interacting with the API are possible:
```javascript
.version(opt, msg, ver)

.version(opt, ver)

.version(ver)

.version()
```
When no parameter is provided, [read-pkg-up](https://github.com/sindresorhus/read-pkg-up) is called to find the parent `package.json`, so the one of the module using yargs, parses its content, and uses the `version` value from there.

_Note_: Currently, there are no tests. I would like to get some ideas how to write the tests first.

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,6 +5,7 @@ var path = require('path')
 var Usage = require('./lib/usage')
 var Validation = require('./lib/validation')
 var Y18n = require('y18n')
+var readPkgUp = require('read-pkg-up')
 
 Argv(process.argv.slice(2))
 
@@ -379,17 +380,18 @@ function Argv (processArgs, cwd) {
 
   var versionOpt = null
   self.version = function (opt, msg, ver) {
-    if (arguments.length === 1) {
+    if (arguments.length === 0) {
+      ver = readPkgUp.sync({cwd: path.join(__dirname + '/..')}).pkg.version
+      opt = 'version'
+    } else if (arguments.length === 1) {
       ver = opt
-      versionOpt = 'version'
-      msg = usage.deferY18nLookup('Show version number')
+      opt = 'version'
     } else if (arguments.length === 2) {
-      versionOpt = opt
       ver = msg
-      msg = usage.deferY18nLookup('Show version number')
-    } else {
-      versionOpt = opt
     }
+
+    versionOpt = opt
+    msg = msg || usage.deferY18nLookup('Show version number')
 
     usage.version(ver || undefined)
     self.boolean(versionOpt)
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -378,11 +379,26 @@ function Argv (processArgs, cwd) {
...
 
   var versionOpt = null
-  self.version = function (ver, opt, msg) {
-    versionOpt = opt || 'version'
-    usage.version(ver)
+  self.version = function (opt, msg, ver) {
+    if (arguments.length === 0) {
+      ver = parent(__dirname).parse().version
+      versionOpt = 'version'
+      msg = usage.deferY18NLookup
```
bcoe: this looks like a bug, should be:

```js
msg = usage.deferY18nLookup('Show version number')
```

There's starting to be some duplicate code here, I would actually refactor to something like this:

```js
self.version = function (opt, msg, ver) {
  if (arguments.length === 0) {
    ver = parent(__dirname).parse().version
    opt = 'version'
  } else if (arguments.length === 1) {
    ver = opt
    opt = 'version'
  } else if (arguments.length === 2) {
    ver = msg
  }

  versionOpt = opt
  msg = msg || usage.deferY18nLookup('Show version number')

  usage.version(ver)
  self.boolean(versionOpt)
  self.describe(versionOpt, msg)
  return self
}
```

bcoe:
@maxrimue this is a great start, here's how I would write a test:

```js
    it('defaults to the version # in the package.json', function () {
        return yargs(['--version'])
        .version()
        .wrap(null)
        .argv
      })
      r.logs[0].should.eql(require('./package.json').version)
    })
```

index.js
maxrimue: @bcoe Good catch, definitely agree there

maxrimue:
@bcoe The tests unfortunately have to be more complicated, which is why I asked, here's why:
`parent-package-json`, if called, starts in one directory (which, in this PR is at `__dirname` because of the parameter), and then goes to the parent folder, searches for a `package.json`, and continues the search in the next parent folders until it finds one. This is why we would need a module requiring yargs from a parent folder.

As you can see in the current commit, I now have added an `if/else` path. If `parent-package-json` doesn't find anything, it returns `false`, if it finds one, it returns methods (like `.parse()`). The tests also fail because the module won't find any `package.json`, so it'll return `false` instead of the desired version.

Do you have any idea how to achieve proper tests for this? Would also be cool to hear from you @nexdrew and @lrlna

maxrimue:
Closed as #343 replaces this pull request

PR #339, opened 2016-01-09
split(' ') is no longer needed in tests

bcoe:
@kellyselden look what your feature request allowed us to do to tests \o/ sign of a great feature idea.

CC: @nexdrew


kellyselden:
@bcoe That's awesome. I'll be one of the first users here once it's released https://github.com/kellyselden/eslint-mocha/blob/master/bin/eslint-mocha.js#L19-L26

PR #343, opened 2016-01-10
default to the version we read in the package using yargs

bcoe:
This is an update to @maxrimue's awesome work adding tests.

I found the test cases slightly tricky to simulate, I ended up:

1. copying the current yargs module to a fixtures directory so that I could test the normal behavior.
2. creating a symlink in the fixtures directory so that I could test the symlink behavior.

I would love thoughts from @sindresorhus, and @nexdrew as to how we might simply this testing logic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1124,18 +1124,18 @@ present script similar to how `$0` works in bash or perl.
 ----------------------------------------
 
 Add an option (e.g. `--version`) that displays the version number (given by the
-`version` parameter) and exits the process. If no option is provided, it will default to `--version`.
-If present, the `description` parameter customizes the description of the version
-option in the usage string. You can also pass no `version` parameter (`.version()`),
-in this case, yargs will parse the `package.json` of your module and use its `version` value.
+`version` parameter) and exits the process.
+
+If no arguments are passed to `version` (`.version()`), yargs will parse the `package.json`
+of your module and use its `version` value. The default value of `option` is `--version`.
 
 You can provide a `function` for version, rather than a string.
-This is useful if you want to use the version from your package.json:
+This is useful if you want to use a version stored in a location other than package.json:
 
 ```js
 var argv = require('yargs')
   .version(function() {
-    return require('../package').version;
+    return require('../lib/version').version;
   })
   .argv;
 ```
--- a/README.md
+++ b/README.md
@@ -1120,13 +1120,14 @@ present script similar to how `$0` works in bash or perl.
 
 `opts` is optional and acts like calling `.options(opts)`.
 
-.version([option], [description], version)
+.version([option], [description], [version])
 ----------------------------------------
 
 Add an option (e.g. `--version`) that displays the version number (given by the
 `version` parameter) and exits the process. If no option is provided, it will default to `--version`.
 If present, the `description` parameter customizes the description of the version
-option in the usage string.
+option in the usage string. You can also pass no `version` parameter (`.version()`),
+in this case, yargs will parse the `package.json` of your module and use its `version` value.
 
 You can provide a `function` for version, rather than a string.
 This is useful if you want to use the version from your package.json:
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1120,13 +1120,14 @@ present script similar to how `$0` works in bash or perl.
...
-.version([option], [description], version)
+.version([option], [description], [version])
 ----------------------------------------
 
 Add an option (e.g. `--version`) that displays the version number (given by the
 `version` parameter) and exits the process. If no option is provided, it will default to `--version`.
 If present, the `description` parameter customizes the description of the version
-option in the usage string.
+option in the usage string. You can also pass no `version` parameter (`.version()`),
+in this case, yargs will parse the `package.json` of your module and use its `version` value.
```
maxrimue: Maybe we should add that you can also just leave out `description` and/or `option` parameters and still pass a `version` parameter, not only like `.version(null, null, version)` but like `.version(desc, ver)` and `.version(ver)`, too

index.js
```diff
@@ -379,24 +380,39 @@ function Argv (processArgs, cwd) {
...
     usage.version(ver || undefined)
     self.boolean(versionOpt)
     self.describe(versionOpt, msg)
     return self
   }
 
+  function guessVersion () {
+    var obj = readPkgUp.sync({
+      cwd: path.resolve(__dirname, '../')
+    })
```
nexdrew: I think the only time this may return an incorrect result is if yargs is a nested dependency - but that shouldn't really happen, right?

test/usage.js
```diff
@@ -874,6 +877,52 @@ describe('usage tests', function () {
...
     })
 
+    describe('default version #', function () {
+      before(function (done) {
+        this.timeout(10000)
+        // create a symlinked, and a physical copy of yargs in
+        // our fixtures directory, so that we can test that the
+        // nearest package.json is appropriately loaded.
+        cpr('./', './test/fixtures/yargs', {
+          filter: /node_modules|example|test|package\.json/
```
nexdrew: To make this more of a "real world" scenario, we probably shouldn't filter out `package.json` (since yargs will have one when installed as a dependency). Perhaps instead we should copy a control `package.json` file with a constant version so we can guarantee its version is not picked up by the test.

index.js
```diff
@@ -379,24 +380,39 @@ function Argv (processArgs, cwd) {
...
+
+    // fallback to process.cwd() this
+    // solves the case of symlinked modules.
+    if (!obj.pkg) {
+      obj = readPkgUp.sync()
+    }
```
nexdrew: What's the use-case for the symlink? Is it for when yargs is a symlinked global dependency (e.g. using `npm link`)? I'm just concerned that falling back to cwd here might return different, unexpected results if a consumer has symlinked their yargs dependency and relies on this functionality unintentionally - in which case they'd be better off handing yargs an explicit version value.

Typically, the only symlink involved in a real-world global install of a consumer would the bin executable on the user's PATH, which would point to a normal `node_modules` install such that the first `readPkgUp` call would work.

index.js
bcoe: I can't think of a scenario where this logic breaks down -- like you say maybe if there's a bin sitting next to a nested `package.json` ... but in this case, the closest `package.json` is probably still the right choice.

index.js
bcoe: there's two reasons why I think it might be worth falling back to `cwd`:

1. without handling this edge-case, we can't test the `yargs.version()` functionality when yargs is linked:

`npm link`
`npm link yargs`

2. when testing I sometimes like to create a bin in the root directory of the yargs folder, this would also not work for `yargs.version()`.

My temptation is to handle the edge-case, and add tests for any weird exceptions we run into in the wild.

what do you think?

test/usage.js
bcoe: I was thinking the same, but opted to be lazy -- I'll update the test accordingly.

PR #346, opened 2016-01-12
alias should be applied to camel-case form of argument

### LINKED ISSUE #341: camel casing problems with .default(..
If i pass an option like `--this-option` yargs auto creates the camel cased version so the argv object includes `this-option` and `thisOption`. 

On the other hand if i specify .default('this-option', 'value') and i don't pass any `--this-option` arg then the `argv` does not contain camelCased version with default value. Only `this-option` is present

bcoe:
fixes #341


nexdrew:
LGTM :+1:

PR #350, opened 2016-01-15
Make yargs use English locales in tests

maxrimue:
This fixes a bug where the unit tests were expecting yargs' output to be in English, which means that tests, checking for equality with prewritten outputs in English, will fail on machines where yargs will detect and use another locale than English.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   },
   "scripts": {
     "pretest": "standard",
-    "test": "nyc --cache mocha --timeout=4000 --check-leaks",
+    "test": "nyc --cache mocha --require ./test/before.js --timeout=4000 --check-leaks",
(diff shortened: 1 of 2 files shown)


lrlna:
The most fascinating bug, yo! 

Glad you noticed this ^.^

package.json
```diff
@@ -33,7 +33,7 @@
   },
   "scripts": {
     "pretest": "standard",
-    "test": "nyc --cache mocha --timeout=4000 --check-leaks",
+    "test": "nyc --cache mocha --require ./test/before.js --timeout=4000 --check-leaks",
```
bcoe: hooray, \o/ this does the trick for you?

maxrimue: @bcoe Yep, all tests passing :+1:

PR #354, opened 2016-01-22
Remove changelog from published NPM package

dthree:
I'm thinking about using Yargs in a [super exciting project](https://www.npmjs.com/package/cash), but due to the nature of the project, performance and package size is a big issue.

Yargs unnecessarily publishes a 44kb changelog - *very* important on the Github repo, but not so much in a published module dependency.

Can we remove this please? You'll make many projects that much snappier. :smiley:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,0 @@
-.idea/
-node_modules/
-*.swp
-test.js
-coverage
-.nyc_output
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,2 @@
+CHANGELOG.md
+
--- a/.npmignore
+++ b/.npmignore
@@ -1,2 +1,7 @@
 CHANGELOG.md
-
+.idea/
+node_modules/
+*.swp
+test.js
+coverage
+.nyc_output


bcoe:
@dthree I was thinking about this the other day, you should probably actually copy the entire `.gitignore` into the `.npmignore` -- `.npmignore` takes precedence, it doesn't combine the two files as far as I know. :+1:

dthree:
Oh really? Didn't know that. Okay one sec! :+1:

dthree:
A good to go now :+1:

bcoe:
@dthree oh my goodness, cash is a really cool idea!

bcoe:
@dthree I'm excited to see the other features that bubble up from your work.

dthree:
Hey thanks - you're awesome and I love your work! :fire:

nexdrew:
So, small problem with this change (a quote from https://docs.npmjs.com/misc/developers):

> The following paths and files are never ignored, so adding them to .npmignore is pointless:
>
> `package.json`
> `README` (and its variants)
> `CHANGELOG` (and its variants)
> `LICENSE` / `LICENCE`

I can confirm this with `npm pack .` against the current `master` branch that includes these changes.

So I think we should perhaps change the filename if we want to exclude it.

bcoe:
I'm not super keen to change the name of `CHANGELOG`, since this is a defacto standard -- at npm I've been known to truncate files as part of a `prepublish` step ... but this feels slight weird.

dthree:
Oh wow. Yeah not a big deal, don't worry about it. :smiley:

bcoe:
@dthree I ran into this same thing at npm, where I wanted to remove a README.md from a module on publish, because it potentially had secrets in it -- there's not a great workflow for this! almost feels like maybe `.npmignore` should be ground truth, but I imagine there are good reasons for the implementation as it is.

dthree:
Yeah, totally understand.

Issue #352, opened 2016-01-18
make yargs configurable


bcoe:
It would be nice to be able to enable and disable some of yargs' more controversial parsing functions, e.g., `--no`, dot-notation.

There are several tickets that request the ability to do so:

https://github.com/bcoe/yargs/issues/345
https://github.com/bcoe/yargs/issues/349

I propose that we make the yargs parser configurable:

* I appreciate how AVA simply loads its configuration from `package.json`, I'd prefer that we start with an approach like this rather than pulling in a library like `rc`.
* the default configuration would be all current parsing options turned on, an individual could then disable features such as `dot-notation` by adding:

```json
{
  "yargs": {
    "dot-notation": false
  }
}
```

btd:
In your case it will make configuration global (the same way you can use environment variables). 
I am :+1: but probably better to make this configurable also per option also.

maxrimue:
@bcoe Should we accept configuration only via `package.json` or just add it to the possibilities of passing configuration? Also, how should yargs know whether to parse the `package.json`'s content or not if it's not the only option, maybe via `.config()`?

lrlna:
I like the idea of having it go with `.config()`; in my mind everything pertaining to my application will be inside my application, including my configurations. 

However, I do feel like that's pretty heavy for the applications users will be building. That's almost like configuring the application every time yargs does it's usual parsing magic. 

I am assuming the way we want `package.json` to work is install only certain parts of yargs (i.e. with disabled `dot-notation`, etc). Meaning the user's application doesn't worry about dealing with configuration, or does that still happen at run-time?

bcoe:
@btd, @kerryChen95 could you give our `4.x` alpha branch a shot, and see if it does the trick for you:

`npm i yargs@alpha --save`

You can now add a stanza to your package.json to disable dot notation (and other parsing features):

```json
{
  "yargs": {
     "dot-notation": false
  }
}
```

See https://github.com/yargs/yargs-parser#configuration for details.

PR #356, opened 2016-01-25
split out parser into its own well tested module

bcoe:
I've split out the parser into a [stand-alone module](https://github.com/yargs/yargs-parser) that handles the configuration bits for us:

See: #352 

You can turn on and off parsing features by using the `yargs` stanza in package.json:

```json
{
  "name": "my-module",
  "yargs": {
     "camel-case-expansion": false
  }
}
```

@nexdrew addressing your comments in https://github.com/yargs/yargs-parser/pull/1, `yargs-parser` allows you to configure the `cwd` which is the approach I was thinking of for loading configuration appropriately.

I'd love to really dial in how the heck we handle loading configuration appropriately in the various contexts:

* testing `yargs-parser` and `yargs` with `npm link`.
* running libraries locally that have `yargs` or `yargs-parser` as a dependency.
* running libraries that have been installed globally with `-g`.

We are trying to solve similar problems with loading a default version from `package.json`, I wonder if there's room for a clever module here -- I bet we can extend on some of the work Sindresorhus has done (his use-case is slightly different since he can be confident AVA is installed as a development dependency, and is run in the context of a directory, i.e., is not installed globally).

Reviewers: @nexdrew, @lrlna


bcoe:
I'm going to go ahead and merge this, @nexdrew feel free to make some edits to the documentation in a future pull request.

I've fixed the issues making Windows tests flaky in another pull request.

Issue #174, opened 2015-06-14
Require a valid command


DominicBarnard:
I've been trying to figure out a clean way to require that a valid command argument has been supplied.
Currently I'm using the `check()` method like this:
```js
    .command('command1', 'A command' ,function(){
        // handler...
    })
    .command('command2', 'Another command' ,function(){
        // handler...
    })
    .demand(1, 'must provide a valid command')
    .check(function(argv, opts){
        if(!argv._[0].match(/command1|command2/)) {
            throw new Error('must provide a valid command');
        }
    })
```
This works just fine except any new command that is added in future also has to be remembered to be added to the regex.
I think it would be great to have a cleaner way to specify this behaviour.  Maybe something like a  `requireCommand` method or some option that can be passed to the `strict` method like:
```js
.strict({requireCommand: true})
```

bcoe:
@DominicBarnard I really like this idea. Would you like to take a crack at adding this, I could also look at working on it later in the week.

DominicBarnard:
@bcoe Sure. I'll try to make time for it this week.

mahnunchik:
+1

vjpr:
+1

paazmaya:
:+1: with :grapes:

jkwuc89:
Would like to see this enhancement added as well. In the meantime, @DominicBarnard suggestion above using .check is a decent workaround.

nfantone:
+1 on this. Nice to have.

bcoe:
@DominicBarnard @nfantone @jkwuc89 @paazmaya see:

https://github.com/bcoe/yargs/pull/357
https://github.com/bcoe/yargs/pull/368

This will be landing in `yargs 4`.

bcoe:
@DominicBarnard, @nfantone, @jkwuc89, yargs 4.x now has support for _strict arguments:_

https://github.com/bcoe/yargs/blob/4.x/test/validation.js#L73

Give it a spin by running:

`npm i yargs@next`

PR #357, opened 2016-01-26
require valid command

lrlna:
Check if `argv` provided is in `commandHandlers` if using `.strict()`.

A WIP pull for #174

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -108,7 +108,6 @@ module.exports = function (yargs, usage, y18n) {
       if (key !== '$0' && key !== '_' &&
         !descriptions.hasOwnProperty(key) &&
         !demanded.hasOwnProperty(key) &&
-        !commandKeys.hasOwnProperty(key) &&
         !aliasLookup.hasOwnProperty(key)) {
         unknown.push(key)
       }
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -114,6 +114,12 @@ module.exports = function (yargs, usage, y18n) {
       }
     })
 
+    argv._.forEach(function (key) {
+      if (!commandKeys.hasOwnProperty(key)) {
+        unknown.push(key)
+      }
+    })
+
     if (unknown.length > 0) {
       usage.fail(__n(
         'Unknown argument: %s',
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -95,6 +95,7 @@ module.exports = function (yargs, usage, y18n) {
     var aliasLookup = {}
     var descriptions = usage.getDescriptions()
     var demanded = yargs.getDemanded()
+    var commandKeys = Object.keys(yargs.getCommandHandlers())
     var unknown = []
 
     Object.keys(aliases).forEach(function (key) {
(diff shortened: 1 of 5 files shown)


lrlna:
I'll open this up again tomorrow when I get the tests to pass :sweat_smile: :palm_tree: Will be better that way, I think :100:

maxrimue:
The tests fail because the parsed command does not end up in the `argv` object, but in the `argv._` array, so `self.unknownArguments()` in `validation.js` does not check against it. Alternatively, you could create another check loop like this:
```javascript
argv._.forEach(function (key) {
  if (!commandKeys.hasOwnProperty(key)) {
    unknown.push(key)
  }
})
```
For me, it got the old tests and your new one passing, although I'm not sure if there's a better way to solve it, it's just what came to my mind right now.

lib/validation.js
```diff
@@ -107,11 +108,18 @@ module.exports = function (yargs, usage, y18n) {
       if (key !== '$0' && key !== '_' &&
         !descriptions.hasOwnProperty(key) &&
         !demanded.hasOwnProperty(key) &&
+        !commandKeys.hasOwnProperty(key) &&
```
maxrimue: Great work, but isn't this line unnecessary now?

lrlna: I concur, it definitely is.

test/validation.js
```diff
@@ -70,6 +70,23 @@ describe('validation tests', function () {
         .argv
     })
 
+    it('fails with invalid command', function (done) {
```
bcoe: this is wonderful \o/

lrlna: yay! :tada:

lrlna:
@bcoe I think the next thing to implement within this scope would be the `.demand(1, ['name', 'age'])`. What do you think? Or should we merge this one step at a time?

lib/validation.js
```diff
@@ -112,6 +113,12 @@ module.exports = function (yargs, usage, y18n) {
       }
     })
 
+    argv._.forEach(function (key) {
+      if (!commandKeys.hasOwnProperty(key)) {
+        unknown.push(key)
```
bcoe: I really appreciate that this implementation means that we don't need to go back and get new translations from people. Great work.

bcoe:
@lrlna I agree, let's make it so that it's possible to specify a count for arguments, and a set of known arguments at the same time. `4.x` is coming along really well :)

lrlna:
@bcoe coolest of beans; will get on that this weekend :ok_hand:

lrlna:
Let me know what you folks think about this: I am thinking it may be a little more elegant doing something other than checking if max is an array. However, this way we are not committing any breaking changes (I don't think?).

test/usage.js
```diff
@@ -58,6 +81,28 @@ describe('usage tests', function () {
           r.logs.should.have.length(0)
           r.exit.should.be.ok
         })
+        it('fails if one command and an argument not on the list are provided', function () {
```
bcoe: could you add two additional test cases:

* one that verifies that it fails if all the required arguments exist, but not enough commands are provided?
* one that verifies that no failure occurs if the required arguments and the required number of commands are provided.

bcoe:
This is 99% of the way there, added a note about two tests I'd love to see -- but otherwise this is looking wonderful.

test/usage.js
lrlna: will do! :100:

lrlna:
@bcoe i got the usage tests :heavy_check_mark: ; still gotta got the validation tests in that we talked about.

bcoe:
@lrlna looking great :+1:

bcoe:
Once the Windows tests pass, I say :shipit:

Issue #363, opened 2016-02-02
Separate command config from command execution


nexdrew:
Allow the configuration/intentions of a command (i.e. its options and usage text) to be defined upfront in a declarative way, separately from the handler that executes when the command is called.

The semantics should work for commands defined within the same module/file (perhaps an object structure or chainable config methods) or for commands defined across different submodules (conforming to expected `exports` or API).

Would be nice to support a convention-over-configuration model for submodules, if possible. Meaning something like auto-discovering command submodules in a configured directory or something.

This probably implies the need for "global" or context-specific options, i.e. options that apply to nested levels, where "global" just means the top level. For starters, this would make sense for `help()` and `version()` options but could also apply to other options shared across commands or nested contexts.

See related #121, #202, #225 (lots of comments), #288

bcoe:
@nexdrew @iarna @maxrimue, okay I think I've got an idea for how to make commands better. How do do you feel about this syntax:

```js
#!/usr/bin/env node

require('./')
  .version()
  .command(
    'push <source> [target]',
    function (yargs) {
      yargs
        .option('name', {
          describe: 'a username'
        })
        .option('password', {
          describe: 'a password'
        })
    },
    function (argv) {
      // this will get called if, help
      // version, validation, etc, passed.
    })
  }
  .version()
  .alias('v', 'version')
  .help()
  .alias('h', 'help')
  .globals(['help', 'version'])
  .argv
```

* rather than getting `yargs` passed in as an argument to the command handling function, you provide two functions:
  * the first function builds the yargs object for the command.
  * the second function will get executed with a parsed argv: as long as `--version`, `--help`, etc hasn't forced an early exist.
* we'll introduce the `globals` feature, which allows you to specify that certain arguments should remain across the commands, e.g., I can call `foo.js push --help`.
* we'll introduce the concept of positional arguments, as @nexdrew suggested.

lrlna:
I like it, @bcoe. It seems to fix a bunch of issues folks have been having with commands. 

We probably will have to do some education about this change though (as I guess with everything under 4.x branch); I don't want people panicking about commands, cause that's my favourite feature :tada: :dancers: 

Aside from that, I am not quite clear on the use of globals (sorry if you've talked about this previously!). Is `.command()` the only place we are planning to use this, or is this a more prominent feature? 

Will look through your #368 next :smile:

nexdrew:
[...]

Issue #362, opened 2016-02-02
Populate argv with named positional arguments from command string


nexdrew:
Related to #234 and #333.

I think this is something `commander` got right and we should model similar support in `yargs` (or `yargs-parser`).

Basically, map positional arguments to option names, translating arguments in angle brackets as required (e.g. `<requiredArg>`) and arguments in square brackets as optional (e.g. `[optionalArg]`).

Something similar to:

```js
var argv = require('yargs')
  .command('push <remote> [branch]', 'Push a local branch to a remote destination', function (yargs, argv) {
    console.log('remote:', argv.remote, '\nbranch:', argv.branch)
  })
  .argv
```

```
$ my-program push origin master
remote: origin
branch: master

$ my-program push origin
remote: origin
branch: undefined

$ my-program push
Commands:
  push <remote> [branch]  Push a local branch to a remote destination

Missing required argument: remote
```

It seems that supporting named positional arguments is really just shorthand DSL for minimal option configuration (besides the flagless parsing), i.e. `<requiredArg>` is equivalent to `yargs.demand('requiredArg')`.

Some implementation details to consider:

1. Does it only apply to `command()`, or does it also apply to `usage()`? We could probably support it in `command()` without a breaking change (3.x), but supporting it in `usage()` would definitely be a breaking change (4.x).
2. Do we allow optional positional arguments to be precede required positional arguments (e.g. `push [optional] <required>`) or even allow them to be interspersed (e.g. `push <required1> [optional1] [optional2] <required2>`), and is this feasible for parsing?
3. Do we allow standard `option()` configuration to apply to named positional arguments, for declaring things like `type` and `default`? And fallback to standard boolean/number/string parsing if no key-based config is given? How does this affect usage/help content?
4. Do we also support variadic arguments like `commander` (e.g. `push <remote> [branch...]`)? Perhaps this is just shorthand for explicitly configuring the option as `type: 'array'`?
5. If we support standard `option()` configuration for named positional arguments, do we also support them as flagged options as well, e.g. such that `push upstream` and `push --remote upstream` are equivalent?

nexdrew:
This could possibly make it easier to implement #201, though we'd probably need to add a `path` option type (or maybe `file` and `directory` option types).

bcoe:
@nexdrew addressed in https://github.com/bcoe/yargs/pull/368, in this first pass:

[...]

PR #368, opened 2016-02-07
working on a new API for commands that creates less redundant code

bcoe:
I'm working on a new API for commands that makes writing them less redundant:

* commands now accept two functions a `builder` for describing command-specific help, and a `handler` that will get argv passed to it.
* the concept of `global` options is being introduced, to eliminate the need for setting the same option over and over again on each command.
* positional argument parsing: `foo <foo> [bar]` is being introduced.

- [x] **Builder/Handler Pattern.**

The definition of commands has now been split into a:

* `builder`: which is used to describe a command.
* `handler`: which is used to handle a command when it is invoked.

```js
var argv = require('yargs')
  .command(
    'push', 
    'push to a URL', 
    function (yargs) {
      // build command specific yargs object.
    },
    function (argv) {
      // handle the parsed options.
   })
   .help('help')
  .argv
```

- [x] **Providing a Module For Commands**

I would like to make it so that you could also provide a module to `command`, you would write a module like this:

```js
exports.builder = function (yargs) {
  return yargs.option('f', {})
}
exports.handler = function (argv) {
// do something with argv.
}
```

And would provide it to a command like this:

```js
yargs.command('foo', 'my awesome foo command', require('./test-module')).
```

I think this is pretty killer, and splitting out the builder and handler makes this possible what do you think @nexdrew?

- [x] **Global Options**

You can now specify `global` options using `.global()` or the `global: true` shorthand on options.

These options will not be reset when passed to commands, by default `help`, `version`, are treated this way (since this is the behavior folks usually want).

- [x] **Parsing Positional Arguments**

see: #363, #362


test/integration.js
```diff
@@ -50,57 +51,59 @@ describe('integration tests', function () {
...
-        fs.symlinkSync(process.cwd(), './test/fixtures/yargs-symlink')
-        return done()
-      })
-    })
-
-    describe('version #', function () {
-      it('defaults to appropriate version # when yargs is installed normally', function (done) {
-        testCmd('./normal-bin.js', [ '--version' ], function (buf) {
-          buf.should.match(/9\.9\.9/)
+  if (!isWindows()) {
```
maxrimue: Do we really need a new dependency for that? Why not just use `process.platform !== 'win32'`?

wraithan:
Big fan of the new style, looking forward to taking advantage of it in a couple little tools I've written.

README.md
```diff
@@ -462,32 +464,89 @@ values stored in `argv._`).  Set `desc` to `false` to create a hidden command.
...
-command being executed (those described in the outer yargs instance).
+```js
+yargs.command('get', 'make a get HTTP request', {
+    url: {
+      default: 'http://yargs.js.org/'
+    }
+  })
+  .help()
+  .argv
+```
```
lrlna: I think it may be better to stick with the same structure of an example that we use above, i.e:

```js
var argv = require('yargs')
  .command('get', 'make a get HTTP request', {
    url: {
      default: 'http://yargs.js.org/'
    }
  })
  .help()
  .argv
```
The change in structure might confuse people (even though you mention that this is a `yargs` instance), and knowing people don't always read text and skip to code, we might run into some trouble :o What do you think?

bcoe: @lrlna I've updated the documentation for **command modules** which I think are the most elegant approach we can direct people towards.

lrlna: @bcoe :cool:, I like it.

pscanf:
@bcoe a maybe-trivial clarification on the API: the intended purpose of the `handler` function is to kick off my command logic execution, correct?

Example:
```js
#!/usr/bin/env node

import yargs from 'yargs';
import do_a from './do_a.js';
import do_b from './do_b.js';

yargs
  .command('do_a', 'Does A', {/* a options */}, do_a)
  .command('do_b', 'Does B', {/* b options */}, do_b);
```

(I ask because in the example you save in a variable the `argv` object, which I guess has little use at that point though)

pluma:
Is there any way to specify the equivalent of the last argument to `demand` (i.e. an error message) when defining commands with positional arguments this way? The default error message is pretty meaningless for mortals.

PR #374, opened 2016-02-12
Pass the original error as the second argument of fail() handler. (fixes #373)

### LINKED ISSUE #373: How to get stack trace in case of an error?
```js
#! /usr/bin/env node

import yargs from 'yargs';

yargs
    .fail((...args) => {
        console.log('args', args);
    })
    .command('lint', 'Report code format errors.', () => {
        throw new Error('test');
    })
    .argv;

```

Executing this code fails with `test` output. I am expecting `.fail()` to be called, but it is not. Is there a way to disable `yargs` error handler?

gajus:


bcoe:
@gajus this is great \o/ two small requests:

* one could you submit the pull against the `v4` branch? we're very close to releasing the next major version of yargs.
* could you add a test in `test/usage.js` for this new behavior?

gajus:
> one could you submit the pull against the v4 branch? we're very close to releasing the next major version of yargs.

@bcoe I think this requires closing this PR and opening a new one. Right?

gajus:
@bcoe Done, https://github.com/bcoe/yargs/pull/382.

Issue #367, opened 2016-02-06
calling help() without arguments should not generate the usage string


bcoe:
I think that rather than `help()` returning a usage string, calling `help()` with no arguments should instead enable `help` for the default argument `--help`.

This is more consistent with the new behavior of `--version`.

If someone wants to show help, they should call `showHelp()`.

What do you think: @nexdrew, @lrlna?

maxrimue:
Definitely a fan of this idea! :+1:

nexdrew:
Agreed, good idea :+1:

lrlna:
seems super reasonable :100:

lijunle:
I love the consistence between `.help()` method with other methods. However, please provide a way to get the help message.

In my case, the app is doing some customized validation after pass, if it fails, an error message with the generic help message is shown. After this change, I cannot get the generic help message.

Any ideas on this?

nexdrew:
@lijunle There are still a couple options, though they may be less convenient:

1. If you just want to print the help message, you can use `yargs.showHelp()` (to print to stderr) or `yargs.showHelp('log')` (to print to stdout).

2. If you want the help message as a string, you can use `yargs.getUsageInstance().help()`.

Either of these would work in, for example, a [custom fail function](http://yargs.js.org/docs/#methods-failfn).

If this isn't what you're looking for (or presents a problem with what you're trying to do), I think it's reasonable to open another issue requesting a new method that reinstates the logic of the 3.x version of `.help()` with no args, possibly named `.getHelp()` or similar. Thanks!

lijunle:
Although `yargs.getUsageInstance().help()` is long and seems using an internal API, it is working. Thanks @nexdrew

PR #376, opened 2016-02-13
Made help() only assign '--help' option if called without args

maxrimue:
This takes care of @bcoe's idea in #367

### DIFF
--- a/README.md
+++ b/README.md
@@ -733,7 +733,7 @@ when displaying usage instructions, e.g.,
 
 ```js
 var yargs = require('yargs')(['--help'])
-  .help('help')
+  .help()
   .group('batman', 'Heroes:')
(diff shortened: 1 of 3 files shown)


bcoe:
@maxrimue I think you can get rid of help entirely, and just change `addHelpOpt` to:

```js
  var helpOpt = null
  self.addHelpOpt = self.help = function (opt, msg) {
    helpOpt = opt || 'help'
    self.boolean(opt)
    self.global(opt)
    self.describe(opt, msg || usage.deferY18nLookup('Show help'))
    return self
  }
```

maxrimue:
@bcoe Good catch! Will add this + docs

bcoe:
:shipit: (once windows tests pass).

PR #382, opened 2016-02-14
Pass the original error as the second argument of fail() handler. (fixes #373)

### LINKED ISSUE #373: How to get stack trace in case of an error?

gajus:
A follow up to https://github.com/bcoe/yargs/pull/374

### DIFF
--- a/index.js
+++ b/index.js
@@ -535,7 +535,7 @@ function Argv (processArgs, cwd) {
       try {
         args = parseArgs(processArgs)
       } catch (err) {
-        usage.fail(err.message)
+        usage.fail(err.message, err)
       }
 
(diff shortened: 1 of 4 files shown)


bcoe:
@gajus awesome, thanks for the contribution :hand:

PR #385, opened 2016-02-14
update 4.x changelog

bcoe:


bcoe:
@mathieumg it will be up on http://yargs.js.org/ once this lands.

mathieumg:
Thanks! :)

PR #391, opened 2016-02-17
Do not include module commands in usage text more than once

nexdrew:
Just need to check for module command _before_ adding it to usage; otherwise the recursion adds it twice. Includes test to guard against regression.


bcoe:
that's an easy fix :) was worried what you described was something bizarre! :shipit:

PR #392, opened 2016-02-17
Do not reset wrap for commands

nexdrew:
Preserve the "global" `wrap()` value so that usage width is consistent for commands that do not wish to override it (particularly when using an options object instead of a builder function). Commands that wish to override it may still do so.

Includes a test for each scenario.


bcoe:
wonderful :shipit:

PR #393, opened 2016-02-17
Use implicit usage string when not set by command builder function

nexdrew:
I like the feature of using the original command string as the `usage()` string for commands that pass an options object instead of a builder function.

I thought it'd be cool to give commands with builder functions the same treatment as long as they don't:

1. parse argv in the builder function itself (in which case a `--help` flag would have already triggered usage output)
2. explicitly set the `usage()` string in the builder function

This is totally subjective and so is up for debate, but I think keeping the different command approaches as similar as possible is a good idea that could lead to more interesting features down the road (e.g. command suggestions, etc).

Also not sure if this "breaks" user expectations based on 3.x functionality, but hey, this is 4.x.

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -60,6 +60,13 @@ module.exports = function (yargs, usage, validation) {
       // a function can be provided, which interacts which builds
       // up a yargs chain and returns it.
       innerArgv = commandHandler.builder(yargs.reset(parsed.aliases))
+      // if the builder function did not yet parse argv with reset yargs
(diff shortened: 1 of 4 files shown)


bcoe:
@nexdrew this seems quite reasonable to me.

bcoe:
@nexdrew give `npm i yargs@next` a shot, it has your fixes.

nexdrew:
@bcoe Looks good, thanks :sunglasses:

PR #395, opened 2016-02-19
Preserve groups containing global keys

### LINKED ISSUE #336: `group` property displays option twice
When I pass in multiple options using `.options({})` and the options in the object have the `group` property, it honors the group but each option is listed twice.

novemberborn:
Heya! I used yargs in a new CLI tool the other day. This tool takes some global options as well as per-command options. I noticed I had to [restore the global option groups](https://github.com/novemberborn/wilee/blob/cdf47828ae17dd9ffd3944def85adf189205f403/src/lib/cli.js#L45) for each command, which I was not expecting.

With this PR I'm proposing that groups containing global keys are preserved within commands. Please let me know if that's a sensible thing for yargs to do!

One concern raised in #336 is the ability to control the order of the groups in each command's help output. Currently this PR puts the groups with global keys first. Possibly the `groups` object could be reordered if the group is redeclared without keys.

I also may be misunderstanding the myriad uses of `yargs.reset()`, though the tests seem to pass aside for one introduced to fix #336. I can fix those tests if people are happy with this approach.

Thanks.

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,6 +20,8 @@ function Argv (processArgs, cwd) {
   var self = {}
   var command = null
   var completion = null
+  var groups = {}
+  var preservedGroups = {}
   var usage = null
   var validation = null
   var y18n = Y18n({
@@ -69,6 +71,19 @@ function Argv (processArgs, cwd) {
       })
     })
 
+    // preserve groups containing global keys
+    preservedGroups = Object.keys(groups).reduce(function (acc, groupName) {
+      var keys = groups[groupName].filter(function (key) {
+        return key in globalLookup
+      })
+      if (keys.length > 0) {
+        acc[groupName] = keys
+      }
+      return acc
+    }, {})
+    // groups can now be reset
+    groups = {}
+
     var arrayOptions = [
       'array', 'boolean', 'string', 'requiresArg',
       'count', 'requiresArg', 'count', 'normalize', 'number'
@@ -103,7 +118,6 @@ function Argv (processArgs, cwd) {
 
     exitProcess = true
     strict = false
-    groups = {}
     completionCommand = null
     self.parsed = false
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -385,7 +396,6 @@ function Argv (processArgs, cwd) {
     return options
   }
 
-  var groups = {}
```
novemberborn: FWIW `reset()` seems to be called before this is reached. `groups` should be declared higher up regardless of this PR.

nexdrew:
@novemberborn Hey thanks! I think this is certainly a reasonable request - the groups for global options should be preserved for commands just like the options themselves are. Since the concept of global options is so new to yargs, I think we just haven't shaken out all the proper uses/bugs for them.

I think this looks good. My only concern is that, per #336 and the failing test, yargs should output the command-specific option groups _before_ it outputs the global option groups, so some reordering seems necessary. Could you possibly work that into this PR and also add a specific test for preserving groups with global options (one that actually uses `.global()` or `global: true` - unlike the failing test :smiley:)?

novemberborn:
@nexdrew cool, thanks! I've (force-)pushed a commit which addresses your concerns. By default the preserved groups are shown last, but you can add to such a group or indeed bump it up in the command builder.

test/usage.js
```diff
@@ -1542,6 +1542,117 @@ describe('usage tests', function () {
...
+        ''
+      ])
+    })
+
+    it('can bump up preserved groups', function () {
+      var r = checkUsage(function () {
+        return yargs(['upload', '-h'])
+          .command('upload', 'upload something', function (yargs) {
+            return yargs
+              .group([], 'Awesome Flags:')
```
nexdrew: This is an interesting "feature". :smile: I like the functionality, but the API is not very intuitive IMO. That being said, I can't really think of a better alternative, so :+1:.

nexdrew:
LGTM :+1:

@novemberborn Thanks for the contribution!

test/usage.js
novemberborn: Heh yea. It's a side-effect of being able to add to preserved groups but I figured it'd be worthwhile adding a test for it.

nexdrew: Without the test, I would not have spotted this. So great idea!

nexdrew:
@bcoe Any objections here?

bcoe:
@novemberborn @nexdrew I love it, will work on getting a release out tomorrow, and on moving yargs to an org.

bcoe:
@novemberborn give `npm i yargs@next` a shot, it has your fixes to `yargs-parser` and the global groups bug.

novemberborn:
@bcoe works great!

Issue #396, opened 2016-02-19
yargs-parser modifies args


novemberborn:
`yargs-parser` modifies the `args` array when parsing, notably when [`--` is separated by `=`](https://github.com/yargs/yargs-parser/blob/78f49ffde9a3305facb66b2faeae862b33d0745e/index.js#L124).

Because `args` is parsed again when a command is found this causes positional arguments within the command to be misaligned.

This behavior doesn't seem documented in `yargs-parser`, so maybe the bug is there, or maybe `yargs` should copy `args` when [calling `Parser.detailed`](https://github.com/bcoe/yargs/blob/e0d55f154084967e095816054bf5f2aec1de7412/index.js#L577).

Happy to provide a fix for either or both projects, just need to know which fix is preferred :smile:

novemberborn:
@bcoe if you're going to get a release out, this should be fixed too. It prevents positional arguments in commands from working if a preceding option uses `=` for its value.

bcoe:
@novemberborn mind adding a failing test? once I have a better idea of what's going on from this, we can talk about which project needs the fix -- I'm guessing it will probably be `yargs-parser` :+1:

novemberborn:
@bcoe failing test: https://github.com/yargs/yargs/pull/399

`yargs-parser` fix: https://github.com/yargs/yargs-parser/pull/9

PR #399, opened 2016-02-21
Parser bug testcase

novemberborn:
Test case for #396. Notice that the case with `'--foo=fOo'` fails, but the case with `'--foo', 'fOo'` passes.


bcoe:
merged into the changelog update.

Issue #403, opened 2016-03-01
how should checks behave for nested commands?


Enet4:
Considering this particular use case:

``` javascript
var argv = require('yargs')
    .command('hunt')
    .boolean('cat')
    .boolean('dog')
    .check(function(argv, opt) {
        // can't take both!
        return !(argv.cat && argv.dog);
    })
    .argv;

console.log(argv);
```

Running it with the arguments `--cat --dog` alone will trigger the check and fail to proceed, as expected. However, when running with `hunt --cat --dog`, the check function isn't called at all. One could argue that I can just specify the check inside the command's builder, but these options apply to all commands, and I wouldn't want to create builders and replicate calls to `check(fn)` to each one of them just for this purpose.

Shouldn't a `check` like this one apply to any (or no) command by default? Otherwise, can we have a means of specifying that the check is global?

maxrimue:
I can confirm this bug. Interestingly, this works with yargs v3, but not with yargs v4.

maxrimue:
I've identified the issue here as follows:
When the `index.js` asks `command.js` for which commands have handler functions ([L619](https://github.com/yargs/yargs/blob/master/index.js#L619)), `command.js` returns all registered commands ([L48](https://github.com/yargs/yargs/blob/master/lib/command.js#L48)), disregarding whether they have handlers or not, which then causes the `index.js` to just run the handler function and not perform any further checks like custom `.check()`'s ([L681](https://github.com/yargs/yargs/blob/master/index.js#L681)).

I will try to find out how to solve this quickly. Thanks for reporting!

bcoe:
@maxrimue @Enet4 with the new approach to command handling here is what my expectation would be:

```js
var argv = require('yargs')
    .command('hunt', 'hunt command', function (yargs) {
      return yargs
        .check(check)
    })
    .boolean('cat')
    .boolean('dog')
    .check(check)
    .argv;

function check () {
  // can't take both!
  return !(argv.cat && argv.dog);
}

console.log(argv);
```

This new command API has the advantage that it's explicit. If we did want different behavior for `hunt` we'd simply not apply the `check`.

I could see an argument being made for `check` being made global ... but I think this might be a bit presumptuous do you have any thoughts @nexdrew?

bcoe:
@jamesr73 @maxrimue let's use this thread to hash out how `check` should behave when nesting commands: does it default to being global? perhaps it's overridden by inner checks?

nexdrew:
[...]

PR #408, opened 2016-03-03
Fix unintentional command execution

maxrimue:
This PR aims at fixing yargs unintentionally running commands that do not have associated handlers, and thus skipping, for example, custom checks, as raised in #403.

Originally, in v3, yargs only executed command functions if available, which means code like this would work:
```javascript
var argv = yargs(['command', '--option', 'lol'])
  .command('command')
  .string('option')
  .check(argv => argv.option !== 'lol')
  .argv
```
^ this should output an error that custom checks weren't met.

However, as v4 introduced a major overhaul of the command API and changed things a bit, that code does not work anymore in said version.

The problem is that yargs attempts to execute handlers of commands that don't exist, because it thinks they do. This, however, does not cause an error, as `command.js`'s if-behaviour does not throw if no handler is available. As a result, yargs finally does not do things like custom checks anymore.

__This PR doesn't actually yet fix things, but is meant to show the problem, and a potential approach to fix it.__

### DIFF
--- a/index.js
+++ b/index.js
@@ -616,7 +616,7 @@ function Argv (processArgs, cwd) {
 
     // if there's a handler associated with a
     // command defer processing to it.
-    var handlerKeys = command.getCommands()
+    var handlerKeys = command.getCommandsWithHandlers()
     for (var i = 0, cmd; (cmd = handlerKeys[i]) !== undefined; i++) {
       if (~argv._.indexOf(cmd) && cmd !== completionCommand) {
         setPlaceholderKeys(argv)
--- a/index.js
+++ b/index.js
@@ -616,11 +616,12 @@ function Argv (processArgs, cwd) {
 
     // if there's a handler associated with a
     // command defer processing to it.
-    var handlerKeys = command.getCommandsWithHandlers()
+    var handlerKeys = command.getCommands()
     for (var i = 0, cmd; (cmd = handlerKeys[i]) !== undefined; i++) {
       if (~argv._.indexOf(cmd) && cmd !== completionCommand) {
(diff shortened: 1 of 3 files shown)


bcoe:
@maxrimue good work :+1: mind rebasing with:

https://github.com/yargs/yargs/pull/413

so that we can get tests running, and I'll give more thorough code review this weekend?

maxrimue:
@bcoe Rebased, however tests will still fail because this PR is not yet solving everything

jamesr73:
How would/should the following use case behave?

I'm interested in setting up multiple commands that all share some common args and add their own on top. I'd like one check common to the global args and possibly individual checks for each command.

```javascript
var argv = require('yargs')
  .options({
      glo: { description: 'Global 1' },
      bal: { description: 'Global 2' }
  })
  .global(['glo', 'bal'])
  .check(function (argv) {
    // check that global args are exclusive
    return (argv.glo && argv.bal);
  })
  .command('bar', 'fuzz', function (yargs) {
    return yargs
      .options({
        com: { description: 'Command 1' },
        mand: { description: 'Command 2' }
      })
      .check(function (argv) {
        // check that command args are exclusive
        return !(argv.com && argv.mand);
      })
  })
  .argv;

console.log(argv);
```

bcoe:
@maxrimue where do we stand on this pull request, I'd like to work on a release of yargs today.

bcoe:
@maxrimue I'm going to put some detailed comments on #403 -- I don't think this is the correct solution to the problem -- it brings us closer to how commands worked in `3.x`, but I think this was too confusing.

Issue #404, opened 2016-03-03
Add translations for type 'number'


maxrimue:
As of v4, [yargs now supports the option type `number`](https://github.com/yargs/yargs/pull/384).
This new type also needs translations to other languages.

This issue is mainly a central list for keeping track of the status for each translation. Because of previous locale contributions, I've also mentioned some people here to ask them for help, however, contributions can of course be done by anyone for any language.

Here's a list of needed translations:
- [x] German (@maxrimue) (#405) 
- [x] Spanish (@zkat) (#417)
- [x] French (@zkat) (#420)
- [x] Indonesian (@rilut) (#425)
- [x] Japanese (@watilde) (#422)
- [x] Korean (@disjukr) (#410)
- [x] Norwegian (@sindresorhus) (fba25734b41bab7fa9c1bf6cee64d5fff83e75a2)
- [x] Polish (@kamilogorek) (#406) 
- [x] Portuguese (@zkat) (#419)
- [x] Portuguese (Brazil) (@zkat) (#418)
- [x] Turkish (@feyzo) (#407)
- [x] Chinese (@qiu8310) (#411) 

#### How to contribute

1. Search for the file for your language in `/locales/` (__If it doesn't exist yet, you're of course also welcome to create one for your language! In this case, use the `en.json` as a template__)
2. Add this line _right after line 7_ (for consistency):
`"number": "<your translation>",`
3. Submit a pull request

Every contribution is very welcome!

kamilogorek:
Polish: https://github.com/yargs/yargs/pull/406

oti:
@watilde thank you!

maxrimue:
Thanks everyone!

PR #411, opened 2016-03-04
Add chinese translations for new type 'number'

qiu8310:
#404


nexdrew:
@qiu8310 Awesome, thanks! :+1:

PR #413, opened 2016-03-04
tweak appveyor config to make tests stop failing

bcoe:


nexdrew:
LGTM. Can we also take the opportunity to stop testing with iojs and start testing with latest?

nexdrew:
:+1:

PR #414, opened 2016-03-04
Allow command modules to define all command properties

nexdrew:
This just expands support for command modules such that the command string and description can also be defined by the module, like so:

```js
// foo.js
module.exports = {
  command: 'foo <bar> [baz]',
  describe: 'foo all the things',
  builder: function (yargs) { return yargs },
  handler: function (argv) {}
}
```

```js
// cli.js
var argv = require('yargs')
  .command(require('./foo.js'))
  .argv
```

A command string must be provided by the `command` property (and it must be a string).

A description is optional. If not provided, the command will be "hidden". If provided, it can be a string or a boolean, and for consistency with the options API it can be provided by a `describe`, a `description`, or a `desc` property (in that order).

I see this as an incremental step toward allowing command modules to be more self-sufficient and auto-discovered.

Note that I added a bunch of "API" tests to `test/command.js` to account for what the `.command()` method should accept as args, but I'm not sure they provide much value. But I figure, why not.

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -6,7 +6,13 @@ module.exports = function (yargs, usage, validation) {
 
   var handlers = {}
   self.addHandler = function (cmd, description, builder, handler) {
-    // allow a module to be provided for a command.
+    // allow a module to define all properties
+    if (typeof cmd === 'object' && typeof cmd.command === 'string' && cmd.builder && typeof cmd.handler === 'function') {
+      self.addHandler(cmd.command, extractDesc(cmd), cmd.builder, cmd.handler)
+      return
+    }
+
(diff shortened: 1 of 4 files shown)


nexdrew:
CI failures seem to be red herrings.

`npm install` failed on Travis, but only for Linux Node 0.12.

AppVeyor failed due to npm 1.x with Node 0.10, which will be fixed by #413.

bcoe:
@nexdrew mind rebasing with:

https://github.com/yargs/yargs/pull/413

should help some of the CI weirdness.

nexdrew:
Rebased

PR #415, opened 2016-03-04
Update Node versions in Travis config

nexdrew:
Replace iojs with latest Node version.

For now, this will increase the range to Node 5 while preserving support back to Node 0.10.

This should help future-proof our testing a bit when new Node versions are released.


nexdrew:
Covered by #413

PR #425, opened 2016-03-09
Add "number" translation for Indonesian

rilut:
As per #404


nexdrew:
@rilut Awesome, thank you!

Issue #426, opened 2016-03-10
Using strict() within the command's options


alinex:
I tried to use a command with it's own additional options (coffee script):

    builder: (yargs) ->
      job = 'test'
      yargs
      .demand 1
      .usage "\nUsage: $0 <job> [options]"
      .option 'xtest',
        alias: 'x'
        type: 'string'
      .group 'x', "#{string.ucFirst job} specific:"
      # help
      .help 'h'
      .alias 'h', 'help'
      .epilogue "For more information, look into the man page."
      #.strict()
    handler: (args) ->
      console.log 'test'
      console.log args

It works fine, the help page looks good but if I use the `strict()` here I get:

    Unknown argument: test

Why is it having a problem with the command name `test` here as an argument?

alinex:
As JavaScript the command module would look like:

    module.exports = {
      builder: function(yargs) {
        var job;
        job = 'test';
        return yargs.demand(1).usage("\nUsage: $0 <job> [options]")
        .option('xtest', {
          alias: 'x',
          type: 'string'
        })
        .group('x', (string.ucFirst(job)) + " specific:")
        .help('h').alias('h', 'help')
        .epilogue("For more information, look into the man page.");
      },
      handler: function(args) {
        console.log('test');
        return console.log(args);
      }
    }

nexdrew:
@alinex I can confirm this bug. The "unknown arguments" validation check induced by `.strict()` is not context- or command-aware, so it's attempting to validate the full `argv._` array against reset command keys.

I think we need to introduce some concept of command context (e.g. levels of nesting, what command is currently being executed, etc) to fix this and to better support nested commands/subcommands. The logic to "reset" things is already there, so we can probably use that to drive this; we just need to add some mechanism to keep track of the resets, I think.

I'll see if I can put together a quick PR to get this ball rolling. Thanks for reporting this!

nexdrew:
This was fixed in [4.4.0](https://github.com/yargs/yargs/blob/master/CHANGELOG.md#v440-20160403-2110-0700). 👍

Issue #353, opened 2016-01-19
offset demand of a subcommand


gajus:
I have an entry file:

```js
#! /usr/bin/env node

import yargs, {
    argv
} from 'yargs';

import {
    fixCommand,
    lintCommand
} from './commands';

yargs
    .help('help')
    .strict()
    .command('fix', 'Fix code format.', fixCommand)
    .command('lint', 'Report code format errors.', lintCommand)
    .argv;
```

My `lintCommand.js` looks like this:

```js
export default (yargs) => {
    let argv;

    argv = yargs
        .help('help')
        .strict()
        .options({
            stdin: {
                default: false,
                describe: 'Used to indicate that subject body will be read from stdin.',
                type: 'boolean'
            }
        })
        .argv;

    if (argv.stdin) {
        // Require no argv._ values.
        yargs
            .demand(0, 0)
            .argv;
    } else {
        // Require at least one argv._ value.
        yargs
            .demand(1)
            .argv;
    }
};

```

When I execute the program such as:

```sh
node ./index.js lint --stdin
```

I get an error:

> Too many non-option arguments: got 1, maximum of 0

The reason for this is that `argv._` value is `[ 'lint' ]`.

How do I remove the command value from `argv._`?

nexdrew:
@gajus Interesting problem. Does something like this work for `lintCommand.js`?

```js
export default (yargs, parentArgv) => {
    yargs
        .help('help')
        .strict()
        .options({
            stdin: {
                default: false,
                describe: 'Used to indicate that subject body will be read from stdin.',
                type: 'boolean'
            }
        });

    if (parentArgv.stdin) {
        // Require no argv._ values.
        yargs.demand(0, 0);
    } else {
        // Require at least one argv._ value.
        yargs.demand(1);
    }

    let argv = yargs.argv;
};
```

gajus:
How would that help? `argv._` is still `['lint']`.

gajus:
Here is how I did it:

```js
import yargs from 'yargs';

export default () => {
    let argv;

    argv = yargs(process.argv.splice(3))
        .help('help')
        .strict()
        .argv;

    console.log('argv._', argv._);
};
```

```sh
node ./index.js lint --stdin foo
> argv._ [ 'best' ]
```

bcoe:
@gajus we've been undertaking a big overhaul of commands. We've introduced the concept of required positional arguments, which I think will solve this problem for you partially:

https://github.com/bcoe/yargs/pull/368

[...]

PR #428, opened 2016-03-11
Introduce command execution context

nexdrew:
This is something I've wanted to do for a while but never had a good grasp on how to implement it. (And, I arguably still don't. :grinning:)

However, issue #426 brought up a good first use-case for maintaining a context of command execution, in order to change the behavior of "unknown argument" validation (induced by `.strict()`) based on if it's running for a (possibly nested) command or not.

As is, this branch fixes that problem - without breaking any existing tests - but is still lacking tests for the scenarios it addresses.

The reason I'm opening this PR now is that I'd like some feedback on the approach/implementation before I put much more effort into it. Basically, I'm adding another piece of state (tied to the yargs singleton) that does **not** get reset for command execution, but rather keeps track of which commands are currently being executed. This `context` object is available to all submodule logic (via the `.getContext()` method), intended to be **modified** at state transitions and **read** for processing logic that is dependent on the current command context. I realize that exposing it in this way leaves it vulnerable to unexpected manipulation (even from user code), but it seems more useful/flexible this way and is consistent with other "internal" APIs as well.

Note that the `context` initially has two fields: a `commands` array representing the currently executing commands (supports nested subcommands) and a `resets` integer representing the total number of times a reset has occurred. I thought the number of resets would be useful, but I ended up not using it since, for command processing, I can rely on the length of the commands array instead. I fully expect additional fields to be added to the `context` in the future, if the need arises for additional state not currently represented.

Also note that this branch already uses the `context` in two different ways - once for validation logic and once for correcting the implicit usage string for subcommands. I also commented a way it might be useful for interpreting `.demand()` for a command, but I shied away from this b/c I honestly don't understand the intent/expectation of demanding a certain number of args within a (sub)command.

At any rate, please let me know your thoughts. I pushed the branch to the main repo so it is available for easy fetching/modification by yargs members.

### DIFF
--- a/yargs.js
+++ b/yargs.js
@@ -47,11 +47,19 @@ function Yargs (processArgs, cwd, parentRequire) {
     )
   }
 
+  // use context object to keep track of resets, subcommand execution, etc
+  // submodules should modify and check the state of context as necessary
+  var context = { resets: -1, commands: [] }
+  self.getContext = function () {
+    return context
+  }
+
   // puts yargs back into an initial state. any keys
   // that have been set to "global" will not be reset
   // by this action.
   var options
   self.resetOptions = self.reset = function (aliases) {
+    context.resets++
     aliases = aliases || {}
     options = options || {}
     // put yargs back into an initial state, this
--- a/lib/command.js
+++ b/lib/command.js
@@ -70,6 +70,9 @@ module.exports = function (yargs, usage, validation) {
     var argv = parsed.argv
     var commandHandler = handlers[command]
     var innerArgv = argv
+    var currentContext = yargs.getContext()
+    var parentCommands = currentContext.commands.slice()
+    currentContext.commands.push(command)
     if (commandHandler.builder && typeof commandHandler.builder === 'function') {
       // a function can be provided, which interacts which builds
       // up a yargs chain and returns it.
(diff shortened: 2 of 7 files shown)


maxrimue:
I really like the idea of this PR and the code looks good so far too :+1:

What's your idea on how `.demand()` could be used with the addition of context?

nexdrew:
@maxrimue Well, when someone does `.demand(1)` from within a command builder function (which I've seen a few times in issues), are they trying to demand one _total_ flagless args or one _additional_ flagless arg? I mean, the command that the builder belongs to will not be processed/executed if at least one flagless arg was not already given (representing the command itself).

So what I'm speculating is that they really mean "when executing this command, require one additional flagless arg", which becomes a context-dependent interpretation of `.demand()`, such that we should probably _increment_ the total number of flagless args demanded by the value specified (e.g. 1 to represent the command and 1 more to represent the arg needed by the command handler) rather than _set_ the total number of flagless args demanded to the value specified. This idea is currently represented by the commented code here: https://github.com/yargs/yargs/blob/command-context/index.js#L254

Otherwise, it doesn't really make sense to `.demand(1)` inside a command builder, because we already know one flagless arg was given.

nexdrew:
Concerning `.demand(1)` within a command builder, perhaps it's just unclear to users that the builder function is not executed until the command is given on the command line?

For an example of my confusion, see issue #353 (which was based on the 3.x API).

Either way, I think there's a mismatch between user expectations and yargs interpretation of command context.

nexdrew:
Now with tests for behavior that was modified to be context-aware! This PR represents initial support for subcommands (commands within commands) in yargs, which includes the following 3 areas:

1. Positional arg population now supports subcommands
2. Strict mode (unknown argument validation) now supports subcommands
3. Default usage text now supports subcommands

There may be other areas not yet addressed (e.g. should `.demand()` be contextual per command?), but these were the major ones I found in testing, and I think this is a good first step.

PR should be ready for final review.

maxrimue:
What does everybody think about whether or not to include context for `.demand()`?

I think including context for `.demand()` makes sense and is probably what people would expect when demanding inside a command.

yargs.js
```diff
@@ -241,7 +249,7 @@ function Yargs (processArgs, cwd, parentRequire) {
 
     if (typeof keys === 'number') {
       if (!options.demanded._) options.demanded._ = { count: 0, msg: null, max: max }
-      options.demanded._.count = keys
+      options.demanded._.count = keys // + context.commands.length
```
bcoe: this comment is accurate? assume it's here to explain the value of `options.demanded._.count`?

divmain:
@maxrimue, I've been following this pull request because I ran into the same questions.  I'd definitely be in favor of contextual `.demand()`.  As a user, it is the behavior that I would expect.  Thanks for the work on this @nexdrew!

bcoe:
@maxrimue I don't object to this one bit, but I think this should be a separate well thought out pull request. This pull already introduces quite a bit of new logic, and I'd rather keep new demand logic separate.

bcoe:
@maxrimue @divmain If there's not already an issue open on the subject, start one going and let's think about what changes to the demand API should be made.

divmain:
Looks like this is covered in #353.

yargs.js
nexdrew: The comment is meant to show what yargs should possibly be doing to make `.demand(num)` contextual. This by itself may not be sufficient though (e.g. error message handling). I can remove it from this PR.

bcoe:
@nexdrew I say :shipit:

bcoe:
@nexdrew give `npm i yargs@next` a shot, I'll promote the release early this week.

PR #432, opened 2016-03-12
run non-singleton version of yargs

### LINKED ISSUE #424: strange interaction when multiple libraries are using the same yargs module
see: https://github.com/bcoe/nyc/issues/187

bcoe:
fixes #424 

Because `require('yargs')` is a singleton, it can cause strange interactions when multiple libraries use the same version of yargs. We've been seeing this problem crop up a few times in `nyc`.

Without breaking the core API, this change presents the option of loading a non-singleton version of yargs;

singleton: `require('yargs');`.
non-singleton: `require('yargs/yargs');`.


bcoe:
@novemberborn mind providing a second set of eyes on this?

elas7:
@bcoe could it be that line 12 in index.js is duplicated from yargs.js?

````javascript
  processArgs = processArgs || [] // handle calling yargs().
````

As it appears after the call to yargs()

PR #434, opened 2016-03-12
remove duplicate line from index.js

bcoe:
thanks @elas7 \o/

If you have a few minutes, try `npm i yargs@next` I want to make sure I haven't broken anything with this refactor.

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,7 +9,6 @@ var exports = module.exports = Argv
 
 function Argv (processArgs, cwd) {
   var argv = yargs(processArgs, cwd, require)
-  processArgs = processArgs || [] // handle calling yargs().
   singletonify(argv)
   return argv
 }


elas7:
@bcoe well, for my use case everything seems to be working fine!

One thing I was thinking about is maybe removing `exports.rebase = yargs.rebase` from index.js. After all, it was only exported for tests. But now we export it from yargs.js, so there's really no need to export it from both places.

The only thing to modify is in the usage.js test. You need to define 
`var rebase = require('../yargs').rebase`. And then replace all 3 uses of `yargs.rebase` with `rebase`

Let me know if that makes sense.

bcoe:
@elas7 seems like a worthwhile change, I doubt anyone is actually using that method other than our tests.

PR #436, opened 2016-03-13
remove duplicated export rebase from index.js

elas7:
As seen in #434


bcoe:
looks good to me, @maxrimue what do you think?

maxrimue:
LGTM :+1:

PR #437, opened 2016-03-13
fix tests that use .command() improperly

### LINKED ISSUE #429: Update tests that use .command() improperly
There are still a handful of tests that use the 3.x `.command()` API like so:

```js
.command('wombat', 'wombat burrows', function (yargs, argv) {
  return argv
})
```

But this is misleading b/c 4.x no longer supports passing `argv` as the 2nd argument to the "function which used to be the handler but is now the builder".

Let's update these tests so they are compatible with the 4.x `.command()` API, either using:
- a builder function and handler function
- a module with `builder` (object or function) and `handler` (function) properties
- or an options object and handler function

lrlna:
I left these to provide only `cmd` and `desc`, as they are only testing `fail` outputs in `validation` and `usage`.

closes #429

### DIFF
--- a/test/usage.js
+++ b/test/usage.js
@@ -86,9 +86,7 @@ describe('usage tests', function () {
         var r = checkUsage(function () {
           return yargs('wombat -w 10 -m 10')
             .usage('Usage: $0 -w NUM -m NUM')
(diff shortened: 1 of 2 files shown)


bcoe:
@nexdrew look good to you, I say :shipit:

PR #438, opened 2016-03-14
[New] if a `demand` message is provided, gsub in “got” and “count”

### LINKED ISSUE #431: default `.demand` message has counts, provided one can't?
How can I do `.demand(n, msg)` such that `msg` can contain a count of the number of non-option arguments provided?

ljharb:
thar she blows, fixing #431!

### DIFF
--- a/README.md
+++ b/README.md
@@ -687,6 +687,8 @@ If `key` is an array, demand each element.
 
 If a `msg` string is given, it will be printed when the argument is missing,
 instead of the standard error message. This is especially helpful for the non-option arguments in `argv._`.
+Every occurrence of "$0" will be replaced with the number of non-option arguments in `argv._`;
+every occurrence of "$1" will be replaced with the number of non-option arguments demanded.
 
 If a `boolean` value is given, it controls whether the option is demanded;
 this is useful when using `.options()` to specify command line parameters.
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -22,6 +22,10 @@ module.exports = function (yargs, usage, y18n) {
         usage.fail(
           __('Not enough non-option arguments: got %s, need at least %s', argv._.length, demanded._.count)
         )
+      } else if (typeof demanded._.maxMsg === 'string') {
+        usage.fail(
+          demanded._.maxMsg.replace(/\$0/g, argv._.length).replace(/\$1/g, demanded._.count)
+        )
       } else {
         usage.fail(
           __('Too many non-option arguments: got %s, maximum of %s', argv._.length, demanded._.max)
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -14,7 +14,9 @@ module.exports = function (yargs, usage, y18n) {
     var _s = argv._.length
 
     if (demanded._ && (_s < demanded._.count || _s > demanded._.max)) {
-      if (demanded._.msg !== undefined) {
+      if (typeof demanded._.msg === 'string') {
+        usage.fail(demanded._.msg.replace(/\$0/g, argv._.length).replace(/\$1/g, demanded._.count))
+      } else if (demanded._.msg !== undefined) {
         usage.fail(demanded._.msg)
       } else if (_s < demanded._.count) {
         usage.fail(
--- a/test/validation.js
+++ b/test/validation.js
@@ -90,6 +90,16 @@ describe('validation tests', function () {
         .argv
     })
 
+    it('fails with substituted counts if "too many" message is defined', function (done) {
+      yargs(['a', 'b', 'c'])
+        .demand(1, 2, undefined, 'got $0, expected $1 ($0 / $1)')
+        .fail(function (msg) {
+          msg.should.equal('got 3, expected 1 (3 / 1)')
+          return done()
+        })
+        .argv
+    })
+
     it('fails in strict mode with invalid command', function (done) {
       yargs(['koala'])
         .command('wombat', 'wombat burrows')
--- a/test/validation.js
+++ b/test/validation.js
@@ -70,6 +70,16 @@ describe('validation tests', function () {
         .argv
     })
 
+    it('fails with substituted counts if msg is defined', function (done) {
+      yargs([])
+        .demand(1, 'got $0, expected $1 ($0 / $1)')
+        .fail(function (msg) {
+          msg.should.equal('got 0, expected 1 (0 / 1)')
+          return done()
+        })
+        .argv
+    })
+
     it('fails in strict mode with invalid command', function (done) {
       yargs(['koala'])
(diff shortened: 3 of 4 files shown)


bcoe:
will land this earl this week.

bcoe:
@lrlna @nexdrew thoughts?

nexdrew:
You can kinda already do this:

```js
require('yargs')
  .updateLocale({
    'Not enough non-option arguments: got %s, need at least %s': 'got %s, expected %s'
  })
  .demand(1)
  .argv
```

```
$ node issue431.js

got 0, expected 1
```

ljharb:
@nexdrew good to know, but that's a very obscure and undocumented way of doing something relatively straightforward :-)

lrlna:
o.O oh cool, didn't know that @nexdrew 

We also already have `demand` to output the same kind of deal when no argument other than `num` is provided. 
```
require("./")
  .command("config", "set up this module")
  .command("send", "send a message")
  .command("start", "start a conversation")
  .demand(2)
  .argv
```

```bash
$ node issue431.js 

Commands:
  config  set up this module
  send    send a message
  start   start a conversation

Not enough non-option arguments: got 0, need at least 2

```

I guess this is for :sparkles: ultra fancy :sparkles:  messages folks might want ?

nexdrew:
@ljharb Agreed. It's not _entirely_ undocumented ([see here](http://yargs.js.org/docs/index.html#methods-updatestringsobj)), but it _is_ obscure because you basically have to look in the locale files of yargs source code to know what your options are. The cool thing about it is that you can override just about **any** yargs-defined string this way, but it does have limitations and is arguably not straightforward.

That being said, I'm not opposed to this pull. I'd like to avoid doing the same thing in a bunch of different ways, but this change (1) seems simple and straightforward, (2) is somewhat orthogonal to i18n, and (3) makes for a nicer API. So looks good to me! :+1:

ljharb:
@lrlna yes, exactly - i wanted to make a message that described them better than "non-option arguments", but still include the actual vs expected counts.

lrlna:
@ljharb I am all for it :+1:

ljharb:
@bcoe friendly ping! it's halfway through the week :-)

bcoe:
@ljharb (CC: @lrlna) working on code reviewing this, an edge-case popped up.

What should we do if you are demanding arguments, or arguments and positional arguments?

```js
#!/usr/bin/env node
var argv = require('./')
  .demand(['a', 'b'], "pork chop sandwhiches $0")
  .argv;

console.log(argv)
```

or

```js
#!/usr/bin/env node
var argv = require('./')
  .demand(1, ['a', 'b'], "pork chop sandwhiches $0")
  .argv;

console.log(argv)
```

ends up outputting:

```sh
Missing required arguments: a, b
pork chop sandwhiches $0
```

Is there something else we should be outputting? Perhaps this is just an edge-case, i.e., `$0 ... $x` is only replaced for positional arguments?

ljharb:
@bcoe good question. I don't think it would be sensible to output counts in that case, and outputting the array of received arguments isn't as useful, because the user already knows which explicit named args they provided.

elas7:
I think there's another edge case. 

There are 2 messages that can be displayed when the count is invalid:

- `Not enough non-option arguments: got %s, need at least %s`
- `Too many non-option arguments: got %s, maximum of %s`

So, if a custom message is provided, it won't make too much sense in one of those cases.

A solution may be to find a way to provide two custom error messages. Or to provide `$2` as `demanded._.max` to the custom message, so as to be able to create a message that handles both cases. Or to use @nexdrew method of calling `updateLocale()`

ljharb:
aha, i didn't realize there was a "max" message. That means that even the current API of accepting one message doesn't make sense.

Would you be amenable if I updated the PR so `.demand(count[, notEnoughMessage[, tooManyMessage]])` was the signature?

bcoe:
@ljharb it seems like we've uncovered a bit of a hornet's nest. I've got a busy day at the npm mines today, but I'm going to reflect on this problem -- In a perfect world we'd make providing a custom message a bit easier, without breaking the existing API.

ljharb:
@bcoe what would you think about the `.demand(count[, notEnoughMessage[, tooManyMessage]])` signature, but using `updateLocale` internally? That way the `%s` would be consistent, and it would just be sugar for the less intuitive approach of `updateLocale`?

ljharb:
I've updated the PR to include a message override param in `demand` for "too many".

At the moment it still uses the `$0`/`$1` notation, but if indicated, I'll update it to use the translation internals accordingly.

bcoe:
@ljharb @nexdrew okay, `demand` overriding is a tire fire is my conclusion ... I started putting some work into merging this today -- and realized another edge-case (which is why the build is currently failing with test-coverage issues). We need to cover this case:

`.demand(['arg', 'arg'], 1, 2, 'min message', 'max message', 'argument message?')`

I propose that rather than adding the new `maxMessage` argument (which complicates things further) I propose that we deprecate `.demand()`, and introduce two new methods: `demandCommand`, `demandOption`.

`.demandCommand(min, max, minMsg, maxMsg)`
`.demandCommand(min, minMsg)`
`.demandOption('option', 'msg')`
`.demandOption(['option', 'option'], 'msg')`

I was against this at first, but at the end of the day I think splitting out the logic like this will start to put out the tire fire ... perhaps we can even refactor `.demand()` to call `demandCommand` and `demandOption`.

ljharb:
Sounds good - I'll work on a PR to add the two new methods.

bcoe:
@ljharb would love to get this polished, happy for some other folks in the yargs org to take on the work 👍 but also happy to wait for your contribution.

ljharb:
Sorry about the delay, time's gotten away from me :-) if it's done in the meantime that's OK, otherwise I'll try to get something up soon.

bcoe:
@ljharb nothing to apologize for, I appreciate the contribution so far. if anyone else contributes to this feature, we should make sure it's made as a pull-request to your branch so that you can coordinate with them.

bcoe:
@maxrimue @lrlna, etc, does anyone else want to help see this over the finish line? we went down a rabbit hole with the function signature of `demand`.

lrlna:
hii! imma look into it :v:, @bcoe //cc @maxrimue

bcoe:
@lrlna @ljharb bumping this pull; if we don't have any time to think about how we should fix this historic API right now, perhaps we should create a tracking ticket in issues and close this for the time being? Just so we don't have pull requests kicking around for too long :)

ljharb:
I'm fine with that - as long as something remains open :-)

lrlna:
I've opened an issue to track this and will close this pull request for now. Hope to finish the PR by the end of the day 🎉

PR #441, opened 2016-03-16
Remove duplicated option keys from arrayOptions

elas7:
It wasn't causing any issue, but this should save a few iterations in `arrayOptions.forEach`


nexdrew:
@elas7 LGTM! Thanks!

bcoe:
@elas7 good catch.

PR #445, opened 2016-03-18
Check for commands before validating argv._ in strict mode

### LINKED ISSUE #444: Unable to use required positional arguments in strict mode
For parsing my args I want to use strict mode (not allow passing of any extra args).
I have 1 required position argument and some required and optional options.

Test configuration (`test.js`):
```js
#! /usr/bin/env node
'use strict';

const argv = require('yargs')
  .required(1, 'Provide required positional argument')
  .options({
    optional: {
      describe: 'optional option'
    },
    required: {
      describe: 'required option',
      required: true
    }
  })
  .strict()
  .argv;

console.log(argv);
```

Running this script with different options:
```bash
$ ./test.js
Provide required positional argument

$ ./test.js positional
Missing required argument: required

./test.js positional --required
Unknown argument: positional

./test.js positional --required --optional
Unknown argument: positional
```

Removing `strict()` statement fixes the situation:
```bash
./test.js positional --required
{ _: [ 'positional' ],
  required: true,
  '$0': 'test.js',
  optional: undefined } 
```

nexdrew:
Fixes #444

Note that `commandKeys` is an array, not an object, so I replaced the use of `hasOwnProperty()` with `indexOf()` to account for inherent Array properties like `length`.

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -125,11 +125,13 @@ module.exports = function (yargs, usage, y18n) {
       }
     })
 
-    argv._.forEach(function (key) {
-      if (!commandKeys.hasOwnProperty(key)) {
-        unknown.push(key)
-      }
-    })
+    if (commandKeys.length > 0) {
+      argv._.forEach(function (key) {
+        if (commandKeys.indexOf(key) === -1) {
+          unknown.push(key)
+        }
+      })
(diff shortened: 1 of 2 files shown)


bcoe:
:shipit:

PR #446, opened 2016-03-20
update yargs to 4.4.0

bcoe:
@nexdrew here's an update to the changelog that:

* pulls in a few fixes/new features (that actually went out in 4.3.2).
* updates all the dependencies @elas7 made some slick fixes in yargs-parser.

I don't think it's a huge deal that some of this went out as `4.3.2` went out already; not like it was a breaking release -- perhaps post-hoc we unpublish `4.3.2`, but might be overkill.

Going forward perhaps we should start trying out the same changelog approach you're advocating at npm; I've been undertaking this process pretty manually:

* I pull in a bunch of changes over a span of time
* I document in the changelog as I go.
* I usually cut a release from the changelog branch.


nexdrew:
@bcoe Sorry if I jumped the gun on 4.3.2. I reviewed the commits since 4.3.1 and didn't see anything beyond minor fixes except the Italian locale. I wanted to get that regression fixed ASAP since it was potentially blocking folks from upgrading to 4.x.

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 ## Change Log
 
+### v4.4.0 (2016/3/20 15:07 +07:00)
+
+- [446](https://github.com/yargs/yargs/pull/446) update yargs-parser, y18n, nyc, cliui (@bcoe)
+- [#445](https://github.com/yargs/yargs/pull/445) strict mode was failing if no commands were registered (@nexdrew)
+- [#443](https://github.com/yargs/yargs/pull/443) adds Itallian translation \o/ (@madrisan)
```
nexdrew: "Italian" has one too many "l"s. :)

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
...
+- [#443](https://github.com/yargs/yargs/pull/441) remove duplicate keys from array options configuration (@elas7)
```
nexdrew: Looks like the link text has the wrong number, should be 441.

nexdrew:
I was gonna try to finish up the command context stuff tonight, but that can wait for the next release. It will need a bit of testing.

bcoe:
@nexdrew I think you're right, we could probably call a new language a patch release (this was what I felt constituted a minor).

Let's just release the library updates as `4.3.3`, and boom our changelog isn't a lie :+1:

nexdrew:
@bcoe You're right, a new language should constitute a minor bump. I acted out of impulse to get a fix out the door, probably shouldn't have done that. I guess I could have released from a different branch. Not sure what to do when we want to get a hotfix out but new features have already landed in master. (I also deliberately neglected to update the changelog b/c I was only thinking of the fix.)

I'm cool with whatever you want to do. I don't want to break protocol with how we usually do things.

bcoe:
@nexdrew no worries, I think you did the right thing -- that was a pretty major bug and it was good to get the patch out. I'm excited that we might be able to automate the changelog process :)

PR #448, opened 2016-03-22
showHelp getting the emission function itself

osher:
Motive: in tests I don't want yargs to print anything to the console. But I do want to catch it and see that it try.
So - when I pass it the handler itself - let it write to this handler.

Consider the following use-case scenario:

```
var yargs = require('yargs');
var options = { ... }; 
module.exports = function args(process, log)  {
   var args = yargs.usage("foo").help().options(options).parse(process.argv);

   if (!log) log = console.log; //injected in test time, defaults to console.log;

   if (args.help) {
        yargs.showHelp( log )
        process.exit();
        return;
   }

   return args
}

```

then, in tests

```
var sut = require('../lib/args')

describe("lib/args", function() {
   describe("when provided with help switch", function() {
        var logs = [];
        var mockProcess = { env: { ... }, argv: [ .... ], exit: function() { ... }  };
        var xUsage =  /^Usage:  ......  (whatever)    /
        before(function() {
             var args = sut( mockProcess, mockLog )
        })
        it("should call exit", function() { ... } );
        it("should output help message", function() { 
             logs.length.should.eql(1);
             logs[0].should.match(xUsage)
        });
        function mockLog(msg) {
             logs.push(msg);
        }
   })
})
```

### DIFF
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -321,9 +321,8 @@ module.exports = function (yargs, y18n) {
 
   self.showHelp = function (level) {
     if (!level) level = 'error'
-    var emit = 'function' == typeof level ? level : console[ level ]
-    
-    if ('function' == typeof emit) emit(self.help());
+    var emit = typeof level === 'function' ? level : console[ level ]
+    emit(self.help())
   }
 
   self.functionDescription = function (fn) {
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -320,8 +320,10 @@ module.exports = function (yargs, y18n) {
   }
 
   self.showHelp = function (level) {
-    level = level || 'error'
-    console[level](self.help())
+    if (!level) level = 'error'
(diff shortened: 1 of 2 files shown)


osher:
Mmm.. let me add tests too

bcoe:
@osher is this motivated by you wanting to test libraries that use yargs as a dependency?

bcoe:
@osher bump, mind giving me a little more context about the problem this solves? Looks great though, just curious :+1:

bcoe:
@osher thanks for the great contribution, this will go out in the next release.

osher:
Sorry for the late response - however - you got me just right :+1: 

Thanks for the fast merge.
Looking forward for the next release :-)

bcoe:
@osher give `npm i yargs@next` a shot, I'll promote the release early this week.

osher:
:+1

PR #450, opened 2016-03-25
Fix bash completion when COMP_WORDS has a null word

elas7:
A little background: When executing a completion function, Bash sets the `COMP_WORDS` array containing the individual words in the command line. If the cursor is at the end of the line (after the last word and separated by a blank), an **additional null word** is added to `COMP_WORDS`. That way the completion function can tell if the user is trying to complete the last word, or a new word after it.

The issue here is that we aren't sending the last null word to yargs because bash's word splitting removes unquoted nulls.

This PR should fix it. It uses the **“${name[@]}”** form to send the words, which preserves them all.

### DIFF
--- a/completion.sh.hbs
+++ b/completion.sh.hbs
@@ -10,10 +10,10 @@ _yargs_completions()
     local cur_word args type_list
 
     cur_word="${COMP_WORDS[COMP_CWORD]}"
-    args=$(printf "%s " "${COMP_WORDS[@]}")
+    args=("${COMP_WORDS[@]}")
 
     # ask yargs to generate completions.
-    type_list=`{{app_path}} --get-yargs-completions $args`
+    type_list=$({{app_path}} --get-yargs-completions "${args[@]}")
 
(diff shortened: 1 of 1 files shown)


nexdrew:
@elas7 Can you provide an example of completion output before and after this change? I'm not seeing any difference in my testing, but I'm probably missing something.

elas7:
@nexdrew Sure. Let's say you have this in `test.js`:

```javascript
#!/usr/bin/env node
var argv = require('yargs')
    .describe('foo', 'foo description')
    .command('com', 'com description')
    .completion().argv
```

If you do `$ ./test.js --f`<kbd>TAB</kbd>, you should get completed to `--foo`

But if you do this: `$ ./test.js --f`` `<kbd>TAB</kbd> (blank space after --foo), you should get completed to `com`.

Before this fix, you got completed to `--foo` in the last case, because yargs thought you were trying to complete the `--f` word.

nexdrew:
@elas7 Verified. Thanks!

bcoe:
@elas7 give `npm i yargs@next` a shot, I'll promote the release early this week.

PR #454, opened 2016-03-26
Fix demand() when second argument is an array

elas7:
Specifying a command count and required arguments at the same time causes the required arguments to be set as the 'max' of argv._

For example, calling `.demand(1, ['foo', 'bar'])` causes this to be set in options.demanded: `_: { count: 1, msg: undefined, max: [ 'foo', 'bar' ] } }`

This doesn't really fails because, when doing the validation, comparing a number to an array returns true most of the time on Javascript. But we shouldn't be relying on undocumented behavior if possible.

There is actually an edge case that fails: When calling `.demand(1, ['1'])`, it will fail is the user provides more than one command. I added a test for that.

### DIFF
--- a/test/usage.js
+++ b/test/usage.js
@@ -300,6 +300,18 @@ describe('usage tests', function () {
           'pork chop sandwiches'
         ])
       })
+
+      it("shouldn't interpret the second argument as a max when it is an array", function () {
+        var r = checkUsage(function () {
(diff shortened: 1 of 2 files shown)


bcoe:
@elas7 give `npm i yargs@next` a shot, I'll promote the release early this week.

PR #457, opened 2016-03-30
Improve $0 in Windows

elas7:
I made some changes to the way the `$0` variable is created, so that the Windows version matches the Linux one:
  - It ignores `argv[0]` if it ends in "node.exe" or "iojs.exe"
  - It allows to use the rebased version of `argv[1]` if it uses Windows style paths

Now, the tricky part is that I had to make platform dependent tests, because the behavior of `node.path.relative` changes depending on platform, and there's not a straightforward way to mock it. Let me know what you think.

### DIFF
--- a/test/usage.js
+++ b/test/usage.js
@@ -2112,7 +2112,7 @@ describe('usage tests', function () {
 
       it('is not resolved to the relative path if it is larger, using Windows paths', function () {
         mockProcessArgv(['node', 'C:\\script.js'], function () {
-          yargs([], 'C:\\very\\long\\current\\directory\\').$0.should.equal('\\script.js')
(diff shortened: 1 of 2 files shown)


bcoe:
@elas7 this looks great, did you run on a Windows VM? (or on your own Windows computer?).

elas7:
@bcoe It was on a Windows VM (Windows 7)

elas7:
Missed a failing test on Windows in the last commit. (I don't know why it wasn't found by the CI)

Here's the fix.

bcoe:
@elas7 weird, this makes me a bit concerned about our AppVeyor configuration

We should try to purposefully break tests on AppVeyor at some point, and see if we can figure out why it's not catching certain things.

bcoe:

PR #460, opened 2016-03-30
Add missing features to README docs

nexdrew:
1. Include `command` and `describe` props for command modules (since #414)
2. Include `Error` instance as 2nd arg for fail handler (since #382)
3. Sync up locale list with locale files

Also added missing `"number"` string to American English and Italian locales. (I think my Italian translation is correct. :smiley:)


bcoe:
you rock @nexdrew, boom let's conventional-changelog this up \o/

PR #461, opened 2016-03-31
Avoid repeating options in completion

elas7:
This makes sure that an option won't be suggested for completion if the option, or any of its aliases, is already in the provided arguments.

Pretty much the same thing that was done for command completion in https://github.com/yargs/yargs/pull/183

I figure this is a good start, but it only checks options in **long form**. Checking for short form depends on whether 'short-option-groups' is enabled or not, and would need to duplicate parts of the parser logic, so it is a bit tricky.

I also had to modify slightly the integration test that checks that the completion for '--help' works. Now it won't complete if the current word is '--help', so I made it try to complete '--h'


bcoe:
@elas7 awesome, great work tackling some of these completion issues.

bcoe:

PR #462, opened 2016-03-31
If invoking .config() without parameters, set a default option

elas7:
Before this, if you called `.config()` without parameters, an option named `'undefined'` is set to receive the config file. I propose to set the default name `'config'`.

This way it works just like `.help()` and `.completion()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -622,12 +622,14 @@ var argv = require('yargs')
   .argv;
 ```
 
-<a name="config"></a>.config(key, [description], [parseFn])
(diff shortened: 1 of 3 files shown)


maxrimue:
LGTM :+1:

bcoe:
@elas7 awesome \o/ I'm going to start using [conventional-changelog](https://github.com/stevemao/conventional-changelog-angular/blob/master/convention.md) for yargs; this merge is the first kick of the can.

bcoe:

PR #463, opened 2016-03-31
[WIP] New method to apply command modules matching glob pattern

nexdrew:
(Sorry for another WIP pull, but I wanted to propose this to the team.)

My proposal is to add a new `.commands(glob, opts)` method that:

- finds command submodules matching the given glob pattern (via new dependency [`glob`](https://github.com/isaacs/node-glob)) relative to the bin module executing yargs (via existing use of `require-main-filename`)
- requires the found submodules (another "dynamic" use of require)
- applies the submodules as commands (via `command.addHandler()`)

For programs with many commands, this allows authors to modularize them into their own files and easily apply all of them to yargs with a single concise method, like so:

```js
require('yargs')
  .commands('cmd*') // finds all nested submodules
                    // like cmd-foo.js and lib/whatever/cmd-bar.js
  .help()
  .argv
```

They can even add new commands without changing the yargs configuration at all!

The initial logic in this branch works, but it needs tests.

I'm looking for feedback on the following:

1. What do you think about the method name `.commands()`?
2. Do you see any problems with adding `glob` as a dependency?
3. Any advice for writing tests that work outside the context of yargs?

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1,3 +1,6 @@
+var globSync = require('glob').sync
+var path = require('path')
+
 // handles parsing positional arguments,
 // and populating argv with said positional
 // arguments.
@@ -56,6 +59,16 @@ module.exports = function (yargs, usage, validation) {
       else parsedCommand.demanded.push(cmd.replace(bregex, ''))
     })
     return parsedCommand
+  }
+
+  self.findModules = function (glob, req, cwd, opts) {
+    opts = opts || {}
+    opts.nodir = true
+    if (!opts.cwd) opts.cwd = cwd
+    if (typeof opts.matchBase === 'undefined') opts.matchBase = true
+    globSync(glob, opts).forEach(function (file) {
+      self.addHandler(req(path.resolve(opts.cwd, file)))
+    })
   }
 
   self.getCommands = function () {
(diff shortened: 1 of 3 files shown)


nexdrew:
It has come to my attention 😉 that we could achieve an acceptable subset of this functionality by using the much lighter-weight `require-directory` instead of `glob` - the main difference being support for a single directory at-a-time instead of a potentially complex glob pattern.

I'm happy to make this change. I will either update this PR or close it in favor of a new one.

I could still use feedback on the API method name though.

maxrimue:
@nexdrew Sounds cool :+1: 
I think it's a good idea to use `require-directory` instead. It should be okay to make this change on this PR. Concerning the method name, `commands` could sound like a shorthand to create multiple commands with the same properties but a different name (`'cmd*'`). I would suggest something like `.importCommands`, to make it clear that we're talking about importing things here.

nexdrew:
@maxrimue Yeah, I think you're right about the name, especially given that in some cases the API uses plural method names as aliases for singular method names (e.g. `.option()` and `.options()`).

Thanks for the feedback! Anyone else have strong feelings about `.importCommands()`? I was kinda thinking `.commandDir()` or something.

bcoe:
@nexdrew in the name of keeping yargs' package size as small as possible, what if we add this feature but opt for using require-directory instead; I'm also thinking, if we're opting for this simple of an approach, perhaps we just implement similar functionality to require-directory in our `/lib` folder? can be convinced either way.

nexdrew:
Closing in favor of #494.

PR #469, opened 2016-04-07
fix(default): Remove undocumented alias of default()

elas7:
Remove an alias of the `default()` method, called 'defaults', that was added to avoid usage of
a reserved word (https://github.com/yargs/yargs/commit/1bd4375e11327ba1687d4bb6e5e9f3c30c1be2af)

This isn't actually needed because a method name is a fair use of reserved keywords in Javascript.

### DIFF
--- a/yargs.js
+++ b/yargs.js
@@ -311,8 +311,6 @@ function Yargs (processArgs, cwd, parentRequire) {
     return self
   }
 
-  self.defaults = self.default
-
   self.describe = function (key, desc) {
     options.key[key] = true
     usage.describe(key, desc)


bcoe:
@elas7 unfortunately, this is a breaking change and I bet a few folks have stumbled into using `defaults` rather than `default`. I'm fine with deprecating the usage, but I think this will need to wait for the next major.

bcoe:
@elas7 rather than this change, let's either document this alias or document and indicate that this alias is deprecated and will go away in `v5.0.0`?

bcoe:
@elas7 this has landed in `next` along with several of your changes to `yargs-parser`:

Issue #335, opened 2016-01-08
Isolate completion functionality from shell integration


abelnation:
Just like you can use yargs to parse any string (not just `process.argv`), it would be great to be able to get completions for any string.  Currently, it looks like `completions.js` relies on `process.argv` explicitly.

bcoe:
@abelnation don't suppose you would be willing to take a stab at this?

rwjblue:
Looks like this was addressed by https://github.com/yargs/yargs/pull/470.

bcoe:
@rwjblue thanks for helping us perform bookkeeping on old issues.

PR #470, opened 2016-04-08
feat(completion): allow to get completions for any string, not just process.argv

elas7:
Add new method `.getCompletion(args, done)` which receives an array of strings representing a line to complete, and calls the `done` callback with the resulting completions.

Update the completion tests to pass the arguments explicitly to yargs. We can't pass them by mocking `process.argv` now because `process.argv` is only checked on `index.js`, which is called before the tests can apply the mock.

### DIFF
--- a/README.md
+++ b/README.md
@@ -859,6 +859,29 @@ var argv = require('yargs')
   .argv
 ```
 
+.getCompletion(args, done);
+---------------------------
+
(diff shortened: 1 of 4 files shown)


bcoe:
@elas7 this is great, this should address @abelnation's request in #335?

elas7:
@bcoe that´s right, it addresses #335

PR #471, opened 2016-04-09
feat(validation): Add .skipValidation() method

### LINKED ISSUE #453: Add ability to process specified options prior to .demand()
Recently migrated from `optimist` and was very happy with your drop-in replacement. Glancing the documentation, saw there were new capabilities and desired to delegate all option processing to this module. Our app, **abao**, has a single required parameter, so I modified the code as such:
```js
var argv = yargs
    .usage("Usage: \n  abao </path/to/raml/file> [OPTIONS]" +
    "\n\nExample: \n  " + "abao api.raml --server http://api.example.com")
    .options(Abao.options)
    .wrap(80)
    .demand(1, "abao: Must specify path to RAML file")
    .argv;
```

But I discovered that once I added `.demand()`, other *one and done* options no longer process correctly. For example, if I invoke application as such:
```bash
$ abao --version
```

It now displays usage and error for missing argument. Same for `--help` and a couple other
"output'n'quit" style options that don't *actually* require the argument needed for the app's primary functionality.
.
In order to make use of the `.demand()` functionality, I need to be able to specify certain options
that should be handled **before** `.demand()` processing begins, for those not *actually* requiring the argument needed for the app's primary functionality.

elas7:
`.skipValidation()` takes a key. If the key is present, validation won't be performed.

Closes #453

### DIFF
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -232,6 +232,7 @@ describe('yargs dsl tests', function () {
         defaultDescription: {},
         choices: {},
         requiresArg: [],
+        skipValidation: [],
         count: [],
         normalize: [],
         number: [],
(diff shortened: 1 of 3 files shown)


test/yargs.js
```diff
@@ -886,6 +887,43 @@ describe('yargs dsl tests', function () {
...
+          .demand(1)
+          .fail(function (msg) {
+            return done()
+          })
+          .skipValidation(['skip', 'reallySkip'])
+          .argv
+      argv.koala.should.equal(true)
+    })
+
+    it('allows key to be specified with option shorthand', function () {
```
bcoe: it would be worthwhile to add one additional test that shows validation failing if one skipped argument is provided, but another argument is provided that causes it to still become a failure scenario.

bcoe:
@elas7 this seems great to me, is there a specific open-issue this addresses (for book keeping).

plroebuck:
@bcoe, #453.

bcoe:
@elas7 I appreciate that we can move `helpOrVersion` to a more generic name.

nexdrew:
Sorry I have been delinquent in my reviews and am thus a little late to the game here. I don't really have a problem with `.skipValidation()`, but it's not the same as short-circuiting, which is what I thought the goal of #453 was.

If validation is skipped via special flag, execution just continues without an indication to the program of what scenario to run. Perhaps skipping validation is more flexible than short-circuit functionality, but I kinda thought executing a custom function and then exiting the process (per `.exitProcess()` setting) would be more helpful/popular than just skipping validation.

@bcoe @elas7 @plroebuck Thoughts?

PR #478, opened 2016-04-12
fix(pkgConf): fix aliases issues in .pkgConf()

elas7:
Update `.pkgConf()` to use the new `configObjects` option from yargs-parser, this way the aliases issues are solved.

Change wording of `.pkgConf()` tests and description as it no longer works by setting defaults, and add some further tests.


bcoe:
@elas7 great work.

PR #479, opened 2016-04-12
feat(configuration): Add .configObject() method to pass configuration objects to yargs

### LINKED ISSUE #423: Suggestion: allow passing an object instead of a configuration file path to yargs
Let's say you want to have some internal logic in your application to parse configuration files, and then send the result object to yargs.

Right now, you could do it by defining a custom parser in config(). But I really don't need to take a filename as an input, nor having it as an user option available in the CLI. Here is an example using config():

```javascript
yargs
.config('settings', function() {
  return {'a':1, 'b':2}
})
.default('settings', 'fakepath')
```

If a way to directly pass an object is implemented, it is important for the object to have the same order of precedence as a config file passed to config(), so that it is bellow command line args, but above defaults.

Thanks!

elas7:
Similar to .config(), but it receives an explicit configuration object instead of setting a key for a file path to a JSON configuration object.

Closes #423

### DIFF
--- a/README.md
+++ b/README.md
@@ -674,6 +674,28 @@ var argv = require('yargs')
   .argv
 ```
 
+.configObject(obj)
+------------------
+
+Similar to [`config()`](#config), it receives an explicit configuration object.
+
+`obj` will be parsed and its properties will be set as arguments.
+
+```js
+var argv = require('yargs')
+  .configObject({foo: 1, bar: 2})
+  .argv
+console.log(argv)  
+```
+
+```
+$ node test.js
+{ _: [],
+  foo: 1,
+  bar: 2,
+  '$0': 'test.js' }
+```
+
 <a name="count"></a>.count(key)
 ------------
 
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -684,6 +684,17 @@ describe('yargs dsl tests', function () {
     })
   })
 
+  describe('configObject', function () {
+    it('allows a configuration object', function () {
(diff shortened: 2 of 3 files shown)


bcoe:
I like your modification to yargs-parser, introducing the concept of a config-object; this fits with the fact that we're trying to make that library agnostic of implementation details, such as whether your config is coming from the file-system.

I'm less convinced that we need to add this additional option to the already fairly large yargs API, what's the advantage of this approach over simply defining default values on your options?

I just want to be careful, because as we introduce more features to yargs it becomes harder for folks to dive in and learn how to use it...

Here's an alternative API idea, what if we overload `.config()` so that if it receives an `object` it uses the object as the config settings, and if it receives a string it assumes this represents a `key`? I'd be happier with this compromise. CC @nexdrew @maxrimue @lrlna

elas7:
@bcoe that makes sense.

This approach is useful if an application has internal logic to parse configuration files, for example by using 'rc', which looks for `.config` files in several directories. 

It could be done by defining default values, but then we can run into the same problems with aliases that happened to `.pkgConf()`, and we would also have those values documented on `--help`.

Of course, as you say, this is a very small use case, and it can very well be worth to avoid enlarging the API. If we do add it, I agree that overriding `.config()` is the cleanest way.

bcoe:
@elas7 awesome, seems like a good compromise:

* it's a logical extension to `.config()`.
* we keep the API surface the same.

If you're happy I'm happy.

PR #480, opened 2016-04-13
feat(configuration): Allow to directly pass a configuration object to .config()

elas7:
Continues from #479


nexdrew:
This looks good! Nice work @elas7!

PR #481, opened 2016-04-13
docs(default): Document deprecated alias of default()

elas7:
Continues from #469


nexdrew:
LGTM 👍 Thanks @elas7!

PR #482, opened 2016-04-15
Overhauled travis configuration

maxrimue:
This PR makes the `.travis.yml` significantly slimmer by using the official, automatic way of testing with Node.js provided by Travis. While this configuration basically makes Travis do the same thing, the configuration now leaves the set-up part entirely to Travis.

This PR also enables testing on macOS again as the required nvm package is now [officially available for all macOS images](https://github.com/travis-ci/travis-ci/issues/2311#issuecomment-205549262). This new enhancement of support by Travis also is yet another reason why we don't need the "C workaround" for enabling Node.js testing on the macOS platform anymore.


.travis.yml
```diff
@@ -1,20 +1,13 @@
...
-  - NODE_VERSION="0.10"
-  - NODE_VERSION="0.12"
-  - NODE_VERSION="node"
-  - NODE_VERSION="4"
+
+node_js:
+  - "0.10"
+  - "0.12"
+  - "4"
+  - "5"
```
nexdrew: Can we keep `"node"` here (instead of `"5"`)? That will guarantee we always test with the latest version available in nvm.

nexdrew:
This is awesome! Thanks for staying on top of this.

.travis.yml
maxrimue: @nexdrew Sure, updated :smile:

bcoe:
@maxrimue great work, make it so \o/

Issue #492, opened 2016-05-02
commands: we should make builder, handler, description, command optional


bcoe:
It's not the case that we always wand a builder function for a command, I propose that we make each of these parameters optional.

bcoe:
@vjpr will address https://github.com/yargs/yargs/issues/509 here.

PR #494, opened 2016-05-03
Add .commandDir(dir) to API to apply all command modules from a relative directory

nexdrew:
This replaces #463, opting to require modules from a single directory at a time via `require-directory` (instead of supporting globs).

The implementation is slightly trickier than I originally thought, but it allows a consumer to do things like the following.

### Example setup

**directory structure**:

```
cli.js
commands/
  people.js
  places.js
  things.js
  common_cmds/
    details.js
    search.js
  people_cmds/
    call.js
  places_cmds/
    nav.js
```

**cli.js**:

```js
#!/usr/bin/env node
require('yargs')
  // apply people.js, places.js, things.js but not subdirectories
  .commandDir('commands')
  .help()
  .wrap(40)
  .argv
```

**commands/people.js**:

```js
exports.command = 'people <command>'
exports.desc = 'Work with people'
exports.builder = function (yargs) {
  // apply commands in subdirectories
  return yargs.commandDir('common_cmds').commandDir('people_cmds')
}
exports.handler = function (argv) {}
```

**commands/places.js**:

```js
exports.command = 'places <command>'
exports.desc = 'Work with places'
exports.builder = function (yargs) {
  // you can even share the same subcommands e.g. common_cmds
  return yargs.commandDir('common_cmds').commandDir('places_cmds')
}
exports.handler = function (argv) {}
```

**commands/things.js**:

```js
exports.command = 'things <command>'
exports.desc = 'Work with things'
exports.builder = function (yargs) {
  return yargs.commandDir('common_cmds')
}
exports.handler = function (argv) {}
```

**commands/common_cmds/details.js**:

```js
exports.command = 'details <name>'
exports.desc = 'Get details by name'
exports.builder = {}
exports.handler = function (argv) {
  // parent command accessible via argv._ as always
  console.log('getting %s details for name: %s', argv._[0], argv.name)
}
```

### Example CLI

```
$ ./cli.js --help
Commands:
  people <command>  Work with people
  places <command>  Work with places
  things <command>  Work with things

Options:
  --help  Show help            [boolean]
```

```
$ ./cli.js people --help
cli.js people <command>

Commands:
  details <name>  Get details by name
  search <query>  Search via query
  call <person>   Call someone

Options:
  --help  Show help            [boolean]
```

```
$ ./cli.js people details       
cli.js people details <name>

Options:
  --help  Show help            [boolean]

```

```
$ ./cli.js people details Andrew
getting people details for name: Andrew
```

### API

This adds one method to the API: `.commandDir(dir, [opts])`

`dir` is a directory name which can/should be relative to the module calling the method.

`opts` is an options object that is passed to `require-directory`, allowing consumers to customize the behavior. In the initial implementation, the following opts are valid:

- `opts.recurse`: boolean, default `false`

    Should yargs require command modules from all nested directories? The default is `false` so subcommands can be nested instead of flattened.

- `opts.extensions`: array of strings, default `['js']`

    The types of files to include when requiring.

- `opts.visit`: function

    A synchronous function that is called for each command module encountered. Accepts `commandObject`, `pathToFile`, and `filename` as arguments. Returns `commandObject` to include the command; any falsey value to exclude/skip it.

- `opts.include`: RegExp or function

    Whitelist certain modules. See [require-directory whitelisting](https://www.npmjs.com/package/require-directory#whitelisting) for details.

- `opts.exclude`: RegExp or function

    Blacklist certain modules. See [require-directory blacklisting](https://www.npmjs.com/package/require-directory#blacklisting) for details.

### Notes

This also includes a change that allows a command module to only define `command` and `description`, which better supports requiring JSON modules (though I'm not sure why someone would want to do that) and fulfills (at least) a part of #492.

I considered adding support for using the filename of a command module as its `command` string if the module itself doesn't define `command`, but I think it'd be nice to support this for _all_ command modules (i.e. not just when using `.commandDir()`), so I'd like to tackle that separately/subsequently.

I don't have strong feelings about the method name `commandDir`. Would be happy to change it if folks like something else better. As always, feedback is welcomed!

### Still lacking

- [x] Tests!
- [x] README update

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -9,11 +9,13 @@ module.exports = function (yargs, usage, validation) {
 
   var handlers = {}
   self.addHandler = function (cmd, description, builder, handler) {
-    // TODO allow objects that only define command and description
-    
-    // allow a module to define all properties
-    if (typeof cmd === 'object' && typeof cmd.command === 'string' && cmd.builder && typeof cmd.handler === 'function') {
-      self.addHandler(cmd.command, extractDesc(cmd), cmd.builder, cmd.handler)
+    // allow modules that define (a) all properties or (b) only command and description
+    if (typeof cmd === 'object' && typeof cmd.command === 'string') {
+      if (cmd.builder && typeof cmd.handler === 'function') {
+        self.addHandler(cmd.command, extractDesc(cmd), cmd.builder, cmd.handler)
+      } else {
+        self.addHandler(cmd.command, extractDesc(cmd))
+      }
       return
     }
 
--- a/lib/command.js
+++ b/lib/command.js
@@ -1,3 +1,4 @@
+var path = require('path')
 var requireDirectory = require('require-directory')
 
 // handles parsing positional arguments,
@@ -39,12 +40,31 @@ module.exports = function (yargs, usage, validation) {
     }
   }
 
-  self.addDirectory = function (dir, req, mainFilename, opts) {
+  self.addDirectory = function (dir, context, req, mainFilename, opts) {
     opts = opts || {}
+    // dir should be relative to the command module
+    dir = path.join(context.dirs[context.commands.join('|')] || '', dir)
+    // disable recursion to support nested directories of subcommands
+    if (typeof opts.recurse !== 'boolean') opts.recurse = false
+    // exclude 'json', 'coffee' from require-directory defaults
+    if (!Array.isArray(opts.extensions)) opts.extensions = ['js']
+    // allow consumer to define their own visitor function
     var parentVisit = typeof opts.visit === 'function' ? opts.visit : function (o) { return o }
+    // call addHandler via visitor function
     opts.visit = function (obj, joined, filename) {
       var visited = parentVisit(obj, joined, filename)
-      if (visited) self.addHandler(visited)
+      // allow consumer to skip modules with their own visitor
+      if (visited) {
+        // check for cyclic reference
+        // each command file path should only be seen once per execution
+        if (~context.files.indexOf(joined)) return visited
+        // keep track of visited files in context.files
+        context.files.push(joined)
+        // map "command path" to the directory path it came from
+        // so that dir can be relative in the API
+        context.dirs[context.commands.concat(parseCommand(visited.command).cmd).join('|')] = dir
+        self.addHandler(visited)
+      }
       return visited
     }
     requireDirectory({ require: req, filename: mainFilename }, dir, opts)
--- a/lib/command.js
+++ b/lib/command.js
@@ -1,3 +1,5 @@
+var requireDirectory = require('require-directory')
+
 // handles parsing positional arguments,
 // and populating argv with said positional
 // arguments.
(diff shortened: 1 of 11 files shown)


bcoe:
@nexdrew I love this work, don't suppose we could get a test in the mix :)

AGresvig:
Yes, this is exactly what I had hoped for - I think the chosen approach makes perfect sense and complements the functionality provided by command modules really well. Love it.
Please let this find it's way into master in the near future.

Great work @nexdrew

nexdrew:
@AGresvig Thanks! I'll get some tests and docs written up within the next couple days so we can merge this safely.

nexdrew:
@bcoe @maxrimue I believe this is ready for final review.

lib/command.js
```diff
@@ -6,9 +9,13 @@ module.exports = function (yargs, usage, validation) {
...
-    // allow a module to define all properties
-    if (typeof cmd === 'object' && typeof cmd.command === 'string' && cmd.builder && typeof cmd.handler === 'function') {
-      self.addHandler(cmd.command, extractDesc(cmd), cmd.builder, cmd.handler)
+    // allow modules that define (a) all properties or (b) only command and description
+    if (typeof cmd === 'object' && typeof cmd.command === 'string') {
+      if (cmd.builder && typeof cmd.handler === 'function') {
+        self.addHandler(cmd.command, extractDesc(cmd), cmd.builder, cmd.handler)
+      } else {
+        self.addHandler(cmd.command, extractDesc(cmd))
+      }
```
maxrimue: I'm a little confused here, is allowing commands without builder and handler functions an essential feature for `.commandDir()` or is this unrelated?

maxrimue:
Looks awesome, well done! Could we perhaps make some variables constants since there are some good reasons for doing so in yargs in general as shown in #502?

lib/command.js
nexdrew: @maxrimue Good question. This could certainly be done in a separate PR (maybe as a precursor) but is just something I ran into when testing with `require-directory`.

A couple reasons:

1. yargs currently blows up when you give it a command module/object that does not have all the expected properties of a command module - this is something that should probably be fixed in general
2. `require-directory` looks for and loads `json` files as well as `js` and `coffee` modules by default, which I have overridden to be just `js` (since a JSON "command module" cannot support a `handler` function by itself), but if folks want to define "commands" via JSON files, then they will need this change to support an option of `extensions: ['js', 'json']` (or similar)

I don't really think no. 2 is a big concern, unless folks are dynamically generating their commands as JSON (not sure why you would), but is more of a "why not?".

I view no. 1 as a bigger problem that we should address, either before or with #492. So figured I'd kill two birds with one stone here. I can easily be persuaded otherwise.

nexdrew:
> Could we perhaps make some variables constants since there are some good reasons for doing so in yargs in general as shown in #502?

@maxrimue Ooh, good catch! I think there are probably a couple places I could use `const` in here. I'll review that and push whatever changes are compatible with tests.

nexdrew:
@maxrimue Swapped `var` for `const` in commit 0d14096.

maxrimue:
@nexdrew Awesome! LGTM

lib/command.js
bcoe: Let's address this in #429 👍

package.json
```diff
@@ -19,6 +19,7 @@
     "os-locale": "^1.4.0",
     "pkg-conf": "^1.1.2",
     "read-pkg-up": "^1.0.1",
+    "require-directory": "^2.1.1",
```
bcoe: I appreciate that this only pull in once dependency in total \o/

bcoe:
@nexdrew please give `npm cache clear; npm i yargs@next` a spin I'm feeling ready to publish `4.8.0` once windows tests are passing.

bcoe:
@AGresvig @maxrimue would love to also have you try out the next candidate release too 👍 

AGresvig:
@bcoe OK will do!

Issue #497, opened 2016-05-09
Help stops before finishing to write in Node v6.1.0


nicola:
```js
var yargs = require('yargs')
const program = yargs
  .usage('usage: $0 [options]')

// program.option('hello', {description: 'test option'})
for (var i = 0; i < 20; i++) {
  program.option('hello_' + i, {description: 'this is a medium sized description, good!'})
}

program
  .help('help')
  .wrap(null)

const argv = program.argv

if (argv._.length < 1) {
  yargs.showHelp()
}
```

Here is my little command:

```
$ node example2.js
usage: example2.js [options]

Options:
  --hello_0   this is a medium sized description, good!
  --hello_1   this is a medium sized description, good!
  --hello_2   this is a medium sized description, good!
  --hello_3   this is a medium sized description, good!
  --hello_4   this is a medium sized description, good!
  --hello_5   this is a medium sized description, good!
  --hello_6   this is a medium sized description, good!
  --hello_7   this is a medium sized description, good!
  --hello_8   this is a medium sized description, good!
  --hello_9   this is a medium sized description, good!
  --hello_10  this is a medium sized description, good!
  --hello_11  this is a medium sized description, good!
  --hello_12  this is a medium sized description, good!
  --hello_13  this is a medium sized description, good!
  --hello_14  this is a medium sized description, good!
  --hello_15  this is a medium sized description, good!
  --hello_16  this is a medium sized description, good!
  --hello_17  this is a medium sized description, good!
  --hello_18  this is a medium sized description, good!
  --hello_19  this is a medium sized description, good!
  --help      Show help  [boolean]
```

All correct, but when I use the flag `--help`, it breaks:

```
$ node example2.js --help
usage: example2.js [options]

Options:
  --hello_0   this is a medium sized description, good!
  --hello_1   this is a medium sized description, good!
  --hello_2   this is a medium sized description, good!
  --hello_3   this is a medium sized description, good!
  --hello_4   this is a medium sized description, good!
  --hello_5   this is a medium sized description, good!
  --hello_6   this is a medium sized description, good!
  --hello_7   this is a medium sized description, good!
  --hello_8   this is a medium sized description, good!
  --hello_9   this is a medium sized description, good!
  --hello_10  this is a medium sized description, good!
  --hello_11  this is a medium sized description, good!
  --hello_12  this is a medium sized description, good!
  --hello_13  this is a medium sized description, good!
  --hello_14  this is a medium sized description, good!
[...]

PR #498, opened 2016-05-10
fix: defer exiting, don't panic w/ .exit()

eljefedelrodeodeljefe:
Fixes: https://github.com/yargs/yargs/issues/497

Please don't take this too seriously, I have never used yargs, but am playing with a solution to the problem.

This would pass tests. I am not sure whether it is API compatible.

### DIFF
--- a/yargs.js
+++ b/yargs.js
@@ -11,6 +11,12 @@ var readPkgUp = require('read-pkg-up')
 var pkgConf = require('pkg-conf')
 var requireMainFilename = require('require-main-filename')
 var objFilter = require('./lib/obj-filter')
+var EventEmitter = require('events')
+var ee = new EventEmitter()
+
+ee.on('custom_exit', function (res) {
+  // no-op
+})
 
 var exports = module.exports = Yargs
 function Yargs (processArgs, cwd, parentRequire) {
@@ -651,7 +657,7 @@ function Yargs (processArgs, cwd, parentRequire) {
     if (completionCommand && ~argv._.indexOf(completionCommand) && !argv[completion.completionKey]) {
       self.showCompletionScript()
       if (exitProcess) {
-        process.exit(0)
+        return ee.emit('custom_exit', 0)
       }
     }
 
@@ -667,7 +673,7 @@ function Yargs (processArgs, cwd, parentRequire) {
         })
 
         if (exitProcess) {
-          process.exit(0)
+          return ee.emit('custom_exit', 0)
         }
       })
       return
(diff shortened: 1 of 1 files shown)


eljefedelrodeodeljefe:
Had to push a second commit for 0.10 builds

eljefedelrodeodeljefe:
added commit for nextTick. Also a test on Windows is breaking for v6. We'll see.

addaleax:
Opened a PR for the windows issue at https://github.com/tapjs/spawn-wrap/pull/21. :)

eljefedelrodeodeljefe:
Awesome! CI Failure then unrelated...

bcoe:
executing on `next_tick` will definitely break the contract that yargs has with existing implementors, the expectation is that yargs would exit before the first line of user-code executes if an event has occurred that should cause yargs to terminate.

As usual, a busy week at npm but I will have some hacking time this weekend to land some Node 6 fixes with people's help 👍

eljefedelrodeodeljefe:
I dropped the commit then. I leave it be and defer to you w/ the fix then. Also reworded PR and commit to have friendlier language cc @addaleax

addaleax:
Yeah, sorry, I haven’t kept track of the discussion in Node core on how exiting is best handled in cases like this, so I probably won’t be much of a help here. And I’m not even sure it’s possible to do this correctly while still keeping the contract with dependents of yargs as mentioned by @bcoe… but I might be mistaken.

eljefedelrodeodeljefe:
Yeah no worries. Just mentioned you for the rename. Also I think your assumption is right.

bcoe:
@eljefedelrodeodeljefe moving some of the conversation here so I'm not spamming 1,000,000 people; I've got a hacking weekend planned this weekend, and will work on getting yargs patched such that it's backwards compatible.

Would love your help coming up with what the API should look like for `5.x`, definitely want to help people move in the correct direction.

eljefedelrodeodeljefe:
Yeh, ditto. It's not easy looking at Node's source, tbh. I am gonna examine it from all sides. We'll see...

bcoe:
@eljefedelrodeodeljefe closing in favor of this:

https://github.com/yargs/yargs/pull/501

1. I know I'm moments away from exiting, so it won't effect the upstream program to set `stdout/stderr` to blocking.
2. It ensures the program exits before the program using yargs has the opportunity to write any output.

eljefedelrodeodeljefe:
👍

PR #499, opened 2016-05-10
chore: use const as a semantic tool

jameswomack:
Philosophy behind this: http://womack.io/2016/01/05/Const-ant-Craving/
tldr;
* `const` has been supported in Node.js since at least 0.4
* `const` makes code more readable—it's clear a value won't change
* standard helps w/ this by pointing out when `const` is misused

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,14 +1,14 @@
 // classic singleton yargs API, to use yargs
 // without running as a singleton do:
 // require('yargs/yargs')(process.argv.slice(2))
-var yargs = require('./yargs')
+const yargs = require('./yargs')
 
 Argv(process.argv.slice(2))
 
 module.exports = Argv
 
 function Argv (processArgs, cwd) {
-  var argv = yargs(processArgs, cwd, require)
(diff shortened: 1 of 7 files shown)


lrlna:
That's cool! I think we were talking about whether or not we would want to be transitioning more into ES6 stuff back in February (can't remember what the jury decided on that). This could be a nice transition over ✨

bcoe:
@jameswomack this is great 👍 I have a big block of time to work on OSS on this weekend, and will land the heck out of tickets.

nexdrew:
TIL `const` goes back to Node 0.6. Wow.

Nice writeup @jameswomack. Good work!

jameswomack:
Thanks everyone. Yeah, I found [a Chromium/V8 issue](https://bugs.chromium.org/p/v8/issues/detail?id=327) regarding basic `const` support that's dated Apr 24, 2009!

bcoe:
rebased here: https://github.com/yargs/yargs/pull/502

PR #501, opened 2016-05-14
fix: make stdout flush on newer versions of Node.js

bcoe:
see #497

### DIFF
--- a/lib/blocking-io.js
+++ b/lib/blocking-io.js
@@ -1,0 +1,6 @@
+module.exports = function () {
+  if (typeof process.stdout._handle.setBlocking === 'function') {
+    process.stdout._handle.setBlocking(true)
+    process.stderr._handle.setBlocking(true)
+  }
+}
--- a/lib/blocking-io.js
+++ b/lib/blocking-io.js
@@ -1,5 +1,5 @@
 module.exports = function () {
-  if (typeof process.stdout._handle.setBlocking === 'function') {
+  if (process.stdout._handle && typeof process.stdout._handle.setBlocking === 'function') {
     process.stdout._handle.setBlocking(true)
     process.stderr._handle.setBlocking(true)
   }
--- a/lib/blocking-io.js
+++ b/lib/blocking-io.js
@@ -1,6 +1,0 @@
-module.exports = function () {
-  if (process.stdout._handle && typeof process.stdout._handle.setBlocking === 'function') {
-    process.stdout._handle.setBlocking(true)
-    process.stderr._handle.setBlocking(true)
-  }
-}
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -1,5 +1,6 @@
 // this file handles outputting usage instructions,
 // failures, etc. keeps logging in one place.
+var blockingIO = require('./blocking-io')
 var cliui = require('cliui')
 var decamelize = require('decamelize')
 var stringWidth = require('string-width')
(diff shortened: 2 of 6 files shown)


lib/blocking-io.js
```diff
@@ -0,0 +1,6 @@
+module.exports = function () {
+  if (process.stdout._handle && typeof process.stdout._handle.setBlocking === 'function') {
+    process.stdout._handle.setBlocking(true)
+    process.stderr._handle.setBlocking(true)
+  }
```
Fishrock123: Solution seems fine to me.

One note: If `_handle` exists it's always a `net.Socket` (or inherits from it), and thus `setBlocking()` should always exist.
(This has not been changed in years.)

Fishrock123: Actually, i'm curious, does this actually work when done retroactively?

If you have async writes and they you do this does it immediately cause a blocking flush on call? If you don't know, that's fine, I'll investigate.

bcoe: @Fishrock123 trying to go make sure yargs supports all the way back to `0.8` currently:

```js
Benjamins-MBP:yargs benjamincoe$ node --version
v0.8.28
Benjamins-MBP:yargs benjamincoe$ node
> process.stdout._handle.setBlocking
undefined
```

was hoping it would work retroactively, and tried this initially, it does not appear to. You need to have set blocking to true before you begin writing.

lib/blocking-io.js
```diff
@@ -0,0 +1,6 @@
+module.exports = function () {
+  if (process.stdout._handle && typeof process.stdout._handle.setBlocking === 'function') {
```
kzc: You have to check and call setBlocking for stdout and stderr separately. Otherwise it leads to this problem:
```
$ cat stdio.js
if (process.stdout._handle && typeof process.stdout._handle.setBlocking === 'function') {
  process.stdout._handle.setBlocking(true)
  process.stderr._handle.setBlocking(true)
}
process.stdout.write('Hello stdout.\n');
process.stderr.write('Hello stderr.\n');
process.exit(1);
```
```
$ node -v
v6.0.0
```
```
$ node stdio.js
Hello stdout.
Hello stderr.
```
```
$ node stdio.js >/dev/null
Hello stderr.
```
```
$ node stdio.js 2>/dev/null

# no output
```
```
$ node stdio.js 2>eee
$ cat eee
/tmp/stdio.js:3
  process.stderr._handle.setBlocking(true)
                        ^

TypeError: Cannot read property 'setBlocking' of undefined
```
See the fix used in uglify-js: https://github.com/mishoo/UglifyJS2/pull/1061/files

lib/blocking-io.js
Fishrock123: Looks like it was added in v0.11.2: https://github.com/nodejs/node/commit/20176a98416353d4596900793f739d5ebf4f0ee1 in response to https://github.com/nodejs/node-v0.x-archive/issue/3584

Fishrock123: I suppose this doesn't ever manifest as an issue before 0.11.2 then, but I haven't tested it.

lib/blocking-io.js
Fishrock123: Oh, yeah. Good catch.

bcoe: @Fishrock123 @kzc pulled the logic into its own module, gives us a nice central place to address any other edge-cases that come up, without the code hanging out in a bunch of random libraries:

https://github.com/yargs/set-blocking

PR #511, opened 2016-05-20
Rename zh.json to zh_CN.json

andyhu:
This is for simplified Chinese so the correct locale should be zh_CN. For traditional Chinese it should be something like zh_HK or zh_TW.


bcoe:
@andyhu thank you. I have a concern though, this could break existing apps that were relying on the `zh` locale name; perhaps we should just have two locale files? `zh.json` and `zh_CN.json`?

andyhu:
Hi @bcoe, I have the same concern. However I'm a Chinese native speaker and I rarely use the locale string 'zh' since it's quite different between Traditional Chinese and Simplified Chinese but as Chinese we can understand both of them. Or probably change it until next major release after something like zh_TW is available?

bcoe:
@andyhu let's just keep `zh.json` and `zh_CN.json` and in the next major we'll drop `zh.json`.

andyhu:
thanks @bcoe

bcoe:
@andyhu mind submitting a pull with both files, then I can give you credit for the work when I merge.

bcoe:
@andyhu thank you for the contribution \o/ closing in favor of https://github.com/yargs/yargs/issues/519 which keeps zh.json, and adds zh_CN.json.

bcoe:
@andyhu give this a spin, I've landed your changes finally:

PR #525, opened 2016-06-01
fix: drop unused camelcase dependency fixes #516

### LINKED ISSUE #516: Drop the camelcase dependency
Looks like an artifact from pre-`yargs-parser` extraction. Testing shows it can be safely dropped from `yargs`.

bcoe:


maxrimue:
LGTM :shipit:

Issue #486, opened 2016-04-20
global options are not affected by environment variables when using sub commands


kumar303:
When declaring a sub command that also has its own options, environment variables no longer affect global options. Is my configuration wrong?

````javascript
var argv = require('yargs')
  .env('MY_PROGRAM')
  .options({
    'f': {
      alias: 'fruit-thing',
      type: 'boolean',
      global: true,
    },
  })
  .command('cmd', '', function(yargs) {
    return yargs.options({
      'sub': {
        type: 'boolean',
      }
    });
  })
  .argv
console.log(argv);
````
I would expect this to print `{fruitThing: true}` but it does not:
````
MY_PROGRAM_FRUIT_THING=true node script.js cmd
````
However, if I comment out `yargs.options(...)` in the sub command, it works.

This is only limited to environment variables. The following works as expected:
````
node script.js --fruit-thing cmd
````

nexdrew:
@kumar303 Good find!

I think the problem is that the `envPrefix` is not treated as "global" - in the context of a command, it has been [reset to `undefined`](https://github.com/yargs/yargs/blob/v4.6.0/yargs.js#L116), meaning env vars will be ignored for that command.

To workaround this, you would have to duplicate your `.env('MY_PROGRAM')` config within each command builder.

I could've sworn there was already an issue for this, but I can't find it now. (Perhaps I'm just remembering conversations from other channels.)

I think the correct fix would be to preserve the `envPrefix` for commands, like a global option. Thoughts?

kumar303:
That fix sounds good to me. It seems like it should apply to sub commands by default. I can use the workaround in the meantime, thanks.

PR #526, opened 2016-06-01
Added documentation about commands not inheriting configurations

maxrimue:
As it seems, it's still unclear how commands actually behave when it comes to configuration. This is probably because of that information missing in the documentation.

These new lines are meant to make clear that commands do not inherit the configuration of their parent context, which means that users have to re-apply configurations they want to keep across scopes.

What does everyone think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -467,6 +467,9 @@ values stored in `argv._`).  Set `desc` to `false` to create a hidden command.
 Hidden commands don't show up in the help output and aren't available for
 completion.
 
+Note that commands will **not** automatically inherit the configuration of their
+parent context (eg. a parent command or global context). 
+
 Optionally, you can provide a `builder` object to give hints about the
 options that your command accepts:
 
--- a/README.md
+++ b/README.md
@@ -467,9 +467,6 @@ values stored in `argv._`).  Set `desc` to `false` to create a hidden command.
 Hidden commands don't show up in the help output and aren't available for
 completion.
 
-Note that commands will **not** automatically inherit the configuration of their
-parent context (eg. a parent command or global context). 
-
 Optionally, you can provide a `builder` object to give hints about the
 options that your command accepts:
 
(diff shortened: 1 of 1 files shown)


nexdrew:
I think we should list the configuration we expect to be applied to commands, something like:

- options marked `global` from any parent context
- `.help()` and `.version()` options (by default)
- `.wrap()` config

And there may be a couple others we should add to this, like [`.env()`](https://github.com/yargs/yargs/issues/486) and possibly [`.check()`](https://github.com/yargs/yargs/issues/403) (though I think we agreed the latter should not be inherited, and we should fix #403 with documentation added here).

bcoe:
@nexdrew perhaps submit a patch against @maxrimue's work, or come to a consensus in slack?

README.md
```diff
@@ -467,6 +467,9 @@ values stored in `argv._`).  Set `desc` to `false` to create a hidden command.
 Hidden commands don't show up in the help output and aren't available for
 completion.
 
+Note that commands will **not** automatically inherit the configuration of their
```
bcoe: Let's move this documentation below

> Optionally, you can provide a `builder` object to give hints about the options that your command accepts:

And let's make it sound a bit more positive, pointing folks towards the `globals` option 👍 

> By default only [global](#global) commands are inherited by the yargs instance passed to the builder.

maxrimue: @bcoe Did you mean `...only [global](#global) `**`options`**` are...`?

I like this approach, though I think mentioning inheritance of parent options (if you meant that) is only half of the story, since this PR is also heading to inform people that commands need to be reconfigured and don't just use the configuration of their parent context, such as `.strict()` needs to be re-applied to every sub command.

So let's combine informing about inheritance of the configuration, and the inheritance of options (also mentioning the global method), like this:

```
Note that commands will not automatically inherit _configuration or options_ of their parent context. 
This means you'll have to re-apply configuration if necessary, and make options global manually using 
the [global](#global) method.
```

bcoe: :shipit: works for me.

maxrimue:
I've tried to address your concerns with the latest commit, @nexdrew and @bcoe.

What do you think?

nexdrew:
@maxrimue Works for me, thanks!

README.md
```diff
@@ -481,6 +481,14 @@ yargs.command('get', 'make a get HTTP request', {
   .argv
 ```
 
+Note that commands will not automatically inherit configuration _or_ options
+of their parent context. This means you'll have to re-apply configuration
+if necessary, and make options global manually using the [global](#global) method.
+
+Additionally, the [`help`](#helpoption) and [`version`](#versionoption)
```
nexdrew: Looks like we might want to add explicit anchors for `.help()` and `.version()` in the README since these links don't seem to work in GitHub or npmjs.com.

maxrimue: @nexdrew Fixed!

PR #527, opened 2016-06-06
feat(command): derive missing command string from module filename

nexdrew:
Mentioned in #494, this PR makes a `command` string optional for module-based commands, using the base filename of the `require()`d module for the command when the `command` string is not specified/exported.

Not only does this get us one step closer to implementing #492, but it also paves the way for supporting an "auto" mode for `commandDir()`, where conventions can be used to determine subcommand hierarchies and command names can be defined by filenames of submodules.

In this implementation, a `command` string is derived by filename via the following:

- For `.command()`: filename is looked up in `require.cache` via the new [`which-module`](https://github.com/nexdrew/which-module) dependency
- For `.commandDir()`: filename is already available via the `require-directory` API

This logic only applies when the command module does not export a `command` string. For non-module command objects (ones that are not `require()`d), a descriptive error is thrown if a `command` string is not given.

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1,5 +1,7 @@
 const path = require('path')
+const inspect = require('util').inspect
 const requireDirectory = require('require-directory')
+const whichModule = require('which-module')
 
 // handles parsing positional arguments,
 // and populating argv with said positional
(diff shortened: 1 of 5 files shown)


bcoe:
Looks good to me :shipit:

PR #533, opened 2016-06-29
Show better output on command handler error

yofreke:
I had a similar issue to https://github.com/yargs/yargs/issues/373

My subcommands were throwing errors, and yargs was eating them and only printing the message (if any).  No stack trace, very sad.

This adds a field to errors resulting from `commandHandler.handler()` calls, `_yargsErrorName `.  The presence of this field causes usage to print the `_yargsErrorName` along with the original `err.stack`.

### Example:
Lets assume that the error I am throwing in my handler is `throw new Error('boop')` 

Old (not helpful):
``` bash
# Missing demand arg
$ my-command
<HELP TEXT>

# With demand arg, handler throws error
$ my-command --required-arg
<HELP TEXT>

boop
```

New (yes helpful):
``` bash
# Missing demand arg
$ my-command
<HELP TEXT>

# With demand arg, handler throws error
$ my-command --required-arg
<HELP TEXT>

COMMAND_HANDLER_ERROR: Error: boop
    at <STACK>
```

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -163,7 +163,12 @@ module.exports = function (yargs, usage, validation) {
     populatePositional(commandHandler, innerArgv, currentContext)
 
     if (commandHandler.handler) {
-      commandHandler.handler(innerArgv)
+      try {
+        commandHandler.handler(innerArgv)
+      } catch (err) {
+        err._yargsErrorName = 'COMMAND_HANDLER_ERROR'
+        throw err
+      }
     }
     currentContext.commands.pop()
     return innerArgv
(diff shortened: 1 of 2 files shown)


bcoe:
@yofreke I don't want to start tacking additional parameters on errors specifically for commands, I think this would need to be a larger discussion that we open an issue around and discuss -- we've also been fairly careful to not print messages directly to the console in yargs, folks can register a method for handling errors, which may or may not print to `stdout` or `stderr`:

https://github.com/yargs/yargs#failfn

Why don't you open a ticket, and we can discuss placing some additional context on error objects, and what this would look like? If we think it's something worth doing, I'd like to provide additional contextual information for more than just commands.

yofreke:
I was pretty sure this PR would get denied, but wanted to kick off a discussion about it anyways.  Moving discussion to a new issue!

Issue #529, opened 2016-06-13
.commandDir(dir) does not work when the yargs definition script is `require`-ed from other file


popkirby:
While I'm building a tool with `.commandDir(dir)`, an error occured when I require `lib/cli.js` and execute `bin/tool.js`.

`node lib/cli.js` works properly.

Directory tree:
```
.
├── bin
│   └── tool.js
├── lib
│   ├── cli.js
│   └── commands
│       └── foo.js
└── package.json

```

bin/tool.js:
```
#!/usr/bin/env node
require('../lib/cli.js');
```

lib/cli.js:
```
'use strict';
require('yargs')
.commandDir('commands')
.argv;
```

lib/commands/foo.js:
```
'use strict';
exports.command = 'foo';
exports.describe = 'description';
exports.builder = {};
exports.handler = function (argv) { console.log('foo') };

```

The error message is:
```
fs.js:945
  return binding.readdir(pathModule._makeLong(path), options.encoding);
                 ^

Error: ENOENT: no such file or directory, scandir '/Users/popkirby/Documents/yargs-bug/bin/commands'
    at Error (native)
    at Object.fs.readdirSync (fs.js:945:18)
    at requireDirectory (/Users/popkirby/Documents/yargs-bug/node_modules/require-directory/index.js:59:6)
    at Object.self.addDirectory (/Users/popkirby/Documents/yargs-bug/node_modules/yargs/lib/command.js:72:5)
    at Object.Yargs.self.commandDir (/Users/popkirby/Documents/yargs-bug/node_modules/yargs/yargs.js:211:13)
    at Object.<anonymous> (/Users/popkirby/Documents/yargs-bug/lib/cli.js:4:2)
    at Module._compile (module.js:541:32)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
```

nexdrew:
@popkirby I can confirm this is an issue, thanks for reporting! The problem is that yargs is interpreting a top-level `.commandDir()` call as relative to `require.main.filename`, which in this case is `/Users/popkirby/Documents/yargs-bug/bin/tool.js`. I will need to think about a proper solution.

In the meantime, you can workaround this by using `__dirname`. E.g. change **lib/cli.js** to this:

```js
'use strict';
var join = require('path').join;
require('yargs')
  .commandDir(join(__dirname, 'commands'))
  .argv;
```

leesei:
[...]

PR #542, opened 2016-07-07
feat: use readme from latest published version instead of master

nexdrew:
Instead of having Flatdoc render the README from master, let's query the registry for the `latest` published version of yargs and tell Flatdoc to use the corresponding tag in GitHub.

This should fix the [problem where ](https://github.com/yargs/yargs/issues/529#issuecomment-226997791) [our docs are out-of-sync ](https://github.com/yargs/yargs/issues/529#issuecomment-228461862) [with what's actually published](https://github.com/yargs/yargs/issues/529#issuecomment-229407166).

Note that this introduces a dependency on an external service (https://jsonp.afeld.me/) in order to wrap the JSON from the public npm registry in a JSONP response, to get around the lack of CORS support in the registry. We could use [YQL](https://developer.yahoo.com/yql/) as a "more stable" alternative, but that would require [attribution](https://developer.yahoo.com/attribution/) and I would prefer to support the [awesome open source](https://github.com/afeld/jsonp) service by @afeld instead. (If you agree, please go star the repo!)

### DIFF
--- a/index.html
+++ b/index.html
@@ -22,9 +22,13 @@
 
   <!-- Initializer -->
   <script>
-    Flatdoc.run({
-      fetcher: Flatdoc.github('bcoe/yargs')
-    });
+    // use https://jsonp.afeld.me/ to wrap registry json in jsonp
+    $.get('https://jsonp.afeld.me/?url=https://registry.npmjs.org/yargs', function (data) {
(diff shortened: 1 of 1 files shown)


bcoe:
🚢

Issue #521, opened 2016-05-30
benchmarks \o/


bcoe:
Someone on the [Internetz](https://twitter.com/RijnhardH/status/736199577360031745) complained that yargs is a slow ... you know what, they're probably right we've never made an effort to run many benchmarks (we're getting to the point of maturity where we probably should). 

_Let's run some benchmarks on yargs, comparing it against some other parsers as a baseline:_

- [ ] optimist
- [ ] commander
- [ ] argparse

_Let's fix a few of the glaring performance issues:_

 * I have my suspicions as to where the most likely performance bottlenecks would be (*cough* [cliui](https://github.com/yargs/cliui))

**This work could be pretty fun.**

thoughts, @nexdrew @maxrimue, @lrlna?

rijnhard:
Hi @bcoe thanks for the tweet. 

as promised, heres the benchmarking code I used: https://github.com/rijnhard/node-crappy-benchmarks

rijnhard:
Feel free to point out issues, also I would just like to give you some credit, for taking this approach. My tweets can be intentionally offensive, but I do have a lot of respect for open source maintainers, and yargs is a fantastic library.

bcoe:
@rijnhard no worries, this has actually given me the opportunity to learn a bit more about v8 CPU profiling -- unfortunately I haven't found any of the quick wins I was hoping for so far. I've benchmarked `cliui` at `2795hz`: that's pretty darn fast, and I couldn't find any quick performance tweaks.

I'll move on to profiling some other parts of yargs, since that didn't immediately jump out as the culprit.

rijnhard:
@bcoe i have updated with a bench of argparse. its slightly slower then Commander, but twice as fast as yargs

nexdrew:
Perhaps we should focus on [`yargs-parser`](https://github.com/yargs/yargs-parser) for the moment? There's definitely a lot of looping and RegExp checks in there.

shakyShane:
I switched to yargs for https://github.com/BrowserSync/browser-sync, but am considering switching away again do to these speed issues.

The elephant is in the room is the fact that from a cold boot, yargs requires 95 individual js files/modules just to get running - this should be addressed before anyone spends time reworking loops/regexes

```js
// require yargs, get the require cache, trim cwd
const yargs   = require('yargs');
const path    = require('path');
const keys    = Object.keys(require.cache);
const trimmed = keys.map(key => path.relative(path.join(process.cwd(), 'node_modules'), key));

[...]

PR #544, opened 2016-07-09
fix: lazy-load package.json and cache. get rid of pkg-conf dependency.

bcoe:
we dug into some of the performance concerns with yargs in #521, and came up with an initial quick win.

In this pull request we:

* eliminate `pkg-conf` in favor of using `read-pkg-up` exclusively.
* we lazy-load `read-pkg-up` (the require time for modules was leading to some of yargs' bad performance benchmarks).
* we cache the `package.json` the first time we load it.

- [x] published this to `next` tag and made sure that `pkgConf` and yargs configuration continues working as expected.

### DIFF
--- a/yargs.js
+++ b/yargs.js
@@ -7,8 +7,6 @@ const path = require('path')
 const Usage = require('./lib/usage')
 const Validation = require('./lib/validation')
 const Y18n = require('y18n')
-const readPkgUp = require('read-pkg-up')
-const pkgConf = require('pkg-conf')
 const requireMainFilename = require('require-main-filename')
 const objFilter = require('./lib/obj-filter')
 const setBlocking = require('set-blocking')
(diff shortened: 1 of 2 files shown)


bcoe:
reviewers: @nexdrew, @marneborn

bcoe:
tested this thoroughly on the `next` branch, please feel free to provide commentary after the fact -- I want to get us to a `4.8.0` release this afternoon, and start working on some of the `5.0.0` breaking changes.

yargs.js
```diff
@@ -345,19 +344,30 @@ function Yargs (processArgs, cwd, parentRequire) {
...
+      conf = obj[key]
       options.configObjects = (options.configObjects || []).concat(conf)
     }
 
     return self
   }
 
+  var pkg = null
+  function pkgUp (path) {
+    if (pkg && !path) return pkg
```
nexdrew: There's a slight bug with this logic. If `pkgUp` is called first with a `path` (as in `.pkgConf('kewl', process.cwd())`) and then subsequently without a path (as in `.version()`), the parsed pkg object returned will be from the call with the path, which is most likely the wrong package.json file.

To fix, we should probably cache a map of path -> pkg object and return the cached object that matches the given path (normalized to `'__no_path__'` or something similar).

PR #548, opened 2016-07-11
fix(commandDir): make dir relative to caller instead of require.main.filename

### LINKED ISSUE #529: .commandDir(dir) does not work when the yargs definition script is `require`-ed from other file

nexdrew:
Pull in [`get-caller-file`](https://github.com/stefanpenner/get-caller-file) to always interpret the directory given to `.commandDir()` as relative to the caller. This actually helps simplify the logic a good bit since I no longer need to keep track of a command-to-dir mapping between calls.

The `require` is deliberately lazy, so we only load the dependency if `.commandDir()` is called.

There are a few different packages that provide the "get filename of caller" functionality, but I chose `get-caller-file` because it was the only one that already had passing tests for the range of Node we need and includes Windows tests ([`caller`](https://github.com/totherik/caller) seems to be popular but is lacking the fuller range of testing, and [`caller-path` seems to be broken on Node 4+](https://github.com/sindresorhus/caller-path/issues/1)). An added benefit is that `get-caller-file` also seems to be the smallest and only pulls in one dependency.

Fixes #529.

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -42,10 +42,8 @@ module.exports = function (yargs, usage, validation) {
     }
   }
 
-  self.addDirectory = function (dir, context, req, mainFilename, opts) {
+  self.addDirectory = function (dir, context, req, callerFile, opts) {
     opts = opts || {}
-    // dir should be relative to the command module
-    dir = path.join(context.dirs[context.commands.join('|')] || '', dir)
     // disable recursion to support nested directories of subcommands
     if (typeof opts.recurse !== 'boolean') opts.recurse = false
     // exclude 'json', 'coffee' from require-directory defaults
@@ -62,14 +60,11 @@ module.exports = function (yargs, usage, validation) {
         if (~context.files.indexOf(joined)) return visited
         // keep track of visited files in context.files
         context.files.push(joined)
-        // map "command path" to the directory path it came from
-        // so that dir can be relative in the API
-        context.dirs[context.commands.concat(parseCommand(visited.command || commandFromFilename(filename)).cmd).join('|')] = dir
         self.addHandler(visited)
       }
       return visited
     }
-    requireDirectory({ require: req, filename: mainFilename }, dir, opts)
+    requireDirectory({ require: req, filename: callerFile }, dir, opts)
   }
 
   // lookup module object from require()d command and derive name
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   "dependencies": {
     "cliui": "^3.2.0",
     "decamelize": "^1.1.1",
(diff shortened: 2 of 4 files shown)


lib/command.js
```diff
@@ -62,14 +60,11 @@ module.exports = function (yargs, usage, validation) {
         if (~context.files.indexOf(joined)) return visited
         // keep track of visited files in context.files
         context.files.push(joined)
-        // map "command path" to the directory path it came from
-        // so that dir can be relative in the API
-        context.dirs[context.commands.concat(parseCommand(visited.command || commandFromFilename(filename)).cmd).join('|')] = dir
```
bcoe: I love how much this simplifies things; except for the fact that yargs' documentation is terrifyingly verbose at this point, I'm really feeling like the command functionality is knocking it out of the park -- phew, that was hard work wasn't it.

bcoe:
love it :shipit:

PR #549, opened 2016-07-11
feat(command): builder function no longer needs to return the yargs instance

### LINKED ISSUE #522: make command builder work without returning
I like the new commandDir feature but I have been having a bit of trouble getting a few of the other functions to work. The one I bumped into were `default` and `demand`. I am new to yargs, so I may just be 'holding it wrong'.

It starts with a cli.js
```js
#!/usr/bin/env node

const yargs = require( 'yargs' );

yargs // eslint-disable-line no-unused-expressions
  .commandDir( 'commands' )
  .help( 'h' )
  .alias( 'h', 'help' )
  .wrap( yargs.terminalWidth() )
  .version()
  .alias( 'version', 'v' )
  .argv
```

I have a folder called `commands` which holds two files: new.js and generate.js

```js
// new.js
exports.command = 'new <project>';
exports.description = 'description';
exports.builder = ( yargs ) => {
  yargs.example( 'example' )
    .option( 'B', {
      alias: 'b',
      description: 'description'
    } )
    .demand( 'B' ) //  or demand( 1, 'B' ) or demand( 1, [ 'B' ] )
    .strict();
};
exports.handler = ( argv ) => {
  // Handle
};
```
I expect it to skip the handle and present the user with the help menu when I type `cli new app`, but it goes right into the handler function (which breaks because I never expect argv.B to be empty)

In the second file, I have set the default values for two of the options but their respective keys on the argv object returned in the handler function contain no values if they aren't explicitly set. So: `cli generate file name` => `argv.p` and `argv.b` will be empty. 

```js
// generate.js
exports.command = 'generate <file> <name>';
exports.description = 'description';
exports.builder = ( yargs ) => {
  yargs.example( 'examplee' )
    .option( 'p', {
      alias: 'P',
      description: 'description',
      choices: [ 'firstChoice', 'secondChoice' ],
      default: 'firstChoice'
    } )
    .option( 'b', {
      alias: 'B',
      description: 'description',
      choices: [ 'firstChoice', 'secondChoice' ],
      default: 'firstChoice'
    } )
};

exports.handler = ( argv ) => {
  // Handle
};
```

Could I be doing something wrong here or should this work?

Thanks, Martien

nexdrew:
Fixes #522.

~~I'm not quite sure about the intentions of what should be allowed to be returned by the builder function. Returning the `yargs` instance or nothing should now work, but I wonder if we ever intended to allow it to return something like `yargs.argv`? If so, I think we will need a new issue for that.~~

Otherwise, this PR seems to work fine, but it could possibly result in unexpected behavior in user CLIs if they were previously not returning the `yargs` instance, either by mistake or not. For that reason, this change may be better suited for 5.x.

Note that I skipped the `populates argv with placeholder keys when passed into command handler` unit test because this change breaks it and it no longer seems to be valid (options must now be marked `global: true` in order to apply to the argv of a command handler, be it a placeholder or populated value). We should either remove it (if everyone agrees the test is no longer valid), or fix it by using `global: true` (which is tested elsewhere).

**UPDATE**: With commit 739a4f4, a builder function can also call `yargs.argv`, with or without returning. I think this is a good idea b/c I've seen a few folks [try to do this](https://github.com/yargs/yargs/issues/565) and it makes the API much less rigid. 😎

### DIFF
--- a/test/command.js
+++ b/test/command.js
@@ -493,4 +493,14 @@ describe('Command', function () {
     var argv = y.parse(['foo', 'bar'])
     argv.second.should.equal('bar')
   })
+
+  // addresses https://github.com/yargs/yargs/issues/522
+  it('does not require builder function to return', function () {
+    var argv = yargs('yo')
+      .command('yo [someone]', 'Send someone a yo', function (yargs) {
+        yargs.default('someone', 'Pat')
+      })
+      .argv
+    argv.should.have.property('someone').and.equal('Pat')
+  })
 })
--- a/test/command.js
+++ b/test/command.js
@@ -499,8 +499,32 @@ describe('Command', function () {
     var argv = yargs('yo')
       .command('yo [someone]', 'Send someone a yo', function (yargs) {
         yargs.default('someone', 'Pat')
+      }, function (argv) {
+        argv.should.have.property('someone').and.equal('Pat')
       })
       .argv
     argv.should.have.property('someone').and.equal('Pat')
+  })
+
+  it('allows builder function to parse argv without returning', function () {
+    var argv = yargs('yo Jude')
+      .command('yo <someone>', 'Send someone a yo', function (yargs) {
+        yargs.argv
+      }, function (argv) {
+        argv.should.have.property('someone').and.equal('Jude')
+      })
+      .argv
+    argv.should.have.property('someone').and.equal('Jude')
+  })
+
+  it('allows builder function to return parsed argv', function () {
+    var argv = yargs('yo Leslie')
+      .command('yo <someone>', 'Send someone a yo', function (yargs) {
+        return yargs.argv
+      }, function (argv) {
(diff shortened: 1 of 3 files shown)


test/command.js
```diff
@@ -491,4 +491,13 @@ describe('Command', function () {
     output.should.include('bar')
     output.should.not.include('foo')
   })
+
+  it('does not require builder function to return', function () {
+    var argv = yargs('yo')
+      .command('yo [someone]', 'Send someone a yo', function (yargs) {
```
bcoe: how is `yo` app doing, have they raised a _1 billion $$$_ series B yet?

bcoe:
Is the solution this simple, I could have sworn I thought of an edge-case that would bite us and forced the `5.x` release; I'll look at this pull when I'm fresh on the weekend -- I think we're really close to `5.x` either way.

PR #552, opened 2016-07-13
fix: cache pkg lookups by path to avoid returning the wrong one

nexdrew:
Fixes the small bug mentioned here: https://github.com/yargs/yargs/pull/544/files#r70469421

Includes test 😎


bcoe:
@nexdrew wonderful 🎉 glad you caught this.

Issue #539, opened 2016-07-06
Better error reporting for subcommands


yofreke:
Continuation from: https://github.com/yargs/yargs/pull/533

------

>we've also been fairly careful to not print messages directly to the console in yargs

I totally understand, but with the current implementation you are sending the very minimally helpful (if at all) error message to stderr.  This looks almost like a bug to an end user, certainly not a thought out design decision (in my example the text "boop" is appended to the user defined help text).

>folks can register a method for handling errors

The problem with `.fail()` is that you have to register it on every command / subcommand.  This is super cumbersome for an end user who just wants to see a call stack.  I will have to dig up the example that was failing for us, but we also found that the fail handler would not be called, regardless of where it was registered (will try to find the example, and maybe open a separate issue regarding that).

I really love using yargs, it makes building a CLI 10x simpler.  The problem is that it really falls on its face when a command throws an unexpected error.  Depending on how cryptic the error message is (does it even look like an error when appended to help text?), and how savvy the programmer is (do they crawl the yargs docs to find `.fail`? Do they realise that this is an error in their code and not yargs, since they see the same error output either way?), this is they turn away and start using something else.

The real proposition is this:

- If there is an error parsing cli args
  - throw an error explicitly saying that
  - show help text
  - highlight the argument that threw an error
- If there is a missing cli arg:
  - throw an error explicitly saying that
  - show help text
  - highlight the missing arg
- If there is an error running the command handler logic
  - throw an error explicitly saying that
  - do not show red herring help text
- If there is an error anywhere in user code
  - show a full stack trace
  - do not show red herring help text

evanderkoogh:
I completely agree that the reporting of failures is pretty much broken now.
I was very much confused when just getting the message when things went wrong. Went through all the docs, found the fail handler, implemented it and turns out it doesn't get called.

I use the following config:
```
const yargs = require('yargs')
yargs
	.command(require('./cmds/deploy.js'))
	.command(require('./cmds/build.js'))
	.demand(1, 'Need to specify at least an command')
	.fail(function(msg, err) {
		console.log('blah');
		if(err) {
			throw err
		}
		console.log(msg);
		process.exit(1);
	})
	.strict()
	.help()
	.wrap(100)
	.argv

```
I have tried adding:

```
exports.fail = function(msg, err) {
	console.log('blah');
	if(err) {
		throw err
	}
	console.log(msg);
	process.exit(1);
}
```

to the deploy.js file, but that one also does not get called.

[...]

PR #553, opened 2016-07-13
feat: apply .env() globally

### LINKED ISSUE #486: global options are not affected by environment variables when using sub commands

nexdrew:
Fixes #486.

This means that a configured `envPrefix` will not be reset for nested commands.

This could potentially be considered a breaking change since env vars that used to be ignored for the execution of command handlers may no longer be ignored. 😮 Otherwise, this could be considered a fix. 😕

### DIFF
--- a/test/command.js
+++ b/test/command.js
@@ -547,6 +547,23 @@ describe('Command', function () {
     })
   })
 
+  // see: https://github.com/yargs/yargs/pull/553
+  it('preserves top-level envPrefix', function () {
+    process.env.FUN_DIP_STICK = 'yummy'
+    process.env.FUN_DIP_POWDER = 'true'
+    yargs('eat')
+      .env('FUN_DIP')
+      .global('stick') // this does not actually need to be global
+      .command('eat', 'Adult supervision recommended', function (yargs) {
+        return yargs.boolean('powder').exitProcess(false)
+      }, function (argv) {
+        argv.should.have.property('powder').and.be.true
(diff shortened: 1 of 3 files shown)


bcoe:
@nexdrew I'm comfortable calling this a fix, I think this is closer to what the expected behavior would be -- can't think of a case where you'd want to drop `.env` variables on their face.

bcoe:
@nexdrew alternatively, we could just call this the first `5.x` patch, what do you think. There's also:

https://github.com/yargs/yargs/issues/539

which relates to the fact that `fail()` should probably be global -- I'd be pretty comfortable with trying to get to a `5.x` release this weekend ... and if we quickly want to release a `6.x` release afterwards, so be it.

PR #556, opened 2016-07-15
add config lookup for .implies()

### LINKED ISSUE #547: .implies() ignores parser preferences
According to the readme, you can influence the behaviour of yargs-parser by [providing a custom config](https://github.com/yargs/yargs#configuration). One thing you can configure is boolean negation, which is turned on by default, and means that `--no-foo` gets parsed as `foo: false`. By turning this off, `--no-foo` would get parsed as `no-foo: true`. This configuration and the expected outcome however is ignored by yargs' `.implies()`.

`implies()` expects two keys and requires the second to be true if the first is, too. It has extended capabilities because it also supports boolean negation, so it always sees a `--no-` not as part of the key that you want to check itself. This should however only be the case if `boolean-negation` wasn't turned off in the parser configuration. This is not caused by a bug, but rather by a missing feature, because a configuration lookup was never implemented at this point. See the necessary lines [here](https://github.com/yargs/yargs#configuration).

maxrimue:
This will eventually fix #547 

This PR is still a WIP as the new added test passes regardless of whether `boolean-negation` is turned on or off, however, I'd love to hear what folks think about the general approach already.

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -219,8 +219,8 @@ module.exports = function (yargs, usage, y18n) {
     const implyFail = []
 
     Object.keys(implied).forEach(function (key) {
-      let booleanNegation
-      if (yargs.getOptions().configuration['boolean-negation'] === 'false') {
+      var booleanNegation
+      if (yargs.getOptions().configuration['boolean-negation'] === false) {
         booleanNegation = false
       } else {
         booleanNegation = true
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -221,6 +221,7 @@ module.exports = function (yargs, usage, y18n) {
     Object.keys(implied).forEach(function (key) {
       var num
       const origKey = key
+      const booleanNegation = yargs.getOptions().configuration['boolean-negation']
       var value = implied[key]
 
       // convert string '1' to number 1
@@ -230,7 +231,7 @@ module.exports = function (yargs, usage, y18n) {
       if (typeof key === 'number') {
         // check length of argv._
         key = argv._.length >= key
-      } else if (key.match(/^--no-.+/)) {
+      } else if (key.match(/^--no-.+/) && !booleanNegation) {
         // check if key doesn't exist
         key = key.match(/^--no-(.+)/)[1]
         key = !argv[key]
@@ -244,7 +245,7 @@ module.exports = function (yargs, usage, y18n) {
 
       if (typeof value === 'number') {
         value = argv._.length >= value
-      } else if (value.match(/^--no-.+/)) {
+      } else if (value.match(/^--no-.+/) && !booleanNegation) {
         value = value.match(/^--no-(.+)/)[1]
         value = !argv[value]
       } else {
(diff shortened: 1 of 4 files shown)


test/validation.js
```diff
@@ -57,6 +57,23 @@ describe('validation tests', function () {
         })
         .argv
     })
+
+    it('should ignore the --no- part if boolean negation is disabled', function (done) {
```
bcoe: If the test isn't async, you can just not accept `done` as an argument, and then drop the line:

```js
return done()
```

bcoe: rather than:

`it('should ignore the --no- part if boolean negation is disabled')`

I might put:

`it('does not treat --no- as a special case, if boolean negation is disabled')`,

Had to read the current test description a few times, I might just be nit-picking though :)

bcoe:
@maxrimue this is looking quite clean, great work \o/

really curious as to why you're not able to get the test to fail! I'm sure it will be something silly.

maxrimue:
@bcoe 
I discovered some logical mistakes and fixed them now. However, there is one big problem: I discovered that `.config()` and `.pkgConf()` actually can't be used to set the parser configuration. Yargs only uses `pkgUp` to determine the configuration that is then handed over to the parser ([as seen here](https://github.com/yargs/yargs/blob/master/yargs.js#L639)).

So the only problem left now is to change the configuration in the test that I've created to `boolean-negation: false`... Am I missing something, or should we introduce a coding way to change the parser configuration?

lib/validation.js
```diff
@@ -219,6 +219,12 @@ module.exports = function (yargs, usage, y18n) {
     const implyFail = []
 
     Object.keys(implied).forEach(function (key) {
+      let booleanNegation
```
bcoe: this doesn't work in older versions of Node.js, so you'll want to use var.

bcoe:
@maxrimue turning on and off yargs' parsing options is not handled with `pkgConf`:

the `yargs` stanza in your application can be used to turn on and off settings for yargs itself, whereas `pkgConf` allows you to expose a key for your own libraries to manage configuration:

* the library [nyc](https://github.com/istanbuljs/nyc) uses `pkgConf` to configure an `nyc` flag that provides configuration.
* whereas, a library that uses yargs, and wants to turn off some of its parsing settings would use the `yargs` stanza in their package.json to specify this.

for your test, rather than using `pkgConf`, try specifying a `cwd` for yargs:

`yargs(['--foo', '--bar', 'false'], './test/fixtures')`

^ by doing this, it should load the `yargs` stanza from your configuration file.

maxrimue:
@bcoe Had to change a line to make the `pkgUp()` call aware of the customisable `cwd` variable, but thanks to your suggestion everything works now! Ready for final review.

bcoe:
@maxrimue this looks great to me, there's another pull of @nexdrew's I'd like to land (on yargs-parser) and a fix that I'd like to land for nyc that I just opened, then I'll publish all three fixes to `next` and we can try things out on a few projects.

bcoe:
@maxrimue give this a spin:

Issue #455, opened 2016-03-26
Sync error of handler should throw a global error


satazor:
```
// foo.js
'use strict';

const yargs = require('yargs');

yargs
.help('help').alias('help', 'h')
.command('foo', 'Foo command is awesome', () => {
    console.log('foo handler');
    throw new Error('This error will show below the help');
})
.argv;
```

If you now run `node foo` it will show the usage and the error below.

<img width="716" alt="screen shot 2016-03-26 at 21 39 45" src="https://cloud.githubusercontent.com/assets/1017236/14062557/8585162c-f39b-11e5-8f4c-603e4e8a8f64.png">

I was expecting the error to be thrown globally and not being handled by yargs.

yargs version: 4.3.2

nexdrew:
@satazor In yargs 4.x, the first function (3rd argument) is a builder (configuration) and you should specify a second function (4th argument) as the handler.

What happens if the error is thrown from the handler function instead? Does it then do what you expect?

```js
yargs.command('foo', 'Foo awesome', (yargs) => {
  return yargs
}, (argv) => {
  throw new Error('error message')
})

```

satazor:
Sorry, I made a mistake in the example. Though the output of the sample code above with the handler still has the same output:

```js
'use strict';

const yargs = require('yargs');

yargs
.help('help').alias('help', 'h')
.command('foo', 'Foo awesome', (yargs) => yargs, (argv) => {
    console.log('foo handler');
    throw new Error('error message');
})
.argv;
```

<img width="679" alt="screen shot 2016-03-26 at 21 58 14" src="https://cloud.githubusercontent.com/assets/1017236/14062640/e11c0192-f39d-11e5-9ddd-06eed319470c.png">

nexdrew:
Ok, I see. If you want to customize this behavior, I think you need to define your own `.fail()` function.

```js
yargs.fail((msg, err) => {
  // do something here, maybe throw err
})
```

Let us know how that works for you.

satazor:
I see, but is this the correct default behaviour? I was not expecting `yargs` to be handling handler errors. For instance, the handler contains a programming error, the consumer of the CLI would be looking at help usage + `Cannot access X of undefined`.

AFAIK, this kind of stuff was ok in v3 but changed in v4.

nexdrew:
[...]

Issue #427, opened 2016-03-10
Allow customising of "extra" help


donaldpipowitch:
When I use certain things like choices or default values the outputted help looks quite convoluted to me. It would be nice to optionally hide that output of format it differently.

E.g. I have this:

```
Options:
  -l, --log-level  Set log level
      [choices: "silly", "verbose", "info", "warn", "error", "silent"] [default:
                                                                         "info"]
  -v, --version    Show version number                                 [boolean]
  -h, --help       Show help                                           [boolean]

```

Maybe something like this is nicer:

```
Options:
  -l, --log-level  Set log level (default: "info")
                   [ "silly", "verbose", "info", "warn", "error", "silent" ]
  -v, --version    Show version number                                 [boolean]
  -h, --help       Show help                                           [boolean]
```

```
Options:
  -l, --log-level  Set log level (defaults to "info" or use "silly", "verbose", "warn", "error", "silent")
  -v, --version    Show version number                                 [boolean]
  -h, --help       Show help                                           [boolean]
```

```
Options:
  -l, --log-level  Set log level (defaults to "info")
  -v, --version    Show version number                                 [boolean]
  -h, --help       Show help                                           [boolean]
```

```
Options:
  -l, --log-level  Set log level (info,silly,verbose,warn,error,silent)
  -v, --version    Show version number                                 [boolean]
  -h, --help       Show help                                           [boolean]
```

lrlna:
@donaldpipowitch I am really liking your suggestions! A more readable version is always :100: for me. 

This could work as a verbose version for our help output, for sure.

KohPoll:
+1

skygragon:
exactly, the choices line is too long to read, why not break them into different lines to make our eyes  feel better?

PR #562, opened 2016-07-17
feat: provide suggestions if unknown command is called

### LINKED ISSUE #496: recommend similar commands
Some command-line interfaces recommend similar commands if the command entered is not understood (and a command is required).

This would be a slick extension to our command handling.

maxrimue:
This closes #496.

This PR adds a method `.recommendCommands()`, that, if turned on, will use a very small library called [didyoumean](https://www.npmjs.com/package/didyoumean) to determine and show similar named, registered commands, if a command gets called that is unknown to yargs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1347,6 +1347,12 @@ as a configuration object.
 `cwd` can optionally be provided, the package.json will be read
 from this location.
 
+.recommendCommands(boolean)
+---------------------------
+
+If this feature is turned on, yargs will provide suggestions for similiar commands if a command is called
+that wasn't defined in your configuration. By default, this feature is disabled.
+
 .require(key, [msg | boolean])
 ------------------------------
 .required(key, [msg | boolean])
--- a/README.md
+++ b/README.md
@@ -1352,6 +1352,7 @@ from this location.
 
 If this feature is turned on, yargs will provide suggestions for similiar commands if a command is called
 that wasn't defined in your configuration. By default, this feature is disabled.
+Alternatively, you can also pass no argument to enable this feature.
 
 .require(key, [msg | boolean])
 ------------------------------
--- a/locales/de.json
+++ b/locales/de.json
@@ -34,5 +34,5 @@
   "Path to JSON config file": "Pfad zur JSON-Config Datei",
   "Show help": "Hilfe anzeigen",
   "Show version number": "Version anzeigen",
-  "Did you mean %s?": "Meinten Sie %s?"
+  "Did you mean %s?": "Meintest du %s?"
 }
--- a/locales/de.json
+++ b/locales/de.json
@@ -33,5 +33,6 @@
   "Invalid JSON config file: %s": "Fehlerhafte JSON-Config Datei: %s",
   "Path to JSON config file": "Pfad zur JSON-Config Datei",
   "Show help": "Hilfe anzeigen",
-  "Show version number": "Version anzeigen"
+  "Show version number": "Version anzeigen",
+  "Did you mean %s?": "Meinten Sie %s?"
 }
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -270,6 +270,17 @@ describe('yargs dsl tests', function () {
         .argv
     })
 
+    it('recommends a similiar command if no command handler is found', function () {
+      var r = checkOutput(function () {
+        yargs(['boat'])
+          .command('goat')
+          .recommendCommands(true)
+          .argv
+      })
+
+      r.logs[0].should.match(/Did you mean 'goat'/)
+    })
+
     it("skips executing top-level command if builder's help is executed", function () {
       var r = checkOutput(function () {
         yargs(['blerg', '-h'])
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -270,7 +270,7 @@ describe('yargs dsl tests', function () {
         .argv
     })
 
-    it('recommends a similiar command if no command handler is found', function () {
+    it('recommends a similar command if no command handler is found', function () {
       var r = checkOutput(function () {
         yargs(['boat'])
           .command('goat')
@@ -279,6 +279,28 @@ describe('yargs dsl tests', function () {
       })
 
       r.logs[0].should.match(/Did you mean goat/)
+    })
+
+    it('does not recommend a similar command if a "false" is passed as a parameter', function () {
+      var r = checkOutput(function () {
+        yargs(['boat'])
+          .command('goat')
+          .recommendCommands(false)
+          .argv
+      })
+
+      r.logs.should.be.empty
+    })
+
+    it('does not recommend a similiar command if no similar command exists', function () {
+      var r = checkOutput(function () {
+        yargs(['foo'])
+          .command('nothingSimilar')
+          .recommendCommands()
+          .argv
+      })
+
+      r.logs.should.be.empty
     })
 
     it("skips executing top-level command if builder's help is executed", function () {
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -278,7 +278,7 @@ describe('yargs dsl tests', function () {
           .argv
       })
 
-      r.logs[0].should.match(/Did you mean 'goat'/)
+      r.logs[0].should.match(/Did you mean goat/)
     })
 
     it("skips executing top-level command if builder's help is executed", function () {
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -274,7 +274,7 @@ describe('yargs dsl tests', function () {
       var r = checkOutput(function () {
         yargs(['boat'])
           .command('goat')
-          .recommendCommands(true)
+          .recommendCommands()
           .argv
       })
 
--- a/yargs.js
+++ b/yargs.js
@@ -10,6 +10,7 @@ const Y18n = require('y18n')
 const requireMainFilename = require('require-main-filename')
 const objFilter = require('./lib/obj-filter')
 const setBlocking = require('set-blocking')
+const didYouMean = require('didyoumean')
 
 var exports = module.exports = Yargs
 function Yargs (processArgs, cwd, parentRequire) {
@@ -125,6 +126,7 @@ function Yargs (processArgs, cwd, parentRequire) {
 
     exitProcess = true
     strict = false
+    recommendCommands = false
     completionCommand = null
     self.parsed = false
 
@@ -285,6 +287,12 @@ function Yargs (processArgs, cwd, parentRequire) {
 
   self.getDemanded = function () {
     return options.demanded
+  }
+
+  var recommendCommands
+  self.recommendCommands = function (value) {
+    recommendCommands = !!value
+    return self
   }
 
   self.requiresArg = function (requiresArgs) {
@@ -661,6 +669,11 @@ function Yargs (processArgs, cwd, parentRequire) {
       if (~handlerKeys.indexOf(cmd) && cmd !== completionCommand) {
         setPlaceholderKeys(argv)
         return command.runCommand(cmd, self, parsed)
+      } else {
+        if (recommendCommands) {
+          const similiarCommand = didYouMean(cmd, handlerKeys)
+          console.log('Did you mean \'' + similiarCommand + '\'?')
+        }
       }
     }
 
--- a/yargs.js
+++ b/yargs.js
@@ -125,7 +125,6 @@ function Yargs (processArgs, cwd, parentRequire) {
     command = command ? command.reset() : Command(self, usage, validation)
     if (!completion) completion = Completion(self, usage, command)
 
-    exitProcess = true
     strict = false
     recommendCommands = false
     completionCommand = null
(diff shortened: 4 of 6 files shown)


yargs.js
```diff
@@ -661,6 +669,11 @@ function Yargs (processArgs, cwd, parentRequire) {
       if (~handlerKeys.indexOf(cmd) && cmd !== completionCommand) {
         setPlaceholderKeys(argv)
         return command.runCommand(cmd, self, parsed)
+      } else {
+        if (recommendCommands) {
+          const similiarCommand = didYouMean(cmd, handlerKeys)
+          console.log('Did you mean \'' + similiarCommand + '\'?')
```
bcoe: let's pull this into our translation files, and make sure we support multiple languages 👍 -- we'll need to start passing the `y18n` dependency into this module like we do for `usage`, etc.

test/yargs.js
```diff
@@ -270,6 +270,17 @@ describe('yargs dsl tests', function () {
         .argv
     })
 
+    it('recommends a similiar command if no command handler is found', function () {
```
bcoe: this is such a clean implementation! great work.

locales/de.json
```diff
@@ -33,5 +33,6 @@
   "Invalid JSON config file: %s": "Fehlerhafte JSON-Config Datei: %s",
   "Path to JSON config file": "Pfad zur JSON-Config Datei",
   "Show help": "Hilfe anzeigen",
-  "Show version number": "Version anzeigen"
+  "Show version number": "Version anzeigen",
+  "Did you mean %s?": "Meinten Sie %s?"
```
bcoe: hooray! wonder what a good way to request more translations would be? perhaps reach back out to some of the original folks ... might be worth eyeballing if we want any more new text in the application first though, get it done in one chunk.

maxrimue: Agreed. I'm not 100% sure about the German translation I provided, for your information, the "Sie" means "you", but there's actually two you's in German, there's "Sie" and "du" which I could both use here, "Sie" is really formal, "du" really personal, what do you think?

bcoe: Did you mean is fairly conversational in English, so I'd lean towards more personal.

yargs.js
```diff
@@ -661,6 +671,11 @@ function Yargs (processArgs, cwd, parentRequire) {
       if (~handlerKeys.indexOf(cmd) && cmd !== completionCommand) {
         setPlaceholderKeys(argv)
         return command.runCommand(cmd, self, parsed)
+      } else {
+        if (recommendCommands) {
+          const similiarCommand = didYouMean(cmd, handlerKeys)
```
nexdrew: Perhaps we should defer the `require('didyoumean')` until it's actually used here. That would avoid loading an unused dependency in cases where the CLI doesn't even use commands.

nexdrew: Did you mean `similarCommand` 😉 ?

yargs.js
```diff
@@ -661,6 +671,11 @@ function Yargs (processArgs, cwd, parentRequire) {
...
+          console.log(__('Did you mean %s?', similiarCommand))
```
nexdrew: This can output `Did you mean null?`, we should probably avoid that.

nexdrew: We should probably move this out of the `argv._` loop to avoid outputting multiple "did you mean"s. I would think the first encountered non-null recommendation would take precedence.

yargs.js
```diff
@@ -287,6 +291,12 @@ function Yargs (processArgs, cwd, parentRequire) {
     return options.demanded
   }
 
+  var recommendCommands
+  self.recommendCommands = function (value) {
+    recommendCommands = !!value
```
nexdrew: I'd be tempted to interpret `. recommendCommands()` (with no arg) as enabling this functionality. The only other API method we have that only takes a boolean/truthy/falsey value is `detectLocale` which is enabled by default (so calling `.detectLocale()` there doesn't really make sense).

Perhaps we could change this line to be `recommendCommands  = value !== false` instead, so that it's enabled for no args or any arg that is not explicitly `false`.

nexdrew:
Should we consider doing `if (exitProcess) process.exit(1)` when recommending similar commands? Or perhaps attempt a tighter integration with `.strict()` by deferring the recommendation or appending it to the error message?

yargs.js
```diff
@@ -125,6 +128,7 @@ function Yargs (processArgs, cwd, parentRequire) {
 
     exitProcess = true
     strict = false
+    recommendCommands = false
```
nexdrew: Should `recommendCommands` be considered global, i.e. should it be preserved for commands instead of reset to false?

nexdrew:
Note that this functionality currently doesn't support nested subcommands very well. In the following example, it actually outputs `Did you mean null?` for every command level (potentially multiple times), even when the correct commands are given:

```js
require('yargs')
  .command('one', 'one', function (yargs) {
    return yargs.command('two', 'two', function (yargs) {
      return yargs.command('three', 'three', {}, function (argv) {
        console.log('three handler')
      }).recommendCommands(true)
    }).recommendCommands(true)
  })
  .recommendCommands(true)
  .argv
```

```
$ node recommend.js one two three
Did you mean null?
Did you mean null?
Did you mean null?
three handler
```

I think the check for similar commands needs to be context-aware somehow, so that we don't attempt to suggest a command at the wrong level.

maxrimue:
Okay, so I've updated the code a little, with the following changes:
- The whole suggestion logic now happens inside a global function (makes it reusable if needed)
- Aside from booleans, you can now provide no parameter to `.recommendCommands()` as well (this would trigger it to be activated)
- The suggestions now only are provided if `didYouMean` finds a similar command
- `similiar` typo has been fixed

There are two things I didn't implement yet because I'd like to hear the opinions from others about it first:
- Should we exit if recommendation happens? (Only `if strict`, or `if exitProcess`?)
- Should the `recommendCommands` setting apply globally or be reset? 

I'll also add some more tests.

bcoe:
## Random Thought

@nexdrew, @maxrimue this is a bit of an aside:

I keep reading random tweets, etc., with folks complaining about yargs having too many dependencies:

https://twitter.com/mcclure111/status/754325494754078721

This relates somewhat to folks' complaints about yargs performance.

## Where I'm Going With This

Whenever we look to pull in a new library, like `didyoumean`, or `require-dir`, I'm slightly hesitant due to these sorts of comments.

@nexdrew in the past, you've suggested the idea of making yargs more modular, what do you think this would look like? I wonder if in `yargs@5.x` we could split functionality up a bit more, make more if it optional.

nexdrew:
@bcoe I started documenting/mapping-out the interface/contracts between the yargs API and its submodules to try and get a better idea about what overrides/plugins might look like, but I'm afraid there's quite a bit of coupling that would have to be undone, which is probably the wrong approach.

IMO the things folks most want to cut out or customize are probably (in no particular order):

1. Cut out package.json lookups, for performance reasons (make them opt-in)

    Potentially removes the need for `read-pkg-up` and `require-main-filename`.

2. Customize help text, mostly layout/formatting but possibly text and everything else too

    Potentially removes the need for `cliui`, `decamelize`, `string-width`, and `window-size`. Some level of text customization is currently possible via `.updateLocale()`, but customizing other things like formatting of examples and [display of choices](https://github.com/yargs/yargs/issues/427) is not.

3. Customize error handling

    You can pretty much do this with a custom `.fail()` handler, but [it's not currently global for commands](https://github.com/yargs/yargs/issues/539) and [can be less than ideal](https://github.com/yargs/yargs/issues/455).

4. Make i18n support opt-in, since most folks probably don't need it

    Potentially removes the need for `os-locale` and `y18n`.

5. Make support for commands opt-in

    Potentially removes the need for `get-caller-file`, `require-directory`, and `which-module`. I really want to make support for command-based CLIs top notch in yargs, but I'd bet a lot of CLIs don't need it (i.e. do their own `argv._` check) or only use flagged args/options.

6. Make support for bash completion opt-in

    Most CLIs probably don't use it.

7. Customize parsing logic

    I actually think yargs does a really good job at this, but we need to improve on coercion and handling of arrays. I think a system that lets you swap out `yargs-parser` for something else would be cool, but I don't think that's very practical.

Perhaps if we could identify the smallest possible API that covers 80% of CLIs, we could then work out a system where the other 20% is covered by explicit opt-ins (which includes most of the larger features listed above). I'm thinking that would take care of most people's gripes about yargs, and it would probably also help us develop a clearer interface for overrides/plugins.

My spur-of-the-moment ideas of what this might look like in code is:

```js
// start with minimal API, no opt-ins
var yargs = require('yargs')
  .option('f', {
    alias: 'file',
    describe: 'A file to load'
  })

// add some opt-ins: Exhibit A
require('yargs-commands')(yargs)
  .command('load', 'Load a file', {}, (argv) => {
    console.log('loading', argv.file)
  })
require('yargs-completion')(yargs)
  .completion()
require('yargs-i18n')(yargs)

// add some opt-ins: Exhibit B
yargs
  .withCommands(require('yargs-commands'))
  .command('load', 'Load a file', {}, (argv) => {
    console.log('loading', argv.file)
  })
  .withCompletion(require('yargs-completion'))
  .completion()
  .withI18n(require('yargs-i18n'))

// add some opt-ins: Exhibit C
yargs
  .use(require('yargs-commands'))
  .use(require('yargs-completion'))
  .use(require('yargs-i18n'))
  .command('load', 'Load a file', {}, (argv) => {
    console.log('loading', argv.file)
  })
  .completion()
```

Not sure which, if any, make for the best system. I'm not crazy about the idea of splitting up yargs into multiple opt-in packages, but I think it's the only way to cut down on the number of dependencies, and thus overall size, of "core" yargs. If it was only performance that mattered, then we could just start lazy-loading as much as possible and make certain defaults (like package lookups) be enabled via explicit API method.

Other ideas welcomed.

 these opt-in features out into their own packages would allow us to cut down on the number of dependencies needed in "core" yargs, but we could possibly still include all deps and just lazy-load them

yargs.js
```diff
@@ -287,6 +290,25 @@ function Yargs (processArgs, cwd, parentRequire) {
...
+      recommendCommands = true
+    } else {
+      recommendCommands = false
+    }
+    return self
+  }
+
+  self.recommendCommand = function (cmd, handlerKeys) {
+    if (recommendCommands) {
+      const similarCommand = require('didyoumean')(cmd, handlerKeys)
```
nexdrew: I think we play with different `threshold` option values here, since the default doesn't seem to support short names very well, like `one` or `list`.

yargs.js
```diff
@@ -287,6 +290,25 @@ function Yargs (processArgs, cwd, parentRequire) {
...
+      if (similarCommand) {
+        console.log(__('Did you mean %s?', similarCommand))
```
nexdrew: I'm not sure I would do the `console.log` here. If the purpose of this method is reusability, then it would probably make more sense to just return the value and let the consumer log. Otherwise, we might not need this method.

README.md
```diff
@@ -1347,6 +1347,12 @@ as a configuration object.
 `cwd` can optionally be provided, the package.json will be read
 from this location.
 
+.recommendCommands(boolean)
```
nexdrew: We should probably note that the boolean arg is optional, and calling this without it enables the functionality.

yargs.js
nexdrew: I think a threshold of `0.7` seems to work best for short command names:

```js
const didyoumean = require('didyoumean')
didyoumean.threshold = 0.7
const similarCommand = didyoumean(cmd, handlerKeys)
```

nexdrew:
Based on my testing, I think the `recommendCommands` setting should definitely be global.

yargs.js
maxrimue: @nexdrew I agree that to make `self.recommendCommand` more reusable inside yargs, it makes sense to just return the determined value. However, if `.recommendCommands` is enabled, I'd still log it to the console since I think the idea behind this method is to assist the user. I can't really think of an advantage when `.recommendCommands()` just returns you a recommended command and you still have to log it, since that's what I think will happen anyway. What do you think?

bcoe:
----
@nexdrew I think we should, for posterity, record this great writeup somewhere.

My gut is that, in the short term, trying to lazy-load as much as possible continues to be a great first step -- if we did decide to further break out the yargs API ... I'd still want the core install to be fairly _batteries included_.

Take the the `i18n` support -- I definitely have seen yargs embraced in foreign markets, I think that little considerations like `y18n` go a long way.
--- 

@maxrimue this is looking great, I'm going to tweak a couple minor things on this branch and land it as I push us towards `5.x` over the next couple weeks.

nexdrew:
@bcoe I'll create a new issue for past and future discussion of breaking up yargs into a set of smaller, composable modules.

In terms of `.recommendCommands()`, I think it will pretty much be good to go if we change it to output __at most__ one recommendation per run. (I also think it would be smart to make it context-aware, instead of looking at all possible command args on each invocation of `parseArgs`, but that can be modified later.)

PR #567, opened 2016-07-21
fix out-of-date examples in yargs docs

lrlna:
Since we switched how this worked in the winter; also reflects how we document [demand here](https://github.com/yargs/yargs#demandcount-max-msg)

easy peasy ☀️


nexdrew:
Actually, with `.command()`, yargs will not

1. reset itself,
2. reparse argv, and
3. perform validation

when "running" the command unless some kind of builder is given, either `(yargs) => yargs` or `_.identity` or `{}` at a minimum. AFAICT, this has been true since yargs 4.0.0, meaning that this documented example:

```js
var argv = require('yargs')
  .command('install', 'tis a mighty fine package to install')
  .demand(1, ['w', 'm'])
  .strict()
  .argv
```

has probably never worked properly - at least not when `install` is given on the command line. I also notice that there are no tests that use both `.command()` and `.demand()` that expect failure due to required options. The only thing I see with this combination is `no failure occurs if the required arguments and the required number of commands are provided.` in test/usage.js.

As yargs stands today, I think the minimal example is something like this:

```js
var argv = require('yargs')
  .command('install', 'tis a mighty fine package to install', {})
  .demand(1, ['w', 'm'])
  .global(['w', 'm'])
  .strict()
  .argv
```

And `.strict()` is only useful for making sure the first non-flagged argument is `install`.

Similarly, I think the line_count.js example should be something like this:

```js
#!/usr/bin/env node
var argv = require('yargs')
    .usage('Usage: $0 <command> [options]')
    .command('count [options]', 'Count the lines in a file', {})
    .option('f', {
      alias: 'file',
      describe: 'Load a file',
      required: true,
      nargs: 1,
      global: true
    })
    .demand(1)
    .example('$0 count -f foo.js', 'count the lines in the given file')
    .help().alias('h', 'help')
    .epilog('copyright 2015')
    .argv

var fs = require('fs')
var s = fs.createReadStream(argv.file)

var lines = 0
s.on('data', function (buf) {
    lines += buf.toString().match(/\n/g).length
})

s.on('end', function () {
    console.log(lines)
})
```

```
$ node line_count.js 
Usage: line_count.js <command> [options]

Commands:
  count [options]  Count the lines in a file

Options:
  -f, --file  Load a file                                             [required]
  -h, --help  Show help                                                [boolean]

Examples:
  line_count.js count -f foo.js  count the lines in the given file

copyright 2015

```

```
$ node line_count.js count --file line_count.js
28
```

```
$ node line_count.js count -f line_count.js
28
```

lrlna:
well didn't I open a can of warms with our docs! 

I think we do have a [couple](https://github.com/yargs/yargs/blob/master/test/validation.js#L119) of [tests](https://github.com/yargs/yargs/blob/master/test/validation.js#L109) for this: 

> there are no tests that use both .command() and .demand() that expect failure due to required options.

Unless you're talking about something else and I missed the point 🙊 


And 👍  on the `line-count.js` example; I should have probably just refactored it while I was at it. It's pretty darn messy. And doing both demands in one example is a little bit unnecessary, you're right.

nexdrew:
@lrlna Neither of those tests use the `.command()` method, but if you add `.command('marsupial')` to [this one](https://github.com/yargs/yargs/blob/v4.8.1/test/validation.js#L109), then it fails (times out). :/

More clearly though, we should have a test like this, which is similar to the line_count.js example:

```js
it('fails when a required argument is missing with command given', function () {
  var failMsg
  yargs('count')
    .demand(1, ['f'])
    .command('count', 'Get line count of file')
    .fail(function (msg) {
      failMsg = msg
    })
    .argv
  expect(failMsg).to.equal('Missing required argument: f')
})
```

A test like this currently fails unless you modify it to be this:

```js
it('fails when a required argument is missing with command given', function () {
  var failMsg
  yargs('count')
    .demand(1, ['f'])
    .global('f')
    .command('count', 'Get line count of file', function (yargs) {
      return yargs.fail(function (msg) {
        failMsg = msg
      })
    })
    .argv
  expect(failMsg).to.equal('Missing required argument: f')
})
```

Without a test like that, we're not accurately testing the line_count.js example.

Before yargs 4, a `.command()` without a builder was just used for help text, but it seems we changed that - either intentionally or not - without going back and checking all our documented examples. :/ I don't think any one person is to blame here, it's just something that happens when  making big changes.

Maybe now that we have an awesome yargs website, we should look into breaking the documentation into separate pages and not rely so much on one big README? Just thinking out loud here.

bcoe:
@nexdrew @lrlna oh my goodness, I couldn't be more in agreement that our README is _too_ big; I've been thinking with IstanbulJS, it would be good to have targeted documentation based on some common scenarios (https://github.com/istanbuljs/istanbuljs.github.io/issues):

* a section on the simplest possible usage.
* perhaps a section on commands.
* a section on configuration.

lrlna:
Ok, I renamed this PR. Let's just fix all of the outdated examples we have in readme here. That's first step. 

Next step is fixing the whole documentation thing on the website. We've been talking about it forever, let's actually start making it happen.

@nexdrew , hmmm you know I think I was fixing that at the same time as @bcoe was fixing commands and we were pulling it in at the same time pretty much. My tests were totally based on pre-4.x version of yargs, which is basically butts and is confusing now. Woops! 
But also, `demand` is a freakin' mess.

PR #580, opened 2016-08-07
feat: adds recommendCommands() for command suggestions

bcoe:
working from @maxrimue's branch for `recommendCommands()`, and implementing @nexdrew's code-review.

This pull request gets us to `v1` of command recommendations (thanks @maxrimue for the hard work) addressing some of the nits in #562 (@maxrimue sorry for @nexdrew and I providing such a difficult review).

Here's a breakdown of some of the changes made:

* rather than pulling in another dependency, I simply added a helper module for calculating levenshtein distance: this is a fairly simple algorithm, and I could imagine us using it in other places -- might as well dance around adding another dependency to yargs.
* we now handle the edge-case that @nexdrew outlined with nested arguments significantly better, by avoiding making recommendations until after we've walked through all the command handlers.
* we now recommend commands using the validation module, this allows us to piggyback off existing functionality in `usage.js`, and `validation.js`.
* we now use the same method signature as `strict()` for `recommendCommands()` rather than trying to be overly fancy with argument overriding.

reviewers: @maxrimue @nexdrew I plan on stubbornly marching us towards `yargs@5.x` over the next week, appreciate your patience.

### DIFF
--- a/yargs.js
+++ b/yargs.js
@@ -287,6 +287,25 @@ function Yargs (processArgs, cwd, parentRequire) {
     return options.demanded
   }
 
+  var recommendCommands
+  self.recommendCommands = function (value) {
+    if (value === undefined || !!value === true) {
+      recommendCommands = true
+    } else {
+      recommendCommands = false
+    }
+    return self
+  }
+
+  self.recommendCommand = function (cmd, handlerKeys) {
+    if (recommendCommands) {
+      const similarCommand = require('didyoumean')(cmd, handlerKeys)
+      if (similarCommand) {
+        console.log(__('Did you mean %s?', similarCommand))
+      }
+    }
+  }
+
   self.requiresArg = function (requiresArgs) {
     options.requiresArg.push.apply(options.requiresArg, [].concat(requiresArgs))
     return self
@@ -675,6 +694,7 @@ function Yargs (processArgs, cwd, parentRequire) {
       return argv
     }
 
+<<<<<<< HEAD
     if (argv._.length) {
       // check for helpOpt in argv._ before running commands
       // assumes helpOpt must be valid if useHelpOptAsCommand is true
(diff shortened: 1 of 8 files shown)


bcoe:
TODO: sort by length before applying algorithm, so that we take the least-changes/longest word first.

yargs.js
```diff
@@ -707,6 +713,10 @@ function Yargs (processArgs, cwd, parentRequire) {
         }
       }
 
+      // recommend a command if recommendCommands() has
+      // been enabled, and no commands were found to execute
+      if (recommendCommands && argv._.length) validation.recommendCommands(argv._[argv._.length - 1], handlerKeys)
```
nexdrew: It looks like this is called even when `argv._` only contains valid commands or arguments (called as part of running the command); it just potentially doesn't have any commands to match against.

nexdrew: The `argv._.length` check is actually redundant.

nexdrew:
I would probably change [lines 706 - 718](https://github.com/yargs/yargs/blob/e8914a56cacec64eab8e3531b07fd519ce447b0f/yargs.js#L706-L718) to this:

```js
      // if there's a handler associated with a
      // command defer processing to it.
      var handlerKeys = command.getCommands()
      if (handlerKeys.length) {
        var firstUnknownCommand
        for (var i = 0, cmd; (cmd = argv._[i]) !== undefined; i++) {
          if (~handlerKeys.indexOf(cmd) && cmd !== completionCommand) {
            setPlaceholderKeys(argv)
            return command.runCommand(cmd, self, parsed)
          } else if (!firstUnknownCommand && cmd !== completionCommand) {
            firstUnknownCommand = cmd
          }
        }

        // recommend a command if recommendCommands() has
        // been enabled, and no commands were found to execute
        if (recommendCommands && firstUnknownCommand) {
          validation.recommendCommands(firstUnknownCommand, handlerKeys)
        }
      }
```

What do you think?

PR #582, opened 2016-08-09
feat: interpret demand() numbers as relative to executing command

### LINKED ISSUE #451: Make .demand() depend on context?
This question was raised in #428.

As of right now, `.demand()`, when called, does not care about the context it was called in, meaning that if you `.demand(1)` inside a command, it won't throw, because the the command itself satisfies `.demand()` already:
```javascript
var yargs = require('yargs')(['lol'])
        .command('lol', '', (yargs) => yargs.demand(1)) // Won't throw
        .argv;
```

However, you could argue that if used inside a command, `.demand()` should ignore the command itself so that the upper example won't work anymore.

How does everybody think about this? Should `.demand()` be contextual, stay as it is, or even become more configurable to allow both?

nexdrew:
Fixes #451.

Previously, min and max values passed to `.demand()` were interpreted as the _total_ number of non-option arguments, which included the arguments representing the command(s) itself.

With this PR, the min and max values are now interpreted as the _additional_ number of non-option arguments, relative to the currently executing command.

Here's an example:

```js
require('yargs')
  .command(
    'lint',
    'Lint a single file',
    (yargs) => yargs.demand(1, 1),
    (argv) => console.log('linting file:', argv._[1])
  )
  .argv
```

Before this PR, the command would not work as expected (i.e. would not accept any flagless args) because the `lint` command itself would satisfy the min and max on the command line.

```console
$ program lint
linting file: undefined

$ program lint example.js
program lint

Too many non-option arguments: got 2, maximum of 1

$ program lint one.js two.js
program lint

Too many non-option arguments: got 3, maximum of 1
```

After this PR, the command works by requiring a single flagless arg on the command line besides the `lint` command itself.

```console
$ program lint
program lint

$ program lint example.js
linting file: example.js

$ program lint one.js two.js
program lint

Too many non-option arguments: got 2, maximum of 1
```

When `.demand()` is used outside of a command builder, the functionality remains the same.

Note that this is a **BREAKING CHANGE** since users relying on the previous (now deemed incorrect) functionality will need to change the values they pass to `.demand()` within command builders, decrementing them by the "nested/depth count" of the command.

Also note that this **does not** change the contents of `argv._` within the command handler; `argv._` still contains the command key(s) it took to execute the command(s). (Note the use of `argv._[1]` in the above example.)

We could possibly change this (which would be another breaking change), but I'd rather leave it this way to potentially support command handlers that are reusable across multiple commands (allowing the handler to slightly change its functionality based on the command keys present in `argv._`).

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -11,18 +11,19 @@ module.exports = function (yargs, usage, y18n) {
   // arguments were provided, i.e., '_'.
   self.nonOptionCount = function (argv) {
     const demanded = yargs.getDemanded()
-    const _s = argv._.length
+    // don't count currently executing commands
+    const _s = argv._.length - yargs.getContext().commands.length
 
     if (demanded._ && (_s < demanded._.count || _s > demanded._.max)) {
       if (demanded._.msg !== undefined) {
         usage.fail(demanded._.msg)
       } else if (_s < demanded._.count) {
         usage.fail(
-          __('Not enough non-option arguments: got %s, need at least %s', argv._.length, demanded._.count)
+          __('Not enough non-option arguments: got %s, need at least %s', _s, demanded._.count)
         )
       } else {
         usage.fail(
-          __('Too many non-option arguments: got %s, maximum of %s', argv._.length, demanded._.max)
+          __('Too many non-option arguments: got %s, maximum of %s', _s, demanded._.max)
         )
       }
     }
--- a/test/validation.js
+++ b/test/validation.js
@@ -135,6 +135,30 @@ describe('validation tests', function () {
         })
         .argv
     })
+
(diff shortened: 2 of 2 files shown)


mathieumg:
Just got bit by this not being marked as a breaking change in the v5 changelog, must have slipped through the cracks! :laughing:

nexdrew:
@mathieumg Whoops, you're right, looks like we forgot to use the `BREAKING CHANGE` marker in the merged commit, sorry about that!

bcoe:
@nexdrew @mathieumg mind updating the CHANGELOG for v5 post-hoc, save others :)

mathieumg:
@bcoe Done!

PR #595, opened 2016-08-17
Added appveyor support for modern node versions

amilajack:


bcoe:
@amilajack "node" should be pulling in whatever version of Node is most modern on Travis; so I don't know if it's worth explicitly adding `6` into the `.travis.yml`.

Having said that, it's interesting that things failed on `6.4.0`; which is what your tests ran on.

bcoe:
Sure enough, there's a regression between `v6.3.1` and `v6.4.0.

maxrimue:
@amilajack As @bcoe explained, we already tests against Node.js v6 with our current config, but we definitely appreciate you taking the time to create this pull request, so thank you anyway :+1:

PR #597, opened 2016-08-17
Fix tests when run without TTY

addaleax:
- Fix how the process/io mocking in `test/helpers/utils.js` handles synchronous uncaught exceptions by adding a `finally` handler resetting the mocked state.
- Skip two tests that would throw an exception when the test output is redirected to a non-TTY (e.g. `npm test | cat`).

Ref: https://github.com/nodejs/node/issues/8138


test/helpers/utils.js
```diff
@@ -40,18 +40,27 @@ exports.checkOutput = function (f, argv, cb) {
 
     f()
   } else {
-    result = f()
+    try {
+      result = f()
+    } finally {
+      reset()
```
nexdrew: If `f()` doesn't throw, `reset()` is called twice. But I guess twice is better than never. ;)

nexdrew:
LGTM 👍 Thanks, @addaleax!

test/helpers/utils.js
addaleax: Yeah, figured as much… do you feel like this is something you’d like to see addressed or are you fine with it this way?

addaleax:
Is the Windows failure something known or should I take a closer look at it?

addaleax:
Oh… the Windows test failure is real, it just didn’t show up in CI because the Windows tests have the exact same problem CITGM was having, too.

test/helpers/utils.js
nexdrew: Seeing as how this code doesn't change much and the operation is idempotent, I'm fine leaving it this way.

addaleax:
Ok, I’m out of luck for the current AppVeyor failure, that probably has to be left for someone with proper access to a Windows machine…

maxrimue:
@addaleax I can look into it tomorrow :+1:

addaleax:
Nice, thanks. :)

maxrimue:
@addaleax I haven't fully figured out the problem yet, but I know for sure already that tests only fail in the CI, those two failing tests do not fail on my local machine, and they also fail on every Node.js from 0.10 to 6

addaleax:
Turns out also skipping a third test for the combination of Windows & Node v0.10 does the trick. I think that’s okay? At least everything’s green now except for the known problems with Node 6.4.0! :tada:

maxrimue:
@addaleax Great! :+1:

nexdrew:
CI is green, I vote squash and merge

addaleax:
Cool, thanks!

Issue #540, opened 2016-07-06
command should ignore spaces


hax:
Currently extra spaces will be treat as a required positional arguments which is very confuse.

```js
yargs.command('cmd   [arg]')
```

```sh
$ test cmd

Commands:
  cmd   [arg]

Not enough non-option arguments: got 0, need at least 2


$ test cmd a b

{ _: [ 'cmd' ], '$0': '', '': 'b' }
```

bcoe:
👍 sounds like something we should definitely fix.

lrlna:
hey @charlotteis , can we get this tweeted on [YourFirstPr](https://twitter.com/yourfirstpr) ? ✨

Charlotteis:
@lrlna https://twitter.com/yourfirstpr/status/755733795207282688 ✨

najamkhn:
@lrlna @hax on it. :)

jmbyerly:
Hi Yargs-ers! So I think I've fixed the bug, but as a first-time contributor I wanted to check in first before submitting a PR.

When reproducing the bug, I noticed that a similar problem occurs when the user passes in a required positional argument with extra spaces on either side of it: it returns a command with the required argument that was actually passed in the command string and an additional required argument whose value is an empty string for every extra blank space in the command string. 

For the test I wrote to show that the bug is fixed, I used the code snippet given above when the issue was opened, the one without a required positional argument but with an extra space. The bug fix solves both problems but I was wondering if I should write a second test for the related case I described, or if I write one test, which one I should use.

This is actually my first attempted contribution to an OSS project of any kind other than my own, so I figured it was best to play it safe and ask lots of questions. Thanks for the guidance. Love the project!

maxrimue:
Hey @jmbyerly, sounds awesome! I think you can definitely submit your PR with both tests. As you can see, we have _a lot_ of tests in yargs to always make sure everything works as expected in various situations, so it's definitely appreciated if you add more than just one test to the PR, since that increases the number of situations we test for!

Lastly, we don't really care too much about having too many tests like some other folks might do, we actually embrace it. Thanks for looking into this issue!

jmbyerly:
Okay, great! I'll submit a pull request for this tonite then. Thanks!

bcoe:
thanks @jmbyerly!

Issue #596, opened 2016-08-17
Different util.inspect output making tests fail


maxrimue:
As #595 has shown, builds now fail with Node.js v6.4.0.

As it turns out, the tests fail because of [this one test](https://github.com/yargs/yargs/blob/master/test/yargs.js#L421). It tests for yarg's output when a command is supplied that does not have a name. When yargs is about to output that error, it also uses `util.inspect` to show the content of the supplied command object [(reference)](https://github.com/yargs/yargs/blob/master/lib/command.js#L70).

The test output shows that this very output is expected to be slightly different. The original object is:
```javascript
{ desc: 'A command with no name',
  builder: [Function],
  handler: [Function] }
```

Interestingly, using `util.inspect()` on this object generates different outputs on different minors of Node.js v6:

v6.3.1:
```
'{ desc: \'A command with no name\',\n  builder: [ [Function: Function] ],\n  handler: [ [Function: Function] ] }'
```

v6.4.0:
```
'{ desc: \'A command with no name\',\r\n  builder: [ [Function: Function] ],\r\n  handler: [ [Function: Function] ] }'
```

I'm not sure if this is intended behaviour, so I'll report it on the Node.js repo.

maxrimue:
Turns out this is an intended change, so we have to change our tests a little to make them work with both results

maxrimue:
Update: This change in Node.js v6.4.0 is now being considered breaking by some maintainers, and a pull request has been proposed to revert these changes. [(PR)](https://github.com/nodejs/node/pull/8141#issuecomment-240428265)

PR #600, opened 2016-08-18
#540 - Changed parsing of the command string to ignore extra spaces

jmbyerly:
Changed the parsing of the command string to ignore extra spaces between positional arguments, rather than create an additional required positional argument with a value of an empty string for each extra space. Wrote a test to confirm the fix.

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -85,7 +85,8 @@ module.exports = function (yargs, usage, validation) {
   }
 
   function parseCommand (cmd) {
-    var splitCommand = cmd.split(/\s/)
+    var extraSpacesStrippedCommand = cmd.replace(/\s{2,}/g, ' ')
+    var splitCommand = extraSpacesStrippedCommand.split(/\s/)
     var bregex = /\.*[\][<>]/g
     var parsedCommand = {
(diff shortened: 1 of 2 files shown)


maxrimue:
This looks great, thanks for working out the fix! :+1: Btw, if you wonder why Travis failed, it's not because of your code but because of #596

jmbyerly:
Ah okay, yeah I was wondering about that. I checked out some of the other open pull requests and saw that they were throwing the same error so I was guessing (and hoping!) that it was unrelated to my code. Thanks for confirming, that's definitely a relief! Glad my first pull request went off without a hitch!

bcoe:
@jmbyerly great contribution, tests much appreciated.

bcoe:
@jmbyerly could you please give this a spin, I'd like to roll this out soon to get rid of a deprecation notice:

thanks for the contribution.

jmbyerly:
@bcoe Apologies, I'm new to coding and am a bit unclear as to what you mean. Should I run this on my local copy of the repo and verify that it's running without the deprecation notice? I'm happy to help out any way that I can.

PR #602, opened 2016-08-19
ci: test with node 6 once 6.4.0 is patched

### LINKED ISSUE #596: Different util.inspect output making tests fail

nexdrew:
Reverts the Travis and AppVeyor changes deliberately introduced with #597.

It is expected that tests with Node 6 will fail until version 6.4.0 is patched with this reversion: nodejs/node#8143

This PR serves as a reminder to revert the temporary testing change pinned to Node 6.3.1.

Fixes #596.

Thanks to @maxrimue for originally reporting the error, and thanks to @addaleax for fixing our tests with citgm and AppVeyor!


addaleax:
btw, I know this isn’t communicated suuuper-well, but Node v5 is already past its end of life, as far as Node core is concerned.

nexdrew:
@addaleax Hmm, good point. I'll drop Node 5 with this too. :+1:

addaleax:
v6.5.0 is out, so kicking off a new CI with this should return everything to green :)

nexdrew:
@addaleax Thanks for the heads up, will push a new commit dropping Node 5 very soon.

nexdrew:
Looks like it's gonna fail with Node 6.5.0 since the Function descriptors now include the function name (e.g. `builder: [Function: builder]` instead of `builder: [Function]`).

Time to relax the regex methinks.

addaleax:
Oh, yeah, I forgot that… the Node.js core tests had to be fixed up for that, too. You’re probably right.

nexdrew:
Tests are green again so I'm gonna squash-and-merge this. Thanks to @addaleax and @maxrimue for your help!

maxrimue:
Great, thanks @nexdrew :tada:

PR #604, opened 2016-08-21
docs: mention config file names must end in .json

wmhilton:
This one drove me crazy until I finally figured it out. Because [the default configParser implementation](https://github.com/yargs/yargs-parser/blob/cc4dc5663bf9a29527127cf575d37b461c1c47a2/index.js#L426) loads the file using node's `require`, it breaks if your config file is named `.my-app.conf` (for instance). Short term solution: just mention that! (It's easy enough to workaround using a custom parser function.)


bcoe:
@wmhilton thanks for the doc update 👍

PR #607, opened 2016-08-26
fix: upgrade standard, and fix appveyor config so that it works with newest standard

bcoe:


maxrimue:
@bcoe What do you think about this part of [my comment](https://github.com/yargs/yargs/pull/606#issuecomment-242171547)?
> What I was wondering btw., should we maybe drop tests for Default Azure too? Since AppVeyor only lists Windows Server 2012 R2 under available operating systems, I think Default Azure just currently means Windows Server 2012 R2, too, which would mean we run the same tests twice.

I saw we have the same approach in `yargs-parser`'s AppVeyor config, too.

bcoe:
@maxrimue seems reasonable to also default to Azure.

PR #610, opened 2016-08-29
Defer windowWidth() to improve perf for non-help usage

nolanlawson:
Hi :wave:, I'm a big fan of yargs, so after talking with @lrlna about https://github.com/yargs/yargs/issues/521, I wanted to try my hand at improving the overall runtime of `require('yargs')`.

I started by profiling with the Chrome dev tools using `node-nightly` and the `--inspect` option, where I noticed that `windowWidth()` is a very expensive function:

screenshot 2016-08-29 09 12 12

Since `windowWidth()` only has to be calculated when we need the column width, e.g. when the user has typed `--help` or `-h`, I figured we could lazily evaluate it. That's what this PR does.

Here is a very small benchmark to demonstrate the perf improvement:

```bash
time for i in {1..500}; do node -e 'require("./index")'; done
```

Running in node 5.12.0 on a 2013 Macbook Air, I got the following results (real time, averaged):

* **Before:** 122.14ms
* **After:** 108.746ms

Since that includes the cost of the Node process itself, we can also use `performance-now` to measure just the runtime of `require('yargs')`:

```bash
for ((i=0;i<500;i++)); do node -e 'var now = require("performance-now"); var s = now(); require("./index"); var e = now() - s; console.log(e)'; done | awk '{ sum+=$1 } END { print sum/NR }'
```

This gives: 

* **Before:** 41.3002ms
* **After:** 28.5307ms

So in rough terms `require('yargs')` should be about ~30% faster with this PR, with an average of ~12ms shaved off each invocation of `require('yargs')`. (Obviously it doesn't make any difference when doing `-h` or `--help`, but my hunch is that that's not the average use case.)

Please let me know if I missed something in the style/coverage/etc. or if this PR is unfeasible for some other reason I didn't notice. Hope this helps!

### DIFF
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -153,7 +153,7 @@ module.exports = function (yargs, y18n) {
 
       commands.forEach(function (command) {
         ui.div(
-          {text: command[0], padding: [0, 2, 0, 2], width: maxWidth(commands) + 4},
+          {text: command[0], padding: [0, 2, 0, 2], width: maxWidth(commands, theWrap) + 4},
           {text: command[1]}
         )
       })
@@ -227,7 +227,7 @@ module.exports = function (yargs, y18n) {
         ].filter(Boolean).join(' ')
 
         ui.span(
-          {text: kswitch, padding: [0, 2, 0, 2], width: maxWidth(switches) + 4},
+          {text: kswitch, padding: [0, 2, 0, 2], width: maxWidth(switches, theWrap) + 4},
           desc
         )
 
@@ -248,7 +248,7 @@ module.exports = function (yargs, y18n) {
 
       examples.forEach(function (example) {
         ui.div(
-          {text: example[0], padding: [0, 2, 0, 2], width: maxWidth(examples) + 4},
+          {text: example[0], padding: [0, 2, 0, 2], width: maxWidth(examples, theWrap) + 4},
           example[1]
         )
       })
@@ -267,7 +267,7 @@ module.exports = function (yargs, y18n) {
 
   // return the maximum width of a string
   // in the left-hand column of a table.
-  function maxWidth (table) {
+  function maxWidth (table, theWrap) {
     var width = 0
 
     // table might be of the form [leftColumn],
@@ -284,7 +284,6 @@ module.exports = function (yargs, y18n) {
 
     // if we've enabled 'wrap' we should limit
     // the max-width of the left-column.
-    var theWrap = getWrap()
     if (theWrap) width = Math.min(width, parseInt(theWrap * 0.5, 10))
 
     return width
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -96,9 +96,21 @@ module.exports = function (yargs, y18n) {
     epilog = msg
   }
 
-  var wrap = windowWidth()
+  var wrapSet
+  var wrap
   self.wrap = function (cols) {
+    wrapSet = true
     wrap = cols
+  }
+
+  function getWrap () {
+    // lazily call windowWidth() because it's very expensive,
+    // and only needs to be called if the user wants to show usage/help
+    if (!wrapSet) {
+      wrap = windowWidth()
+    }
+    wrapSet = true
+    return wrap
   }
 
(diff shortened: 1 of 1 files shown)


maxrimue:
This. is. awesome.
Thank you so much @nolanlawson :tada:

lib/usage.js
```diff
@@ -96,11 +96,23 @@ module.exports = function (yargs, y18n) {
     epilog = msg
   }
 
-  var wrap = windowWidth()
+  var wrapSet
```
nexdrew: I would initialize this to `false`, for clarity.

lib/usage.js
```diff
@@ -96,11 +96,23 @@ module.exports = function (yargs, y18n) {
...
     wrap = cols
   }
 
+  function getWrap () {
+    // lazily call windowWidth() because it's very expensive,
+    // and only needs to be called if the user wants to show usage/help
+    if (!wrapSet) {
+      wrap = windowWidth()
+    }
+    wrapSet = true
```
nexdrew: Can we make this the first line under the `if (!wrapSet)` condition instead? Seems equivalent but more efficient that way.

lib/usage.js
```diff
@@ -122,9 +134,10 @@ module.exports = function (yargs, y18n) {
         return acc
       }, {})
     )
+    var theWrap = getWrap()
```
nexdrew: To avoid redundant calls to `getWrap()` in the private `maxWidth` function, we could just pass `theWrap` to it as an additional argument.

nexdrew:
Besides my nit-picky inline comments, this LGTM! Thanks @nolanlawson!

lib/usage.js
nolanlawson: agreed, good catch!

lib/usage.js
nolanlawson: Sure thing, probably won't affect perf but makes the code clearer anyway.

lib/usage.js
nolanlawson: OK, will do!

nolanlawson: On second thought, looks like it would add `getWrap()` to 3 other places in the code, which is a big ugly, so gonna suggest we just leave it as-is. Cost of calling `getWrap()` multiple times should be minimal.

nexdrew: Not following you. The other 3 places are just calls to `maxWidth(something)` which all happen within the public `.help()` method. I'm suggesting we stick with the one call here (line 137) and then pass `theWrap` as a 2nd argument to `maxWidth(something, theWrap)` for the other 3 (lines 156, 230, and 251), which would remove the need to call `getWrap()` again within the `maxWidth` function itself.

Think that will make the code too ugly or brittle?

nolanlawson: Ah okay, gotcha now. In that case, I would go one step further and just calculate `theWrap` at the top of the `.help()` method. Then there's no need to add a 2nd argument to `maxWidth()`.

nexdrew: Well, `maxWidth` is called but not _defined_ within `.help()`, that's why the extra arg would be necessary. But if we just moved it, then yes. Either way, I'm all for it.

Sorry to pick nits here. I really do appreciate the contribution!

nolanlawson: Ah jeez yeah, I see that now. OK, second arg is not too bad then! And no prob with the nitpicks; maintenance costs matter. :)

nexdrew:
Me likey. 😺 Thanks again, @nolanlawson, this is great!

bcoe:
@nolanlawson thank you very much for digging into this.

lrlna:
woot!! Thanks so much @nolanlawson ✨ ✨

bcoe:
@nolanlawson @lrlna want to give this a spin?

`npm cache clear; yargs@6.0.0-alpha.1`?

These performance changes go out with the next release (would love the extra set of eyes).

lrlna:
@bcoe i am using that on a refactor of one of my projects right now; it's all good so far and I'll keep an eye :D

PR #611, opened 2016-08-29
Lazily bind functions in singletonify

nolanlawson:
Less of a dramatic speedup than #610, but seems to shave off about 2-3ms from the total runtime of `require('yargs')`.

Since the `inst` object contains a lot of keys (>=67 it seems) which are mostly functions, calling `Function.prototype.bind()` for each one at startup can be expensive. Instead, we can just create a getter for each function that uses a closure, i.e. avoiding `Function.prototype.bind()` entirely.

Using this mini-benchmark:

```bash
for ((i=0;i<500;i++)); do node -e 'var now = require("performance-now"); var s = now(); require("./index"); var e = now() - s; console.log(e)'; done | awk '{sum+=$1} END { print sum/NR }'
```

I get these results in node 6.5.0 on a 2013 Macbook Air:

* **Before:** 43.2434
* **After:** 39.7419

Profiling with `node-nightly --inspect`, this is what it looks like before:

screenshot 2016-08-29 13 59 07

And after:

screenshot 2016-08-29 14 00 02

(Note that these screenshots were just from a single run, so those ~6ms and ~3ms numbers will vary, which is why for the benchmark I used 500 iterations.)

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,8 +24,13 @@ function singletonify (inst) {
   Object.keys(inst).forEach(function (key) {
     if (key === 'argv') {
       Argv.__defineGetter__(key, inst.__lookupGetter__(key))
+    } else if (typeof inst[key] === 'function') {
+      // bind to inst, but use a closure to avoid Function.prototype.bind() cost
+      Argv.__defineGetter__(key, function () {
+        return inst[key]
+      })
     } else {
-      Argv[key] = typeof inst[key] === 'function' ? inst[key].bind(inst) : inst[key]
+      Argv[key] = inst[key]
     }
   })
 }
--- a/index.js
+++ b/index.js
@@ -24,12 +24,7 @@ function singletonify (inst) {
   Object.keys(inst).forEach(function (key) {
     if (key === 'argv') {
       Argv.__defineGetter__(key, inst.__lookupGetter__(key))
-    } else if (typeof inst[key] === 'function') {
-      // bind to inst, but use a closure to avoid Function.prototype.bind() cost
-      Argv.__defineGetter__(key, function () {
-        return inst[key]
-      })
-    } else {
+    } else { // no need to bind these functions; they don't ever use `this`
       Argv[key] = inst[key]
     }
   })


nolanlawson:
Actually just realized we can probably cut out defineGetter entirely. Current implementation is not actually binding; possible we need to add a test to catch that. Will iterate and get back to you.

bcoe:
@nolanlawson curious to see your findings with this one; I think there's a chance in this case that the `4ms` improvement might not be worth the clarity that we use with the usage of `bind`.

Either way, let's see if we can get a test-case around the regression you saw while experimenting? Definitely open to landing this if we find a significant performance improvement, and identical behavior.

nolanlawson:
OK so upon further investigation, it turns out that none of the bound functions actually use `this` (I checked with `/this/.test(func.toString())`), the tests all pass with or without binding, and the speedup does still appear on the order of 2-3ms, with the additional benefit of tighter code. I'll add a new commit to demonstrate.

nolanlawson:
So I guess the only question here is whether the speed boost is worth some potential risk in case a function is ever introduced that uses `this`, but I guess in that case the test should fail immediately, so hopefully that's not such a big risk.

nolanlawson:
Just re-ran the test with 1000 iterations on latest master (cbc36360b5abab42384f42c8ffab7083a28b1961) using:

```bash
for ((i=0;i<1000;i++)); do node -e 'var now = require("performance-now"); var s = now(); require("./index"); var e = now() - s; console.log(e)'; done | awk '{sum+=$1} END { print sum/NR }'
```

Got an average time in ms of **32.4986** vs **31.1284**, so yeah, not necessarily a huge speedup, but code is simpler as well now. Happy to concede to your judgment here; we may have reached the point of diminishing returns. :)

bcoe:
@nolanlawson maybe I'm being a bit too conservative; I don't honestly remember the original intentions for keeping functions bound to the parent context. However, for `1.5ms` performance gain, I think my preference would be to maintain the status quo -- this was still a worthwhile experiment however, if we'd seen a more significant performance boost, we'd have taken the chance.

nolanlawson:
That's fine! I agree 1.5ms is pretty small potatoes. Let's just close the PR and say it was an interesting experiment then. :)

PR #613, opened 2016-08-31
Also pass yargs to the fail handler

mkawalec:
This is needed to display a complete help message for a command that failed in a top-level `fail` handler

### DIFF
--- a/README.md
+++ b/README.md
@@ -1055,15 +1055,17 @@ yargs have been validated.
 
 Method to execute when a failure occurs, rather than printing the failure message.
 
-`fn` is called with the failure message that would have been printed and the
-`Error` instance originally thrown, if any.
+`fn` is called with the failure message that would have been printed, the
+`Error` instance originally thrown and yargs state when the failure
+occured. 
 
 ```js
 var argv = require('yargs')
-  .fail(function (msg, err) {
+  .fail(function (msg, err, yargs) {
     if (err) throw err // preserve stack
     console.error('You broke it!')
     console.error(msg)
+    console.error('You should be doing', yargs.help())
     process.exit(1)
   })
   .argv
--- a/test/usage.js
+++ b/test/usage.js
@@ -476,26 +476,23 @@ describe('usage tests', function () {
 
       it('gives the ability to log a per-command error message if failure occurs in a command', function () {
         var r = checkUsage(function () {
-          try {
-            return yargs
-              .command('foo', 'desc', {
-                bar: {
-                  describe: 'bar command'
-                }
-              }, function (argv) {
-                throw new Error('blah')
-              })
-              .fail(function (message, error, yargs) {
-                console.log(yargs.help())
-              })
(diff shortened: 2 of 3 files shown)


bcoe:
@mkawalec love it. 👍

mkawalec:
@bcoe: Should I change something, or can we merge this? :)

bcoe:
@mkawalec sorry for the slow turnaround! looks great, but could I perhaps get you to add a test? Partially because it would be neat to see how you plan on using this functionality 👍

mkawalec:
I'll add all of that early next week, thanks for the review.

bcoe:
@mkawalec would love to land this, any news on that test :)

mkawalec:
@bcoe: I've added the test, but it fails. The reason for that is that the exception thrown in command handler is not caught and so `fail` is never called. However, in a program I have that uses yargs the `fail` handler is called if an exception is thrown in the command handler. Is this a problem with the test harness or my mistake somewhere?

bcoe:
@mkawalec when I have some time this weekend I'll take a look at the harness 👍

bcoe:
@mkawalec _this weekend_ has quickly become _this following weekend_, for which I apologize; I haven't forgotten about this pull.

mkawalec:
No worries @bcoe, I know the pain well :)

README.md
```diff
@@ -1055,15 +1055,17 @@ yargs have been validated.
 
 Method to execute when a failure occurs, rather than printing the failure message.
 
-`fn` is called with the failure message that would have been printed and the
-`Error` instance originally thrown, if any.
+`fn` is called with the failure message that would have been printed, the
+`Error` instance originally thrown and yargs state when the failure
```
bcoe: slight editing nit, let's change this to:

`fn` is called with the failure message that would have been printed, the
`Error` instance originally thrown, and the yargs instance when the failure occurred.

test/usage.js
```diff
@@ -473,6 +473,31 @@ describe('usage tests', function () {
         r.logs.should.deep.equal(['foo'])
         r.should.have.property('exit').and.be.false
       })
+
+      it('gives the ability to log a per-command error message if failure occurs in a command', function () {
```
bcoe: I finally had the time to debug this test for you, here's the code I pulled together:

```js
      it('is invoked with yargs instance as third argument', function () {
          return yargs('foo')
            .command('foo', 'desc', {
              bar: {
                describe: 'bar command'
              }
            }, function (argv) {
              throw new Error('blah')
            })
            .fail(function (message, error, yargs) {
              yargs.showHelp()
            })
            .exitProcess(false)
            .wrap(null)
            .argv
        })

        r.errors[0].should.contain('bar command')
      })
```

bcoe: @mkawalec awesome 👍

mkawalec:
@bcoe please rereview :)

PR #615, opened 2016-08-31
Clean up indentation in options block

paradox41:
I am looking at adding something to yargs and in the process of getting familiar with the codebase, I came across [this block of code.](https://github.com/yargs/yargs/blob/master/yargs.js#L405-L441) 

At first glance, due to indentation, it looks like nested `if/else if` blocks. This can be confusing when reading the code. Only after further examination did I notice that these blocks actually aren't connected in the same way `if/else if` blocks are. 

I don't see this style being used elsewhere, it doesn't seem to be a violation of standard style, and I feel like this improves readability.

Obviously, this is a style choice, so if people disagree I am happy to close this PR.

### DIFF
--- a/yargs.js
+++ b/yargs.js
@@ -404,39 +404,69 @@ function Yargs (processArgs, cwd, parentRequire) {
 
       if (demand) {
         self.demand(key, demand)
-      } if ('config' in opt) {
+      }
+
+      if ('config' in opt) {
         self.config(key, opt.configParser)
-      } if ('default' in opt) {
+      }
+
+      if ('default' in opt) {
         self.default(key, opt.default)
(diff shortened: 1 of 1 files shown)


TomNeyland:
oh wow I didn't even realize those weren't `else if`s

:+1: for readability

bcoe:
thanks for the patch; I wonder if it would be worth starting a discussion on https://github.com/feross/standard, feels like this could be something that our linter catches.

paradox41:
That is a good idea actually. Thanks for the suggestion.

feross:
Yeah, this is a good find. Will try to make sure `standard` catches this.

PR #624, opened 2016-09-14
make opts object not required anymore for .option()

maxrimue:
Usually, yargs always requires options to be featuring an object for customisation like so:
```javascript
yargs()
  .option('x', {})
  .argv
```

This however seemed strange to me, as said customisation never is needed, you can also use options without any further customisation.

This Pull Request removes the check for whether an object is given to allow for declaration of options like this:
```javascript
yargs()
  .option('x')
  .argv
```

Why would we do that? Because it the object isn't needed anyway, requiring it makes little sense, and it might be confusing to people. Also, some people (like me) might prefer to write down all options in the yargs configuration to make clear which ones exist, also making yargs list them in the `argv` object even if not passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1082,8 +1082,8 @@ For example:
 
 ```js
 require('yargs')
-  .option('foobar', {})
-  .option('foobaz', {})
+  .option('foobar')
+  .option('foobaz')
   .completion()
   .getCompletion(['./test.js', '--foo'], function (completions) {
     console.log(completions)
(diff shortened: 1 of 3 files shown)


bcoe:
@maxrimue love it! sorry I've been slightly AWOL. `ETOOMUCHOSS`

bcoe:
@maxrimue want to give this a spin?

This feature goes out with the next release 👍

PR #633, opened 2016-09-26
chore: upgrade yargs-parser

bcoe:
BREAKING CHANGE: coerce is now applied as a final step after other parsing is complete

CC: @maxrimue


maxrimue:
@bcoe added the documentation, what do you think?

bcoe:
@maxrimue looks great, just made a few minor edits. thans! 👍

PR #634, opened 2016-09-26
chore: add greenkeeper ignore rule for string-width, which no longer Node 0.10

bcoe:
I'd like to continue supporting Node 0.10 in the medium term, since this is a widely used utility library.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,5 +70,10 @@
   "license": "MIT",
   "engine": {
     "node": ">=0.10"
+  },
+  "greenkeeper": {
+    "ignore": [
+      "string-width"
+    ]
   }
 }


maxrimue:
@bcoe LGTM :+1: 

When do you think should we stop supporting `0.10`? Perhaps the first of January next year would be ideal, that's the time when official support for `0.12` is dropped, and since there's always been more people using `0.10` than `0.12` (since `0.10` was current way longer) it seems reasonable to me to drop support for both there.

I understand that wide support is a great attitude for a utility library, but beyond official support, server operators really have to move on, for security reasons alone, and individuals should've switched by now, as it's usually way easier for them.

bcoe:
@maxrimue this is a great question, one we've been asking quite a bit at npm. @seldo recently crunched some numbers and I believe saw that 8.5% of npm users are still on Node 0.10; I think we should hold off on dropping `0.10` support until this # is a bit lower; maybe set a goal like `2%` or `3%`?

PR #635, opened 2016-09-26
Replace lodash.assign with object-assign

yandavid:


yandavid:
For context, this will allow consumers with node versions > 0.12 to use native `Object.assign` and remove the deprecated lodash.assign npm package warning.

bcoe:
@yandavid thanks for bringing this to our attention! this bit us in a few yargs dependencies too.

Since we only use the module in one place, I've opted to instead add the logic to yargs itself:

https://github.com/yargs/yargs/pull/641

This lets us drop a dependency; _we've been trying to strike a balance between yargs consisting of a small # of modules, and using modules where appropriate._

Hope this is just the first of many contributions from you 👍

bcoe:
@yandavid  mind giving this a spin? I'd like to roll this out soon to get rid of a deprecation notice:

npm cache clear; yargs@6.0.0-alpha.1

thanks for your help on this.

yandavid:
@bcoe Looks good to me - the deprecated notice is gone and thing still work. 👍

PR #640, opened 2016-09-27
Added getRawOptions() method to get back reusable options

moos:
This would be useful for composite/wrapper modules to easily integrate options of sub-components.

```
var yargs = require('yargs')
var sub = require('sub').cli

yargs()
  .option(sub.yargs.getRawOptions())
  .option(...)
```

### DIFF
--- a/yargs.js
+++ b/yargs.js
@@ -51,6 +51,11 @@ function Yargs (processArgs, cwd, parentRequire) {
     return context
   }
 
+  var rawOptions = {}
+  self.getRawOptions = function () {
+    return rawOptions
+  }
+
   // puts yargs back into an initial state. any keys
   // that have been set to "global" will not be reset
   // by this action.
@@ -397,6 +402,7 @@ function Yargs (processArgs, cwd, parentRequire) {
         opt = {}
       }
 
+      rawOptions[key] = opt
       options.key[key] = true // track manually set keys.
 
       if (opt.alias) self.alias(key, opt.alias)
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -1439,6 +1439,29 @@ describe('yargs dsl tests', function () {
       expect(msg).to.equal('ball')
       expect(err).to.exist
     })
+
+    it('should save & reuse raw options', function () {
+      var opts = {
+        looks: {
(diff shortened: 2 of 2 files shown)


bcoe:
@moos mind elaborating on your use-case a little bit; currently we do have commands, which you can use to compose a nested CI, and you can flag your options as global so that they're inherited by each command. I wonder if this would do the trick for you?

yargs.js
```diff
@@ -397,6 +402,7 @@ function Yargs (processArgs, cwd, parentRequire) {
         opt = {}
       }
 
+      rawOptions[key] = opt
```
bcoe: one concern I have with this, is that there are other ways that parameters can be set on options, .e.g., using the chaining API:

```js
yargs
  .describe('a', 'my a option')
  .default('a', 99)
```
I think you would do better to use the method `yargs.getOptions`; which already exists, and is used for similar purposes in the utility libraries. This returns the options object which is ultimately executed by yargs' parser.

Do you think this would solve your use-case for you; if so, would you perhaps be willing to contribute some documentation, I would greatly appreciate it.

moos:
Hmm... wasn't aware of global() -- not sure it would work though.  In my use case, `sub` is a completely independent component that has its own cli.  But that cli isn't executed directly from the parent.  The parent simply wants to expose the options available in `sub` in its own cli.

The concrete use case is sitespeedio/sitespeed.io#1225.  I have for now solved the issue by exposing the options in the `sub`  ([browsertime ](https://github.com/sitespeedio/browsertime)), but as a general use case, this may not be feasible.

Also, I believe the existing getOptions() isn't readily re-usable, as simply as passing the results back into yargs.

Hope that explains it better. Totally OK if you don't feel this is a common-enough use case.

bcoe:
@moos I feel this is edge-case enough that I'd like to hold off on adding to the main yargs API. Having said this, I think you could do something clever and extend yargs with this new functionality, rather than forking whole codebase  -- I'll see if I can pull together some sample code for you, and will update this issue with it.

moos:
.Fair enough -- look forward to seeing your extend 'hack'.

PR #641, opened 2016-09-30
fix: drop lodash.assign

bcoe:
lodash.assign is now deprecated; given how simple our use-case is, I opted to simply add the functionality to yargs itself.

### DIFF
--- a/lib/assign.js
+++ b/lib/assign.js
@@ -1,0 +1,15 @@
+// lazy Object.assign logic that only works for merging
+// two objects; eventually we should replace this with Object.assign.
+module.exports = function assign (defaults, configuration) {
+  var o = {}
+  configuration = configuration || {}
+
+  Object.keys(defaults).forEach(function (k) {
+    o[k] = defaults[k]
+  })
+  Object.keys(configuration).forEach(function (k) {
+    o[k] = configuration[k]
+  })
+
+  return o
+}
--- a/package.json
+++ b/package.json
@@ -15,7 +15,6 @@
     "cliui": "^3.2.0",
     "decamelize": "^1.1.1",
     "get-caller-file": "^1.0.1",
-    "lodash.assign": "^4.2.0",
     "os-locale": "^1.4.0",
     "read-pkg-up": "^1.0.1",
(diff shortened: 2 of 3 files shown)


jdalton:
There's also [object.assign](https://github.com/es-shims/object.assign) so you can shim to ease to use `Object.assign`. There's a variety of ways to load it with and without augmenting the built-in.

jdalton:
It's also worth noting that the `lodash.assign` deprecation is just a console log in the npm install. Nothing else is changed.

jdalton:
I pulled the deprecation method because gulp threated to pull out all of lodash due to the message. It's not worth that much drama tbh.

bcoe:
@jdalton I can understand your perspective; nightly canary builds for Chrome, and other browsers, have worked wonderfully, and in the past few years this approach to releasing software has moved forward browser technologies phenomenally.

At the same time, older versions of Node.js come installed on a lot of machines, e.g., `sudo yum install nodejs`, and a lot of new users might not be savvy enough to keep their engine up-to-date...

In both yargs and nyc, I'm trying to figure out a balance of supporting legacy versions of Node.js, and continuing to move forward with the ecosystem. There are interesting challenges, for instance npm itself will likely support Node 0.10 for quite some time and is a user of `nyc` (add to this, `nyc` is a power user of `yargs`).

I plan to chat with @othiym23, and the rest of the CLI team, to figure out an appropriate cadence for dropping Node 0.10 support.

I apologize if I contributed to drama in the twitersphere.

jdalton:
>  I can understand your perspective; nightly canary builds for Chrome, and other browsers, have worked wonderfully, and in the past few years this approach to releasing software has moved forward browser technologies phenomenally.

Nothing canary about it.  I put a log to encourage folks to start using syntax which is supported by the soon-to-be min bar of Node 4 and 6. That's kinda the point of [deprecation](https://en.wikipedia.org/wiki/Deprecation) vs. discontinuing. That said, that specific deprecation was a little premature ;P

> I apologize if I contributed to drama in the twitersphere.

Oh boy, more than twitersphere.

Issue #503, opened 2016-05-15
Unparse arguments


MeanwhileMedia:
This is more of a feature request (as I don't currently see this feature in docs). Might we have a way to convert parsed args back into string format?
One use case could be when a process uses yargs to parse it's args, then manipulates those args during it's lifespan, but later needs to instantiate a child process that receives the args.

bcoe:
@MeanwhileMedia you should be able to access `process.argv` directly, even after yargs has parsed the string; none of the values are changed in place.

Let me know if this does the trick for you.

MeanwhileMedia:
Thanks for the quick response @bcoe. 
My problem is that I call yargs...argv when the process first starts. The object that yargs returns (argsObj) is passed around and manipulated by that process. argsObj becomes a representation of the process's state. At this point, process.argv becomes "outdated".

Eventually, in my app, the process will need to pass it's state into a new child process.  This is why I want to convert argsObj back into a string that I can throw into a cli command.

I"ve build a simple solution to unparse, but I'm currently only traveling a single level deep (see below). It might be nice for yargs to support this in a more extensive manner.

```
      var unparsedString = ''
      _.each(argsObj, function(value, key){
        //We currently only support one layer deep of nested args (dot notation).
        if(_.isArray(value)){
          unparsedString = unparsedString+' --'+key+' '+value.join(' ')
        } else if(_.isObject(value)){
          _.each(value, function(subValue, subKey){
            unparsedString = unparsedString+' --'+key+'.'+subKey+'='+subValue
          })
        } else {
          unparsedString = unparsedString+' --'+key+' '+value
        }
      })
```

jdalton:
This might be helpful: [moxystudio/yargs-unparser](https://github.com/moxystudio/yargs-unparser).

bcoe:
@MeanwhileMedia as @jdalton mentions, we're unlikely to ever add this to yargs itself, but we now have the program `yargs-unparser` 😄

PR #645, opened 2016-10-01
Further fixes to skipValidation functionality

faazshift:
I previously made a change to fix the broken `skipValidation` functionality of `yargs` (see #619). However, as I noted [here](https://github.com/yargs/yargs/pull/619#issuecomment-250929482), the functionality still had issues. The only clean way to fix those issues and get this functionality working how it is supposed to work required more information to be provided from `yargs-parser` (see yargs/yargs-parser#60). I've created a PR on that repo that is depended upon by this PR. If those changes are approved, merged, and released, I can update the dependency version for this PR. Until then, this PR will be expected to fail CI tests.

### DIFF
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -1054,7 +1054,7 @@ describe('yargs dsl tests', function () {
 
     it('allows having an option that skips validation but not skipping validation if that option is not used', function () {
       var skippedValidation = true
-      yargs(['--no-skip'])
+      yargs(['--dont-skip'])
           .demand(5)
           .option('skip', {
             skipValidation: true
@@ -1064,6 +1064,23 @@ describe('yargs dsl tests', function () {
           })
           .argv
       expect(skippedValidation).to.equal(false)
+    })
+
+    it('allows skipping validation with non-boolean, aliased options', function () {
+      var skippedValidation = true
+      yargs(['--do-skip'])
+          .demand(5)
+          .option('d', {
+            alias: 'do-skip',
+            type: 'string',
+            default: 'yes',
+            skipValidation: true
+          })
+          .fail(function (msg) {
+            skippedValidation = false
+          })
+          .argv
+      expect(skippedValidation).to.equal(true)
     })
   })
 
--- a/yargs.js
+++ b/yargs.js
@@ -712,6 +712,7 @@ function Yargs (processArgs, cwd, parentRequire) {
     const parsed = Parser.detailed(args, options)
     const argv = parsed.argv
     var aliases = parsed.aliases
+    var processed = parsed.processedKeys
 
     argv.$0 = self.$0
     self.parsed = parsed
@@ -823,8 +824,8 @@ function Yargs (processArgs, cwd, parentRequire) {
 
     // Check if any of the options to skip validation were provided
     if (!skipValidation && options.skipValidation.length > 0) {
-      skipValidation = Object.keys(argv).some(function (key) {
-        return options.skipValidation.indexOf(key) >= 0 && argv[key] === true
+      skipValidation = processed.some(function (key) {
+        return options.skipValidation.indexOf(key) >= 0
       })
     }
 


bcoe:
@faazshift for how I would expect `--skip-validation` to work, changing the logic in yargs.js to:

```js
    // Check if any of the options to skip validation were provided
    if (!skipValidation && options.skipValidation.length > 0) {
      skipValidation = Object.keys(argv).some(function (key) {
        return options.skipValidation.indexOf(key) >= 0 && argv[key]
      })
    }
```

Does the trick, this works for the following scenarios:

`node test.js --skip-validation`: skips validation.
`node test.js --no-skip-validation`: does not skip validation.
`node test.js --skip-validation yes:` skips validation.

basically, I don't think we need to know which keys were set during the parsing process, and which were included in the original `process.argv` to handle this logic; unless I'm missing something.

Having said this! I like the idea of yargs storing more information about what arguments were set by defaults, config, etc., vs which arguments came in via command line -- before implementing this, however, I'd like to have a pressing need, so that it can shape the API. We've had this feature open for months which might require similar functionality:

https://github.com/yargs/yargs/issues/503

Please correct me if I've missed something; but for the time being I think we could get away with a simpler patch 👍

faazshift:
@bcoe When configuring options, `skipValidation` can optionally be used in order to skip the validation process if that option is passed on the command line. It should work regardless of whether the option is a boolean, string, etc. And it should work regardless of what the passed/default value of the option ends up being, as long as the option was passed on the command line. This is my understanding of the purpose `skipValidation` should be serving.

However, if an option is given a `default` value, it will always be present in the args returned by `yargs-parser`. Thus validation will always be skipped, even if the argument wasn't present on the command-line. Additionally, the current (and previous) logic is comparing against the passed/default value of the option, not just whether it was present on the command line. So if I have an option:

```JS
var args = yargs()
  .demand(5)
  .option('test', {
    type: 'boolean',
    skipValidation: true
  }).argv
```
This will be the behavior:
```BASH
 $ node test.js --test # Skips validation
 $ node test.js --no-test # Doesn't skip validation due to boolean-negation, not due to lack of arguments
 $ node test.js --blah=true # Skips validation
 $ node test.js --blah=false # Doesn't skip validation
```
To me that behavior seems unreliable and problematic at best. If a validation-skipping option is passed on the command line, regardless of its value, it should skip validation. However, without knowing exactly which arguments were _actually_ passed on the command line, it's impossible to account for all the edge cases. Hence why I proposed returning that information from yargs/yargs-parser. There's probably be a better way than what I did in my PR on yargs-parser, but to cleanly, reliably produce the expected functionality, I do think that information needs to be returned from yargs-parser somehow. Until then, this functionality will continue to be unreliable.

bcoe:
@faazshift I appreciate your point of view, but I think the argument is a bit of a slippery slope; we trust how yargs has opted to parse input for all other special option types, e.g., `help`, `version`, `completion`. I don't think we should treat `skipValidation` a special case.

If a user explicitly enters: `./app.js --no-skip-validation`, it's certainly a very weird thing for them to choose to type! but I think the correct thing to do is to not skip validation 😛 

I would like to hear the points of view of a few other contributors though: CC: @nexdrew, @addaleax, @maxrimue, @lrlna.

faazshift:
@bcoe The `help` and `version` options are handled a bit more manually by `yargs` where validation skipping is concerned. Even still, I personally think it makes more sense for the functionality to always happen based on the existence of an option on the command line, not the truthiness of the parsed argument (whether provided or defaulted).

Some better use-case examples:
```BASH
 $ node test.js --run-step=0 # Doesn't skip validation because it's not truthy
 $ node test.js --run-step=1 # Skips validation because it is truthy
 $ node test.js --service-enabled=false # Doesn't skip validation because it's not truthy
 $ node test.js --service-enabled=true # Skips validation because it is truthy
```
I personally think that with all of the examples above, if `skipValidation: true`, validation should be skipped. However, maybe I'm alone in this opinion, in which case I think the docs should more clearly explain the existing functionality. I, too, would like to see the opinions of others on this.

bcoe:
@faazshift no one else has chimed in yet; guess it's a busy week :)

I very much do agree that `skipValidation` could use better documentation. Personally, I continue to think it should work like any other command line argument, but I don't think we do a good job of describing the functionality.

I see it as working well for a CLI that has functionality like the following:

----
`git checkout master`

`WARN: you're not on a clean branch I'm not doing what you want`

`git --force checkout master`

__performs the action.__
---

The `force` option would have been specified in yargs with `.skipValidation('force')`. 

I can imagine a user that defaults `--force` to `true`, perhaps in a configuration file, because they usually find these types of messages annoying. 

I can also imagine this same user performing a risky operation, to be safe they might enter: `git dangerous-operation --force=false`.

Perhaps we can come to a compromise on this feature, where we start by hammering out better documentation?

faazshift:
As `skipValidation` is **not** a command line argument, but rather a behavior modulator for configured options, it still doesn't make sense to me to break separation of concerns in this way. If I configure an option, and specify `skipValidation: true`, I expect `yargs` validation to be skipped when that option is present on the command line, not if that argument's value happens to be truthy (which is completely unreliable and prone to problems). Other options passed when configuring an option in `yargs` modulate functionality in a consistent, reliable manner, so I don't see why `skipValidation` should be treated differently. Also, in the example given, with `--force=false`, shouldn't `yargs` only be concerned with parsing? If an option is configured with `skipValidation: true`, the presence of that option on the command line should disable `yargs` validation agnostic of the value (if any) provided on the command line. Handling of `--force=false` should not be the responsibility of `yargs`. It should be up to the developer of the specific app to handle that use-case.

Just my thinking on the matter. But perhaps I'm alone.

The current documentation does **not** accurately portray the current way the code works. Hence I assumed it to be a bug/oversight in the first place.

Anyhow, I would still love to hear the opinions of others.

nexdrew:
Personally, I'm not too keen on the current "skip validation" functionality in yargs, [as noted here](https://github.com/yargs/yargs/pull/471#issuecomment-214335199), but I guess the current state of the world is that one or more options, _when found_, is supposed to suddenly override all validation of other configured options

The argument here seems to be about the definition of _when found_.

On one hand, we could define it as _when the option is given on the command line, regardless of its value_.

On the other hand, we could define it as _when the option's value is truthy, regardless of where that value comes from_.

The precedent seems to be in favor of the latter, for the following reasons:

- When `.help()` is configured, yargs does not display help and exit when `--no-help` is given. (You can also do `.default('help', true)` which would have to be disabled with `--no-help`.)
- When `.version()` is configured, yargs does not display the version and exit when `--no-version` is given. (Ditto on `.default('version', true)`.)
- Ignoring the option's value seems to be contradictory to the change made in #619, which requires that a "skip validation" option must be a boolean and must have a value of `true` in order to take effect. This change was merged and [released with 6.0.0](https://github.com/yargs/yargs/blob/master/CHANGELOG.md#600-2016-09-30).

Obviously, _precedence_ doesn't equal _correctness_, but I don't think anyone has complained about this being incorrect until now. So if we want to change the interpretation of `.skipValidation()` options, I think we should also do so for `.help()` and `.version()`, which is not a huge stretch but is probably not a popular use-case we need to spend time on.

When it comes to different use-cases, some are obviously simpler than others, and I think we should leave `.skipValidation()` to only apply to the simplest use-case: basically that it implies/forces `type: boolean` and only skips validation when the option's value is `true` - and **this is something we should document**.

For any use-cases more advanced than that, I think there are more elegant/flexible ways to solve the problem. Offhand, authors could either define a custom `.check()` function or perhaps encapsulate the desired functionality in a command (where non-global options/validation does not apply). Otherwise they may need to rethink/refactor their CLI/logic.

I don't think the advanced examples above make much sense for `.skipValidation()`. Things like `--force` or `--run-step=0` should be handled in app logic and should probably _not_ affect option validation. The argument that "yargs should only do the parsing" kinda makes sense here but is a little off-base, because I think "parsing" here means "make sure the parsed options conform to my expecations/requirements", such that `--run-step=saturday` is probably invalid and should fail-fast.

Another thing I noticed while testing `.skipValidation()` is that it doesn't skip positional arg validation for commands. This is [yet another argument](https://github.com/yargs/yargs/pull/586#issue-170989449) for moving positional arg support down to yargs-parser and is something we should probably fix in one way or another.

All that being said, I do agree that exposing more meta information from yargs-parser could be useful, so I'd prefer to judge the merits of yargs/yargs-parser#60 (which I think is a good idea) separate from this PR (which is probably not what we want).

bcoe:
@faazshift @nexdrew as you suggest, let's move this discussion over to yargs-parser. I'd love to dig up a few issues in the backlog that would be served by us surfacing more information from the parser.

PR #646, opened 2016-10-02
feat: reworking yargs API to make it easier to run in headless environments, e.g., Slack

bcoe:
after discussion today, @nexdrew and I settled on what I think is a pretty clever extension to the `parse()` function, as a simple API extension that supports running yargs in headless mode (yargs with no terminal output).

----

A callback `fn` can optionally be provided as the second argument to `.parse()`.
If a callback is given, it will be invoked with three arguments:

1. `err`: populated if any validation errors raised while parsing.
2. `argv`: the parsed argv object.
3. `output`: any text that would have been output to the terminal, had a
  callback not been provided.

```js
// providing the `fn` argument to `parse()` runs yargs in headless mode, this
// makes it easy to use yargs in contexts other than the CLI, e.g., writing
// a chat-bot.
yargs()
  .command('lunch-train <restaurant> <time>', function () {}, function (argv) {
    api.scheduleLunch(argv.restaurant, moment(argv.time))
  })
  .parse(bot.userText, function (err, argv, output) {
    if (output) bot.respond(output)
  })
```

----

Long story short, `parse()` now accepts an optional `fn`, which if provided will:

1. prevent text from being output to the terminal.
2. prevent commands from being invoked if an error is thrown, or text has been output to the terminal.
3. get invoked with `err`, `argv`, `output` (allowing output to be routed somewhere elsewhere, .e.g., a chat-room).

reviewers: @nexdrew, @maxrimue, @lrlna

### DIFF
--- a/README.md
+++ b/README.md
@@ -1137,7 +1137,7 @@ Give some example invocations of your program. Inside `cmd`, the string
 present script similar to how `$0` works in bash or perl.
 Examples will be printed out as part of the help message.
 
-.exitProcess(enable)
+<a name="exitprocess"></a>.exitProcess(enable)
 ----------------------------------
 
 By default, yargs exits the process when the user passes a help flag, uses the
@@ -1516,6 +1516,8 @@ parser.parse(bot.userText, function (err, argv, output) {
   if (output) bot.respond(output)
 })
 ```
+
+***Note:*** Providing a callback to `parse()` disables the [`exitProcess` setting](#exitprocess) until after the callback is invoked.
 
 .pkgConf(key, [cwd])
 ------------
--- a/README.md
+++ b/README.md
@@ -1506,36 +1506,15 @@ If a callback is given, it will be invoked with three arguments:
 // providing the `fn` argument to `parse()` runs yargs in headless mode, this
 // makes it easy to use yargs in contexts other than the CLI, e.g., writing
 // a chat-bot.
-yargs()
+const parser = yargs
   .command('lunch-train <restaurant> <time>', function () {}, function (argv) {
     api.scheduleLunch(argv.restaurant, moment(argv.time))
   })
-  .parse(bot.userText, function (err, argv, output) {
-    if (output) bot.respond(output)
-  })
-```
+  .help()
 
-If you are running yargs in a long-lived process, e.g., parsing incoming messages
-to a web-server, it's important that you do not execute yargs as a singleton:
-
-
-_so, don't do this!_
-
-```js
-yargs
-  .option('a')
-  .parse('--a b')
-```
-
-_instead, instantiate an instance of yargs to handle each request:_
-
-```js
-var yargs = require('yargs/yargs')
-yargs()
-  .option('a')
-  .parse('--a b', function (err, argv, output) {
-  // handle the incoming chat message.
-  })
+parser.parse(bot.userText, function (err, argv, output) {
+  if (output) bot.respond(output)
+})
 ```
 
 .pkgConf(key, [cwd])
--- a/README.md
+++ b/README.md
@@ -1488,12 +1488,33 @@ Valid `opt` keys include:
     - `'number'`: synonymous for `number: true`, see [`number()`](#number)
     - `'string'`: synonymous for `string: true`, see [`string()`](#string)
 
-.parse(args)
+.parse(args, [fn])
 ------------
 
 Parse `args` instead of `process.argv`. Returns the `argv` object.
-
 `args` may either be a pre-processed argv array, or a raw argument string.
+
+A callback `fn` can optionally be provided as the second argument to `.parse()`.
+If a callback is given, it will be invoked with three arguments:
+
+1. `err`: populated if any validation errors raised while parsing.
+2. `argv`: the parsed argv object.
+3. `output`: any text that would have been output to the terminal, had a
+  callback not been provided.
+
+ > pro tip: providing the `fn` argument to `parse()` runs yargs in headless mode, this
+   makes it easy to use yargs in contexts other than the CLI, e.g., writing
+   a chat-bot.
+
+```js
+yargs()
+  .command('lunch-train <restaurant> <time>', function () {}, function (argv) {
+    api.scheduleLunch(argv.restaurant, moment(argv.time))
+  })
+  .parse(bot.userText, function (err, argv, output) {
+    if (output) bot.respond(output)
+  })
+```
 
 .pkgConf(key, [cwd])
 ------------
--- a/README.md
+++ b/README.md
@@ -1494,7 +1494,7 @@ Valid `opt` keys include:
 Parse `args` instead of `process.argv`. Returns the `argv` object.
 `args` may either be a pre-processed argv array, or a raw argument string.
 
-A `context` object can optionally as the second argument to `parse()`, providing a
+A `context` object can optionally be given as the second argument to `parse()`, providing a
 useful mechanism for passing state information to commands:
 
 ```js
--- a/README.md
+++ b/README.md
@@ -1502,11 +1502,10 @@ If a callback is given, it will be invoked with three arguments:
 3. `output`: any text that would have been output to the terminal, had a
   callback not been provided.
 
- > pro tip: providing the `fn` argument to `parse()` runs yargs in headless mode, this
-   makes it easy to use yargs in contexts other than the CLI, e.g., writing
-   a chat-bot.
-
 ```js
+// providing the `fn` argument to `parse()` runs yargs in headless mode, this
+// makes it easy to use yargs in contexts other than the CLI, e.g., writing
+// a chat-bot.
 yargs()
   .command('lunch-train <restaurant> <time>', function () {}, function (argv) {
     api.scheduleLunch(argv.restaurant, moment(argv.time))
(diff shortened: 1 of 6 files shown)


test/yargs.js
```diff
@@ -1494,4 +1494,206 @@ describe('yargs context', function () {
...
+        .argv
+
+      argv._.should.include('completion')
+    })
+
+    it('delegates to custom exit when making completion suggestions', function (done) {
+      var completions = ''
+      var argv = yargs(['--get-yargs-completions'])
+        .completion('completion', function (current, argv) {
+          return new Promise(function (resolve, reject) {
```
nexdrew: Perhaps we should add `require('es6-promise').polyfill()` to this file (`test/yargs.js`) as well.

lib/validation.js
```diff
@@ -187,18 +187,21 @@ module.exports = function (yargs, usage, y18n) {
...
       } catch (err) {
         usage.fail(err.message ? err.message : err, err)
+        continue
       }
-    })
+
+      if (!result) {
+        usage.fail(__('Argument check failed: %s', f.toString()))
+      } else if (typeof result === 'string' || result instanceof Error) {
+        usage.fail(result.toString())
```
nexdrew: If `result instanceof Error`, maybe we should pass it as the 2nd argument to `usage.fail()` here.

yargs.js
```diff
@@ -669,6 +670,42 @@ function Yargs (processArgs, cwd, parentRequire) {
...
+  var exitFn = process.exit
+  var exiting = false
+  self.exit = function (fn) {
+    // if a custom exit handler is provided we emit
+    // the event on the next tick of the event loop, giving
+    // the handler access to the parsed argv.
+    exitFn = function (code) {
+      if (exiting) return
+      exiting = true
+      process.nextTick(function () {
```
nexdrew: If validation fails, the following will execute (when they wouldn't before) because this is asynchronous:

1. a command handler
2. any top-level logic that runs after parsing

yargs.js
```diff
@@ -716,139 +753,147 @@ function Yargs (processArgs, cwd, parentRequire) {
...
+      if (e === ExitEarly) exitFn(ExitEarly.code)
+      else throw e
     }
 
-    setPlaceholderKeys(argv)
-
-    return argv
+    // if a custom exit handler has been provided,
+    // for API consistency always trigger it.
+    if (exitFn !== process.exit) exitFn(0)
```
nexdrew: This is called before validation of positional args for a command, so the current `exiting` check isn't quite sufficient (since, if positional arg validation fails, the custom exit handler would be given a code of 0).

nexdrew:
I think an API that's more "transactional", i.e. exchange one input for one output, would be more ideal than registering a custom logger and exit handler. My idea is that the input is a string or array representing a message or args to parse/execute, and the output is an object containing the following (somewhat similar to the way the `checkOutput()` test helper function works):

1. an overall success/fail indicator, like an integer exit code
2. an `Error`, if one was thrown (could possibly be used instead of an exit code)
3. a parsed `argv` object
4. buffered text (that would typically be output to the console/terminal) as a string or array of strings

@bcoe What do you think? How would you like to proceed?

lrlna:
@bcoe this is so ace 💯 think of all the things people can build!!

yargs.js
```diff
@@ -669,6 +692,42 @@ function Yargs (processArgs, cwd, parentRequire) {
...
+
+  // we use a custom logger that buffers output,
+  // so that we can print to non-CLIs, e.g., chat-bots.
+  var silent = false
+  var _logger = {
+    log: function () {
+      var args = Array.prototype.slice.call(arguments)
+      if (!silent) console.log.apply(console, args)
+      hasOutput = true
+      output += args.join(' ')
```
nexdrew: For consistency with `console.log` and `console.error`, I think we need to include line feeds too, something like:

```js
if (output.length) output += '\n'
output += args.join(' ')
```

Same for `_logger.error`. Otherwise the spacing in `output` is not _exactly_ what would have been output to the console. What do you think?

yargs.js
```diff
@@ -669,6 +692,42 @@ function Yargs (processArgs, cwd, parentRequire) {
...
+    }
+  }
+  self._getLoggerInstance = function () {
+    return _logger
+  }
+  // has yargs output an error our help
+  // message in the current execution context.
+  self._hasOutput = function () {
+    return !!hasOutput
```
nexdrew: Since this is a private variable that is always a boolean, we could get away with just `return hasOutput` here.

yargs.js
```diff
@@ -382,9 +385,29 @@ function Yargs (processArgs, cwd, parentRequire) {
...
 
+  var parseFn = null
   self.parse = function (args, shortCircuit) {
+    // by providing a function as a second argument to
+    // parse you can capture output that would otherwise
+    // default to printing to stdout/stderr.
+    if (typeof shortCircuit === 'function') {
+      parseFn = shortCircuit
+      shortCircuit = null
+      silent = true
```
nexdrew: We could possibly replace `silent` checks with `self._hasParseCallback()`.

README.md
```diff
@@ -1488,13 +1488,33 @@ Valid `opt` keys include:
...
+1. `err`: populated if any validation errors raised while parsing.
+2. `argv`: the parsed argv object.
+3. `output`: any text that would have been output to the terminal, had a
+  callback not been provided.
+
+```js
+// providing the `fn` argument to `parse()` runs yargs in headless mode, this
+// makes it easy to use yargs in contexts other than the CLI, e.g., writing
+// a chat-bot.
+yargs()
```
nexdrew: I think we need to really emphasize the use of `require('yargs/yargs')` and constructing/configuring a new `yargs()` instance for every call to `parse(msg, cb)`. It wasn't clear to me that this was necessary, but testing discovered that it was (due to the interesting way config is `reset()` for commands).

yargs.js
bcoe: Great point 👍 adding.

README.md
ceejbot: Interesting-- this will mean that bot authors will need to make a little factory to make a new yargs object & configure it every time they want to parse some input. This adds a little boilerplate but the docs are clear that you need it, so no problem.

nexdrew:
I think this looks great, nice work!

For future enhancements, I agree it would be better to refactor state management (of chained configuration and context) into its own submodule, to hopefully clean up all the individual pieces of state that are modified in different places right now.

I also think it would be nice to register a single `parse()` callback once, something like `yargs.onParse(fn)`, which would affect all subsequent calls to `yargs.parse(msg)`, since apps (like chatbots) will probably use the same callback logic for every message.

Otherwise, I think this PR certainly makes yargs viable for "single config and multiple parses" functionality desirable in non-CLI use-cases. (Many apologies for being so picky about this and for dragging this out for much longer than you wanted!) Great job!

README.md
bcoe: @scriptdaemon I love this idea, and think it would make things a lot more powerful ... I think I have an idea as to how to add this functionality without too major of an API change. What do you think of this:

```
yargs
  .command('foo')
  .parse('foo --bar', {user: 'bcoe}, function (err, argv, output) {
   
  })
```

where `{user: "bcoe"}` would be treated like passing in a configuration file.

bcoe:
@scriptdaemon I implemented your idea, allowing a context configuration object to be passed to `parse()`, let me know what you think. I'll be working on landing this shortly.

scriptdaemon:
@bcoe Looks great. Since it works basically the same as `.config()`, that would imply it should act as an object of defaults  as well if I am correct (i.e. I should be able to use the same name as an optional positional parameter, and if a user supplies a value for that parameter when issuing a command, overwrite the context value with the supplied value, yes?) I like that implementation, because that takes care of needing the context for both the builder and the handler for each command.

bcoe:
@scriptdaemon I hadn't thought of this side-effect, but yes that totally works:

```js
const yargs = require('./')
const parser = yargs
  .command('lunch-train [restaurant]', 'start lunch train', function () {}, function (argv) {
    console.log(argv.restaurant, argv.time)
    // api.scheduleLunch(argv.restaurant, moment(argv.time))
  })
  .help()

parser.parse('lunch-train', {restaurant: "Rudy's"}, function (err, argv, output) {
  if (output) bot.respond(output)
})```

File: yargs.js [hunk 9]
```diff
@@ -443,8 +438,8 @@ function Yargs (processArgs, cwd, parentRequire) {
     var parsed = parseArgs(args, shortCircuit)
     if (parseFn) {
       parseFn(exitError, parsed, output)
+      self.reset()
```
nexdrew: @bcoe This is a lot more (unnecessary) overhead than just calling `resetForNextParse()`, meaning `reset()` will be called N+1 times (where N is the number of commands) for `yargs.parse(msg, cb)`.

bcoe: see: https://github.com/yargs/yargs/pull/670

kevupton:
> ```js
> 
> ```

@bcoe 
When I am running console log inside my handler, it is being output to the console instead of being passed to the output variable, is this expected ?

bcoe:
@kevupton yes, we don't actually override `console` in yargs, we just skip the steps that output to the console within yargs itself.

Instead of outputting to `console.log` perhaps you can populate a variable with text in the commands?

PR #647, opened 2016-10-04
feat: initial support for command aliases

nexdrew:
Allows a command to have aliases if given as an array, e.g.

```js
.command(['start', 'run', 'go'], 'Start the app', builder, handler)
```

The first element in the array is treated as the canonical command, which supports positional args; the remaining elements are considered aliases. Aliases inherit their positional args from the canonical command and, thus, any positional args defined for an alias are ignored.

For example, all of the following are equivalent and only the `<name>` required positional arg will be parsed/validated:

```js
.command(['hi <name>', 'hello'], 'Say hi', builder, handler)
.command(['hi <name>', 'hello <name>'], 'Say hi', builder, handler)
.command(['hi <name>', 'hello <someone> [others..]'], 'Say hi', builder, handler)
```

For help text, command aliases are displayed in a new right-aligned column after the canonical command and description, e.g.

```console
$ node test.js help
Commands:
  start      Start the app                                    [aliases: go, run]
  hi <name>  Say hi                                             [aliases: hello]

Options:
```

TODO:

- [x] Tests
- [x] ~~Add `"aliases:"` to all locales~~ Will create a new issue for this
- [x] Docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -525,7 +525,11 @@ var argv = require('yargs')
 .command(module)
 ----------------
 
-Document the commands exposed by your application.
+Define the commands exposed by your application.
+
+`cmd` should be a string representing the command or an array of strings
+representing the command and its aliases. Read more about command aliases in the
+subsection below.
 
 Use `desc` to provide a description for each command your application accepts (the
 values stored in `argv._`).  Set `desc` to `false` to create a hidden command.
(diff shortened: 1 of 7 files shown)


bcoe:
@nexdrew for the locale files; feel free to open an issue and not let this block landing this feature.

nexdrew:
This should be ready for final review. Added the ability for command modules to export an `aliases` property so they don't _have_ to use an array for the `command` property, because why not.

@bcoe @maxrimue Let me know what you think.

maxrimue:
LGTM, nice work 👍 

We should probably open a new issue to call for translations of `aliases` once this PR is merged. Btw, the German translation would be **Aliase**, maybe you can implement this right into this PR @nexdrew?
I think we also still need to call for translations for `Did you mean?`, could do this in the same issue.

nexdrew:
Oh good point! Thanks @maxrimue, I'll do that.

PR #658, opened 2016-10-11
feat(command): add camelcase commands to argv

### LINKED ISSUE #653: camel-case commands
For options, e.g., `--foo-bar`, we populate the camel-case `fooBar` variable on the argv object. For commands we're not that clever:

`.command('foo <api-token>')` will result in only `api-token` being populated on the argv object; it would be nice if we were clever enough to populate both `api-token` and `apiToken`.

ssonal:
Populate argv object with camelcase variants of hyphenated arguments.

Fixes #653

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -1,5 +1,6 @@
 const path = require('path')
 const inspect = require('util').inspect
+const camelCase = require('camelcase')
 
 // handles parsing positional arguments,
 // and populating argv with said positional
@@ -185,6 +186,11 @@ module.exports = function (yargs, usage, validation) {
       if (demand.variadic) argv[demand.cmd] = argv._.splice(0)
       else argv[demand.cmd] = argv._.shift()
       postProcessPositional(yargs, argv, demand.cmd)
+      if (/-/.test(demand.cmd)) {
+        var cc = camelCase(demand.cmd)
+        if (typeof argv[demand.cmd] === 'object') argv[cc] = argv[demand.cmd].splice(0)
+        else argv[cc] = argv[demand.cmd]
+      }
     }
 
     while (optional.length) {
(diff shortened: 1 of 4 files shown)


lib/command.js
```diff
@@ -185,6 +186,11 @@ module.exports = function (yargs, usage, validation) {
       if (demand.variadic) argv[demand.cmd] = argv._.splice(0)
       else argv[demand.cmd] = argv._.shift()
       postProcessPositional(yargs, argv, demand.cmd)
+      if (/-/.test(demand.cmd)) {
+        var cc = camelCase(demand.cmd)
```
bcoe: I think it might be worthwhile pulling this into a helper function, e.g., `addCamelcaseExpansions ()`? since the logic is the same in optional and required arguments.

lib/command.js
```diff
@@ -194,6 +200,11 @@ module.exports = function (yargs, usage, validation) {
       if (maybe.variadic) argv[maybe.cmd] = argv._.splice(0)
       else argv[maybe.cmd] = argv._.shift()
       postProcessPositional(yargs, argv, maybe.cmd)
```
bcoe: we should also apply `postProcessPositional()` to both the normal form, and expanded form of the argument; perhaps we could make `postProcessPositional()` take an array? and `addCamelcaseExpansions ()` could return an array of either: the original argument, or the original argument plus the camel-case expansion?

We'd need to add one more test, verifying that `coerce` is applied to the camel-case version of the key (added along-side the existing coerce test).

nexdrew: I agree with @ssonal about the order of the logic; otherwise we'd be doing double duty calling coercion functions for both the normal and expanded form of the argument.

I also agree with @bcoe that we should include another test to make sure coercion is applied to both forms.

bcoe:
@ssonal great work: really clean tests, love the feature 👍 look forward to many more contributions from you.

PR #668, opened 2016-10-14
Suggestions to improve headless yargs

nexdrew:
This adds 3 things to the existing `headless-yargs` branch:

1. Changes `.parse(msg, cb)` so that the original `exitProcess` setting is reinstated after callback invocation
2. Resets the `parseFn` after each call such that you can mix and match `.parse(msg, cb)` calls (where `cb` is invoked) and `.parse(msg)` calls (where no callback is invoked)
3. Adds a note to `parse()` docs about implicitly (and temporarily) disabling the `exitProcess` setting when providing a callback

Includes tests to verify the functionality of 1 and 2.

reviewer: @bcoe

### DIFF
--- a/README.md
+++ b/README.md
@@ -1137,7 +1137,7 @@ Give some example invocations of your program. Inside `cmd`, the string
 present script similar to how `$0` works in bash or perl.
 Examples will be printed out as part of the help message.
 
-.exitProcess(enable)
+<a name="exitProcess"></a>.exitProcess(enable)
 ----------------------------------
 
 By default, yargs exits the process when the user passes a help flag, uses the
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1517,6 +1517,8 @@ parser.parse(bot.userText, function (err, argv, output) {
 })
 ```
 
+***Note:*** Providing a callback disables the [`exitProcess` setting](#exitProcess) until after the callback is invoked.
```
bcoe: Providing a callback to `parse()`, rather than providing a callback.

nexdrew: Good point, added in commit c424c3c

PR #669, opened 2016-10-14
fix: preserve top-level, reset-able state for execution of parse callback

nexdrew:
Otherwise, if you `parse(msg1, cb)` a message that executes a command and then `parse(msg2, cb)` a message that *doesn't* execute a command, the second invocation will have lost all top-level state like `.strict()` and `.demand(1)`.

Added test for this scenario.

I kinda consider this a hack until we can encapsulate all state into a new submodule, but it works. 🎉 

reviewer: @bcoe


bcoe:
@nexdrew looks much cleaner now that we're pulling logic into `freeze()`.

Issue #671, opened 2016-10-16
a context object with circular references will explode the stack on parse


ceejbot:
I'm attempting to use the new headless parsing feature to write a Slack bot, following the example in the [pirate joe repo](https://github.com/yargs/pirate-joe/blob/master/server.js). If I pass a context object to the `parse()` command that happens to include circular references (as many node http request clients or messages do), yargs will never succeed in parsing it. Eventually the process will exceed the maximum call stack size and crash.

In my particular case, I made an RTM client using the `@slack/client` module and hung it onto the context object. That should suffice to repro.

nexdrew:
Closing since this was fixed by #675. Thanks for reporting!

PR #675, opened 2016-10-16
fix: stop applying parser to context object

bcoe:
We were applying yargs' parser to the context object (finding aliases, etc.). Unfortunately, folks often want to pass in objects with cyclical relationships to the parser, e.g., `http.request` which makes the parser blow up quite spectacularly.

We now carve out an exception for the `parse()` context object, such that it doesn't run through yargs' parser.

@ceejbot @zkat


bcoe:
fixes #671

Issue #534, opened 2016-06-29
Request: opposite of "implies"


ljharb:
`implies('x', 'y')` means that if "x" is supplied, "y" _must_ be supplied.

I'd like a way to designate that if "x" is supplied, "y" must _not_ be supplied.

Perhaps it could be called ".conflicting(x, y)"

slx2005:
+1, this sounds like a good idea, a case would be if choices were set, but you have another arg which could be an override to these choices, 

```
args = require('yargs')
.option('f', {
    alias: 'fruit',
    describe: 'choose fruit.',
    choices: ['orange', 'apple', 'banana'],
})
.option('c', {
    alias: 'custom-fruit',
    describe: 'set a custom fruit.'
})
.conflicting('c', 'f');
```

bcoe:
this seems reasonable to me. we have enough slight breaking changes queued up, and API fixes, that I'd like to start marching towards a 5.x release of yargs.

Perhaps we can add this as one of the goals here: https://github.com/yargs/yargs/issues/519

hax:
I also suggest support `-c | -f` notation in command pasing:

```js
yargs.command('test [-c | -f] <args..>')
```

But the basic problem is command parsing only support postional arguments now, I wish it support options too. Should I open a separate issue?

ljharb:
@hax as a new syntactic shorthand for "conflicting"? Is that an argument notation that's used elsewhere? If nothing else uses that convention I don't think yargs should be adding it.

madcampos:
@ljharb answering about this convention:
>Arguments separated by the '|' ( <vertical-line>) bar notation are mutually-exclusive. Conforming applications shall not include the '|' symbol in data submitted to the utility. Alternatively, mutually-exclusive options and operands may be listed with multiple synopsis lines.

From: [POSIX docs](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html) (got it from [Wikipedia](https://en.wikipedia.org/wiki/Command-line_interface#cite_ref-6))

Also, if it's still open for taking I want to help.

ljharb:
@madcampos thanks, so it seems to be a notational convention, not something typed by the user. that makes sense.

nexdrew:
TIL the `.implies()` method interprets any keys prefixed with `--no-` as "when this key is _not_ given", such that you can make two arguments mutually exclusive.

Here's an example:

```js
const argv = require('yargs')
[...]

PR #689, opened 2016-10-24
Implements mutually exclusive args (conflicts)

madcampos:
Addresses #534.

Basic implementation of mutually exclusive arguments option, so if the user supplies both arguments the command fails.
The new option is implemented with the signature: `.conflicts(key1, key2)`

This adds a new message `Arguments %s and %s are mutually exclusive` and it's pt and pt-br translations.

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -277,6 +277,28 @@ module.exports = function (yargs, usage, y18n) {
     }
   }
 
+  var conflicting = {}
+  self.conflicts = function (key1, key2) {
+    if (typeof key1 === 'string' && typeof key2 === 'string' && key1 !== '_' && key2 !== '_') {
+      conflicting[key1] = key2
+      conflicting[key2] = key1
+    }
+  }
+
+  self.getConflicting = function () {
+    return conflicting
+  }
+
+  self.conflicting = function (argv) {
+    var args = Object.getOwnPropertyNames(argv)
+
+    args.forEach(function (arg) {
+      if (conflicting[arg] && args.indexOf(conflicting[arg]) !== -1) {
+        usage.fail(__('Arguments %s and %s are mutually exclusive', arg, conflicting[arg]))
+      }
+    })
+  }
+
   self.recommendCommands = function (cmd, potentialCommands) {
     const distance = require('./levenshtein')
     const threshold = 3 // if it takes more than three edits, let's move on.
@@ -299,6 +321,7 @@ module.exports = function (yargs, usage, y18n) {
       return globalLookup[k]
     })
     checks = []
+    conflicting = {}
     return self
   }
 
@@ -307,10 +330,12 @@ module.exports = function (yargs, usage, y18n) {
     frozen = {}
     frozen.implied = implied
     frozen.checks = checks
+    frozen.conflicting = conflicting
   }
   self.unfreeze = function () {
     implied = frozen.implied
     checks = frozen.checks
+    conflicting = frozen.conflicting
     frozen = undefined
   }
 
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -277,11 +277,28 @@ module.exports = function (yargs, usage, y18n) {
     }
   }
 
+// checks conflicts if foo => !bar
   var conflicting = {}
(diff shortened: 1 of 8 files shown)


bcoe:
@madcampos lovely; might take until this weekend to land, but at first glance this looks great.

bcoe:
@madcampos I apologize for the slow turnaround on this, was busily working on a release for `nyc`; will switch gears and do some yargs work this week.

madcampos:
@bcoe no problem ;)
Hope everything is alright in the pr, i think the only conflit is with the translation files (as i was too lazy/forgetfull and didin't create a specific branch for each issue).

bcoe:
@madcampos mind rebasing, I'm gradually working through an OSS backlog.

madcampos:
Rebased and got no conflicts! 🦄🌈

lib/validation.js
```diff
@@ -277,6 +277,28 @@ module.exports = function (yargs, usage, y18n) {
     }
   }
 
+  var conflicting = {}
+  self.conflicts = function (key1, key2) {
+    if (typeof key1 === 'string' && typeof key2 === 'string' && key1 !== '_' && key2 !== '_') {
```
bcoe: I recommend that we don't add an special handling of `_` for the time being -- if someone complains that they tried to set `_` as a conflicting arg and it didn't work we can:

1. figure out what the heck they expected it to do, and potentially add support.
2. or, instead add a warning about this.

test/validation.js
```diff
@@ -71,6 +71,34 @@ describe('validation tests', function () {
     })
   })
 
+  describe('conflicts', function () {
+    it('fails if both arguments are supplied', function (done) {
```
bcoe: tests! thank you.

lib/validation.js
```diff
@@ -277,6 +277,28 @@ module.exports = function (yargs, usage, y18n) {
...
+      conflicting[key1] = key2
```
bcoe: `implies` can accept an object rather than two keys, and will iterate over the object setting multiple implications at the same time; I think `conflicts` should have the same API 👍

lib/validation.js
madcampos: That looks weird... i don't know why I did this last part of the test... to me it looks like some leftover (I tried some approaches and ended up with the simplest one)

test/validation.js
```diff
@@ -71,6 +71,34 @@ describe('validation tests', function () {
...
+    it('should not fail if only one argument is supplied', function () {
+      yargs(['-b'])
+          .conflicts('f', 'b')
+          .fail(function (msg) {
+            expect.fail()
+          })
+          .argv
+
+      yargs(['-f'])
+      .conflicts('f', 'b')
```
maxrimue: Can we get some indentation here? 😄

bcoe:
@madcampos I'm going to try to get a release of yargs out tonight; no pressure, but if you get these fixes made made be able to sneak this in.

test/validation.js
madcampos: 😂 ok, noted!

madcampos:
@bcoe @maxrimue updated to accept an array as argument. But it's giving an parsing error on ruining the tests for some reason I couldn't figure out. 😢😞

I was thinking about the object as argument use case, how should this be dealt with?
For arrays it simply maps all the args recursively, but what should the value of an object's key represent?

I thought of using the object representation as the argument and it's value, for example: `.conflitcs('f', {'bar': 'baz'})` would only fail if `test -f --bar=baz`, and `text -f --bar=qux` would be ok.

bcoe:
@madcampos I've been on a bit of an open source hiatus lately, just starting to get back in the swing of things; I'll work on getting some tests around this.

bcoe:
Closing in favor of https://github.com/yargs/yargs/pull/741, which:

* makes a couple tiny tweaks to your work, to make `conflicts` and `implies` pretty much identical APIs.
* gets tests passing.

@madcampos thanks a ton for getting this 99% of the way there, this is a feature request that's been kicking around forever.

madcampos:
You welcome.

Issue #618, opened 2016-09-02
Cannot create an error to say "too many arguments"


kumar303:
I want a command with extraneous arguments to raise an error. For example:

    node script.js args-not-supported

This should produce some kind of error, I guess like:

````
Options:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]

Too many non-option arguments
````

Is this possible? I thought the following code would do it but it does not produce an error at all.

````javascript
var yargs = require('yargs');
yargs
  .help('help')
  .demand(0)
  .strict()
  .argv;
````

nexdrew:
@kumar303 Try `.demand(0, 0)` instead.

nexdrew:
Also, you can customize the error message with a 3rd argument, like `.demand(0, 0, 'Extraneous arguments are not allowed.')`

kumar303:
d'oh! Thanks, I guess I didn't read the docs well enough. It seems weird that `demand(0)` wouldn't do anything at all though since it's non-sensical without a max also of 0. Anyway, now that I know it's not a bug, feel free to close the issue.

bcoe:
@kumar303 this is definitely a part of the API that is confusing, we have this open issue:

https://github.com/yargs/yargs/issues/504

Would love your feedback and thoughts on these potential API improvements.

Issue #566, opened 2016-07-21
demand does not play well with command


zyy7259:
I typed the code from section [http://yargs.js.org/docs/#examples-yargs-is-here-to-help-you](http://yargs.js.org/docs/#examples-yargs-is-here-to-help-you)

```js
#!/usr/bin/env node
var argv = require('yargs')
    .usage('Usage: $0 <command> [options]')
    .command('count', 'Count the lines in a file')
    .demand(1)
    .example('$0 count -f foo.js', 'count the lines in the given file')
    .demand('f')
    .alias('f', 'file')
    .nargs('f', 1)
    .describe('f', 'Load a file')
    .help('h')
    .alias('h', 'help')
    .epilog('copyright 2015')
    .argv;

var fs = require('fs');
var s = fs.createReadStream(argv.file);

var lines = 0;
s.on('data', function (buf) {
    lines += buf.toString().match(/\n/g).length;
});

s.on('end', function () {
    console.log(lines);
});
```

And execute `node line_count.js count`, but `demand('f')` does not give the hint `Missing required arguments: f`

When I comment out the `.command` line, the `demand('f')` works.

`yargs` version: `4.8.1`

lrlna:
hey @zyy7259 ! I apologize for the faulty example here. When demanding both a command and an argument, you would be [using the syntax](http://yargs.js.org/docs/#methods-demandcount-max-msg) of `demand(1, ['f'])`. 

So the example above would look like:

```
#!/usr/bin/env node
var argv = require('./')
  .usage('Usage: $0 <command> [options]')
  .command('count', 'Count the lines in a file')
  .example('$0 count -f foo.js', 'count the lines in the given file')
  .demand(1, ['f'])
  .alias('f', 'file')
  .nargs('f', 1)
  .describe('f', 'Load a file')
  .help('h')
  .alias('h', 'help')
  .epilog('copyright 2015')
  .strict()
  .argv;

var fs = require('fs');
var s = fs.createReadStream(argv.file);

var lines = 0;
s.on('data', function (buf) {
    lines += buf.toString().match(/\n/g).length;

});

s.on('end', function () {
    console.log(lines);

});
```

[...]

Issue #504, opened 2016-05-15
separate demand into demandOption and demandCommand


lrlna:
As @ljharb was fixing #431 with #438 , we've discovered a bit of a black hole that is `demand`. 

The suggested solution at this point is to work off `demand` being separated in to two different functions: `demandOption` and `demandCommand`, clear detailed in [this comment](https://github.com/yargs/yargs/pull/438#issuecomment-200545670). 

I am currently working on separating the two. To avoid breaking too many things, I am hoping to keep both `require` and `demand` as a alias for `demandOption` until we decide to completely get rid of `demand`. 

Let's use this thread to track any discussion we might have about `demand` and the proposed separation.

kumar303:
This sounds good to me. I was pointed this way because I was trying to do `demand(0, 0)` to throw an error when extraneous arguments are provided. I agree that `demand()` could do with a split.

It's sort of a nit-pick but I would call `demandCommand()` `demandArg()` instead because not all arguments are commands (they could be file paths, etc).

lrlna:
@bcoe @nexdrew @maxrimue you *won't* believe this, buttttttt I actually had time to work this separation out this weekend. lmk if y'all think this might still be useful, and I'll PR a WIP branch before I do anything else \o/

maxrimue:
@lrlna That's awesome 🎉! I do think this still would be useful, so go ahead!

PR #699, opened 2016-11-06
separate demand API into demandOption and demandCommand

lrlna:
We've spoken quite a bit about having a separate demand API for both options and commands, specifically in #504, #438, #566, #618. 

This PR is creating this separation with: 

`demandCommand`,
```
demandCommand(min, max, minMsg, maxMsg)
```
or alternatively just the `min`,
```
demandCommand(min, minMsg)
```
and `demandOption`,
```
demandOption(['opt', 'anotherOpt'], msg)
```
or alternatively just,
```
demandOption('opt', msg')
```

I would love another pair of eyes on this, since I might have missed some aspects or use cases of this API, and I want to make sure I don't forget/break stuff \o/

cc @maxrimue @nexdrew @bcoe

### DIFF
--- a/README.md
+++ b/README.md
@@ -1028,13 +1028,24 @@ require('yargs')
     alias: 'p', 
     describe: 'provide a path to file'
   })
-  .option('spec, {
+  .option('spec', {
     alias: 's', 
     describe: 'program specifications'
   })
   .demand(['run', 'path'], 'Please provide both run and path arguments to work with this tool')
   .help()
   .argv
+```
+which will provide the following output:
+```bash
+Options:
+  --run, -r   run your program                [required]
+  --path, -p  provide a path to file          [required]
+  --spec, -s  program specifications
+  --help      Show help                        [boolean]
+        
+  Missing required arguments: run, path
+  Please provide both run and path arguments to work with this tool
 ```
 
 If a `boolean` value is given, it controls whether the option is demanded;
@@ -1043,22 +1054,34 @@ this is useful when using `.options()` to specify command line parameters.
 ```javascript
 // demand individual options within the option constructor
 require('yargs')
-  .option('run', {
-    alias: 'r', 
-    describe: 'run your program',
-    demand: true
-  })
-  .option('path', {
-    alias: 'p', 
-    describe: 'provide a path to file', 
-    demand: true
-  })
-  .option('spec, {
-    alias: 's', 
-    describe: 'program specifications'
+  .options({
+    'run': {
+      alias: 'r', 
+      describe: 'run your program',
+      demand: true
+    },
+    path': {
+      alias: 'p', 
+      describe: 'provide a path to file', 
+      demand: true
+    },
+    'spec': {
+      alias: 's', 
+      describe: 'program specifications'
+    }
   })
   .help()
   .argv
+```
+which will provide the following output:
+```bash
+Options:
+  --run, -r   run your program                                       [required]
+  --path, -p  provide a path to file                                 [required]
+  --spec, -s  program specifications
+  --help      Show help                                               [boolean]
+        
+Missing required arguments: run, path
 ```
 
 <a name="demandCommand"></a>.demandCommand(min, [minMsg])
@@ -1078,9 +1101,20 @@ require('yargs')
       console.log(`setting ${argv.key} to ${argv.value}`)
     }
   })
-  .demandCommand(1)
+  // provide a minimum demand and a minimum demand message
+  .demandCommand(1, 'You need at least one command before moving on')
   .help()
   .argv
+```
+which will provide the following output:
+```bash
+Commands:
+  configure <key> [value]  Set a config variable         [aliases: config, cfg]
+  
+Options:
+  --help  Show help                                                   [boolean]
+    
+You need at least one command before moving on
 ```
 
 <a name="describe"></a>.describe(key, desc)
--- a/README.md
+++ b/README.md
@@ -160,7 +160,7 @@ area.js:
 #!/usr/bin/env node
 var argv = require('yargs')
     .usage('Usage: $0 -w [num] -h [num]')
-    .demand(['w','h'])
+    .demandOption(['w','h'])
     .argv;
 
 console.log("The area is:", argv.w * argv.h);
@@ -188,7 +188,7 @@ demand_count.js:
 ````javascript
 #!/usr/bin/env node
 var argv = require('yargs')
-    .demand(2)
+    .demandCommand(2)
     .argv;
 console.dir(argv);
 ````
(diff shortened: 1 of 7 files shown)


bcoe:
@lrlna at first glance this looks great, I think it really simplifies the demand API to not try to overload things so much!

I believe this will call for a `7.x` release?

lrlna:
@bcoe yea, this is definitely a 7.x release. We should wait until we have more things for it though, eh?

README.md
```diff
@@ -1005,45 +1005,117 @@ displaying the value in the usage instructions:
...
+```javascript
+// demand individual options within the option constructor
+require('yargs')
+  .options({
+    'run': {
+      alias: 'r', 
+      describe: 'run your program',
+      demand: true
+    },
+    path': {
```
maxrimue: Little `'` missing here 😱

lib/usage.js
```diff
@@ -128,12 +128,14 @@ module.exports = function (yargs, y18n) {
   self.help = function () {
     normalizeAliases()
 
-    var demanded = yargs.getDemanded()
+    var demandedOption = yargs.getDemandedOption()
+    var demandedCommand = yargs.getDemandedCommand()
```
maxrimue: Maybe we should rename those to make them plurals? Like `demandedOptions`, `getDemandedOptions` etc.?

lib/usage.js
```diff
@@ -303,15 +305,15 @@ module.exports = function (yargs, y18n) {
   // make sure any options set for aliases,
   // are copied to the keys being aliased.
   function normalizeAliases () {
-    var demanded = yargs.getDemanded()
+    var demandedOption = yargs.getDemandedOption()
```
maxrimue: Same plural thing ^

lib/validation.js
```diff
@@ -10,21 +10,27 @@ module.exports = function (yargs, usage, y18n) {
   // validate appropriate # of non-option
   // arguments were provided, i.e., '_'.
   self.nonOptionCount = function (argv) {
-    const demanded = yargs.getDemanded()
+    const demandedCommand = yargs.getDemandedCommand()
```
maxrimue: Same plural thing ^

lib/validation.js
```diff
@@ -73,13 +79,13 @@ module.exports = function (yargs, usage, y18n) {
 
   // make sure all the required arguments are present.
   self.requiredArguments = function (argv) {
-    const demanded = yargs.getDemanded()
+    const demandedOption = yargs.getDemandedOption()
```
maxrimue: Same plural thing ^

lib/validation.js
```diff
@@ -107,7 +113,7 @@ module.exports = function (yargs, usage, y18n) {
   self.unknownArguments = function (argv, aliases) {
     const aliasLookup = {}
     const descriptions = usage.getDescriptions()
-    const demanded = yargs.getDemanded()
+    const demandedOption = yargs.getDemandedOption()
```
maxrimue: Same plural thing ^

lib/usage.js
lrlna: Yea, so, I kept typing the plural version all the time. I feel like you're right, and we should probably keep this at plural; seems to make more lexical sense.

README.md
lrlna: oh ofc, it's in the readme :/

lrlna:
@maxrimue and done 🎉

yargs.js
```diff
@@ -301,42 +301,58 @@ function Yargs (processArgs, cwd, parentRequire) {
     return self
   }
 
-  self.demand = self.required = self.require = function (keys, max, msg) {
+  self.demandOption = self.demand = self.required = self.require = function (keys, max, msg) {
```
lrlna: I am wondering if we should still keep `self.demand` as a fallback (?) for `demandOption`, or just completely rename it to `demandOption`

maxrimue: If I get this right, `demandOption` only accepts specific option names instead of just an integer for a number of minimum options, like `demandCommand`, right? (Even though the title of `demandOption` still says `count` as a parameter)

In that case, `demandOption` _does not replace `demand`_, which is why I think `demand` should be removed even as an alias, because that could cause confusion.

lrlna: @maxrimue that is an interesting point. I feel like `demand` traditionally accepted `keys` as `number` only if it was a command, otherwise we would set it up as `demand: true` in option config. If it was a `string` we would dealing with an option. Please correct if I misunderstood that somehow. 

Which means `demand` is effectively the same as `demandOption`, because all I did is take out the `keys` as `number` logic and put it into `demandCommand`. Does that make sense?

maxrimue: @lrlna Yeah that sounds right to me. I'm just meaning, if you took out logic of the `demandOption` compared to the old `demand`, `demandOption` doesn't have all the features the old `demand` did.

If folks now (or rather then) download 7.x or something and get no error because `.demand()` still exists, but something like `.demand(1)` doesn't work anymore, that might cause confusion.

Maybe I'm just not getting something, but apart from that, it might also be bad to keep it because of the naming stuff. Like, `demandOption` and `demandCommand` are very expressive names, but I would find it hard to remember which `demand` was an alias of. What do you think?

lrlna: > `demandOption` doesn't have all the features the old `demand` did.
@maxrimue yea, you're right; people will be expecting the same behaviour that we had before. let me change this

bcoe:
@lrlna @maxrimue looks like this is coming together really well 👍 haven't had time to dig into things too thoroughly yet -- first things is first I'm going to try to patch some of the little bug issues that have floated in over the past couple weeks.

I can understand the argument for keeping the method signature of `demand()` the same, as long as `demandCommand` and `demandOption` have a simpler signature -- _but_, I think we should perhaps trim down the documentation for `demand` instead pointing people towards `demandOption` and `demandCommand`.

https://github.com/yargs/yargs/blob/master/yargs.js#L309

What would we need to do to land this, looks like we might be able to re-add some tests if we're going to keep the method signature close to the same? and add a couple more tests?

lrlna:
@bcoe I am thinking we should just get rid of `demand` entirely. @maxrimue is right, these two methods are a lot more descriptive and `demand` no longer has the same features which might definitely confuse folks. 

I *do* need to add more tests, but I need a few more days 'cause I am prepping a demo for node interactive and it's a bit of a mayhem atm \o/

I also don't mind waiting until we have a couple more breaking changes to start a `7.x` release. But let me know what y'all think too.

bcoe:
@iarna @maxrimue I have a few OSS cycles (or at least I'm making a few OSS cycles) so I'm taking a stab at addressing the last remaining items on this ticket here: https://github.com/yargs/yargs/pull/740

PR #704, opened 2016-11-13
chore: add camelcase to greenkeeper ignore, until we drop 0.10 support

bcoe:


bcoe:
@nexdrew agreed, soon!

PR #718, opened 2016-11-27
Use https for image URL at README.md

simison:
I just noticed this was breaking https at https://www.npmjs.com/package/yargs

https://i.imgur.com/4WFGVJ9.png


bcoe:
@simison thanks for the contribution; the tests are breaking due to new standard rules, addressed here:

https://github.com/yargs/yargs/pull/717

Issue #720, opened 2016-11-28
Accept yargs-parser configuration directly instead of just via package.json


laggingreflex:
The `configuration` object that yargs passes to yargs-parser can only be configured via `package.json`. 

https://github.com/yargs/yargs/blob/3fe9b8fdbd53da8841cc001cffe7f7794100031a/yargs.js#L824-L825
```
options.configuration = pkgUp(cwd)['yargs'] || {}
```

User should be able to pass this configuration directly through yargs, such as:
```
require(yargs)(...).configuration({...})
```

bcoe:
@laggingreflex I think it would confuse the API to have both a `.config()` for configuring your own libraries `configuration()` for configuring yargs' settings.

Out of curiosity, what's your specific use-case, such that you don't want to configure in the `package.json`.

laggingreflex:
https://github.com/laggingreflex/yargs-test-720

nexdrew:
This appears to be a bug [introduced with 556 here](https://github.com/yargs/yargs/pull/556/commits/fb5c4896d916ba6a9c326842200ce646012e5287#diff-6dd17733300c85ca1a41786abcb4a1c9R639). In order to find the package.json file closest to `require.main.filename`, the line in `parseArgs` that calls `pkgUp(cwd)` should be calling `pkgUp()` instead (i.e. should not use `cwd`), which we do in `guessVersion()` [here](https://github.com/yargs/yargs/blob/51af0a640ba3df0f86b9ed1dd01ade018af6b279/yargs.js#L634).

I can confirm the problem and the fix, but applying the fix gives me these test results:
<img width="561" alt="screen shot 2016-12-05 at 10 41 11 am" src="https://cloud.githubusercontent.com/assets/1929625/20891119/70be8422-bad7-11e6-8865-f42d5e2abd56.png">

Thus, I need to look into the test that's supposed to be covering that line in lib/validation.js.

bcoe:
@laggingreflex a fix for this is going out in the next release of yargs; it's a pretty big one and I would love the extra help testing. Here's how you can get a sneak peek:

`npm cache clear; npm i yargs@next`

Let me know if this does the trick for you.

bcoe:
@laggingreflex I confirmed that the latest yargs works with the example `yargs-test-720`; thanks for writing this! it made testing quite easy:

```bash
Benjamins-MBP-2:~ benjamincoe$ yargs-test-720 -x 3 -x 4 -x 99
99
```

[...]

PR #726, opened 2016-12-05
fix: do not use cwd when resolving package.json for yargs parsing config

nexdrew:
Fixes a bug introduced with #556, demonstrated by https://github.com/laggingreflex/yargs-test-720.

Also, while investigating changes from 556, I found that validation logic for the `.implies()` method interprets a prefix of `--no-` on a key as a tiny DSL for "when this key is _not_ given". From this, we can actually ignore the `boolean-negation: false` setting for the `.implies()` method because `boolean-negation` affects _parsing_ but not the _DSL of implications_. To demonstrate this, I added some tests (with comments) showing how the DSL is interpreted.

The following use-cases are supported for `.implies()`:

1. `bar: 'foo'` = use of `--bar` requires that `--foo` be given
2. `bar: '--no-foo'` = use of `--bar` requires that `--foo` *cannot* be given
3. `bar: 'noFoo'` = use of `--bar` requires that `--noFoo` (or `--no-foo` with `"boolean-negation": false`) be given
4. `bar: '--no-noFoo'` = use of `--bar` requires that `--noFoo` (or `--no-foo` with `"boolean-negation": false`) *cannot* be given

In all use-cases, the `boolean-negation` config setting _only changes how `--no-foo` (or `--no-ANYTHING`) is parsed in `argv`_, it does not change how the implications are interpreted.

Thus, I also reverted 556 in its entirety.

Note that this fixes the bug found by #720, but it technically doesn't close 720 because I have not attempted to change the API to allow parsing config to be defined in any other way besides a `"yargs"` stanza in package.json.

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -223,12 +223,6 @@ module.exports = function (yargs, usage, y18n) {
     const implyFail = []
 
     Object.keys(implied).forEach(function (key) {
-      var booleanNegation
-      if (yargs.getOptions().configuration['boolean-negation'] === false) {
-        booleanNegation = false
-      } else {
-        booleanNegation = true
-      }
       var num
       const origKey = key
       var value = implied[key]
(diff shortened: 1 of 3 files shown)


bcoe:
@nexdrew I tested with `yargs-test-720` installed with `-g` and from the folder. worth noting that it would have always worked when run form the containing directory, which is how we missed this.

bcoe:
@nexdrew this is going out in the next release of yargs, would love your help testing:

PR #727, opened 2016-12-06
feat: support for positional argument aliases

bcoe:
When migrating one of npm, Inc's internal chat-bots to yargs, I ran into some help strings that were documented like so:

`delete <user|org>    delete a user or organization`

Given that it's not a huge amount of work, I thought it would be worth adding this syntax to yargs' command API, in the form of positional argument aliases.

CC: @ceejbot, @nexdrew

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -108,12 +108,12 @@ module.exports = function (yargs, usage, validation) {
       if (/\.+[\]>]/.test(cmd) && i === splitCommand.length - 1) variadic = true
       if (/^\[/.test(cmd)) {
         parsedCommand.optional.push({
-          cmd: cmd.replace(bregex, ''),
+          cmd: cmd.replace(bregex, '').split('|'),
           variadic: variadic
         })
       } else {
         parsedCommand.demanded.push({
-          cmd: cmd.replace(bregex, ''),
+          cmd: cmd.replace(bregex, '').split('|'),
           variadic: variadic
         })
       }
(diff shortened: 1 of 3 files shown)


iarna:
I will immediately make use of this when you land it. =D

nexdrew:
I have one or two suggestions, will submit them as a review here soon.

lib/command.js
```diff
@@ -108,12 +108,12 @@ module.exports = function (yargs, usage, validation) {
       if (/\.+[\]>]/.test(cmd) && i === splitCommand.length - 1) variadic = true
       if (/^\[/.test(cmd)) {
         parsedCommand.optional.push({
-          cmd: cmd.replace(bregex, ''),
+          cmd: cmd.replace(bregex, '').split('|'),
```
nexdrew: Would be cool to allow whitespace between the aliases, e.g.

```js
yargs.command('get <username | email> [ password ]', 'fetch a user by username or email.')
```

We can make this happen with the following 4 changes:

1. Change line 99 to this:

    ```js
    var splitCommand = extraSpacesStrippedCommand.split(/\s+(?![^\[]*\]|[^<]*>)/)
    ```

2. Change line 108 to this:

    ```js
    if (/\.+[\]>]/.test(cmd.replace(/\s/g, '')) && i === splitCommand.length - 1) variadic = true
    ```

3. Change line 111 to this:

    ```js
    cmd: cmd.replace(/\s/g, '').replace(bregex, '').split('|'),
    ```

4. Change line 116 to this:

    ```js
    cmd: cmd.replace(/\s/g, '').replace(bregex, '').split('|'),
    ```

All tests still pass this way. Let me know what you think.

test/command.js
```diff
@@ -807,4 +807,33 @@ describe('Command', function () {
     argv.should.have.property('someone').and.equal('world')
     commandCalled.should.be.true
   })
+
+  describe('positional aliases', function () {
+    it('allows an alias to be defined for a required positional argument', function () {
+      var parser = yargs('yo')
```
nexdrew: Small nit: change one of these tests to use `.argv` instead of `.parse()`.

For the tests that use `.parse()`, you can remove the args string passed to `yargs()`.

bcoe:
@iarna wow, 23 days already :) would love the extra help testing this; `6.6.0` is looking like a pretty big release:

Issue #735, opened 2016-12-17
.describe with object results in [object Object] key


Janpot:
Following code:

```js
const yargs = require('yargs');
console.log(yargs.describe({ x: 'y' }).argv);
```

prints:

```
{ _: [], '$0': '', '[object Object]': undefined }
```

bcoe:
@Janpot thank you! this will be out in the next release.

PR #736, opened 2016-12-17
Fix [object Object] property on describe with object

Janpot:
fixes https://github.com/yargs/yargs/issues/735


bcoe:
@Janpot your fix will be going out with the next release of yargs, is a fairly big one so would love the help testing, to get  sneak peek run:

Thanks for your contribution.

Issue #659, opened 2016-10-11
Configuration parser called twice


codedot:
```
$ cat test.js
function load(path)
{
        console.warn("warning: %s not found", path);
        return {};
}

require("yargs")
        .config("config", load)
        .global("config")
        .command("command")
        .argv;
$ node test --config /tmp/missing/file command
warning: /tmp/missing/file not found
warning: /tmp/missing/file not found
$ 
```

joelpinheiro:
Hi. I made a pull request with my proposed solution.

nexdrew:
This is a problem concerning the relationship between argv parsing (via yargs-parser) and "lazy" command execution (via yargs), as described below:

1. argv parsing is executed with configured options
2. yargs determines if any parsed args represent a command
3. if so, that command's "builder" is run to configure the options for that command
4. repeat steps 1-3 until no more (nested) commands are found

Since command configuration is lazy/dynamic in this approach, it is necessary to do argv parsing more than once (N+1 times, in fact, where N represents the number of nested subcommands given on the command line). And since the [config file loading/parsing](https://github.com/yargs/yargs-parser/blob/924b014ac25ecbaa2a01f40c2100512a68bdba1d/index.js#L426-L452) is handled by the argv parsing logic in yargs-parser (which has no concept of commands), it is executed each time argv parsing occurs (assuming the config option is valid for that command). When the config option is marked as global, it becomes valid for every command and every level of parsing/execution.

The short-term workaround is to not use a global config option, but rather define it as an option only at the "innermost" commands. For the example above, it would look more like this:

```js
function load (path) {
  console.warn("warning: %s not found", path)
  return { hello: 'world' }
}
require('yargs')
  .command('command', 'Description', yargs => yargs.config('config', load), argv => {
    // by this time, load has been called and its result has been applied to argv
    console.log(argv.config) // path given on command line
    console.log(argv.hello) // value from load function
  })
  .argv
```

I'm not quite sure what the best long-term solution is, but I think we need to add some concept of commands down to the yargs-parser level, which would allow us to parse positional args there (giving them 1st-class citizen status) as well as defer/cache config file loading/parsing as necessary. Perhaps we also need to take a harder look at changing the command API to better support a fully declarative DAG instead of a lazily-built one.

codedot:
[...]

PR #738, opened 2016-12-23
Configuration parser called twice (659)

joelpinheiro:
Added a new verification (cmd.localeCompare('command')).


joelpinheiro:
Regarding the .command:
> Note that commands will not automatically inherit configuration or options of their parent context. This means you'll have to re-apply configuration if necessary, and make options global manually using the global method.

To apply the JSON configuration to command I need to use global method. The problem is that if I re-apply the configuration the load function will be executed twice. So, technically as far as I know, this is the correct behaviour (displaying two times the error).

Now, what should be the best strategy? Should I verify if the config was already called and prevent it to be called again? Should I define the global in the very first stage? Thank you.

nexdrew:
Unfortunately the fix for #659 is not this easy. Please see my comments there.

PR #749, opened 2017-01-04
fix: line wrapping for examples without desc

### LINKED ISSUE #697: if no description is provided, allow first argument to example to take up entire line
Hi,

I'm not sure what the problem is, but over on [etcher](https://etcher.io) today I found a problem where yargs seemed to be applying the line-wrapping over-enthusiastically?
https://github.com/resin-io/etcher/issues/810

@jviotti managed to work around it, but I figured it would only be 'polite' to also report the problem upstream ;-)

maxrimue:
This PR fixes #697. Now, if an example is defined without a description, the usage example will take up the entire line instead of just 50% of it.

### DIFF
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -261,10 +261,24 @@ module.exports = function (yargs, y18n) {
       })
 
       examples.forEach(function (example) {
-        ui.div(
-          {text: example[0], padding: [0, 2, 0, 2], width: maxWidth(examples, theWrap) + 4},
-          example[1]
-        )
+        if (example[1] === '') {
+          ui.div(
+            {
+              text: example[0],
+              padding: [0, 2, 0, 2]
+            }
+          )
+        } else {
+          ui.div(
+            {
+              text: example[0],
+              padding: [0, 2, 0, 2],
+              width: maxWidth(examples, theWrap) + 4
+            }, {
+              text: example[1]
+            }
+          )
+        }
       })
 
       ui.div()
--- a/test/usage.js
+++ b/test/usage.js
@@ -1337,6 +1337,25 @@ describe('usage tests', function () {
       })
     })
 
+    it('should not wrap left-hand-column if no description is provided', function () {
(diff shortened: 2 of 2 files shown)


lib/usage.js
```diff
@@ -261,10 +261,24 @@ module.exports = function (yargs, y18n) {
...
       examples.forEach(function (example) {
-        ui.div(
-          {text: example[0], padding: [0, 2, 0, 2], width: maxWidth(examples, theWrap) + 4},
-          example[1]
-        )
+        if (example[1] === '') {
+          ui.div(
+            {
+              text: example[0],
+              padding: [0, 2, 0, 2]
```
bcoe: I think rather than having no wrap here, you should use `theWrap`; this way the example will still wrap onto multiple lines at the width boundary, but will take up a full line.

maxrimue: @bcoe So you mean I should add a width property like in line 276? 😄

bcoe: @maxrimue I think you want ` width: maxWidth(theWrap) + 4`.

maxrimue: @bcoe I've looked at this again but it seems that even without the width property it takes up the full line:
bildschirmfoto 2017-01-10 um 14 48 51

Isn't that what you meant with `but it will take up a full line`?
Using the width property like you suggested makes it wrap after each character:
bildschirmfoto 2017-01-10 um 14 44 41

Anyway, what I besides would suggest is perhaps adding a newline after each example to make it more distinguishable:
bildschirmfoto 2017-01-10 um 14 50 53

This can be achieved very easily with a bottom padding of 1.

bcoe:
@maxrimue was hacking all weekend landing various tickets for 7.x; this work should be on `master` now 👍 

would love your help testing, _like everything!_

FYI: meet @pahdo, and @stevezhu, they were helping land a ton of features this weekend at HackIllinois! 🎉

PR #765, opened 2017-01-21
feat: introduce custom yargs error object

bcoe:
BREAKING CHANGE: yargs will no longer aggressively supress errors, allowing errors that are not generated internally to bubble.

CC: @iarna, @satazor curious as to what you think of this implementation.

### DIFF
--- a/lib/completion.js
+++ b/lib/completion.js
@@ -46,8 +46,9 @@ module.exports = function (yargs, usage, command) {
       if (handlers[args[i]] && handlers[args[i]].builder) {
         const builder = handlers[args[i]].builder
         if (typeof builder === 'function') {
-          const y = builder(yargs.reset())
-          return y ? y.argv : null
+          const y = yargs.reset()
+          builder(y)
+          return y.argv
         }
       }
     }
--- a/lib/completion.js
+++ b/lib/completion.js
@@ -44,7 +44,11 @@ module.exports = function (yargs, usage, command) {
     var handlers = command.getCommandHandlers()
     for (var i = 0, ii = args.length; i < ii; ++i) {
       if (handlers[args[i]] && handlers[args[i]].builder) {
-        return handlers[args[i]].builder(yargs.reset()).argv
+        const builder = handlers[args[i]].builder
+        if (typeof builder === 'function') {
+          const y = builder(yargs.reset())
+          return y ? y.argv : null
+        }
       }
     }
 
--- a/lib/yerror.js
+++ b/lib/yerror.js
@@ -1,11 +1,10 @@
-const util = require('util')
-
 function YError (msg) {
   this.name = 'YError'
   this.message = msg || 'yargs error'
   Error.captureStackTrace(this, YError)
 }
 
-util.inherits(YError, Error)
+YError.prototype = Object.create(Error.prototype)
+YError.prototype.constructor = YError
 
 module.exports = YError
--- a/lib/yerror.js
+++ b/lib/yerror.js
@@ -1,0 +1,11 @@
+const util = require('util')
+
+function YError (msg) {
+  this.name = 'YError'
+  this.message = msg || 'yargs error'
+  this.stack = (new Error()).stack
+}
+
+util.inherits(YError, Error)
+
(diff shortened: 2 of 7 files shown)


bcoe:
@JaKXz you were mentioning you'd be happy to start helping out on yargs as well? Would love the extra set of 👀 on this.

lib/completion.js
```diff
@@ -44,7 +44,11 @@ module.exports = function (yargs, usage, command) {
     var handlers = command.getCommandHandlers()
     for (var i = 0, ii = args.length; i < ii; ++i) {
       if (handlers[args[i]] && handlers[args[i]].builder) {
-        return handlers[args[i]].builder(yargs.reset()).argv
+        const builder = handlers[args[i]].builder
```
bcoe: this logic was creating an exception that we were quietly ignoring. It came up when the builder was an object rather than a function, or the builder returned nothing -- because this error wasn't a `YError` tests started failing.

test/completion.js
```diff
@@ -97,15 +97,13 @@ describe('Completion', function () {
...
             })
-            .argv
           })
           .command('cmd2', 'second command', function (subYargs) {
             subYargs.options({
               opt2: {
                 describe: 'second option'
               }
             })
-            .argv
```
bcoe: this test wasn't using the command api appropriately.

lib/yerror.js
```diff
@@ -0,0 +1,11 @@
+const util = require('util')
+
+function YError (msg) {
+  this.name = 'YError'
+  this.message = msg || 'yargs error'
+  this.stack = (new Error()).stack
```
iarna: I think this last line would be better as:

```
  Error.captureStackTrace(this, YError)
```
By using `captureStackTrace` you won't get the `YError` constructor in the call stack.

yargs.js
```diff
@@ -867,7 +868,8 @@ function Yargs (processArgs, cwd, parentRequire) {
       try {
         args = parseArgs(processArgs)
       } catch (err) {
-        usage.fail(err.message, err)
+        if (err instanceof YError) usage.fail(err.message, err)
```
iarna: Ah, so this is the magic moment!

lib/completion.js
satazor: So this isn't actually related to YError but a bug that was discovered because of it?

lib/yerror.js
```diff
@@ -0,0 +1,11 @@
+const util = require('util')
```
satazor: Any reason for not using https://www.npmjs.com/package/create-error-class?

yargs.js
satazor: Yep!

lib/yerror.js
iarna: That's a lotta code to do a very little thing… Ben's implementation is 9 lines (a third of which are whitespace) of very simple js, versus … 62 lines. Now concise isn't automatically better, but personally I find what's here way easier to grok. A lot of the weirdness in those 62 lines are about browser compatibility, is that a thing that yargs cares about?

bcoe: @satazor @iarna hadn't thought of the front-end use-case; I know in the past a couple people had used something like browserify to load yargs in the browser. It might be worth switching from `util.inherits` to something like:

```js
MyError.prototype = Object.create(Error.prototype);
MyError.prototype.constructor = MyError;
```

Which I lifted rom the MDN docs.

@satazor to answer your question; I get a lot of pressure from the user-base to keep yargs to as few dependencies as possible:

https://twitter.com/guillaumeQD/status/807917750425440256

More so than a lot of libraries I work on, this has motivated me to opt for less dependencies where possible -- I think people are extra sensitive about library-size, because yargs is something you tack onto a project to add CLI functionality, it's not seen as a core part of the application logic.

satazor: 👍

JaKXz: This was my only nitpick, thanks for bringing it up @satazor - here's the nodejs discussion on the discouragement of using `util.inherits`: https://github.com/nodejs/node/issues/4179

Issue #764, opened 2017-01-20
.conflicts for a global options ignored in command, unless explicitly added to command handler


rasmuslp:
I'm trying to define some options, which should be mutually exclusive, but the `.conflicts` option specified on top-level seems to be ignored when a command is provided. Is this the intended behaviour?

#### v1

```JS
// v1: Plain .conflicts
require('yargs')
    .command({
        command: 'a'
    })
    .option('b', {
        global: true
    })
    .option('c', {
        global: true
    })
    .conflicts('b', 'c')
    .argv;
```

If I run v1 with no command, yargs stop me. But if I add the command, it doesn't:
``` bash
$ node v1.js -b 2 -c 3
Arguments b and c are mutually exclusive

$ node v1.js a -b 2 -c 3
# No output
```

#### v2
```JS
// v2: Explicitly added .conflicts to command builder
require('yargs')
    .command({
        command: 'a',
        builder: function (yargs) {
            return yargs
                .conflicts('b', 'c');
        }
    })
    .option('b', {
        global: true
    })
    .option('c', {
        global: true
    })
    .conflicts('b', 'c')
    .argv;
```

In v2 yargs stops me:
``` bash
$ node v2.js -b 2 -c 3
Arguments b and c are mutually exclusive

$ node v2.js a -b 2 -c 3
Arguments b and c are mutually exclusive
```

bcoe:
@rasmuslp this is currently intended behavior; but we're in discussion regarding changing the default behavior of options in yargs@7. Please join the conversation in #742 and provide your two cents ... at this point I'm pretty convinced that options should default to being global.

rasmuslp:
Alright, thank you for the response and the pointer.

bcoe:
@rasmuslp I just released a candidate release of yargs 7.x, which switches most settings to being applied globally -- I think this makes the API less confusing.

I would love your help testing out the new release against some of your libraries, before we push things out to a wider audience:

`npm cache clear; npm i yargs@next`

Let me know what you think.

Issue #762, opened 2017-01-19
Bug: .check ignored when .command is specified and complied with


giacecco:
I'm using version 6.6.0. The code below intentionally triggers an error:

```
var argv = require('yargs')
    .check((argv, opts) => { throw(new Error("Error!")); return false; })
    .argv;
console.log("Hello world!");
```

```
$ node test.js

Error!
$ 
```

... but if I add a ```.command``` directive, it does not:

```
var argv = require('yargs')
    .command("foo")
    .check((argv, opts) => { throw(new Error("Error!")); return false; })
    .argv;
console.log("Hello world!");
```

```
$ node test.js foo
Hello world!
$ 
```

Oddly enough, if I don't provide the specified command, then ```.check``` works again:

```
$ node test.js bar
Commands:
  foo

Error!
$
```

bcoe:
@giacecco I believe this is working as expected, because checks are not applied globally by default; rather they are applied at the level they're executed.

So this would have the behavior you expect:

```js
var argv = require('yargs')
    .command("foo", (yargs) => {
      yargs.check((argv, opts) => { throw(new Error("Error!")); return false; })
    })
    .check((argv, opts) => { throw(new Error("Error!")); return false; })
    .argv;
console.log("Hello world!");
```

This re-registers the check on the instance of yargs that is executed for the command.

_Now!_ I understand the fact that many of these settings are not applied globally (to all commands)_ is confusing. We're in discussion on `yargs@7.x` #742, to change this default behavior. Would love to have you add your thoughts.

giacecco:
@bcoe thanks, but I've tried your code and it does not work. Of course I am not as knowledgeable as you of yargs, but there is something wrong in the ```yargs``` object handed over to the handler function. ```check``` simply does not exist there. Try the code below:

```
var argv = require('yargs')
    .command("foo", "[some description]", { }, (yargs) => {
        yargs.check((argv, opts) => { throw(new Error("Error 1!")); return false; })
    })
    .check((argv, opts) => { throw(new Error("Error 2!")); return false; })
    .argv;
console.log("Hello world!");
```

You will get:

```
$ node test.js foo
test.js foo

yargs.check is not a function
$
```

I don't have the permissions to re-open this issue.

bcoe:
@giacecco the third argument is the `builder`, the fourth argument is the `handler` (you can optionally choose to react to parsed options for a command here rather than at the top-level of the application:

```js
var argv = require('yargs')
[...]

PR #766, opened 2017-01-22
feat: rethink how options are inherited by commands

bcoe:
BREAKING CHANGE: by default options, and many of yargs' parsing helpers will now default to being applied globally; such that they are no-longer reset before being passed into commands.

This is to address the fact that the existing behavior has created a lot of confusion: #659, #762, #764, etc.

This is going to be a pretty major change, and would appreciate as much feedback as possible as it's built out.

@nexdrew I'm going to merge your work from #734 into this.


bcoe:
@nexdrew, @JaKXz, @maxrimue, @lrlna this is getting much closer to being finished, and I would love some feedback.

I've combined a lot of yargs' configuration logic into `populateParserHintObject`, `populateParserHintArray`; This both cuts down on repeated code, and provides a single location to enable options/settings globally.

Along the way I've been cleaning up some API inconsistencies, and correcting a few bugs. For starters, it would be wonderful to have folks run this branch against some of their projects, specifically projects that use commands heavily (CC: @nexdrew I know you have a few).

lib/command.js
```diff
@@ -131,47 +131,62 @@ module.exports = function (yargs, usage, validation) {
...
+
     if (!yargs._hasOutput()) populatePositionals(commandHandler, innerArgv, currentContext, yargs)
 
     if (commandHandler.handler && !yargs._hasOutput()) {
       commandHandler.handler(innerArgv)
     }
+
+    // we apply validation post-hoc, to so that custom
+    // checks get passed populated positional arguments.
+    yargs._runValidation(innerArgv, aliases)
```
bcoe: the little bit of refactoring in this method was to hold off on applying validation until after we've populated variadic values -- this was largely so that one can provide a `.check()` at the top level, and have it passed the appropriate post-processed object after commands are parsed.

yargs.js
```diff
@@ -993,27 +987,31 @@ function Yargs (processArgs, cwd, parentRequire) {
...
-        validation.implications(argv)
-        validation.conflicting(argv)
+        self._runValidation(argv, aliases)
       }
     }
 
     return setPlaceholderKeys(argv)
   }
 
+  self._runValidation = function (argv, aliases) {
```
bcoe: `_runValidation` was pulled into a helper, to make it easier to run optionally -- this was so that we could hold off on running it until the last moment in the case of commands.

README.md
```diff
@@ -1112,9 +1107,9 @@ Options:
 Missing required arguments: run, path
 ```
 
-<a name="demandCommand"></a>.demandCommand(min, [minMsg])
+<a name="demandCommand"></a>.demandCommand([min=1], [minMsg])
```
maxrimue: Very awesome, I love this one

lib/command.js
```diff
@@ -131,47 +131,62 @@ module.exports = function (yargs, usage, validation) {
...
+      aliases = innerYargs.parsed.aliases
     }
+
     if (!yargs._hasOutput()) populatePositionals(commandHandler, innerArgv, currentContext, yargs)
 
     if (commandHandler.handler && !yargs._hasOutput()) {
       commandHandler.handler(innerArgv)
     }
+
+    // we apply validation post-hoc, to so that custom
```
maxrimue: tiny typo here

README.md
```diff
@@ -418,6 +418,9 @@ Check that certain conditions are met in the provided arguments.
 If `fn` throws or returns a non-truthy value, show the thrown error, usage information, and
 exit.
 
+`global` indicates whether or not `check()` should be enabled both
```
JaKXz: nit: I think 

> indicates whether `check()` should be enabled...

reads a little clearer, since I don't have to think about negations and my pyrate brain don't hurt.

yargs.js
```diff
@@ -172,56 +172,129 @@ function Yargs (processArgs, cwd, parentRequire) {
...
+
+  self.skipValidation = function (keys) {
+    populateParserHintArray('skipValidation', keys)
+    return self
+  }
+
+  function populateParserHintArray (type, keys, value) {
+    keys = [].concat(keys)
+    keys.forEach(function (key) {
+      self.global(key)
```
nexdrew: Instead of explicitly marking everything as global, I kinda figured it would be easier to just invert the reset logic to assume that everything is global unless explicitly marked as _non_-global.

In that approach, we could perhaps change `options.global` to be `options.local`, treat `yargs.global(opts, false)` as a synonym for a new API method `yargs.local(opts)`, and only worry about resetting local options in the reset method. (I think positional args for a command would be the only options to default as local.)

Thoughts? Is my view on the matter overly simplistic?

yargs.js
```diff
@@ -172,56 +172,129 @@ function Yargs (processArgs, cwd, parentRequire) {
...
+      })
+      builder(temp)
+    } else if (typeof key === 'object') {
+      // an object of key value pairs: {'x': parse () {}, 'y': parse() {}}
+      Object.keys(key).forEach(function (k) {
+        builder(k, key[k])
+      })
+    } else {
+      // a single key value pair 'x', parse() {}
+      self.global(key)
```
nexdrew: Similarly, this approach means that you _must_ call `.global(opts, false)` _last_ to disable an option from being global.

E.g. code like this will not have the desired effect because the call to `.coerce()` makes the option global again:

```js
require('yargs')
  .describe('opt', 'My local option')
  .string('one')
  .global('opt', false)
  .coerce('opt', opt => 'local_' + opt)
  .argv
```

This is probably not a big deal, but we should either mention this caveat in the README/docs or go with another approach that doesn't have this caveat.

Issue #742, opened 2016-12-29
Version 7.0.0


maxrimue:
Again, another ticket to have all discussions and info on the next major in one thread. Feel free to propose (or make) changes. Everyone is invited to discuss here!

## Breaking

- [x] ~Remove deprecated `.demand()` in favour of `.demandCommand()` and `.demandOption()` as introduced in v6.6.0 (#740)~ _since we're breaking a ton of stuff in this release, I'm tempted to leave `.demand()` for now (I don't see a reason to add an extra annoyance for folks upgrading)`.
- [x] Alert misuse of yargs API (#739)
- [x] Don't handle program errors within yargs (#455)
- [x] Apply options defined in global scope globally by default
- [x] Rearranged order of precedence for args (#627)
- [x] Rearranged order of precedence for context/default values (#724)
- [x] default commands, finally!

Additionally, we wanted to pull of a readme overhaul. Perhaps a major would be a great time to do this, or we postpone this. Also, we discussed adding certain tutorials to the yargs website ([yargs.github.io/#26](https://github.com/yargs/yargs.github.io/issues/26)), what's the status on that?

cc: @bcoe @nexdrew @lrlna @jameswomack @elas7

xstable:
You also need to remove it from documentation. 
Today I run into the demandCommand()-Error because I give it a try like descriped in the docu.

Can you explain what to use now instead of demandCommand?

lrlna:
nice one @maxrimue ! ✨ 

errrr about tutorials -- i tried and horribly failed; i can't write cohesively thought out things for the life of me :(

maxrimue:
Thanks @lrlna! ✨ 

How about we simply work on this together in the dedicated issue? We could collect ideas first on how to position the tutorials, what they should be about etc., and then eventually write them together in a wip pull request, what do you think? I'm not experienced in this either but I'm sure together we can do diss 😁

wtgtybhertgeghgtwtg:
Will `yargs@7.0.0` keep support for `node@0.12.0` or require `node@4.0.0`?

maxrimue:
@wtgtybhertgeghgtwtg Our plan originally was to drop support for older Node.js versions early this year. Perhaps this next major would be a good time to do that, as both v0.10 and v0.12 are not maintained anymore and insecure to use. What does everyone think about this?

bcoe:
[...]

PR #767, opened 2017-01-22
yargs 7.x

bcoe:
see: #742


wtgtybhertgeghgtwtg:
Is the minimum required `node` version going to be bumped for `yargs@7.0.0`?

lhecker:
@bcoe I'm currently facing the problem that, if I define subcommands, exceptions thrown by `coerce` don't cause yargs to abort the program anymore. I believe it's because of [those](https://github.com/yargs/yargs/blob/57a8490d3843acad83d387c782efc61e7facf1b7/lib/command.js#L158) [two](https://github.com/yargs/yargs/blob/57a8490d3843acad83d387c782efc61e7facf1b7/lib/command.js#L173) lines. Since they pass `true` as the `_skipValidation` parameter [this](https://github.com/yargs/yargs/blob/57a8490d3843acad83d387c782efc61e7facf1b7/yargs.js#L1048-L1049) block of code will never be executed and thus the value of `parsed.error` is essentially thrown away!

bcoe:
@lhecker I think you're right, `if (parsed.error) throw new YError(parsed.error.message)` should be pulled into the new `_validation` helper -- mind opening a ticket, or _still_ a pull request against the `7.x` branch?

lhecker:
@bcoe I don't really feel that comfortable working on yargs yet as in fact I only started using this library yesterday. For instance I really don't know what this "`_validation` helper" is. 😅 But since the fix is surely very easy to write the easiest way is probably if you fix it yourself - If you don't mind of course. 🙂

bcoe:
@lhecker I'm actually at a hacking event all weekend, working on getting new folks onboarded to yargs ... my point being, if you're interested in contributing, give it a shot 👍 nothing to be embarrassed about.

PR #772, opened 2017-01-29
docs: remove demand examples

### LINKED ISSUE #771: Documentation on options() demand inconsistent
The README example for options includes `demand` as a parameter:

```
var argv = require('yargs')
    .options({
      'f': {
        alias: 'file',
        demand: true,
        default: '/etc/passwd',
        describe: 'x marks the spot',
        type: 'string'
      }
    })
    .argv;
```

but the list of valid parameters doesn't include `demand`:

```
defaultDescription: string, use this description for the default value in help content, see default()
demandOption: boolean or string, demand the option be given, with optional error message, see demandOption()
desc/describe/description: string, the option description for help content, see describe()
```

Is it still valid?  Deprecated?  Should it be removed from the example?

maxrimue:
Replaces `demand` in examples with `demandOption`. Closes #771.

### DIFF
--- a/yargs.js
+++ b/yargs.js
@@ -517,6 +517,7 @@ function Yargs (processArgs, cwd, parentRequire) {
 
       var demand = opt.demand || opt.required || opt.require
 
+      // deprecated, use 'demandOption' instead
       if (demand) {
         self.demand(key, demand)
       }
--- a/README.md
+++ b/README.md
@@ -1081,12 +1081,12 @@ require('yargs')
     'run': {
       alias: 'r',
       describe: 'run your program',
-      demand: true
+      demandOption: true
     },
     'path': {
       alias: 'p',
       describe: 'provide a path to file',
-      demand: true
+      demandOption: true
     },
     'spec': {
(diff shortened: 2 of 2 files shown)


yargs.js
```diff
@@ -523,6 +523,7 @@ function Yargs (processArgs, cwd, parentRequire) {
 
       var demand = opt.demand || opt.required || opt.require
 
+      // deprecated, use 'demandOption' instead
       if (demand) {
         self.demand(key, demand)
```
JaKXz: Just curious, should there be a `console.warn` here?

bcoe: @JaKXz I've been bit a couple times where I have a dependency of a dependency that starts printing out ugly warning messages, that are a pain in the butt to get rid of in the output.

I was thinking, with yargs, it might be worth deprecating in documentation without actually spitting out warning messages.

bcoe:
@maxrimue looks good to me, but perhaps make the pull request against the `7.x` branch, since we're marching towards this release?

PR #775, opened 2017-02-03
Allow configuration to extend other config files

rcoy-v:
This is a solution to https://github.com/istanbuljs/nyc/issues/503. I still need to test this with nyc, but I would like to get feedback early.

This only applies the extends functionality when `.config()` is given an object. Please let me know if there are other cases that should use this as well.


bcoe:
@rcoy-v thanks for this 👍 I'm currently digging out of a mound of OSS tickets, and am looking very much forward to getting to this one.

rcoy-v:
I tested this through nyc with one of our internal projects. It fulfilled my requirement at least for extending configurations.

bcoe:
@rcoy-v my only concern is projects that happen to have an `extends` key in their configuration; @nexdrew is usually pretty good at giving feedback on these edge-cases.

we're marching towards a [7.x release](https://github.com/yargs/yargs/pull/767), so it's a good time to add this feature.

rcoy-v:
I agree, this is essentially adding a new reserved word. I would treat it as a major semver change. I did not update the README in this pr to detail the `extends` keyword. I wanted to wait for feedback.

.editorconfig
```diff
@@ -0,0 +1,6 @@
+root = true
+
+[*.js]
+end_of_line = lf
+indent_style = space
+indent_size = 2
```
maxrimue: this is great, thanks for adding it! Could you perhaps add `insert_final_newline = true`?

bcoe:
@rcoy-v mind rebasing this against the `7.x` branch, and opening this pull against `7.x` rather than `master`? marching towards this release.

rcoy-v:
Closed in favor of #779.

PR #779, opened 2017-02-10
Allow provided config object to extend other configs

rcoy-v:


lib/apply-extends.js
```diff
@@ -0,0 +1,19 @@
+var fs = require('fs')
+var path = require('path')
+var assign = require('./assign')
+
+function applyExtends (config, cwd) {
```
nexdrew: [Tweak 1/3] Add optional third argument here, such as:

```js
function applyExtends (config, cwd, subkey) {
```

lib/apply-extends.js
```diff
@@ -0,0 +1,19 @@
...
+  var defaultConfig = {}
+
+  if (config.hasOwnProperty('extends')) {
+    var pathToDefault = path.join(cwd, config.extends)
+    delete config.extends
+
+    defaultConfig = JSON.parse(fs.readFileSync(pathToDefault, 'utf8'))
+    defaultConfig = applyExtends(defaultConfig, path.dirname(pathToDefault))
```
nexdrew: [Tweak 2/3] Check for and use the optional third argument here:

```js
if (subkey) defaultConfig = defaultConfig[subkey] || {}
defaultConfig = applyExtends(defaultConfig, path.dirname(pathToDefault), subkey)
```

yargs.js
```diff
@@ -304,6 +305,7 @@ function Yargs (processArgs, cwd, parentRequire) {
     argsert('[object|string] [string|function] [function]', [key, msg, parseFn], arguments.length)
     // allow a config object to be provided directly.
     if (typeof key === 'object') {
+      key = applyExtends(key, cwd)
```
nexdrew: [Tweak 3/3] Then we can make use of `applyExtends` in the `pkgConf` method as well:

```js
  self.pkgConf = function (key, path) {
    argsert('<string> [string]', [key, path], arguments.length)
    var conf = null
    var obj = pkgUp(path)

    // If an object exists in the key, add it to options.configObjects
    if (obj[key] && typeof obj[key] === 'object') {
      conf = applyExtends(obj[key], cwd, key) // <-- this line changed
      options.configObjects = (options.configObjects || []).concat(conf)
    }

    return self
  }
```

lib/apply-extends.js
nexdrew: [Question] I wonder if we should guard against circular references? It might be unlikely but would overflow the stack. Perhaps this submodule should keep track of the paths it has already loaded?

lib/apply-extends.js
```diff
@@ -0,0 +1,19 @@
+var fs = require('fs')
+var path = require('path')
+var assign = require('./assign')
+
+function applyExtends (config, cwd) {
+  var defaultConfig = {}
+
+  if (config.hasOwnProperty('extends')) {
+    var pathToDefault = path.join(cwd, config.extends)
```
nexdrew: [Comment] I like [@JaKXz's idea](https://github.com/istanbuljs/nyc/issues/503#issuecomment-274916602) of supporting modules by name as well (e.g. `"extends": "my-lib"` instead of just `"extends": "node_modules/my-lib/some-file"`), but since yargs isn't always loading the original "some-file" and doesn't know its path, I like your general approach as the next best thing.

bcoe:
@rcoy-v I'm going to try to put a few hours in against yargs tomorrow, and would love to work on landing this.

Do you have any time to address @nexdrew's suggestions, I could also work on making a few edits and landing this myself -- your call.

rcoy-v:
I might be able to pick this up again tomorrow, but I can't promise that. I'm fine if you want to make edits to this pull request.

bcoe:
@rcoy-v awesome, I'm really excited about this feature.

test/yargs.js
```diff
@@ -1146,6 +1155,21 @@ describe('yargs dsl tests', function () {
...
+        var argv = yargs
+            .config({
+              extends: './test/fixtures/extends/config_1.json',
+              a: 1
+            })
+          .argv
+
+        argv.a.should.equal(1)
+        argv.b.should.equal(22)
+        argv.z.should.equal(15)
```
JaKXz: 🔥 this is great.

my only nitpick is about the indenting on line 1162.

rcoy-v: fixed

yargs.js
rcoy-v: done

lib/apply-extends.js
rcoy-v: done

lib/apply-extends.js
rcoy-v: I feel yargs is too general of a tool to make assumptions for shorthand extends like `my-lib`. For more context, my use case in nyc probably won't even extend a root `.nycrc` in another package. Rather, it will extend something like `my-lib/configs/.nycrc`.

bcoe:
@rcoy-v @JaKXz I'm at [Hack Illinois](https://hackillinois.org/) this coming weekend; I was thinking getting yargs@7 out over the weekend would be a great task for my team -- tldr; expect to see a bunch of these tickets land on the weekend.

test/yargs.js
```diff
@@ -1411,6 +1433,13 @@ describe('yargs dsl tests', function () {
 
       argv.foo.should.equal('a')
     })
+
+    it('should apply default configurations from extended packages', function () {
+      var argv = yargs().pkgConf('foo', 'test/fixtures/extends/packageA').argv
```
nexdrew: The `path` argument to the `.pkgConf(key, path)` method is optional. When I don't pass a value, I get this error:

```
path.js:7
    throw new TypeError('Path must be a string. Received ' + inspect(path));
    ^

TypeError: Path must be a string. Received undefined
    at assertPath (path.js:7:11)
    at Object.join (path.js:1211:7)
    at applyExtends (/Users/abg/git/nexdrew/github/yargs/lib/apply-extends.js:18:30)
    at Object.Yargs.self.pkgConf (/Users/abg/git/nexdrew/github/yargs/yargs.js:479:14)
    at Object.<anonymous> (/Users/abg/git/nexdrew/github/demo/pr779/index.js:2:4)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
```

Perhaps we should change line 479 of yargs.js to this:

```js
      conf = applyExtends(obj[key], path || cwd, key) // <-- add the `|| cwd` part
```

and maybe add a test for this scenario as well?

Otherwise, this is looking really good!

bcoe:
@nexdrew is this the new argument validation, I think we might find a few edge-cases where we need to loosen the rules a bit.

nexdrew:
@bcoe 

> is this the new argument validation?

Negative. This error is thrown by the builtin Node `path` module. The `argsert` definition for `.pkgConf()` looks correct to me:

```js
argsert('<string> [string]', [key, path], arguments.length)
```

The 2nd string is optional.

JaKXz:
@nexdrew won't `arguments.length` give you a number [which doesn't match the type of optional string]?

nexdrew:
@JaKXz As I understand it, the 3rd argument to `argsert()` should be the number of arguments given. The actual types are checked via the 2nd argument, which is an array representing the actual argument values.

bcoe:
@JaKXz, @nexdrew is correct -- the use of `argsert` in yargs is dead-simple, based on various performance bike-shedding discussions.

nexdrew:
@rcoy-v See [this PR against your branch](https://github.com/rcoy-v/yargs/pull/1) to add the changes I requested. I _think_ that will put this over the finish line. Let me know what you think. Thanks!

lib/apply-extends.js
```diff
@@ -0,0 +1,41 @@
...
+var previouslyVisitedConfigs = []
+
+function checkForCircularExtends (path) {
+  if (previouslyVisitedConfigs.indexOf(path) > -1) {
+    throw new YError("Circular extended configurations: '" + path + "'.")
+  }
+}
+
+function getPathToDefaultConfig (cwd, pathToExtend) {
+  return path.isAbsolute(pathToExtend) ? pathToExtend : path.join(cwd, pathToExtend)
```
rcoy-v: I use `path.isAbsolute` here. According to node docs this was introduced in v0.11.2. I'm not sure where the topic of v0.10 support has landed. Maybe it depends on outcome of latest discussion in #742. If `path.isAbsolute` cannot be used, I will figure something else out.

test/yargs.js
```diff
@@ -1,16 +1,21 @@
...
 require('chai').should()
 
 describe('yargs dsl tests', function () {
   beforeEach(function () {
-    yargs.reset()
+    yargs = require('../')
+  })
+
+  afterEach(function () {
+    delete require.cache[require.resolve('../')]
```
rcoy-v: `yargs.reset()` does not fully reset the yargs instance. Testing configs was not atomic. So I made it that each test has a truly fresh yargs.

test/yargs.js
rcoy-v: Done with your help in 5353a2bb169d20a5396e76e0aa27cf6536de2eaa.

lib/apply-extends.js
scriptdaemon: Would `path.resolve(pathToExtend)` work?

nexdrew: Yes, I believe `path.resolve()` would work here. I actually thought that `path.join()` worked the same way, which is why I said "relative or absolute" in the docs, but apparently it doesn't.

bcoe:
@rcoy-v, @JaKXz, @nexdrew, phew! thanks for all the hard-work on this feature, laded this weekend as part of our big push towards yargs 7.x.

I would love some help testing, before we promote this to a wider audience:

@rcoy-v this means we can _finally_ upstream this functionality to `nyc` 😄

PR #784, opened 2017-02-18
fix: address positional argument strict() bug introduced in #766

bcoe:
this addresses a bug related to positional command arguments in strict mode:

`command <foo> [bar]`

@nexdrew does this solve the issues that you brought up in slack?

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -178,13 +178,13 @@ module.exports = function (yargs, usage, validation) {
       positionalMap = populatePositionals(commandHandler, innerArgv, currentContext, yargs)
     }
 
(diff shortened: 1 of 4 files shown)


bcoe:
@nexdrew checked in one more update; noticed while testing the regression introduced in #766 that we were executing command handlers when validation failed, this was an order of operation bug.

bcoe:
@nexdrew I'm going ahead and landing these couple tweaks, so that I can start adding some tests to your default command branch.

If we find more issues around commands and the `global`/`local` changes, we can land further patches.

Issue #724, opened 2016-12-02
default value interferes with passing context to sub-commands


bcoe:
setting a default value for a global option overrides the context object.

```js
yargs()
          .command('batman', 'batman command', function (yargs) {
            yargs.command('robin', 'robin command', {}, function (_argv) {
              argv = _argv
            })
          })
          .option('state', {
            default: undefined,
            global: true
          })
          .parse('batman robin --what', {
            state: 'grumpy but rich'
          }, function (_err, argv, _output) {})
```

maxrimue:
@bcoe I _think_ I tracked down the problem. It seems like `argv` will _always_ take precedence over any context passed to `.parse()`. For example, this will fail:
```javascript
var parsed = yargs.parse('--foo bar', {
  foo: 'not bar'
})

parsed.foo.should.equal('not bar')
```
If I'm not totally wrong here, it seems that the problem lies in [L875](https://github.com/yargs/yargs/blob/master/yargs.js#L875), where `assign` is called to merge context and argv, _but_ in a way that argv takes precedence. Changing the order in this statement made the aforementioned test case pass.

Isn't the context object meant to take precedence over argv?

maxrimue:
@bcoe I would propose we add a new parameter for the parser like: `Parser.detailed(args, options, state)` so that the parser can apply state in precedence over default values and give back the argv like that. What do you think? Would take on this when I know what to do.

bcoe:
@maxrimue before we add any more configuration options; I think I'd be tempted to just swap the order of precedence. I argue that context should take precedence.

Would you be okay with making this one of the breaking changes in #742 (CC: @nexdrew).

PR #786, opened 2017-02-18
fix: context should override parsed argv

bcoe:
BREAKING CHANGE: context now takes precedence over argv

----
context object should take precedence over parsed values in argv.

fixes: #724 

CC: @scriptdaemon I think this is better behavior for writing chat-bots using yargs.


bcoe:
@stevezhu awesome 👍 thanks for landing this.

Issue #627, opened 2016-09-21
Discuss: Order of precedence for program argument values


jaridmargolin:
## The Problem

The current order of precedence for resolving arguments is as follows:
1. CLI flags
2. Config file (if configured)
3. Env var (if configured)
4. Default values (if configured)

I wanted to start an active discussion around the possibility of swapping the order of precedence for config files and env vars (or less favorably, offering a way to configure the order of precedence):
1. CLI flags
2. Env var (if configured)
3. Config file (if configured)
4. Default values (if configured)
## Why

Precedence... I believe there is already and standard and understood order of precedence in the community. I also believe consistency is paramount. When picking up a tool the less domain knowledge I need have to know the better.

Notable examples:
1. [npm](https://docs.npmjs.com/misc/config) - The package manager for JavaScript
2. [aws](http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#config-settings-and-precedence) - A unified tool that provides a consistent interface for interacting with all parts of AW
3. [docker](https://docs.docker.com/engine/reference/commandline/cli/) - Docker is the world's leading software containerization platform
4. [ansible](http://docs.ansible.com/ansible/intro_configuration.html#environmental-configuration) - An IT automation tool
5. [git](https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables#Committing) - A distributed version control system designed to handle everything from small to very large projects with speed and efficiency.
6. [viper](https://github.com/spf13/viper#why-viper) - Viper is a complete configuration solution for go applications including 12 factor apps

addaleax:
Maybe a bit related: I mentioned on twitter how one thing I like about npm’s config system is that npm does not only respect settings that are available in environment variables, but exports its entire set of configuration settings to subprocesses using env vars.

Maybe yargs applications could do the same thing? That would make a `env var` > `config file` precedence even more useful.

bcoe:
@jaridmargolin environment variables should now be taking precedence over config files, would love your help testing:

`npm i yargs@next --save`

thanks for your help nit picking on this feature.

PR #787, opened 2017-02-18
fix: pull in yargs-parser with modified env precedence

bcoe:
BREAKING CHANGE: environment variables now take precedence over config files.

fixes: #627 

CC: @addaleax, @jaridmargolin


nexdrew:
We should update the docs to clarify this change here: https://github.com/yargs/yargs#envprefix

I'll submit a PR soon.

Otherwise, great work on 7.x @bcoe, @stevezhu, @pahdo, and everyone else!

PR #790, opened 2017-02-21
fix: running parse() multiple failed when help() enabled

### LINKED ISSUE #776: Crash on second call to help command
EDIT: The same error occurs doing anything a second time when using `help()`.

Version:

`6.6.0`

Error:

```
/mnt/c/Users/kwilliams/Documents/Projects/cheevobot/node_modules/yargs/yargs.js:899
          return k.length > 1
                  ^

TypeError: Cannot read property 'length' of undefined
    at /mnt/c/Users/kwilliams/Documents/Projects/cheevobot/node_modules/yargs/yargs.js:899:19
    at Array.filter (native)
    at parseArgs (/mnt/c/Users/kwilliams/Documents/Projects/cheevobot/node_modules/yargs/yargs.js:898:42)
    at Object.Yargs.self.parse (/mnt/c/Users/kwilliams/Documents/Projects/cheevobot/node_modules/yargs/yargs.js:499:18)
    at SteamBot.exec (/mnt/c/Users/kwilliams/Documents/Projects/cheevobot/lib/steam-bot.js:81:18)
    at SteamUser.SteamBot._client.on (/mnt/c/Users/kwilliams/Documents/Projects/cheevobot/lib/steam-bot.js:35:12)
    at emitThree (events.js:116:13)
    at SteamUser.emit (events.js:195:7)
    at SteamUser._emitIdEvent (/mnt/c/Users/kwilliams/Documents/Projects/cheevobot/node_modules/steam-user/components/utility.js:29:12)
    at SteamUser._handlers.(anonymous function) (/mnt/c/Users/kwilliams/Documents/Projects/cheevobot/node_modules/steam-user/components/chat.js:282:9)
```

Code:

```js
'use strict'

// -- Dependencies -------------------------------------------------------------

// Node.js API
const path = require('path')

// Node packaged modules
const confit = require('confit')
const SteamUser = require('steam-user')

// Local modules
const dice = require('./plugins/cmds/dice')
const uptime = require('./plugins/cmds/uptime')

// -- Public Interface ---------------------------------------------------------

class SteamBot {
  constructor () {
    this._client = new SteamUser()
    this._client.on('friendOrChatMessage', (sender, msg, chat) => {
      this.exec(msg, { sender, chat })
    })
  }

  start (cb) {
    confit(path.resolve('config/')).create((err, config) => {
      if (err) return cb(err)

      // Build command parser
      this._parser = yargs()
      this._parser.command({
        'command': '!ch <cmd>', // config.get('cmd'),
        'description': 'test desc', // config.get('description'),
        'builder': yargs => {
          yargs.command(dice(this))
          yargs.command(uptime(this))
          yargs.help()
          yargs.version()
        }
      })

      this._client.logOn(config.get('login'))
      this._client.on('loggedOn', response => {
        this._client.setPersona(SteamUser.EPersonaState.Online)
        cb()
      })
    })
  }

  exec (cmd, context) {
    this._parser.parse(cmd, context, (err, argv, output) => {
      if (err && !output) throw err
      if (output) {
        this._client.chatMessage(context.chat, output)
      }
    })
  }
}

// -- Exports ------------------------------------------------------------------

module.exports = SteamBot

```

Steps to produce:

```
Scriptdaemon: !ch help // First call
CheevoBot:  !ch <cmd> // Description doesn't appear?

Commands:
  dice [roll]  Simulate a dice roll
  uptime       Retrieve bot uptime

Options:
Scriptdaemon: !ch help // Second call
CheevoBot is now Offline. // Crash
```

bcoe:
there was a bug that caused an exception to be raised when `parse()` was called multiple times on the same yargs instance if `help()` is enabled.

fixes #776

### DIFF
--- a/test/command.js
+++ b/test/command.js
@@ -747,6 +747,21 @@ describe('Command', function () {
     counter.should.equal(2)
   })
 
+  // addresses: https://github.com/yargs/yargs/issues/776
+  it('allows command handler to be invoked repeatedly when help is enabled', function (done) {
+    var counter = 0
+    var y = yargs([])
+      .command('foo', 'the foo command', {}, function (argv) {
+        counter++
+      })
+      .help()
+    y.parse(['foo'], function () {})
+    y.parse(['foo'], function () {
+      counter.should.equal(2)
+      return done()
+    })
+  })
+
   // addresses https://github.com/yargs/yargs/issues/522
   it('does not require builder function to return', function () {
     var argv = yargs('yo')
--- a/yargs.js
+++ b/yargs.js
@@ -957,7 +957,7 @@ function Yargs (processArgs, cwd, parentRequire) {
         // consider any multi-char helpOpt alias as a valid help command
         // unless all helpOpt aliases are single-char
         // note that parsed.aliases is a normalized bidirectional map :)
-        var helpCmds = [helpOpt].concat(aliases[helpOpt])
+        var helpCmds = [helpOpt].concat(aliases[helpOpt] || [])
         var multiCharHelpCmds = helpCmds.filter(function (k) {
           return k.length > 1
         })


test/command.js
```diff
@@ -747,6 +747,21 @@ describe('Command', function () {
...
 
+  // addresses: https://github.com/yargs/yargs/issues/776
+  it('allows command handler to be invoked repeatedly when help is enabled', function (done) {
+    var counter = 0
+    var y = yargs([])
+      .command('foo', 'the foo command', {}, function (argv) {
+        counter++
+      })
+      .help()
+    y.parse(['foo'], function () {})
```
JaKXz: related but tangential question: should the fn passed to parse be optional? just curious.

test/command.js
```diff
@@ -747,6 +747,21 @@ describe('Command', function () {
     counter.should.equal(2)
   })
 
+  // addresses: https://github.com/yargs/yargs/issues/776
+  it('allows command handler to be invoked repeatedly when help is enabled', function (done) {
+    var counter = 0
+    var y = yargs([])
+      .command('foo', 'the foo command', {}, function (argv) {
```
JaKXz: I don't know the test code well enough, but, my preference would be a method that is spied on and then asserted that it's been called twice, possibly outside the second `parse` call.

What do you think?

test/command.js
scriptdaemon: It is optional, but specifying something prevents yargs from outputting to stdin (my possibly limited understanding of how it works).

bcoe: @JaKXz, @scriptdaemon is correct; if you provide a function, we assume that you want to hook into output -- rather than the output of parse being printed to stdout.

This is useful for chat-bots, which is @scriptdaemon use-case.

test/command.js
bcoe: @JaKXz I like this idea ... but I also have yet to pull sinon into the codebase. I think pulling in a mocking framework is beyond the scope of this pull, but I do like the idea. a couple questions come to mind:

* is there something lighter-weight than sinon that's popular?
* I think we should probably pull in a stub/mock framework in the context of a larger test cleanup.

JaKXz: @bcoe I'm a big fan of https://github.com/testdouble/testdouble.js/ - and yeah I think it'd do nicely in a larger cleanup.

bcoe: @JaKXz awesome, perhaps we could eyeball the test-suite sometime soon, and see how many places we're using ad-hock spies -- if we're doing it a bunch of times, I say let's try `testdouble.js` out.

PR #797, opened 2017-02-25
fix: extending config files breaks on node 0.10

stevezhu:
issue from #779 

path.isAbsolute() not supported in node 0.10
use path.resolve() instead as mentioned in discussion in issue


phated:
❤️  ❤️  @stevezhu and @bcoe!!!!

PR #808, opened 2017-02-27
docs: env vars now take precedence over config file/objects

nexdrew:
Just update the README/docs to reflect the new order.

See yargs/yargs-parser#81 and #787


bcoe:
@nexdrew 🎸

PR #810, opened 2017-03-02
fix: --help with default command should print top-level help

### LINKED ISSUE #803: it's weird that running `--help` on default command does not show help for all commands

nexdrew:
Fixes #803, see discussion there.

First wrote the failing test, then included the fix.


nexdrew:
@bcoe Haven't done a _whole_ lot of 7.x testing, but what I have done looks good so far.

Issue #824, opened 2017-03-13
Number commands don't work


sputnik13:
Commands defined as numbers do not resolve to the appropriate command internally due to casting of numerical arguments to Numbers.

Test code below

```
// test.js
var argv = require('yargs')
    .command('1', "This won't work", {}, (argv) => {
        console.log('Got command 1');
    })
    .command('one', "This will work", {}, (argv) => {
        console.log('Got command one');
    })
    .demandCommand()
    .help()
    .wrap(72)
    .argv

console.log(argv)
```

Running with text command
```
$ node yargs_test.js one
Got command one
{ _: [ 'one' ], help: false, '$0': 'yargs_test.js' }
```

Running with numeric command
```
$ node yargs_test.js 1
{ _: [ 1 ], help: false, '$0': 'yargs_test.js' }
```

Issue #823, opened 2017-03-11
strict mode does not take into account camel-case for positional arguments


gajus:
After upgrading to `yargs@7.0.2` and `node@v7.7.1` I started getting:

> Not enough non-option arguments: got 0, need at least 1

Command:

```bash
babel-node src/bin/index.js upload-data ./.data.json
```

Code:

Entry program:

```js
#!/usr/bin/env node

import yargs from 'yargs';

// eslint-disable-next-line no-unused-expressions
yargs
  .env('ASDMP')
  .strict()
  .commandDir('commands')
  .help()
  .wrap(80)
  .argv;

```

Command builder:

```js
export const command = 'upload-data <data-file>';
export const desc = 'Updates or inserts information to the database.';

export const builder = (yargs: Object): void => {
  yargs
    .env('ASDMP')
    .strict()
    .options({
      ...commandLineConnectionOptionsShape
    })
    .demandCommand(1, 1);
};

```

gajus:
It appears this is `babel-node` independent (which was my first thought).

```bash
node dist/bin/index.js upload-data ./.data.json

```

gives the same error.

gajus:
The issue goes deeper. Removing `demandCommand` doesn't fix it. The `argv.dataFile` value is not being captured.

gajus:
It is starting to look like this is a dependency of a dependency thats breaking the behaviour. I have just tried downgrading to `v7.0.1` and it doesn't fix the behaviour.

gajus:
Downgrading to v6 makes it work.

bcoe:
@gajus I see the issue, it is a bit of an edge-case (which we should fix):

your initial problem was `.demandCommand`; this is not necessary because you're using non-optional positional arguments `<>`.

The underlying problem now biting you, is that strict does not seem to be taking into account the came-case form of positional arguments:

`exports.command = 'upload-data <datafile>';`

The above works as expected.

gajus:
Understood. Thank you for clarifying.

gajus:
Unless I am doing something wrong, this doesn't work either:

```
babel-node src/bin/index.js upload-data .movies.json
src/bin/index.js upload-data <datafile>

Options:
  --help               Show help                                       [boolean]
  --database-database                                        [string] [required]
  --database-host                                            [string] [required]
  --database-user                                            [string] [required]

Not enough non-option arguments: got 0, need at least 1
```

```js
export const command = 'upload-data <datafile>';
[...]

PR #825, opened 2017-03-13
Number commands don't work

sputnik13:
casting argv._ elements back to string when evaluating for commands to
allow non-string commands to be correctly found and matched

Resolves https://github.com/yargs/yargs/issues/824

### DIFF
--- a/test/command.js
+++ b/test/command.js
@@ -1317,4 +1317,14 @@ describe('Command', function () {
     called.should.equal(false)
     r.errors.should.match(/Missing required argument/)
   })
+
+  it('should support numeric commands', function () {
+    var output = []
+    yargs('1')
+      .command('1', 'numeric command', function (yargs) {
+        output.push('1')
+      })
+      .argv
+    output.should.include('1')
+  })
 })
--- a/yargs.js
+++ b/yargs.js
@@ -983,7 +983,8 @@ function Yargs (processArgs, cwd, parentRequire) {
         var handlerKeys = command.getCommands()
         if (handlerKeys.length) {
           var firstUnknownCommand
-          for (var i = 0, cmd; (cmd = argv._[i]) !== undefined; i++) {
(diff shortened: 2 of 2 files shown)


sputnik13:
why is an increase in coverage considered a fail...  that's puzzling to me

bcoe:
@sputnik13 you're good; coveralls.io switched to looking at branch coverage rather than line coverage which has broken a few things for us.

@nickmerwin I think you'll find quite a few projects bump into this confusion -- do you have a blog post I could reference and tweet, etc.

nickmerwin:
@bcoe we've added a toggle to the repo settings if you'd prefer to disable this:

image

Blog post is coming shortly.

nickmerwin:
@bcoe we've decided to make this opt-in for existing repos so as to minimize confusion; branches will still be analyzed, but not factored into aggregate % unless the setting is checked.

bcoe:
-

sputnik13:
@bcoe any chance we can get reviews on this?  I'm not familiar with your team's processes so I'm just wondering whether nobody's looking at it because of the test failure.

bcoe:
@nickmerwin probably a good call 👍 appreciate the hard work.

@sputnik13 will review soon, appreciate the pull request; with regards to process, there are a few folks who contribute with code reviews, but unfortunately most of us are doing this evenings and weekends -- so the turnaround can take a few days.

With regards to this issue, there's a related ticket that I'd like to fix:

https://github.com/yargs/yargs/issues/823

So I was planning on looking a bit closer at the command parsing logic, to see if we could solve a few categories of bugs at once 😄

sputnik13:
@bcoe understood, and I appreciate the quick responses, I understand it can be a challenge to work on projects on your own time.  Please let me know if there's anything I can do to help.

bcoe:
@sputnik13 haven't forgotten about this, was on a tangent working on conventional-changelog/lerna -- promise to get to this soon.

Issue #827, opened 2017-03-16
demandOption terminates program even if it set to false


bricss:
```demandOption``` set on option property with **falsy** value, throw up error message and terminates program.

```javascript
require('yargs')
  .options({
    'run': {
      alias: 'r',
      describe: 'run your program',
      demandOption: false
    },
    'path': {
      alias: 'p',
      describe: 'provide a path to file',
      demandOption: false
    },
    'spec': {
      alias: 's',
      describe: 'program specifications'
    }
  })
  .help()
  .argv;
```

```shell
Options:
  --run, -r   run your program                                        [required]
  --path, -p  provide a path to file                                  [required]
  --spec, -s  program specifications
  --help      Show help                                                [boolean]

Missing required arguments: run, path
```

bcoe:
@bricss sounds like a bug for sure; I'm probably not going to get too much OSS time until Friday or Sunday of this week -- if you get a chance to take a look, would greatly appreciate a patch.

bricss:
@bcoe working on PR, will keep up to date.

PR #829, opened 2017-03-17
fix(*): fix demandOption ignores false value

bricss:
Bugfix -> https://github.com/yargs/yargs/issues/827


bcoe:
@bricss awesome! thanks for the contribution.

PR #834, opened 2017-03-21
test: Made unit test passing with node-chakracore

kunalspathak:
There was 1 unit test that were failing with node-chakracore because of error message difference between v8 and chakracore. Added logic to generate error on fly and use that to compare the error message.

Ref: nodejs/node-chakracore#189

### DIFF
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -1823,8 +1823,14 @@ describe('yargs dsl tests', function () {
     it('allows an error to be handled by fail() handler', function () {
       var msg
       var err
+      var jsonErrMessage
       yargs('--json invalid')
         .coerce('json', function (arg) {
+          try {
+            JSON.parse(arg)
+          } catch (err) {
+            jsonErrMessage = err.message
+          }
           return JSON.parse(arg)
         })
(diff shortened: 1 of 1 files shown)


bcoe:
@kunalspathak awesome thanks for doing this 👍 

Let me know if I can do anything else to help.

kunalspathak:
Thanks @bcoe  for quick review and merge. I wanted to know if you have some kind of CI setup and if yes, should we add `node-chakracore` validation in it. This will ensure that in future and test changes passes `node-chakracore` as well.

bcoe:
@kunalspathak we use TravisCI and Appveyor currently, can you setup node-chakracore on appveyor perhaps?

kunalspathak:
We could, something similar to what is done at https://github.com/sass/node-sass/pull/1777 . Let me know if you want to do this or i can do when i get chance.

bcoe:
@kunalspathak I would love a patch along these lines; we'll definitely land it faster if you have a chance to take a look -- I doubt I'm as proficient with appveyor and node-chakracore as you are.

PR #840, opened 2017-03-27
Allow command to override global strict

zodern:
This adds a second argument to `strict` to disable it for a command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1809,7 +1809,7 @@ Specify --help for available options
 Specifies either a single option key (string), or an array of options.
 If any of the options is present, yargs validation is skipped.
 
-.strict([enabled=true], [global=true])
+.strict([enabled=true])
 ---------
 
 Any command-line argument given that is not demanded, or does not have a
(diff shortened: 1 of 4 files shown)


bcoe:
@zodern thanks for the patch; I'm not seeing immediately why this additional configuration option is required. Could you provide some information about your specific use-case.

zodern:
I want strict mode enabled for all commands except 2. Instead of enabling it individually for each command, this allows enabling it globally and disabling for individual commands.

Something similar was part of #734, but it was removed in #766.

bcoe:
@zodern gotcha, this makes sense to me now; I'd forgotten about the `global` option being the first parameter for strict.

I'd rather that we get rid of the `global=true` option; I don't think this is particularly useful now that options default to `global`; then we can make the first argument equivalent to `enable`/`disable` which feels more elegant.

This will be a breaking change; but this is fine.

zodern:
@bcoe I removed the global option.

bcoe:
@zodern thank you 👍 will land this in an `8.x` release; I'd like to potentially do a bit more house keeping on the `7.x` branch first -- so might take a little while to merge.

bcoe:
@zodern this change is going out in the next version of yargs. I would love your help testing:

`npm i yargs@next --save` to try it out.

PR #844, opened 2017-03-30
Enable dynamic keys for config extension

### LINKED ISSUE #843: Extends does not work when intended key is different in nested configurations
This is an issue that was discovered in https://github.com/istanbuljs/nyc/pull/541. 

The `extends` functionality does not work when extending continuous files that are meant to be extended from different keys.

#### This is an example from nyc:
Continuously extending nyc configurations from either `.nycrc` or `package.json "nyc"` works. But mixing `extends` between`.nycrc` files and `package.json "nyc"` does not. This is because configuring nyc through package.json, its `extends` will look for the `nyc` key in the extended file. When configuring nyc through .nycrc, it will look at the extended file as a whole.

#### Idea for fix
I don't think we can rely on the `subKey` parameter in [apply-extends](https://github.com/yargs/yargs/blob/master/lib/apply-extends.js#L18). That enforces the same key through all extended configuration files.

Instead, I think we need to introduce a new reserved keyword like `extendsKey` to pair with `extends`. Remove `subKey` from apply-extends, and rather read `extendsKey` from each extended configuration. That allows for dynamic extension of continuously nested configurations.

rcoy-v:
This is an option to resolve #843. Would involve using another reserved keyword, `extendsKey`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1008,7 +1008,41 @@ $ node test.js
   '$0': 'test.js' }
 ```
 
-Note that a configuration object may extend from a JSON file using the `"extends"` property. When doing so, the `"extends"` value should be a path (relative or absolute) to the extended JSON file.
+A configuration object may extend from a JSON file using the `"extends"` property. 
+When doing so, the `"extends"` value should be a path (relative or absolute) to the cwd.
+`"extendsKey"` can be used to further specify a single property to extend from.
+
+```js
+// config.json
+var argv = require('yargs')
+  .config({
+    foo: 1,
+    extends: "./node_modules/lib/config.json",
+    extendsKey: "read"
+  })
+  .argv
+```
+
+```json
+// node_modules/lib/config.json
+{
+  "ignored": {
+    "notUsed": "zed"
+  },
+  "read": {
+    "foo": 2,
+    "bar": 10
+  }
+}
+```
+
+Resulting argv configuration:
+```json
+{
+  "foo": 1,
+  "bar": 10
+}
+```
 
 <a name="conflicts"></a>.conflicts(x, y)
 ----------------------------------------------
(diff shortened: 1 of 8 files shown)


bcoe:
@rcoy-v I wonder if we could hold off on adding an additional configuration option, and figure out default behavior that's intuitive.

It feels like if you have `pkgConf` configured to look at the `nyc` key, you would almost always want the object under this key to be merged with the content of a file such as `.nycrc` that has been configured.

rcoy-v:
I understand not wanting to add more breaking changes with reserved keys. Not the most elegant solution. 

Here is my example of something I would like to do with nyc, but cannot currently:

Now that nyc has basic extending functionality, I have a package that contains a default nyc config intended to be used across a large number of projects. Most of these projects will use the default with no override, so their `.nycrc` is a one-liner.
```json
{"extends": "./node_modules/lib/.nycrc"
```
These projects are already drowning in root level dot-files. It would be nice to reduce the noise there, and move the one-line extends for nyc to package.json.

I can't do that however, because `pkgConf` enforces the `nyc` key, so it doesn't understand the default `nycrc` provided.

Maybe another option for yargs is only use the provided key to `pkgConf` when reading `package.json` files. If an extended file is not a `package.json`, then ignore any given key and read the entire file.

bcoe:
@rcoy-v sorry for dropping this on the floor for a bit; a busy couple months with work, and have neglected open-source.

Here's how I wish things would work, I think it would solve your issue?

`Package A` is a module you publish with your default `.nycrc` configuration.

`Package B` is a module that you wish to have use this default behavior:

* `Package B`: can have a `.nycrc` with `"extends": "./node_modules/lib/.nycrc"`
* `Package B`: can also have the following JSON:

```json
{
  "name": "package-b",
  "nyc": {
    "extends": "./node_modules/lib/.nycrc"
  }
}
```

This would result in just the keys under `nyc` being extended with the parent `.nycrc`, and being passed to `nyc` as configuration.

Would this behavior do the trick?

bcoe:
@rcoy-v bump; would love to land this just want to discuss the specifics of the feature a bit.

I was actually thinking it would be really useful for nyc to publish a few `presets`; e.g., a config that works well with babel.

bcoe:
@rcoy-v I'm closing this in favor of:

https://github.com/yargs/yargs/pull/865

I think what I'd prefer to see, rather than having special logic around the key used for configuration in a package.json, is for folks to start publishing presets for nyc -- e.g., `@istanbul/nyc-babel-config`.

rcoy-v:
Sorry @bcoe I think we missed each other on the timing of this. I'm fine with the approach you went. Thank you for the effort put into this feature!

PR #847, opened 2017-04-01
fix: we shouldn't output help if we've printed a prior help-like message

### LINKED ISSUE #822: recommended command used in conjunction with --help prints twice
```slackboard user <subcommand|help>

Commands:
  update-email <name> <email>           change email to new address for a user
  update-resource <name> <key> <value>  change resource attribute for user
  get <name>                            get user

Options:
  --help  Show help                                                                        [boolean]

Did you mean get?
/usr/bin/nodejs bin/slackboard user <subcommand|help>

Commands:
  update-email <name> <email>           change email to new address for a user
  update-resource <name> <key> <value>  change resource attribute for user
  get <name>                            get user

Options:
  --help  Show help                                                                        [boolean]```

### PR DESCRIPTION
**Pull request by bcoe**
we were accidentally printing the help text twice, if recommendations were made and `exit = false`.

```bash
Commands:
  goat

Options:
  --help  Show help                                                                        [boolean]

Did you mean goat?
Commands:
  goat

Options:
  --help  Show help                                                                        [boolean]
```

Let's not do this, it looks weird!

fixes #822

### DIFF
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -346,8 +346,9 @@ describe('yargs dsl tests', function () {
         .help()
         .recommendCommands()
 
-      parser.parse('boat help', {}, function (_err, argv, output) {
+      parser.parse('boat help', {}, function (err, _argv, output) {
         // it should not have printed the help text twice!
+        err.message.should.equal('Did you mean goat?')
         output.split('Commands:').length.should.equal(2)
         return done()
       })
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -339,6 +339,20 @@ describe('yargs dsl tests', function () {
       r.errors[1].should.match(/Did you mean goat/)
     })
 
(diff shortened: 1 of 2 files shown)


test/yargs.js
```diff
@@ -339,6 +339,20 @@ describe('yargs dsl tests', function () {
...
     })
 
+    // see: https://github.com/yargs/yargs/issues/822
+    it('does not print help message if recommendation has been made', function (done) {
+      const parser = yargs()
+        .command('goat')
+        .help()
+        .recommendCommands()
+
+      parser.parse('boat help', {}, function (_err, argv, output) {
```
JaKXz: It might be redundant, but shouldn't we check that error is null/falsy here? and shouldn't it be `_argv` since we're not using that arg?

bcoe: @JaKXz 👍 done.

PR #848, opened 2017-04-01
chore: add editorconfig

JaKXz:
Easy wins++

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+root = true
+
+[*]
+
+indent_style = space
+indent_size = 2
+
+end_of_line = lf
+charset = utf-8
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,16 @@
+root = true
+
+[*]
+
+# Change these settings to your own preference
+indent_style = space
+indent_size = 2
+
+# We recommend you to keep these unchanged
```
maxrimue: Are these comments intentional? They look like they're from some example usage, they might be confusing to some people, making them think we don't enforce `space` and `2` (though `standard` does)

JaKXz: Good point, I'll remove them.

PR #901, opened 2017-06-19
Fix #849 - make choices correctly observe demandOption setting

### LINKED ISSUE #849: Choices Invalid Value undefined
In version 7 choices started an error if one of the choices was not selected. This is fine but i did have situations in which i didn't want a default and adding undefined to the list of choices didnt make sense either. None of the other argument types are required why is choices.

Thanks.

phanning:
Addresses #849.

Handles `options` map containing `choices` and a combination of `demandOption` and/or `describe`|`description`|`desc`. If `demandOption` is set to `false`, any `choices` values specified will not be required (`undefined` is okay).

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -172,7 +172,8 @@ module.exports = function (yargs, usage, y18n) {
         options.choices.hasOwnProperty(key)) {
         [].concat(argv[key]).forEach(function (value) {
           // TODO case-insensitive configurability
-          if (options.choices[key].indexOf(value) === -1) {
+          if (options.choices[key].indexOf(value) === -1 &&
+              value !== undefined) {
             invalid[key] = (invalid[key] || []).concat(value)
           }
         })
--- a/test/validation.js
+++ b/test/validation.js
@@ -584,6 +584,90 @@ describe('validation tests', function () {
         })
         .argv
     })
+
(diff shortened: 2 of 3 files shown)


brianvoe:
:+1:

bcoe:
@phanning thanks for this pull request 👍 I really appreciate the thorough unit tests.

One tweak I would make however, I think rather than adding the logic to `yargs/yargs.js` we should apply your updated logic during the validation itself [here](https://github.com/yargs/yargs/blob/master/lib/validation.js#L164); I think this will simplify the logic somewhat, and will also ensure that your tests work both when `choices` are passed via the object shorthand to `.option`, and when the `.choices` method is called.

phanning:
@bcoe Good point. I moved the fix to validation.js. I was wondering if it would need something like `&& ((value !== undefined) || (options.demandedOptions.hasOwnProperty(key) && options.demandedOptions(key)))` instead of just `&& value !== undefined`. All the tests are passing with the simpler fix, however.

bcoe:
@phanning thanks for the great contribution, I just made a couple minor tweaks to the tests here:

https://github.com/yargs/yargs/pull/908

(just changed the description slightly, and moved them into the `choices` describe block).

PR #902, opened 2017-06-22
feat: use process.argv0 where possible

keithamus:
This commit adds support for using `process.argv0` where it is available (node 6.4+) as it is much more accurate in getting the command line that the user actually typed, rather than the binary name of the script being called (`process.argv[0]`).

The tests are failing right now because I haven't changed them - I wanted to float this feature to everyone before putting time in to change the tests. Please let me know your thoughts!

Hat-tip to @zkat for doing all the hard research 👍 ✨

### DIFF
--- a/yargs.js
+++ b/yargs.js
@@ -32,21 +32,25 @@ function Yargs (processArgs, cwd, parentRequire) {
 
   if (!cwd) cwd = process.cwd()
 
-  self.$0 = process.argv
-    .slice(0, 2)
-    .map(function (x, i) {
-      // ignore the node bin, specify this in your
-      // bin file with #!/usr/bin/env node
-      if (i === 0 && /\b(node|iojs)(\.exe)?$/.test(x)) return
-      var b = rebase(cwd, x)
-      return x.match(/^(\/|([a-zA-Z]:)?\\)/) && b.length < x.length ? b : x
-    })
-    .join(' ').trim()
+  if (process.argv0) {
+    self.$0 = process.argv0
+  } else {
+    self.$0 = process.argv
+      .slice(0, 2)
+      .map(function (x, i) {
+        // ignore the node bin, specify this in your
+        // bin file with #!/usr/bin/env node
+        if (i === 0 && /\b(node|iojs)(\.exe)?$/.test(x)) return
+        var b = rebase(cwd, x)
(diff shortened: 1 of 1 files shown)


addaleax:
Thank you, this is exactly what we introduced this for in Node! The code changes look good to me.

nexdrew:
Thanks for bringing this up.

I'm not too familiar with `process.argv0`, but for a typical "bin" package that is installed globally, it appears to point to the `node` executable (similar to `process.argv[0]`) instead of the name of the "bin" package (which can typically be determined from `process.argv[1]`). I created a simple example that you can install and run yourself:

```console
$ npm i -g argv0
/usr/local/bin/argv0 -> /usr/local/lib/node_modules/argv0/cli.js
+ argv0@0.0.1
added 1 package in 0.997s
```

```console
$ argv0
{ argv0: 'node',
  argv1basename: 'argv0',
  'yargs$0': '/usr/local/bin/argv0' }
```

You can check the code from this example here: https://github.com/nexdrew/argv0

It seems there's a use-case that's not familiar to me. Is there another way for folks to install and run global "bin" packages? What's the intended use of `process.argv0` for library and CLI/app authors?

bcoe:
@nexdrew thanks for digging into this;

@addaleax + @keithamus any thoughts on @nexdrew's digging, is there anything along the lines of `process.argv1`? should there be 😄

bcoe:
@keithamus tldr; if we can lean on Node rather than applying the logic ourselves, this is awesome (and I appreciate the work).

keithamus:
`argv0` - as is my understanding - is supposed to provide what the user typed in their shell to invoke the command. As the [documentation suggests](https://nodejs.org/api/process.html#process_process_argv0) running `$ exec -a foo node -p 'process.argv0'` produces `foo` as an output. I imagine what is happening with npm's `bin` scripts - is that the shebang line (`#!/usr/bin/env node`) is overtaking that, and so `argv0` becomes `node`.

zkat:
I misunderstood. argv0 will always be the node executable, regardless of how you invoke things

keithamus:
I think we've reached the logical conclusion here. Looks like this is a no-go, I'll close this. Thanks for the feedback everyone!

PR #905, opened 2017-06-25
Fix #899

### LINKED ISSUE #899: Using .conflicts and .option
I'm porting a big project over from using `commander` and I am absolutely loving yargs but encountered this issue.

with yargs 8.0.2: 
```

const yargs = require("yargs");  
yargs('command --foo')
          .command('command', 'a command', {}, function (argv) {})
	  .option('foo', {describe:'A foo foo'})
          .option('bar',{describe: 'A bar bar'})
          .fail(function (msg) {
            console.log(msg);
         
          })
          .conflicts('foo', 'bar')
.argv
```

Results in "Arguments foo and bar are mutually exclusive" despite the fact that "bar" is not specified on the command.   From the doc: "This method can be used to make yargs aware of options that could exist." (i.e. it does not imply that you are setting those options to be true).  If I comment out `.option('bar'...` it works as expected.

I was trying to debug this condition in lib\validation.js but  couldn't find a way to make it work 
```
 if (conflicting[arg] && args.indexOf(conflicting[arg]) !== -1) {
        usage.fail(__('Arguments %s and %s are mutually exclusive', arg, conflicting[arg]))
}
```

 Any insight?

dogboydog:
See description of the problem in #899 

I had to put in some code to ensure that the conflicting keys are actually specified before failing the command. I'm guessing there is some difference in how you build the argv object  depending on whether the `.command()` syntax was used which was causing this discrepancy

I added two new tests for this scenario. Hopefully I didn't break anything. Let me know your thoughts

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -310,9 +310,8 @@ module.exports = function (yargs, usage, y18n) {
     var args = Object.getOwnPropertyNames(argv)
 
     args.forEach(function (arg) {
-      if (conflicting[arg] && args.indexOf(conflicting[arg]) !== -1 && // if there is a conflict defined for this key
-          argv[arg] && argv[conflicting[arg]]) // and both conflicting keys have been specified
-          {
+    // if there is a conflict defined for this key and both conflicting keys have been specified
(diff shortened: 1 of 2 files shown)


bcoe:
@dogboydog have been on a bit of a hiatus, but really appreciate the pull request; starting to dig into it.

PR #907, opened 2017-07-04
fix: defaulting keys to 'undefined' interfered with conflicting key logic

### LINKED ISSUE #899: Using .conflicts and .option

bcoe:
made a couple slight tweaks to @dogboydog's work in https://github.com/yargs/yargs/pull/905

👋 @dogboydog thanks for the great work in #905, I just made a couple small tweaks to the tests and made it so we check for `undefined` rather than any `falsy` value, this way a conflicting key can be set to `false` and it will still trigger the conflict check.

fixes #899

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -310,9 +310,8 @@ module.exports = function (yargs, usage, y18n) {
     var args = Object.getOwnPropertyNames(argv)
 
     args.forEach(function (arg) {
-      if (conflicting[arg] && args.indexOf(conflicting[arg]) !== -1 && // if there is a conflict defined for this key
-          argv[arg] && argv[conflicting[arg]]) // and both conflicting keys have been specified
-          {
+    // if there is a conflict defined for this key and both conflicting keys have been specified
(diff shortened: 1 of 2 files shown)


dogboydog:
Thanks!  It will be nice to have this fixed.

Issue #894, opened 2017-06-11
argv alternative


borela:
Is there a method to call instead of the `argv` getter? In cases where the cli always requires a command(and each command has its handler already set), the `argv` called in the entry point will not be used, making the linter complain of unused expression.

A solution would be having a function `run()` or something like that at the end of the Yargs call chain, is there already such function?

Thanks

miraage:
Running main program implicitly by invoking a getter is a side effect.
This is what all developers across the world should avoid.
Please, add an explicit method to run program, e.g. `run()` or `parse()`.

LinusU:
This should probably be closed now, in Yargs 9.0.0 you can call `.parse()` to get the same behaviour as `.argv`, but with a function call instead of a getter 🎉 👏

mdevils:
I was very confused when I was reading documentation with all those lines ending with `.argv` and it was never used. I guessed that I need to do something with the `argv` field value. Took me some time to figure out that side-effect happens when you access `argv` property.

Maybe documentation should be updated with `.parse()` instead of `.argv`?

PR #915, opened 2017-07-12
add explicit run method

bernardmcmanus:
Addresses https://github.com/yargs/yargs/issues/894 (argv alternative).


bcoe:
👍 I like this.

bcoe:
@bernardmcmanus part of me wonders if we should call this method `.then()` rather than `.run()`; then `.then()` could return a promise that resolves with the parsed arguments (as well as kicking off the parser)?

bcoe:
CC: @miraage, @borela? (what do you think of the above `.then()` idea?)

borela:
I would love to have both options `run` and `then`.

bcoe:
@borela @bernardmcmanus for v1, my preference would be to add `.then()` following this method signature:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then

This would both solve linting issues and address a common feature request (that of yargs providing a more promise-oriented interface).

borela:
@bcoe I thought that using `then` would require me to always pass a function, I forgot to read the notes in the documentation. I would be fine with it always returning a promise.

bernardmcmanus:
@bcoe @borela just to make sure I understand correctly, you're thinking something like:
```js
self.then = function (...args) {
  return Promise
    .resolve(self._parseArgs(processArgs))
    .then(...args);
};
```

IMHO this will have some unintended consequences:
```js
require('yargs')
  // ...
  .then((argv) => {
    // now that I've entered a promise chain my process always exits with status 0,
    // even if I throw an unhandled error, unless I explicitly call process.exit(...)
  });
```

I just don't see a reason for `yargs` to support this pattern considering:
  1. As far as I can tell `_parseArgs` never needs to run asynchronously; and
  2. If someone _really_ wants this behavior, it's easily achieved with
  ```js
  Promise.resolve(
    yargs
      // ...
      .run()
  );
  ```

That said, you obviously know this project much better than I so I will defer to your judgement 😄.

bcoe:
@bernardmcmanus I hear where you're coming from;

I'd like to move towards eventually having a method like `.then()`, which would start moving us towards an API surface that could better support async parsing functionality in yargs ... I'm picturing eventually we support promises for commands, validation, completion, etc. (we do support async functions in some of these places, but it's inconsistent).

My thinking is, if we do eventually have something like `yargs.then((result) => {})` (which would also return a Promise which resolves to `result`, I don't really like the idea of us having `.argv`, `.run()`, and `.then()` all in the API, it starts to get confusing as to which you'd call to parse your darn command line arguments!

I was thinking, maybe we could start moving towards `.then()`, which is an API method I know I'd like to offer, and at the same time would provide a way to step around the linting issues that `yargs.argv` presents.

bernardmcmanus:
@bcoe that's totally fair - thank you for the explanation. I will push an update shortly.

bernardmcmanus:
@bcoe updated

bcoe:
@bernardmcmanus very much appreciate this pull 👍 will work on giving feedback as soon as possible; hopefully we'll land later in the week, or early in the weekend.

bernardmcmanus:
take your time, I'll keep an eye out

bcoe:
@bernardmcmanus @borela I've obviously dropped this on the floor. In an attempt to keep up with OSS obligations, I recently created a community slack channel for yargs (select yargs when joining):

Would love to have you join, and we can see about finally landing this feature.

docs/api.md
```diff
@@ -20,7 +20,27 @@ or use `.parse()` to do the same thing:
...
+```javascript
+require('yargs')
+  .command('thing', 'it does stuff')
+  .then()
+```
+
+In fact, simply resolving a promise with a yargs instance will run the parser:
+
+```javascript
+Promise.resolve(require('yargs')).then(argv => ...)
```
bcoe: must say, I really like this addition to the API; somewhat similar to `.parse([], () => {})` so we might want to consolidate the API a bit more eventually, but I'm happy with landing this for the time being.

MattSturgeon:
@bcoe did you mean to merge this to master? Looks like you merged master into this instead!

yargs.js
```diff
@@ -912,6 +912,11 @@ function Yargs (processArgs, cwd, parentRequire) {
     enumerable: true
   })
 
+  self.then = function (onFulfilled, onRejected) {
+    return new Promise(resolve => resolve(self.argv))
```
bcoe: a couple suggestions to make this feature really rock:

1. I think we should run the `self.argv` in such a way that we catch the error (I would look at the approach that we use with `.parse()`) so that we can reject or resolve the promise depending on whether an error has occurred.
2. I think we should make it so (as with `.parse()`) we suppress output if an `onFulfilled` handler is provided.

having written these two points ... I definitely do think that I'd like to deprecate the `.parse(function())` behavior with this behavior -- I like this much better 👍

bcoe:
closing in favor of #940

PR #922, opened 2017-07-22
Allow implies and conflicts to accept array values

### LINKED ISSUE #845: imply multiple dependencies
```
   var argv = require('yargs')
        .usage('Usage: $0 [options]')
        .option('ssl', {
            default: false,
            description: 'Use SSL?'
            // implies: ['ssl_cert', 'ssl_key'] // this doesn't work... unexpected array
        })
        .implies('ssl', 'ssl_cert')
        .implies('ssl', 'ssl_key') // only this one gets implied
        .option('ssl_key', {default:null, description:'filepath to ssl key'})
        .option('ssl_cert', {default:null, description:'filepath to ssl certificate'})
        .help()
        .alias('h', 'help')
        .argv;
```

`command --ssl`
outputs
> Implications failed:
  ssl -> ssl_key

(no mention of ssl_cert also being implied)

`command --ssl --ssl_key="/foo"`
will not catch that ssl_cert was not passed

### LINKED ISSUE #921: implies and conflicts should accept array values
Closely related to #845 - verging on duplicating it in fact!

[`validation.implies(key, value)`][implies] and [`validation.conflicts(key, value)`][conflicts] should handle `value` being an array of `value`s. Roughly like so:

```js
function (key, value) {
    Object.keys(key).forEach((k) => {
      self.implies(k, key[k])
    })
  } else {
    if (!implied[key]) {
      implied[key] = []
    }
    if (Array.isArray(value)) {
      implied[key].concat(value)
    } else {
      implied[key].push(value)
    }
  }
}
```

This would likely mean modifying anything that reads from the `implied` or `conflicting` objects to handle the new structure, of course.

[implies]: https://github.com/yargs/yargs/blob/086d662f6a904b375c557521f18d12b092ac82f8/lib/validation.js#L227-L238
[conflicts]: https://github.com/yargs/yargs/blob/086d662f6a904b375c557521f18d12b092ac82f8/lib/validation.js#L296-L305

MattSturgeon:
Fixes #921 and fixes #845 

`npm test` passes, but I haven't tried to run yargs with the changes, nor added any new tests (sorry).

I've also not updated the docs.

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -1,4 +1,5 @@
 'use strict'
+const YError = require('./yerror')
 const objFilter = require('./obj-filter')
 const specialKeys = ['$0', '--', '_']
 
@@ -234,7 +235,17 @@ module.exports = function validation (yargs, usage, y18n) {
       })
     } else {
       yargs.global(key)
-      implied[key] = value
+      if (!implied[key]) {
+        implied[key] = []
+      }
+      if (Array.isArray(value)) {
+        value.forEach((i) => self.implies(key, i))
+      } else if (typeof value === 'string') {
+        implied[key].push(value)
+      } else {
+        throw new YError('Invalid second argument passed to implied.' +
+        ' Expected string or array but received ' + typeof value + '.')
+      }
     }
   }
   self.getImplied = function getImplied () {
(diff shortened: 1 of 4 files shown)


bcoe:
@MattSturgeon should have checked open pulls before issues 😛 

thanks for taking a stab at this; why don't you join the slack when you have a moment, and we can coordinate landing this feature.

lib/validation.js
```diff
@@ -318,7 +318,7 @@ module.exports = function (yargs, usage, y18n) {
         conflicting[key] = []
       }
       if (Array.isArray(value)) {
-        conflicting[key].concat(value)
+        conflicting[key] = conflicting[key].concat(value)
```
MattSturgeon:
```js
// Kinda sucks having to replace the array object instead of just mutating it,
// plus we have to type conflicting[key] twice
conflicting[key] = conflicting[key].concat(value)

// About the same speed as concat + assign, but we get to keep the same array
// object. Still kinda ugly though
Array.prototype.push.apply(conflicting[key], value)

// Twice as fast, less hacky than push.apply, but even more verbose
for (var i = 0, l = implied[key].length; i < l; i++) {
    implied[key].push(value[i])
}

// Should be similar to above, but slightly cleaner
// I think this is my favourite.
for (var i in value) {
    implied[key].push(i)
}

// I didn't consider forEach() since it is known to be way slower than a for loop
```

MattSturgeon:
Rebased on master using ES6 style, added a check for valid implies/conflicts `value`s which throws a `YError` if passed - this causes a test to fail, so I'll need to update the tests.

FYI the diff reads much better with [whitespace ignred](https://github.com/yargs/yargs/pull/922/files?w=1) because some of the indention had to change.

lib/validation.js
```diff
@@ -234,7 +235,17 @@ module.exports = function validation (yargs, usage, y18n) {
...
-      implied[key] = value
+      if (!implied[key]) {
+        implied[key] = []
+      }
+      if (Array.isArray(value)) {
+        value.forEach((i) => self.implies(key, i))
+      } else if (typeof value === 'string') {
+        implied[key].push(value)
+      } else {
+        throw new YError('Invalid second argument passed to implied.' +
```
bcoe: the reason that tests are failing is that there's this weird edge-case feature where an argument can imply a certain # of positional arguments in `_`:

```js
yargs.implies({
  foo: 4
})
```

So the second argument can be a number, or a string, or an array with your design. Rather than pulling in the `YError` object and throwing your own error, I would pull in the `argsert` library and do this:

```js
  self.implies = function implies (key, value) {
    argsert('<string|object> [array|number|string]', [key, value], arguments.length)

      Object.keys(key).forEach((k) => {
        self.implies(k, key[k])
      })
    } else {
      yargs.global(key)
      if (!implied[key]) {
        implied[key] = []
      }
      if (Array.isArray(value)) {
        value.forEach((i) => self.implies(key, i))
      } else {
        implied[key].push(value)
      }
    }
  }```

Comment (MattSturgeon) [inline lib/validation.js]: Thanks for the suggestion - it looks much cleaner now.

**Comment by bcoe** [comment]
@MattSturgeon I've released your work to the `next` tag of yargs, I'd love your help testing (since the next release of yargs is pretty large).

`npm i yargs@next --save`

**Comment by Morishiri** [comment]
I've found an issue here.

If you specify array value for `implies` and this implications are failed then we are getting output multiplied by array length.

For example:
```
.option('all', {
    describe: 'My description',
    boolean: true,
    implies: ['svn-url', 'svn-branch', 'svn-user', 'svn-password']
})
```
without specified any implied option will result in this error message:
```
Implications failed:
    all -> svn-url all -> svn-branch all -> svn-user all -> svn-password all -> svn-url all -> svn-branch all -> svn-user all -> svn-password all -> svn-url all -> svn-branch all -> svn-user all -> svn-password all -> svn-url all -> svn-branch all -> svn-user all -> svn-password
```

I will try to create pull request as soon as I will find a moment to do so if you are willing to accept one.

MattSturgeon:
@Morishiri you should probably open a new issue instead of commenting on this one. Are you saying the error message should be cleaned up and less verbose?

Morishiri:
Yes, because now it shows all implications (not only failed) multiplied by implication count. I've created a merge request fixing this. #954 You may want to look into it.

PR #926, opened 2017-07-27
Change engine node version to >=4

sudo-suhas:
One of the breaking changes of `v8.0.0` was requiring node 4 This updates the `engine.node` field in `package.json` which had the wrong value of `>=0.10`

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,6 @@
   ],
   "license": "MIT",
   "engine": {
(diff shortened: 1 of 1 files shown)


bcoe:
@sudo-suhas thanks for the contribution 👍 FYI, if you feel like contributing more, I recently created a community slack for yargs:

PR #927, opened 2017-07-27
BugFix: Exclude positional arguments from completion output

### LINKED ISSUE #818: Bash completion with Positional Arguments
- command: `exports.command = 'test <dir>';`
- exec: `node bin.js --get-yargs-completions`
- output: `test <dir>`
- then the bash completion will output  `test` and `<dir>`

[completion](https://github.com/yargs/yargs/blob/master/lib/completion.js#L59) should be `command[0].split('  ')[0]`

IvanGoncharov:
Fixes #818 + Added test


bcoe:
@IvanGoncharov thanks for this contribution! and sorry for the slow reply, I've been a bit buried in open-source tickets, and also dealing with some stressful personal stuff.

In an effort to speed up development, I've created a community slack for yargs:

Would love to have you join, so that we can coordinate landing this and future patches.

test/completion.js
```diff
@@ -118,7 +118,7 @@ describe('Completion', function () {
 
     it('does not complete hidden commands', function () {
       var r = checkUsage(function () {
-        return yargs(['./completion', '--get-yargs-completions', 'cmd'])
+        return yargs(['./completion', '--get-yargs-completions'])
```
bcoe: curious as to why this change was needed?

bcoe:
@IvanGoncharov thanks for the contribution 👍

PR #937, opened 2017-08-15
refactor: ES6ify

sudo-suhas:
## Changelog
* Add `'use strict'` directive in source and test files. Here's why:
  ```
  λ node --version
  v4.0.0
  
  λ echo let a = 1 > test-node4.js

  λ node test-node4.js
  E:\Projects\experiments\test-node4.js:1
  (function (exports, require, module, __filename, __dirname) { let a = 1
                                                                ^^^
  SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
      at exports.runInThisContext (vm.js:53:16)
      at Module._compile (module.js:413:25)
      at Object.Module._extensions..js (module.js:452:10)
      at Module.load (module.js:355:32)
      at Function.Module._load (module.js:310:12)
      at Function.Module.runMain (module.js:475:10)
      at startup (node.js:117:18)
      at node.js:951:3
  ``` 
* Use arrow functions for callbacks and simple functions(`() => someval`). See below for caveats.
* Use `const`/`let` instead of `var`. See below for caveats.
* Use `${template} strings`
* Wherever possible, set appropriate name for anonymous function. This helps for stack traces.
* Object shorthands (`{ min: min, f: function () {} }` 🡺 `{ min, f() {} }`)
* Alias and reuse `noop`(`function () {}`) in tests

### Caveats
* File `lib/assign.js` was not modified as it is being removed in #936
* Some functions access `arguments`. This doesn't work well with arrow functions. So a few functions have been left alone and not converted to arrow functions.
* In the file `yargs.js`, some variables are accessed using scope hoisting. These have been left as `var`(ex: `exitProcess`).
* Some tests call `this.skip`, `this.timeout`. These haven't been changed to arrow functions.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,4 @@
+'use strict'
 // classic singleton yargs API, to use yargs
 // without running as a singleton do:
 // require('yargs/yargs')(process.argv.slice(2))
@@ -21,7 +22,7 @@ function Argv (processArgs, cwd) {
     to get a parsed version of process.argv.
 */
 function singletonify (inst) {
-  Object.keys(inst).forEach(function (key) {
+  Object.keys(inst).forEach((key) => {
     if (key === 'argv') {
       Argv.__defineGetter__(key, inst.__lookupGetter__(key))
     } else {
--- a/lib/apply-extends.js
+++ b/lib/apply-extends.js
@@ -1,12 +1,14 @@
-var fs = require('fs')
-var path = require('path')
-var YError = require('./yerror')
 
-var previouslyVisitedConfigs = []
+'use strict'
+const fs = require('fs')
+const path = require('path')
+const YError = require('./yerror')
+
+let previouslyVisitedConfigs = []
 
 function checkForCircularExtends (path) {
   if (previouslyVisitedConfigs.indexOf(path) > -1) {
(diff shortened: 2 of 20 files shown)


sudo-suhas:
@bcoe I would urge you to merge this sooner than later to avoid potential merge conflicts

bcoe:
@sudo-suhas point taken 👍 mind rebasing, and I'll work on landing this?

sudo-suhas:
image

plroebuck:
@sudo-suhas Noticed you also used arrow functions with Mocha test code although [discouraged](https://mochajs.org/#arrow-functions) in their documentation. Did you (or @bcoe) experience any problems related to that?

sudo-suhas:
@plroebuck I was not aware that it was discouraged. However, I did not use arrow funcs where `this` was being used in the test:

> Some tests call `this.skip`, `this.timeout`. These haven't been changed to arrow functions.

bcoe:
@plroebuck as @sudo-suhas indicates, we rarely use any of the helpers bound to `this`; so I think we've been okay for the most part -- definitely something to keep in mind as we add more tests.

PR #940, opened 2017-08-22
feat: add .then() method, which resolves a promise

bcoe:
_this branch makes a couple small tweaks to @bernardmcmanus great work in #915. Moving the `.then()` functionality slightly closer to that of `.parse()`_


bcoe:
@bernardmcmanus @borela would love your feedback on this, so that we can work on landing the feature ASAP -- I'm starting to have a few more cycles for OSS again, and am excited to get a new major release of yargs out soon.

borela:
Loved promise syntax :D I'll try to use it on a local project.

docs/api.md
```diff
@@ -1206,6 +1232,19 @@ If `key` is an array, interpret all the elements as strings.
 `.string('_')` will result in non-hyphenated arguments being interpreted as strings,
 regardless of whether they resemble numbers.
 
+<a name="then"></a>.then()
```
bcoe: to make this as powerful as `.parse()` we should allow a context object to be passed in, this is useful for passing state around (as an example we use this for chat bots at npm).

PR #945, opened 2017-08-25
Updated yargs.js to allow both undefined and nulls

mike-weshop:
TypeScript requires the usage of undefined instead of null - this update allows both.

### DIFF
--- a/yargs.js
+++ b/yargs.js
@@ -384,7 +384,7 @@ function Yargs (processArgs, cwd, parentRequire) {
   }
 
   self.demandCommand = function demandCommand (min, max, minMsg, maxMsg) {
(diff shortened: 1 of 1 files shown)


mike-weshop:
The main reason for this, if I use "null" then the TypeScript Compiler fails as it needs undefined instead of null. And if i use undefined then yargs fails as it requires a number or null. This hopefully fixes yargs for people using TS.

bcoe:
@mike-weshop thanks for the contribution; off the top of my head I can't think of why accepting both `undefined` and `null` anywhere would cause issues, we can always follow on with a patch worst case scenario.

bcoe:
@mike-weshop your work is scheduled to go out the door in the next release of yargs. Mind helping me test? You can do so by running:

`npm i yargs@next`

PR #949, opened 2017-09-03
fix: strict mode should not fail for hidden options

### LINKED ISSUE #948: Hidden option is not recognized
An option is not recognised when declared `hidden` by specifying a `falsy` describe.

```js
var yargs = require("yargs");

yargs
.strict()
.option('foo', { type: 'boolean', describe: false })
.parse(['node', 'my-script.js', '--foo']);

// "Unknown argument: foo"
```

```js
var yargs = require("yargs");

yargs
.strict()
.option('foo', { type: 'boolean', describe: 'Foo option' })
.parse(['node', 'my-script.js', '--foo']);

// { foo: true }
```

Am I missing something here?

Thanks!

bcoe:
Address a bug @satazor found related to `strict()` mode firing for hidden options -- in the process I cleaned up strict mode logic for options a bit ... it seems like simply looking at the `aliases` map covers most of our bases.

fixes #948

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -120,26 +120,16 @@ module.exports = function validation (yargs, usage, y18n) {
 
   // check for unknown arguments (strict-mode).
   self.unknownArguments = function unknownArguments (argv, aliases, positionalMap) {
-    const aliasLookup = {}
-    const descriptions = usage.getDescriptions()
-    const demandedOptions = yargs.getDemandedOptions()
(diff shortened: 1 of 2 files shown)


bcoe:
@satazor I've released this to the `next` tag of yargs; would love your help testing since yargs@9.x is shaping up to be a somewhat large release:

`npm i yargs@next --save`

satazor:
I can confirm it works

PR #950, opened 2017-09-06
Run default cmd even if it is the only defined cmd

danez:
This ensures that the default command is executed even if the default command
is the only command that is defined.


bcoe:
@danez good catch 👍 I have a giant backlog on issues, but am gradually landing things. Feel free to bug me in Slack if I'm ever taking a long time on something:

PR #952, opened 2017-09-09
Fix #947 - config and normalise can be disabled with false

### LINKED ISSUE #947: .option('x', { normalize: false }) still normalizes the option
Yargs 8.0.2 still normalizes an argument even if a values of `false` is passed to options.

```javascript
#!/usr/bin/env node
"use strict";
yargs.option('database', { normalize: false });
console.log(yargs.argv.database);
```
Notice how the slash in the URL is removed:
```
> ./normalize-test.js --database mongodb://url
mongodb:/url
```
Is this because the code is just checking for the key presence rather than the actual value?
https://github.com/yargs/yargs/blob/a40cbc95277037b614c6a36b817ad31d4327fdae/yargs.js#L594

brentsnook:


brentsnook:
Travis build seems to have timed out or was cancelled after 6 hours which is weird considering how simple the change was. I don't have permission to kick it off again, can someone please give it a nudge?

bcoe:
@brentsnook this looks reasonable to me, I've kicked the build back off.

PR #881, opened 2017-05-24
feat: middleware

Khaledgarbaya:
OK this is a very WIP and ugly first draft of the middleware feature.
I will need some help to change passing the middleware using a chaining function instead of an array.
For more context check https://github.com/yargs/yargs/issues/876

@bcoe I would love to get some feedback from you.

TODO : 
- [x]  Add documentation
- [x]  Add more tests
- [x]  Cleanup
- [x]  Rebase and squash commits

PR #955, opened 2017-09-14
[feature] add optional errorHandler

### LINKED ISSUE #953: Error Handler not working with `commandDir` ?
Hey,

i have a simple setup:

```js
//index.js

yargs
    .commandDir(`${__dirname}/commands`)
    .argv
```

```js
//commands/foo.js

module.exports = {
    command: "foo",
    desc: "some desc",
    builder: (yargs) => {
        return yargs
            .exitProcess(false)
            .fail((...args) => {
                console.log("handle error");
                process.exit(1);
            });
    },
    handler: () => {
        throw new Error("test");
    }
};
```

```bash
$ node index.js foo

node_modules/yargs/yargs.js:1079
      else throw err
           ^

Error: test
    at Object.handler (commands/foo.js:13:15)
    at Object.self.runCommand (node_modules/yargs/lib/command.js:233:22)
    at Object.Yargs.self._parseArgs (node_modules/yargs/yargs.js:990:30)
    at Object.get [as argv] (node_modules/yargs/yargs.js:927:19)
   ...
```

it never reaches the `fail` function?! what am i doing wrong here?

i simply want to exit with errorCode `1` without displaying the original Error but instead a custom error message.

digitalkaoz:
to be able the handle errors individually thrown by handlers:

```js
require('yargs')
.command(
  '*',
  'example',
  function (yargs) { return yargs },
  function (argv) {
    throw new Error('test')
  },
  function (error) {
    console.log('error catched : ' + error.message)
  }
)
.argv
```

it can handle both, thrown Errors and returned rejected Promises

this fixes #953

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -406,3 +406,18 @@ some of yargs' parsing features:
 
 See the [yargs-parser](https://github.com/yargs/yargs-parser#configuration) module
 for detailed documentation of this feature.
+
+### error Handlers
+
+it is possible to pass an `errorHandler` to commands, to deal with thrown Exceptions or rejected Promises.
+
+```js
+exports.command = 'prune <name> [names..]'
+exports.desc = 'Delete tracked branches gone stale for remotes'
+exports.errorHandler = function(error) {
+  //deal with your error here
+}
+exports.handler = function (argv) {
+  throw new Error("something went wrong");
+}
+```
--- a/docs/api.md
+++ b/docs/api.md
@@ -183,7 +183,7 @@ var argv = require('yargs')
   .argv
 ```
 
-.command(cmd, desc, [builder], [handler])
+.command(cmd, desc, [builder], [handler], [errorHandler])
 -----------------------------------------
 .command(cmd, desc, [module])
 -----------------------------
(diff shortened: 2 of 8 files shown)


bcoe:
@digitalkaoz I think my preference would be to avoid adding another method to `command` at this time (partially because we have another pull request on the wire that uses the 5th argument to `.command` for introducing the [concept of middleware](https://github.com/yargs/yargs/pull/881)).

What I would like would be to make `.fail()` work the way that you expect; at this time things _kind of_ work if you were to throw a `YError` rather than the default `Error` object. All errors generated by yargs itself use this base class for errors. I think what we should probably do, is make it so _if_ you've used the `.fail` method to catch errors, yargs should proceed to start capturing all error objects.

bcoe:
@digitalkaoz did you get a chance to look into using a `YError` any thoughts on what the default behavior should be?

digitalkaoz:
Im currently on vacation, so gimme a few more weeks :)

bcoe:
@digitalkaoz still interested in seeing this over the finish line?

bcoe:
@digitalkaoz having reflected on this a bit, and chatted with @evocateur, I'm not convinced that adding an additional argument to the command API is the best approach to better handling errors that occur in asynchronous commands ... let's start a discussion in #1102

Issue #888, opened 2017-06-01
Feature request: Multiple `usage`


servel333:
Being able to all usage multiple times to demonstrate different ways to call the script makes usage lines more readable and different ways to call the script more obvious, especially with mutually exclusive calling parameters.

Example:

    require('yargs')
    .usage([
      '$0 <type>:<internalId> [<type>:<internalId> [...]]',
      '$0 --type=<type> --id=<internalId>',
      '$0 --type=<type> --internalId=<internalId>',
      '$0 --type=<type> --externalId=<externalId>',
    ].join("\n"))

Could be

    require('yargs')
    .usage('$0 <type>:<internalId> [<type>:<internalId> [...]]')
    .usage('$0 --type=<type> --id=<internalId>')
    .usage('$0 --type=<type> --internalId=<internalId>')
    .usage('$0 --type=<type> --externalId=<externalId>')

Expected output

    ./get-record.js --help
    bin/get-record.js <type>:<internalId> [<type>:<internalId> [...]]
    bin/get-record.js --type=<type> --id=<internalId>
    bin/get-record.js --type=<type> --internalId=<internalId>
    bin/get-record.js --type=<type> --externalId=<externalId>

bcoe:
@servel333 seems like a reasonable improvement to me; feel like taking a stab at a pull request?

Morishiri:
I've created pull request making it possible, You may want to take a look: https://github.com/yargs/yargs/pull/958

servel333:
@Morishiri Thanks! Looks like a good PR.

PR #958, opened 2017-09-18
feat: multiple usage calls are now collected, not replaced

Morishiri:
Feature request #888


bcoe:
@Morishiri thanks for this contribution at first glance this looks really solid.

PR #962, opened 2017-09-25
fix: show hidden options in help

### LINKED ISSUE #851: options without description to not appear in --help
`yargs@7.0.2`

```js
#!/usr/bin/env node

// @flow

import yargs from 'yargs';
import passthroughService from './commands/passthroughService';

yargs
  .command(passthroughService)
  .option('logstash', {
    describe: 'Logstash host:port'
  })
  .help()
  .parse(process.argv.slice(1));

```

```js
// ./commands/passthroughService;

export default {
  builder: (yargs: Object) => {
    return yargs
      .options({
        forward: {
          type: 'string'
        },
        listen: {
          default: 'http://0.0.0.0:3000',
          type: 'string'
        }
      });
// [..]
```

```bash
$ babel-node src/bin/forwardProxy.js passthrough-service --help
Options:
  --logstash  Logstash host:port
  --help      Show help                                                [boolean]
  --listen                             [string] [default: "http://0.0.0.0:3000"]
```

Note the missing `forward` option.

laggingreflex:
Options with `description: false` will now show up help

fixes #851

### DIFF
--- a/test/usage.js
+++ b/test/usage.js
@@ -2314,10 +2314,12 @@ describe('usage tests', () => {
       const r = checkUsage(() => yargs(['--help'])
           .option('answer', {
             type: 'string',
-            choices: ['yes', 'no', 'maybe']
+            choices: ['yes', 'no', 'maybe'],
+            hidden: true
           })
           .option('confidence', {
-            choices: [0, 25, 50, 75, 100]
+            choices: [0, 25, 50, 75, 100],
+            hidden: true
           })
           .wrap(null)
           .argv
@@ -2694,6 +2696,33 @@ describe('usage tests', () => {
         'Options:',
         '  --help     Show help                                                 [boolean]',
         '  --version  Show version number                                       [boolean]',
+        ''
+      ])
+    })
+  })
+
+  describe('hidden options', () => {
+    it('--help should display all options except for hidden ones', () => {
+      const r = checkUsage(() => yargs('--help')
+          .options({
+            foo: {
+              describe: 'FOO'
+            },
+            bar: {},
+            baz: {
+              describe: 'BAZ',
+              hidden: true
+            }
+          })
+          .argv
+        )
+
+      r.logs[0].split('\n').should.deep.equal([
+        'Options:',
+        '  --help     Show help                                                 [boolean]',
+        '  --version  Show version number                                       [boolean]',
+        '  --foo      FOO',
+        '  --bar',
         ''
       ])
     })
--- a/yargs.js
+++ b/yargs.js
@@ -647,7 +647,7 @@ function Yargs (processArgs, cwd, parentRequire) {
       }
 
       const desc = opt.describe || opt.description || opt.desc
-      if (desc) {
+      if (!opt.hidden) {
         self.describe(key, desc)
       }
 


laggingreflex:
Having second thoughts on this. Currently it **shows** options if their description is set to **false**.  (as per [your comment](https://github.com/yargs/yargs/issues/851#issuecomment-301980416) @bcoe). Would it be better if it shows options when description is left blank, and **doesn't** show when it's set to false? This would also be more inline with the current docs (as [pointed out](https://github.com/yargs/yargs/issues/851#issuecomment-293817339) by @gajus)

>exports.describe: string used as the description for the command in help text, **use false for a *hidden* command**

Or could it break existing usage?

gajus:
> Would it be better if it shows options when description is left blank, and doesn't show when it's set to false? This would also be more inline with the current docs (as pointed out by @gajus)

👍

bcoe:
@laggingreflex good catch; I'm with both you and @gajus on this.

Let's show the option if an empty string is provided for the description, hide on `null`, `false`, `undefined`.

Sound reasonable?

gajus:
I still think that an explicit option should be used to make a parameter hidden. All this convention based behaviour is tricky to document/ remember.

As long as option is provided, it should be visible, unless that option contains configuration `{hidden: true}`, i.e.

```js
.options({
  foo: {
    describe: 'FOO'
  },
  bar: {},
  baz: {
    describe: 'BAZ',
    hidden: true
  }
});

```

With this configuration,

* `foo` and `bar` are visible by default
* `baz` is visible if `--yargs-show-hidden` (or whatever) option is provided

The benefits of this approach:

* You can validate that `describe` value is a string
* Hidden options can have a description

bcoe:
@gajus this argument seems reasonable to me, and I think we're inching up on a new major release of yargs any ways, with some of the work I've been doing around command parsing/number coercion.

@laggingreflex thoughts on the options discussed above?

bcoe:
@laggingreflex this looks good to me; any idea why test coverage dropped so much?

laggingreflex:
I was testing with `.only` locally which dropped the coverage. Fixed now. 

Also this doesn't yet implement `--yargs-show-hidden`, I'm working on an alternative for that.

bcoe:
@laggingreflex 👍 I'm working on a pretty big `10.x` release, it would be nice to land this in it (I'd also love help testing the changes on `master` when you have time).

laggingreflex:
Instead of `--yargs-show-hidden` how about this:

```js
yargs.showHelp('log', {showHidden: [true|false]})
```

This gives the user more control. Like maybe they wouldn't want to show hidden commands at all. Or maybe they'd want it behind a custom flag. This way they can implement it themselves. 

It however require that the user implements the `--help` themselves too:

```js
const argv = yargs
  .options({
    baz: {
      describe: 'BAZ',
      hidden: true
    },
    'show-hidden': {
      describe: 'Show hidden options',
      type: 'boolean'
    },
  })
  .help(false)
  .argv

if (yargs.help) {
  yargs.showHelp('log', {hidden: argv.showHidden})
}
```

bcoe:
@laggingreflex I'd be tempted to discuss `--yargs-show-hidden` in a separate issue. I think that this feature is enough of an edge-case that there's benefit to landing this feature first... mainly, I'm on the fence about `--yargs-show-hidden`; for starters I don't like baking the yargs name into a flag, so I'd probably call it something like `--show-hidden` (folks calling this flag feels like an edge-case to me though).

laggingreflex:
I've undone the last commit and I'll move it to a separate issue/pr. 

This now only enables the `{hidden: true}` option to hide the option from `--help` (instead of by omitting the description).

Issue #959, opened 2017-09-18
"command" means "argument"?


sloanlance:
I'm new to yargs after trying "commander" and finding it inadequate.  commander was recommended to me by a colleague, but npm tells me yargs is the most popular package with the features I want.  I've used many CLI parsers over the years with various programming languages.

I'm a little confused by yargs' terminology.  Are "commands" the same as "arguments"?  I thought `command()` was only for program subcommands and their own sets of options.  However, if I want yargs to describe and require a certain number of positional arguments, it seems I need to use `command()` for that as well.

Here's an example from my code:

```js
const argumentErrorMessage = 'Error: Exactly 2 arguments are required';
const arguments = require('yargs')
    .usage('Usage: $0 [options] <specFrom> <specTo>')
    .command('specFrom', 'Specification version to be converted from') // first ARGUMENT
    .command('specTo', 'Specification version to be converted to') // second ARGUMENT
    .demandCommand(2, 2, argumentErrorMessage, argumentErrorMessage) // num. of req. ARGUMENTS (NOT commands)
    .argv._;

[specFrom, specTo] = arguments.slice(0, 2);

console.log('Converting from: "%s".  Converting to: "%s"', specFrom, specTo);
```

This *works*, but I'm not satisfied that the _arguments_ are called "_commands_" in the usage message:

```txt
Usage: example.js [options] <specFrom> <specTo>

Commands:
  specFrom  Specification version to be converted from
  specTo    Specification version to be converted to

Options:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]

Error: Exactly 2 arguments are required
```

Am I using the correct yargs methods for positional arguments, or is there some other way this should be done?

matatk:
I believe that I'm running into the same issue here.  I have been trawling yargs' GitHub issues to see if it is possible to provide a name (i.e. alias) for a positional argument, a la #728 '"flagless" arguments without a command'.  However, that issue was closed in favour of #467 'add support for default commands to yargs'.

It may well be that #467 solves the [issue I'd like to address](https://github.com/mozilla/addons-linter/issues/1550), but it's confusing because it looks like we'd have to code it as if the first positional argument is a command, rather than just setting up an alias name for the first positional argument before any flagged arguments.  It would be great if this could be cleared up (either by the docs, or allowing aliases for positional arguments via the API).

Thanks for yargs, BTW; have been using it for a little while and it rocks :-).

bcoe:
[...]

Issue #570, opened 2016-07-30
How would I describe a positional argument?


wearhere:
Let's say that I'm configuring `yargs` like so:

``` js
var argv = require('yargs')
  .usage('Usage: $0 <sourceDir>')
  .demand(1)
  .argv;
```

How would I `.describe` the `sourceDir` argument?

This [similar issue](https://github.com/yargs/yargs/issues/541) discussed positional arguments to commands, but note that I am not using a command here, but rather would like to describe a top-level positional argument.

bcoe:
@wearhere we've been talking about adding top-level commands, which I think could address this issue well.

popomore:
@bcoe How is the progress of this issue?

atian25:
@bcoe +1 for this

bcoe:
this is now possible with default commands and `.positional()` 🎉 

```js
var argv = require('./')
  .usage('$0 <source-dir>', 'performs operation on source-dir', (yargs) => {
    yargs.positional('source-dir', {
      describe: 'this is the source directory'
    })
  })
  .demand(1)
  .argv;

console.info(argv)
```

Issue #541, opened 2016-07-06
Allow defining positional argument


hax:
Currently describe/number/string... only apply to option, I expect it would also apply to positional argument.

```js
yargs.command('cmd <arg>')
     .number('arg')
     .choices('arg', [2, 3, 4])
```

We also need `arg/args` method like `option/options`

```js
yargs.command('cmd [arg]')
     .arg('arg', {type: 'string', default: 'ok'})
```

Here is the list of methods or keys of arg/args opt I think could apply to arguments:
- boolean
- choices
- config
- configParser
- default
- defaultDescription
- desc/describe/description
- group
- nargs
- normalize
- number
- skipValidation (though I don't understand the use case of this method)
- string
- type
- env (missed in current option/options api)

bcoe:
hey @hax, we actually already support this with the concept of a builder and a handler for commands (these can even be provided in sub-modules, how cool is that!).

we're in discussion to figure out how to make these features more intuitive, but the current pass is pretty well documented here:

https://github.com/yargs/yargs#commandmodule

Would love to have you join in on the conversation [here](https://github.com/yargs/yargs/issues/519) regarding version `5.x` of yargs. We're hoping to make some additional improvements to how commands work.

hax:
@bcoe You mean the 5.x branch already support refine positional argument like option? I will try it.

bcoe:
@hax should be supported in `4.x` with builders and handlers 👍

hax:
@bcoe Sorry, I still can not find how to refine the argument. (with version 4.8.0-candidate)

```js
yargs.command('cmd <arg>')
     .number('arg')
     .choices('arg', [2, 3, 4])
```
As my example in the first comment, I expect to define the the constraint (should be a number, should be in 2, 3, 4) of `<arg>`,  but actually I just defined a `--arg` option.

wearhere:
I'm confused as well, how would you describe `<arg>`? The examples given in https://github.com/yargs/yargs#commandmodule just talk about describing options.

hax:
@wearhere I confirmed that it only describe options, my question is how to solve the requirements of  describing `<arg>`.

bcoe:
@wearhere @hax commands currently only allow for simply positional arguments, but using the commands API you can define options that a command would accept, so:

```js
yargs.command('cmd <foo>')
```

[...]

PR #967, opened 2017-10-05
feat: make positional arguments fully configurable

bcoe:
Introduces the `.positional()` method, which makes positional arguments fully configurable (but wait there's more, positional arguments will also be included separately in help output):

screen shot 2017-10-05 at 12 08 46 am

TODO:

- [x]  make it so we default the settings `demanded` and `array` based on the command string.
- [x] accept the argument `$0` to represent a default command (so that it's more intuitive that commands can be used to handle top-level operations).
- [x] add documentation.
- [x] ~get translations for `Positionals:` (help wanted!).~ we can do this post-hoc.
- [x] finish adding tests.

see: #541, #570, #959, etc

cc: @tkarls, @hax, @sloanlance, @iarna, etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules/
 *.swp
 test.js
 coverage
+package-lock.json
--- a/lib/command.js
+++ b/lib/command.js
@@ -246,54 +246,41 @@ module.exports = function command (yargs, usage, validation) {
     argv._ = argv._.slice(context.commands.length) // nuke the current commands
     const demanded = commandHandler.demanded.slice(0)
     const optional = commandHandler.optional.slice(0)
-    const parseOptions = {
-      array: [],
-      default: {},
-      alias: {}
-    }
     const positionalMap = {}
 
     validation.positionalCount(demanded.length, argv._.length)
 
     while (demanded.length) {
       const demand = demanded.shift()
-      populatePositional(demand, argv, positionalMap, parseOptions)
+      populatePositional(demand, argv, positionalMap)
     }
 
     while (optional.length) {
       const maybe = optional.shift()
-      populatePositional(maybe, argv, positionalMap, parseOptions)
+      populatePositional(maybe, argv, positionalMap)
     }
 
     argv._ = context.commands.concat(argv._)
 
-    postProcessPositionals(argv, positionalMap, parseOptions)
+    postProcessPositionals(argv, positionalMap, self.cmdToParseOptions(commandHandler.original))
 
     return positionalMap
   }
 
-  // populate a single positional argument and its
-  // aliases onto argv.
   function populatePositional (positional, argv, positionalMap, parseOptions) {
-    const cmds = positional.cmd.slice(0)
-    const cmd = cmds.shift()
+    const cmd = positional.cmd[0]
     if (positional.variadic) {
-      parseOptions.array.push(cmd)
-      parseOptions.default[cmd] = []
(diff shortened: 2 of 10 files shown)


lib/command.js
```diff
@@ -239,64 +246,73 @@ module.exports = function command (yargs, usage, validation) {
     argv._ = argv._.slice(context.commands.length) // nuke the current commands
     const demanded = commandHandler.demanded.slice(0)
     const optional = commandHandler.optional.slice(0)
+    const parseOptions = {
```
bcoe: perhaps refactor the population of these parser options into a separate helper based on the command string, since they're also required for `.positional()`.

bcoe:
@jeffijoe @ruimarinho @tkarls @atian25 @popomore @matatk @sloanlance

I just took on this major refactor of how we parse commands in yargs to try to:

1. make commands first class citizens that have most of the same parsing hints as flag options.
2. make the command API a bit less confusing (by improving documentation, and adding a logical command for configuring commands `.positional()`).

I would love your feedback -- I'm feeling pretty excited about this update to commands.

jeffijoe:
@bcoe very nice, great work!

.gitignore
```diff
@@ -4,3 +4,4 @@ node_modules/
 *.swp
 test.js
 coverage
+package-lock.json
```
zkat: 😎

bcoe:
@popomore, @hax, @Morishiri, @zkat, @jeffijoe this work is now available on the candidate release of `yargs@10.0.0`; I would love some help testing (since this is one of the bigger overhauls to yargs' functionality in a while).

You can play with it by running:

`npm i yargs@next`

matatk:
Thanks for working on this; it is looking very promising. I have two questions...

1. It seems that I cannot declare a positional argument without declaring a command.  The following is invalid code.  Is there a way to have positional arguments without having to declare a command?

    ```js
    const argv = require('yargs')
    	.usage('$0 greeting')
		.positional('greeting', {
			describe: 'what you would like the program to say',
			type: 'string'
		})
    	.help()
    	.argv
    
    console.log(argv)

    ```

2. When `argv` gets returned, it comes back as an object such as `{ _: [ 'moo' ], version: false, help: false, '$0': 'index.js' }`. When positional arguments are in use, will we be able to use their name, rather than just `_` to refer to them?  Being able to name the arguments would make code that uses yargs more readable.  (There's [an original issue along these lines](https://github.com/mozilla/addons-linter/issues/1550) that brought me here, and this was the root of it.)

bcoe:
@matatk try this on the most recent branch (npm yargs@next):

```js
const argv = require('./')
  .command('$0 [port]', 'start the server', (yargs) => {
    yargs
      .positional('port', {
        describe: 'port to bind on',
        default: 5000
      })
   })
   .argv
```

I could imagine an alias for this that looks like this:

```js
const argv = require('./')
  .usage('$0 [port]', (yargs) => {
    yargs
      .positional('port', {
        describe: 'port to bind on',
        default: 5000
      })
   })
   .argv
```
I wonder if this might be overkill though -- my feeling is it's probably better to encourage people to move towards using `command('$0')` rather than `.usage()` when they're describing the properties of positional arguments.

matatk:
@bcoe that's brill; just what I was after—super-easy to configure positional for the default command *and* I get to refer to the positional argument by name; thanks :-).

iarna:
@bcoe This looks great! Just be sure to xref in the docs for `.usage()` and I think it's fine to keep them apart. =)

PR #968, opened 2017-10-06
Support for a custom require function

generalov:
The PR to use a custom loader for the commands. Actually all necessary functionality already implemented internally in the Yargs. Thank you guys for the great tool. :)

For example we can use [@std/esm](https://www.npmjs.com/package/@std/esm) as a loader to load the commands written in ES6 modules syntax.
```
const loader = require('@std/esm')(module)

yargs(process.argv.slice(2), undefined, loader)
  .commandDir('cmds', {
    extensions: [ 'js', 'mjs' ]
  })
  .demandCommand()
  .argv
```
cmds/es6.mjs
```
import fs from 'fs'

export const command = 'esm'
export const description = 'The commands can be defined in ES6 modules'

export function handler (argv) {
  console.log("I'm ES6 command!")
}
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,8 +8,8 @@ Argv(process.argv.slice(2))
 
 module.exports = Argv
 
-function Argv (processArgs, cwd) {
-  const argv = yargs(processArgs, cwd, require)
+function Argv (processArgs, cwd, parentRequire) {
+  const argv = yargs(processArgs, cwd, parentRequire || require)
   singletonify(argv)
   return argv
 }


bcoe:
@generalov this seems like a reasonable addition; would it be possible to add a test?

index.js
```diff
@@ -8,8 +8,8 @@ Argv(process.argv.slice(2))
 
 module.exports = Argv
 
-function Argv (processArgs, cwd) {
-  const argv = yargs(processArgs, cwd, require)
+function Argv (processArgs, cwd, parentRequire) {
```
bcoe: I'm thinking maybe we make the second parameter take an object:

```js
{cwd: 'foo', require: func () {}}
```

if a string is provided we'd fallback to the `cwd` behavior.

bcoe:
@generalov do my comments make sense, would love to see this over the finish line soon.

generalov:
@bcoe my apologues for the delay. Could you please advise me that is the best place to put a test?

bcoe:
@generalov it's my turn to apologize for taking a long time to reply! I've been busy with a project at work, and another codebase I've been coming up to speed on.

I would add a test in `test/yargs.js` -- perhaps pull in `@std/esm` as a dev dependency, and you can add a unit test for a real-world example?

bcoe:
@generalov just checking in to see if you're still interested in landing this feature 👍

bcoe:
@genepaul sorry to keep prodding you, but would love to get a unit test added so that we can land this; or perhaps someone else would like to see this over the finish line?

bcoe:
@generalov hey, given that this has been inactive for a while I'm going to go ahead and close this issue ... one thought, if you do need this functionality, is that you could simply use:

```js
yargs(processArgs, cwd, parentRequire || require).argv
```

if you need this functionality. I'd be happy to revisit this feature if you do want to make a case.

PR #970, opened 2017-10-12
chore: switch to find-up from read-pkg-up

bcoe:
`read-pkg-up` has a dependency on [WTFPL](https://github.com/sindresorhus/parse-json/pull/12), which causes licensing issues for npm/Node.js.

`find-up` does pretty much what we need, so I don't see any reason to not simply switch to using it directly CC: @zkat.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,9 +14,9 @@
   "dependencies": {
     "cliui": "^3.2.0",
     "decamelize": "^1.1.1",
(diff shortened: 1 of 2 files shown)


zkat:
LGTM! 🎉

realityking:
In case you want to switch back, sindresorhus/parse-json#12 has been merged & released.

PR #977, opened 2017-10-15
#976 add ko translation

disjukr:
seems commit log has specific pattern now but i don't know how to write commit log about translation.


bcoe:
@disjukr thank you for the translation; I'd probably call this a `chore:` using the http://conventionalcommits.org/ format.

disjukr:
thank you for info

PR #979, opened 2017-10-15
#976 Add ja translation

oti:


bcoe:
@oti @watilde thank you 😄, good catch that we were also missing the mutually exclusive translation.

PR #980, opened 2017-10-15
Add pt-BR translation to #976

madcampos:
Update the translation with the new ones.


bcoe:
@madcampos thank you \o/ glad that we're keeping these updated.

PR #983, opened 2017-10-16
[Discussion] Replacing "$0" with the executing file's base name

lostintangent:
This PR changes the behavior of the `$0` token (when used within the contents of the `usage`, `epilog` and/or `example` methods), so that it displays the executing file's basename, as opposed to the full path. This has the affect of providing default help text that is closer to what the end-user would actually run, since they would likely be running the CLI on their path (or via an `package.json` script), as opposed to using the CLI's entrypoint script path.

This is a breaking change, but since #975 introduces a breaking change for the `usage` method, I figured I'd put this PR out for discussion's sake. I updated all of the tests to ensure they still pass, but there may be other scenarios that this change would break, so I'd love some feedback.

I'm also not entirely sure this is the right behavior (e.h. as opposed to introducing a new token ala `$$0` or something, etc.), and I'm sure this conversation has been had a bunch of times already, so I'd love feedback and/or pointers to an existing/better solution for this. Thanks!

---

**Before this change (using a Yargs-based CLI named `kyte`):**

screen shot 2017-10-15 at 1 42 25 pm

**After this change:**

screen shot 2017-10-15 at 6 16 02 pm

CC @bcoe

### DIFF
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -3,6 +3,7 @@
 // failures, etc. keeps logging in one place.
 const stringWidth = require('string-width')
 const objFilter = require('./obj-filter')
+const path = require('path')
 const setBlocking = require('set-blocking')
 const YError = require('./yerror')
 
@@ -165,7 +166,7 @@ module.exports = function usage (yargs, y18n) {
 
     // the usage string.
     if (usages.length && !usageDisabled) {
-      const u = self.getUsage().replace(/\$0/g, yargs.$0)
+      const u = self.getUsage().replace(/\$0/g, path.basename(yargs.$0))
       ui.div(`${u}`)
     }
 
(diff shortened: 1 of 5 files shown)


bcoe:
@lostintangent I like this idea. I might be tempted to apply the `path.basename` logic in `yargs.js` when we initially populate `$0`, so that we don't need to apply the replacement multiple times.

What does the output look like when you're running a command line script that's not installed globally, e.g., `node ./foo.js --help`?

lostintangent:
It would display the file name (without the path), like so:

screen shot 2017-10-15 at 6 57 38 pm

bcoe:
@lostintangent I like this change, I think it's much more readable.

bcoe:
@lostintangent  I think switching to `basename` is a great idea; let's work on getting this landed so that I can rebase it with https://github.com/yargs/yargs/pull/975, and land a couple more tweaks.

bcoe:
@lostintangent having thought about this a bit, maybe you're right to perform this change in the view logic, rather than in `yargs.js` -- given that we populate `$0` on the argv object that yargs parses, I'm a bit concerned that switching to `basename` might break quite a few folks -- having said this, I'm fine with switching to `basename` in the `usage.js` since it will make the help output much cleaner.

lostintangent:
@bcoe Great timing! I finally got a minute to work on this again :) So is this PR good as-is?

Issue #510, opened 2016-05-19
Allow command handler to accept a promise


vjpr:
This would allow much easier use of ES7 async/await.

``` js
yargs.command('ls', 'List', async function() { ... })
```

**Workaround**

Use IIFE.

``` js
yargs.command('ls', 'List', function () { (async function() { ... })().then() })
```

silenceisgolden:
Could this be accomplished using gulp's [async-done](https://github.com/gulpjs/async-done) library? If so I can try to take a stab at a PR. Thanks!

bcoe:
@silenceisgolden I think we should be able to pull this off without pulling in any dependencies; once the command itself is invoked, its now the application's responsibility to handle things -- so it's perfectly fine for the handler to be async.

I think, basically, we'd just need to update the argument validation and invocation of a command handler to understand a promise. I love this idea (thanks @vjpr) and would love your help with the patch @silenceisgolden.

silenceisgolden:
@bcoe Sounds good, I'll see what I can put together.

eddywashere:
I'd like to see this as well. For now I'm getting by with this https://github.com/eddywashere/yargs-promise.

satazor:
I also ended up doing an user-land solution for this problem just like @eddywashere, but the solution integrates with yargs, instead of changing its internals.

https://github.com/moxystudio/yargs-promise-handler

ianwremmel:
This also seems to get the job done without needing to modify every command builder with a wrapper. That said, native support would be nice.

```js
process.on('unhandledRejection', function onUnhandledRejection(err) {
  var cmd = process.argv.slice(2).join(' ');
  console.error('Failed to execute "' + cmd + '"');
  console.error(err.toString());
  if (argv.debug) {
    console.error(err);
  }
  // eslint-disable-next-line no-process-exit
  process.exit(1);
});
```

zenflow:
To those subscribed: I've submitted PR #1001 to address this. Reviews are appreciated.

quilicicf:
Hi, I've tested the 10.1.0 with a command like this: 
```js
.command([ 'replicate', 'r' ], 'Clones a repository', async (_yargs) => require('./lib/git/replicate').replicate(_yargs)))
```

When the function _replicate_ contains an `await` statement, the parsing fails with this error: 
```
~/forge/Gut/node_modules/yargs/yargs.js:1100
      else throw err
           ^

TypeError: innerYargs._parseArgs is not a function
[...]

PR #1001, opened 2017-11-01
Async command handlers

zenflow:
Addresses issue #510 ~~"Allow command handler to accept a promise"~~ "Support async (i.e. promise-returning) command handlers"

> we'd just need to update the argument validation and invocation of a command handler to understand a promise

This PR handles "update [...] invocation of a command handler to understand a promise". Nothing is done if the promise resolves, but if it rejects, it is handled as ["failure"](https://github.com/zenflow/yargs/blob/master/docs/api.md#failfn) with an `err` but no `msg`.

@bcoe Not sure what needs to be done about "update the argument validation"... It seems like there's no change needed; arguments to command handlers, and validation of said arguments, aren't affected by this feature. This feature deals only with what is *returned* by the command handler (a promise, potentially). Am I missing something?

This PR also makes a slight change to the default failure handler, so that if there's no `msg` but there is an `err`, `err` will be logged instead, which will display the full stack trace for debugging.

Thank you for your work on this project, and taking the time to review this change :)

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -225,7 +225,13 @@ module.exports = function command (yargs, usage, validation) {
 
     if (commandHandler.handler && !yargs._hasOutput()) {
       yargs._setHasOutput()
-      commandHandler.handler(innerArgv)
+      const handlerResult = commandHandler.handler(innerArgv)
+      if (handlerResult && typeof handlerResult.then === 'function') {
+        handlerResult.then(
+          null,
+          (error) => yargs.getUsageInstance().fail(null, error)
+        )
+      }
     }
 
     if (command) {
--- a/test/command.js
+++ b/test/command.js
@@ -1402,4 +1402,17 @@ describe('Command', () => {
       }).to.throw(/.*\.usage\(\) description must start with \$0.*/)
     })
   })
+
+  // addresses https://github.com/yargs/yargs/issues/510
+  it('fails when the promise returned by the command handler rejects', (done) => {
+    const error = new Error()
+    yargs('foo')
+      .command('foo', 'foo command', noop, (yargs) => Promise.reject(error))
+      .fail((msg, err) => {
+        expect(msg).to.equal(null)
+        expect(err).to.equal(error)
+        done()
+      })
+      .argv
+  })
 })
--- a/docs/api.md
+++ b/docs/api.md
@@ -662,10 +662,10 @@ Examples will be printed out as part of the help message.
 <a name="exitprocess"></a>.exitProcess(enable)
 ----------------------------------
 
-By default, yargs exits the process when the user passes a help flag, uses the
-`.version` functionality, or when validation fails. Calling
-`.exitProcess(false)` disables this behavior, enabling further actions after
-yargs have been validated.
+By default, yargs exits the process when the user passes a help flag, the user
+uses the `.version` functionality, validation fails, or the command handler
+fails. Calling `.exitProcess(false)` disables this behavior, enabling further
+actions after yargs have been validated.
 
 <a name="fail"></a>.fail(fn)
 ---------
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -46,9 +46,9 @@ module.exports = function usage (yargs, y18n) {
       if (!failureOutput) {
         failureOutput = true
         if (showHelpOnFail) yargs.showHelp('error')
-        if (msg) logger.error(msg)
+        if (msg || err) logger.error(msg || err)
         if (failMessage) {
-          if (msg) logger.error('')
+          if (msg || err) logger.error('')
           logger.error(failMessage)
         }
       }


test/command.js
```diff
@@ -1402,4 +1402,17 @@ describe('Command', () => {
       }).to.throw(/.*\.usage\(\) description must start with \$0.*/)
     })
   })
+
+  // addresses https://github.com/yargs/yargs/issues/510
+  it('fails when the promise returned by the command handler rejects', (done) => {
+    const error = new Error()
+    yargs('foo')
```
bcoe: I think it might be worth adding a test for the happy path too (the promise resolving).

lib/command.js
```diff
@@ -225,7 +225,13 @@ module.exports = function command (yargs, usage, validation) {
 
     if (commandHandler.handler && !yargs._hasOutput()) {
       yargs._setHasOutput()
-      commandHandler.handler(innerArgv)
+      const handlerResult = commandHandler.handler(innerArgv)
+      if (handlerResult && typeof handlerResult.then === 'function') {
+        handlerResult.then(
+          null,
+          (error) => yargs.getUsageInstance().fail(null, error)
```
bcoe: I would probably just fail with `err.message` rather than `null`, I think your changes in `usage.js` probably wouldn't be needed then?

zcei:
Looks really promising! 👏  (pun intended 😉 )

> Not sure what needs to be done about "update the argument validation"... It seems like there's no change needed

This is a whole different topic, as async argument validation would turn whole `yargs` library into getting async, as the `.argv` cannot be available before the async validation succeeded.

As a workaround, you could throw a specific error (e.g. `UnmatchedAsyncConstraint`) and handle this in the `.fail` method (i.e. show the usage help text), together with your generic failure handling.

So getting this enhancement in without async argument validation would be awesome!

test/command.js
zenflow: Sure, but I can't think of anything to assert in that case... This feature does nothing when the handler result is thenable and it resolves.

I had actually fixed (?) my resolution handler in `lib/command.js` from a `() => {}` noop to `null` so that it would not show as an untested code path in the test coverage reports.

Anyone have a suggestion of what to assert? What do we need to watch out for in this case?

lib/command.js
zenflow: I'm passing null here and making the changes in `usage.js` so that the error **with stack trace** gets logged to the console, similar to what happens when an error is thrown in a sync command handler.

(On that note, maybe we should align how errors are handled in sync vs async command handlers, and handle sync errors as "failures" too?)

It seems important to have the stack trace logged to the console, and it seems like this is the simplest way to achieve it.

I should probably add a test to make sure this happens. I will add one if you agree that that's what we want to happen.

zenflow:
@zcei It's promising, and also thenabling 😉 (Jokes, How long have you been a dad btw? My kids are 8 and 10, Jokes)

> This is a whole different topic, as async argument validation would turn whole yargs library into getting async, as the .argv cannot be available before the async validation succeeded.

Ah, that explains it. Thanks. (D'uh)

Yeah changing yargs to do validation asynchronously seems like it have to have some significant breaking changes, since (correct me if I'm getting this wrong) it must be either always async or always sync, or [Zalgo](http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony) get's released.

Maybe we should open a separate issue for that?

lib/command.js
bcoe: sorry for the slow reply -- this logic seems reasonable to me. Let's add a test for both this and the happy path. To test the happy path, I would just have a command handler that modifies a variable, and assert that it executed as expected:

```js
let asyncHandlerCalled = false
const asyncHandler = new Promise((resolve, reject) => {
  asyncHandlerCalled = true
  return resolve()
})
yargs('foo')
  .command('foo', 'foo command', noop, (yargs) => Promise.reject(error))
```

something like that; you'll need to make sure the test is async, etc.

test/command.js
bcoe: hey, made comment above with recommendation about assertion; I suppose now that the fail handler is called async, you can just call `done()` here after your promise resolves?

bcoe:
@zenflow think you'll have a chance to see this over the finish line? otherwise hopefully I'll have some time over the holiday.

zenflow:
@bcoe Sorry for all the delay - I'm finally free from work for a while now, after a few mad weeks of finishing up, documenting, handing off, etc.

I'll definitely have time to finish this off over the next week :)

bcoe:
@zenflow @zcei please try:

`npm i yargs@next`, which has the async command handler support; if everything is looking good in this release I will promote things soon.

evocateur:
One (un-documented) side-effect of this change is that `.parse()` no longer stops the internal logger from dumping the error (`self._hasParseCallback()`) _or_ implicitly set `.exitProcess(false)`. The former smells like a bug to me, the latter probably wasn't the best choice, and in any case, the only way since v10.1.0 to catch a handler rejection _or_ a yargs validation error (or string 😭) from the same method (when using `.parse()`) is to chain a `fail()` callback before calling `.parse()`.

TL;DR: If you use async command handlers with `.parse()`, you _must_ use a chained `.fail()` callback to ensure you catch _all_ errors thrown during yargs execution.

tyrsius:
@evocateur What does a "chained `fail` callback look like"? Is that chained onto `parse` or the callback?

evocateur:
The yargs API method ".fail()". The parse() method only accepts a callback, it does not return a promise.

remorses:
the typescript types should also be updated

lexoren:
> the typescript types should also be updated

yes i've just run into the same issue

tianyingchun:
what' progress of this issue?

mleguen:
@tianyingchun About updating types?

tianyingchun:
@mleguen oh sorry, paste question to wrong place
my qeustion is ``yargs.parse(argv,callback)` issues call back before handler promise is finished. #1069`
https://github.com/yargs/yargs/issues/1069

PR #1004, opened 2017-11-07
feat: add missing simple chinese locale strings

anhulife:


bcoe:
@anhulife thank you for you contribution.

PR #1009, opened 2017-11-14
document multiple array arguments

Khaledgarbaya:
We should mention that multiple array arguments are all flattened.


bcoe:
@Khaledgarbaya thanks for the edit 👍 I've been a bit AWOL lately with a big project at work; appreciate your contributions.

PR #1015, opened 2017-11-28
Fix for requiresArg with number type

edmBernard:


lib/validation.js
```diff
@@ -57,7 +57,7 @@ module.exports = function validation (yargs, usage, y18n) {
   // make sure that any args that require an
   // value (--foo=bar), have a value.
   self.missingArgumentValue = function missingArgumentValue (argv) {
-    const defaultValues = [true, false, '']
+    const defaultValues = [true, false, '', undefined]
```
bcoe: I find this logic a little weird (it's not your fault, this oddity was already in the codebase). I'm going to do a little bit of investigation when I have some cycles and see if we can figure out a better way to determine if a value has been provided.

PR #1019, opened 2017-12-07
Add failing test for using positional

klippx:
Hello,

We have encountered what appears to be a bug in handling positionals of type string. As you can see from the failed test, yargs is parsing the example as a boolean.

We are unsure if this is a bug or if we are doing someting wrong.

We are happy to provide a fix, if you can confirm this is a bug.

### DIFF
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -1996,6 +1996,21 @@ describe('yargs dsl tests', () => {
       argv.heroes.should.eql(['batman', 'Iron Man'])
     })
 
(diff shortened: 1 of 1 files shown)


klippx:
@bcoe So great that you fixed the issue, thank you for helping us! ❤️

bcoe:
@klippx I've landed the fix to `master`, mind rebasing this branch (you will also need to add `.version(false)` to get things passing).

thanks for the contribution 👍

klippx:
@bcoe Done!

PR #1033, opened 2018-01-01
Use chai 4.x

arunasank:
This fix is required to package chai for Debian.

From a Debian packager:
> (...) in debian, we want to use a single version of chai for all
modules we ship and some modules need chai 4. So we need to port all
modules still using chai < 4 to chain 4.x.

I've updated the tests accordingly, and tested for stylistic differences. Please let me know if anything else is to be done!

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "yargs-parser": "^8.1.0"
   },
   "devDependencies": {
-    "chai": "^3.4.1",
+    "chai": "^4.1.2",
     "chalk": "^1.1.3",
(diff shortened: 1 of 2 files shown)


arunasank:
> And then there's no need to pull in the chai-things helper; other than that, I'm happy to upgrade chai for you, if you'd be kind enough to make this change.

Yep, I was hesitant about adding a new dependency, but I am new to chai land. 🙂 Thanks for the review @bcoe. I have made the necessary changes, and checked that tests pass locally.

PR #1043, opened 2018-01-11
fix: requiresArg should only be enforced if argument exists

### LINKED ISSUE #1041: requiresArg is enforced when option is not passed
index.js
```

yargs.options({
  a: {
    type: "string",
    requiresArg: true
  },
});
console.log(yargs.argv);
```
`node index.js` now fails, but it worked in versions prior 10.0.3.

I ran into the problem when using webpack-dev-server, but I think it probably is not the only thing being affected.

I have spent some time on this, and found the behavior was introduced in https://github.com/yargs/yargs/pull/1015.

bcoe:
fixes #1041 

CC: @TheLarkInn, @shellscape, @jddxf

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -59,13 +59,15 @@ module.exports = function validation (yargs, usage, y18n) {
   self.missingArgumentValue = function missingArgumentValue (argv) {
     const defaultValues = [true, false, '', undefined]
     const options = yargs.getOptions()
-
     if (options.requiresArg.length > 0) {
       const missingRequiredArgs = []
 
       options.requiresArg.forEach((key) => {
-        const value = argv[key]
+        // if the argument is not set in argv no need to check
+        // whether a right-hand-side has been provided.
+        if (Object.keys(argv).indexOf(key) === -1) return
 
+        const value = argv[key]
         // if a value is explicitly requested,
         // flag argument as missing if it does not
         // look like foo=bar was entered.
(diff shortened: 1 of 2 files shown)


lib/validation.js
```diff
@@ -59,13 +59,15 @@ module.exports = function validation (yargs, usage, y18n) {
...
     const options = yargs.getOptions()
-
     if (options.requiresArg.length > 0) {
       const missingRequiredArgs = []
 
       options.requiresArg.forEach((key) => {
-        const value = argv[key]
+        // if the argument is not set in argv no need to check
+        // whether a right-hand-side has been provided.
+        if (Object.keys(argv).indexOf(key) === -1) return
```
jddxf: Why not `argv.hasOwnProperty(key)`?

bcoe: @jddxf no good reason, will update the code accordingly.

evocateur:
Eagerly awaiting this, got bit by it trying to upgrade lerna from yargs 8.x

bcoe:
@evocateur mind taking `yargs@next` for a spin, this should correct the issue; apologies for the regression.

evocateur:
No worries, will do!

evocateur:
@bcoe No joy, pretty sure #1015 was incorrect in adding `undefined` to the `defaultValues` check. I'm hacking locally, will open a PR if I find a solution.

PR #1049, opened 2018-01-18
Add failing test for edge case in #1041

evocateur:
As noted in #1043, related to changes originally made in #1015.

### DIFF
--- a/test/validation.js
+++ b/test/validation.js
@@ -369,46 +369,6 @@ describe('validation tests', () => {
         .argv
     })
 
-    it('fails when a required argument of type number is missing', (done) => {
-      yargs('-w')
-        .option('w', {type: 'number', requiresArg: true})
-        .fail((msg) => {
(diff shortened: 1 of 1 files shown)


test/validation.js
```diff
@@ -475,6 +425,69 @@ describe('validation tests', () => {
     })
   })
 
+  describe('requiresArg', () => {
+    it('fails when a required argument value of type number is missing', (done) => {
+      yargs()
+        .option('w', {type: 'number', requiresArg: true})
+        .parse('-w', (err, argv, output) => {
```
evocateur: I also changed all of these tests to use `parse(...)` instead of `fail().argv` because the `fail()` pattern takes 8 seconds to error when it fails, vs instantly in the `parse()` pattern.

bcoe:
@evocateur I believe we pulled these tests in in your other branch?

evocateur:
Yep!

PR #1050, opened 2018-01-18
fix: Set implicit nargs=1 when type=number requiresArg=true

evocateur:
This fixes the underlying issue in #1015 _without_ breaking nested commands. This PR is built on top of #1049.


evocateur:
The mac builds on Travis seem to be non-functional, wonder why yargs even needs that platform in the matrix?

bcoe:
@evocateur part of me wonders if we could just use `nargs` _for everything_, but going to land this and iterate.

PR #1053, opened 2018-01-20
Fix #969 by expanding context and adding a .yargsOptions function

### LINKED ISSUE #969: Easier way to disable type hints in output
There's a previous question about `[boolean]` in `--help` and `--version` #319, with this suggested workaround:

```
yargs.getOptions().boolean.splice(-2);
```

However, this feels like a workaround indeed :) How about some global switch like this?

```
yargs.disableTypesInOutput().etc;
```

Symbitic:
#969 was a feature request for a way to disable printing type information in the help message. Several maintainers were reluctant to add yet another toggle to the API. I solved this by creating a new function that can be reused for other toggles/value setters. It might even be possible to depreciate several toggles in the current API.

`.yargsOptions` is used for setting internal yargs options to change the default behavior. It allows the user to change properties of the internal `context` object and will throw an error if an unrecognized key is given.

The only key `.yargsOptions` will currently accept is `printTypes`. `printTypes` is a new property I added to `context`. When false, the options will not have their types printed in the help message.

I have added documentation and unit tests for `.yargsOptions` and `printTypes`.

If you do decide to keep `.yargsOptions`, things like `detectLocale`, `exitProcess`, and `strict` could be removed as functions and replaced with keys for `.yargsOptions`.

I choose the name yargsOptions because options and config were already taken. If you can think of a better name, go for it.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1334,3 +1334,12 @@ Format usage output to wrap at `columns` many columns.
 By default wrap will be set to `Math.min(80, windowWidth)`. Use `.wrap(null)` to
 specify no column limit (no right-align). Use `.wrap(yargs.terminalWidth())` to
 maximize the width of yargs' usage instructions.
+
+<a name="yargsOptions"></a>.yargsOptions(options)
+-------------------------------------------------
+
+Change the internal behaviour of yargs.
+
+`options` is an object. Valid keys include:
+
+- `printTypes`: Boolean indicating if types should be printed in the help message (Default = `true`).
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -271,6 +271,7 @@ module.exports = function usage (yargs, y18n) {
         const kswitch = switches[key]
         let desc = descriptions[key] || ''
         let type = null
+        const context = yargs.getContext()
 
         if (~desc.lastIndexOf(deferY18nLookupPrefix)) desc = __(desc.substring(deferY18nLookupPrefix.length))
 
@@ -282,7 +283,7 @@ module.exports = function usage (yargs, y18n) {
         if (~options.number.indexOf(key)) type = `[${__('number')}]`
 
         const extra = [
-          type,
+          context.printTypes ? type : null,
           (key in demandedOptions) ? `[${__('required')}]` : null,
           options.choices && options.choices[key] ? `[${__('choices:')} ${
             self.stringifiedValues(options.choices[key])}]` : null,
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -129,6 +129,64 @@ describe('yargs dsl tests', () => {
     expect(r.errors).to.deep.equal([])
   })
 
+  describe('yargsOptions', () => {
+    it('should show types in the help message by default', () => {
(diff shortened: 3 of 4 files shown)


bcoe:
@Symbitic I like this idea -- for edge-case configuration options such as this, it seems smart to move the behavior into config flags rather than cluttering up the API too much.

_however_, I've actually already started doing this for a few settings in the parser, see: https://github.com/yargs/yargs/blob/master/docs/advanced.md#customizing-yargs-parser; how about, rather than introducing the new `yargsOptions` helper, we instead extend on the already supported parser configuration, allowing hints to also be provided related to display.

Symbitic:
The problem with the parser solution is that it only works when there's a package.json file. In specific use cases (webpack, rollup, electron, pkg, etc) users will still need to change the behavior when the yargs code has already been bundled and is no longer a node module.

On the other hand, perhaps the two could be merged into a hybrid solution.

bcoe:
@Symbitic I like the idea of a hybrid solution that exposes the parser config programmatically and also allows a few display related settings to be toggled -- `yargsConfiguration` might perhaps be a good method name for this, in the library `yargs-parser` the method used is called `configuration`:

```js
var parsed = parser(['--no-dice'], {
  configuration: {
    'boolean-negation': false
  }
})
```

but, my concern is that this is too close to the existing `.config` method ... _thoughts?_

Symbitic:
That was my concern as well. That was why I picked the name yargsConfig, so that users would immediately know that it was for configuring the internal configuration of yargs parsing behavior. You are right, yargsConfiguration would probably be a better name. Some other names to consider might be `.parser()`, `.yargs()`, `.internal()`, `.extend()`, or `.configuration()`. Also possible is to change the yargs function itself to accept the last parameter as a configuration object, e.g. `yargs({})`, `yargs([ '--help' ], {})`.

Those are just some ideas. Either way, I wholeheartedly agree with your concern that the yargs API is becoming too bloated, which is why I think we need a standardized solution for configuring the internal behavior of yargs without dedicated API functions.

bcoe:
@Symbitic having slept on it, what if we went with `.configureYargs`? and this would be appended to (overriding if colliding) any configuration provided in the `package.json` for the parser -- but could also be used for configuring the view -- it should also be the case that providing configuration in the package.json can modify the new view configuration options.

Symbitic:
Fine with me. Do you want me to get started on another pull request or is this something you want to do yourself?

bcoe:
@Symbitic would love your help implementing this, if you're interested 👍 just wanted to chit chat about the cleanest API possible.

Symbitic:
@bcoe Sorry for my radio silence. Sure, I'll be willing to work on this. What do you want to use for chatting?

bcoe:
@Symbitic likewise, sorry for radio silence. I lost a loved one to cancer just over a month ago, and am just starting to get myself back on my feet, and dig myself out of an open-source backlog.

To sync up regarding this feature, there's a chat available here:

I think adding more configuration options to yargs is a good idea, and I'm excited to spec this work out with you.

PR #1054, opened 2018-01-20
feat: requiresArg is now simply an alias for nargs(1)

bcoe:
BREAKING CHANGE: requiresArg now has significantly different error output, matching nargs.

based on @evocateur's clever suggestion, `requiresArg` is now just an alias for `nargs(1)`; simplifying the implementation significantly and moving behavior closer to the parser.

### DIFF
--- a/yargs.js
+++ b/yargs.js
@@ -92,7 +92,7 @@ function Yargs (processArgs, cwd, parentRequire) {
     groups = {}
 
     const arrayOptions = [
-      'array', 'boolean', 'string', 'requiresArg', 'skipValidation',
+      'array', 'boolean', 'string', 'skipValidation',
       'count', 'normalize', 'number'
     ]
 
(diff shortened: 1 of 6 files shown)


yargs.js
```diff
@@ -204,7 +204,7 @@ function Yargs (processArgs, cwd, parentRequire) {
 
   self.requiresArg = function (keys) {
     argsert('<array|string>', [keys], arguments.length)
-    populateParserHintArray('requiresArg', keys)
+    populateParserHintObject(self.nargs, false, 'narg', keys, 1)
```
evocateur: Is calling `self.nargs(keys, 1)` too cute, here?

Both ways seem fine, the tradeoff being a duplicated call to `populateParserHintObject` vs a redundant `argsert` in the nested call...

bcoe: @evocateur it doesn't seem too cute; I like your idea of calling `self.nargs`.

evocateur: I was envisioning it being the entire contents of the method, not in addition to the duplicate argsert+populate call.

PR #1056, opened 2018-01-24
fix: choose correct config directory when require.main does not exist

evocateur:
So, it turns out Jest overrides the module-local `require` and neglects to assign `require.main.filename` (https://github.com/facebook/jest/issues/2150). Maybe I'll even get around to fixing it [here](https://github.com/facebook/jest/blob/be66ccc26ea80e825f2fb4f0f21433832a2c4d97/packages/jest-runtime/src/index.js#L675-L689), but in the meantime, we can work around it by avoiding the "extra" `path.dirname()` when it isn't neccessary (because `require-main-filename` defaults to `process.cwd()`).

Also, I changed a bunch of function parameters to avoid shadowing `path`, it was extremely confusing (and ridiculous to have an inner `require('path').dirname` when `const path = require('path')` is already present at the top of the file).

### DIFF
--- a/lib/apply-extends.js
+++ b/lib/apply-extends.js
@@ -6,9 +6,9 @@ const YError = require('./yerror')
 
 let previouslyVisitedConfigs = []
 
-function checkForCircularExtends (path) {
-  if (previouslyVisitedConfigs.indexOf(path) > -1) {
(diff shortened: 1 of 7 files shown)


evocateur:
This fixes the (untested) fix from #1036, which exposed jest's nefarious activity.

PR #1057, opened 2018-01-24
Add documentation for array()

Nautigsam:
Add some useful documentation for when using array() with positional()

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -63,7 +63,7 @@ so `--foo foo --foo bar` will be parsed as `['foo', 'bar']`
 
 When the option is used with a positional, use `--` to tell `yargs` to stop adding values to the array.
 
-For example: `-foo foo bar -- val` will be parsed as
+For example: `--foo foo bar -- val` will be parsed as
 ```javascript
 {
   foo: ['foo', 'bar'],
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -61,6 +61,16 @@ Tell the parser to interpret `key` as an array. If `.array('foo')` is set,
 Also, if you use the option multiple times all the values will be flattened in one array
 so `--foo foo --foo bar` will be parsed as `['foo', 'bar']`
 
+When the option is used with a positional, use `--` to tell `yargs` to stop adding values to the array.
+
+For example: `-foo foo bar -- val` will be parsed as
```
bcoe: this should be:

`--foo foo bar -- val`

docs/api.md
```diff
@@ -61,6 +61,16 @@ Tell the parser to interpret `key` as an array. If `.array('foo')` is set,
...
+```javascript
+{
+  foo: ['foo', 'bar'],
```
bcoe: I believe this output would actually look something like:

```js
{ _: [ 'val' ],
  foo: [ 'foo', 'bar' ],
}
```

👍

PR #1061, opened 2018-01-27
feat: --show-hidden

laggingreflex:
Adds a `--show-hidden` flag to show options defined with `hidden: true` property

Recap: (#962) Earlier, a fix was made to address the issue where `.options({...})` without a `description` property did not show up up in `--help`. The fix introduced a new `hidden: true` property to truly hide an option from `--help`, and to show all other options (even without `description` property)

Issue with the fix: There was no way to show options defined with `hidden:true` property (they were hidden permanently)

Solution (this patch): This introduces a new flag: `--show-hidden` (pre-configured) which can be used to show those hidden options.

The default key for this new flag may also be custom configured with a new function property `.showHidden('custom-show-hidden')`

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -576,6 +576,11 @@ Describe a `key` for the generated usage information.
 
 Optionally `.describe()` can take an object that maps keys to descriptions.
 
+<a name="hide"></a>.hide(key)
+--------------------
+
+Hides a `key` from the generated usage information. Unless a `--show-hidden` option is also passed with `--help` (see [`showHidden()`](#showHidden)).
+
 .detectLocale(boolean)
 -----------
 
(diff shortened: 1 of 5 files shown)


Morishiri:
The code looks fine, but maybe you could write some UT for that?

laggingreflex:
@Morishiri done

laggingreflex:
@bcoe [done](https://github.com/laggingreflex/yargs/blob/af62bb6af12b00d8fd1b4e7be78fe60feb325433/docs/api.md#showHidden)

kopax:
Hi, I was looking for such feature. I am using 11.0.0, I presume this have not been release yet right ? 
When will it be?

laggingreflex:
[It's released in 11.1.0](../blob/master/CHANGELOG.md#1110-2018-03-04) but there seems to be [an issue](https://github.com/yargs/yargs/issues/1128) regarding latest npm release. 

Try installing/setting 11.1.0 version manually.

gajus:
> Hi, I was looking for such feature. I am using 11.0.0, I presume this have not been release yet right ?
When will it be?

https://github.com/laggingreflex/yargs/blob/master/docs/api.md#showhiddenoption-description

laggingreflex:
You also need to execute [`.showHidden()`](https://github.com/yargs/yargs/blob/master/docs/api.md#showHidden)

kopax:
@laggingreflex so we need to get thooe command documented to everybody?

laggingreflex:
@kopax I'm not sure I understand, could you please elaborate with an example.

kopax:
@laggingreflex 

Sure for example this use case:

I want to have a `--change-copyright-holder` options within a script. I am the only one who can do that, it's my license. I want to keep it hidden to everyone, using the `hidden: true` on my option, I can do it

I want to keep the `--show-hidden` for my self, doing [`.showHidden()`](https://github.com/yargs/yargs/blob/master/docs/api.md#showHidden) will inform users of such hidden option.

Issue #1010, opened 2017-11-14
Question: Filename appearing next to command in help menu


aa7955:
Just a quick question. I am not sure if this is a NodeJS issue or what, but when I run the example yargs code, I get the following in my terminal:

```
example.js <cmd> [args]

Commands:
  example.js hello [name]  welcome ter yargs!

Options:
  --version  Show version number                                       [boolean]
  --help     Show help                                                 [boolean]
```

In the example, the filename is not next to the command `hello` under the `Commands` section. Just wondering how to get the filename out of that section so that it looks like the one from the example:

```
  Commands:
    hello [name]  welcome ter yargs!

```

Morishiri:
It was changed in https://github.com/yargs/yargs/pull/983. It is replacing every `$0` to filename. I guess it cannot be configured not to.

ajalger:
Thanks @Morishiri. This is definitely a fugly change, especially if you print out commands that need more than one input ... arg

Morishiri:
We could actually do something about it, @bcoe can you please drop your opinion on this as I think it should work as it worked before.

bcoe:
We opted to start placing the filename in the command output, because it helps better clarify what's happening when you have top-level positional arguments configured, which in the past would have looked like:

```
Commands:

[foo] <bar>
```

☝️ where you have positional arguments with nothing bound to them. I think that by placing the actual bin being executed as part of the command description we provide clearer help output for applications that have top-level commands configured.

ajalger:
I understand that @bcoe. Just wondering if there is a way to expand the help table then because I am getting arguments on multiple lines like:
```
apiUsersTeamsRoles.js add-user            Add user to an account
  <accountid> <name> <email> [phone]
```

bcoe:
@ajalger that's a pretty long application name and it takes a lot of arguments, even if we exposed an API for increasing the left-hand column width, I think you'd be making the right-hand column look pretty ugly, because you'd end up with something that looks like this:

```
Add 
user to
an account
```

What if you had an alternate bin name that's shorter, like `autr`; and we exposed an API for changing `$0` if we don't already (you might just be able to overwrite `$0`on the yargs object0>

ajalger:
@bcoe yeah this was just an example command-line app. Indeed the filename is longer than expected because it was just used to differentiate one file from another. I wouldn't do that for something that is meant for production.

I'll experiment a little and come up with something that works. Thanks a lot.

bcoe:
[...]

PR #1073, opened 2018-02-19
Remove file extension on usage output

WeeJeWel:
Before, at least on Windows, the `Commands:` section included the file extension, e.g.:

```
C:\Users\emile>athom
athom.js <command>

Commands:
  athom.js app      App related commands
```

With this change, the `.js` will be automatically stripped:

```
C:\Users\emile>athom
athom <command>

Commands:
  athom app      App related commands
```

### DIFF
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -147,7 +147,7 @@ module.exports = function usage (yargs, y18n) {
     normalizeAliases()
 
     // handle old demanded API
-    const base$0 = path.basename(yargs.$0)
+    const base$0 = path.basename(yargs.$0, path.extname(yargs.$0))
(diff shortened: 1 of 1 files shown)


bcoe:
@WeeJeWel I'm on the fence about this; if `athom.js` is set as executable, then the command `athom.js foo` is a functional operation that someone can type in and execute (whereas `athom app` is not).

There's an ongoing conversation happening around how to shorten command descriptions [here](https://github.com/yargs/yargs/issues/1010); I would love your input, maybe we can come to a compromise that makes folks happy.

bcoe:
opted for this solution, which gives even more control over the string displayed in help output:

https://github.com/yargs/yargs/pull/1143

PR #1080, opened 2018-02-26
extend *.rc files in addition to json

### LINKED ISSUE #1035: "extends" should also trigger for files without extension that end in "rc"
It's a common convention for a configuration file to end with the suffix `rc`., e.g., `.nycrc`; it seems like a good extension to the `extends` functionality would be to also support filenames of the form `.*rc`.

Currently `extends` only supports the `.json` extension. fixes https://github.com/istanbuljs/nyc/issues/732

CC: @genepaul

genepaul:
fixes #1035 

This simply updates the regex to include *.*rc files. It does not handle the situation with rc files that are formatted in INI format. That seems to be a separate discussion.

### DIFF
--- a/lib/apply-extends.js
+++ b/lib/apply-extends.js
@@ -21,7 +21,7 @@ function applyExtends (config, cwd) {
 
   if (config.hasOwnProperty('extends')) {
     if (typeof config.extends !== 'string') return defaultConfig
-    const isPath = /\.json$/.test(config.extends)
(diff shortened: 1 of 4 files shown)


genepaul:
I tested this locally with a project that had been broken before because it was using `.nycrc` to extend, and this change fixes it.

bcoe:
@genepaul this seems like a great addition to me, thanks for the contribution.

PR #1095, opened 2018-03-22
Fix #1069

### LINKED ISSUE #1069: `yargs.parse(argv,callback)` issues call back before handler promise is finished.
yargs v11.0.0

When using yargs.parse() with a callback, expected behaviour is for callback to be called after yargs parse has completed. It is also expected that command handlers accept promises in their returns.

Example code:
```javascript

yargs.command({
    command: 'test',
    description: 'a test command',
    handler: () => {
        return new Promise((resolve) => {
            setTimeout(() => {
                console.log('[Promise] finally executed');
                resolve('some result');
            }, 3000);
        });
    }
});

const callback = ()=>{
    console.log('[Yargs] Parse finished');
};

yargs.parse('test', (err, argv, output) => {
    callback(err, output);
});

console.log('[File] End');
```

outputs:
```
$ node index.js
[Yargs] Parse finished
[File] End
[Promise] finally executed 
```
expected:
```
[File] End
[Promise] finally executed 
[Yargs] Parse finished
```

Context: I'm using yargs inside REPL and need the callback to fire _after_ the command has finished processing, so REPL can continue correctly.

Possibly linked issues:
#510 : Adds in the Async handler support, but does not halt `parse` execution or pass promise back to user.
#918 : possibly a feature request for this?
#564 : Unanswered question on how to achieve this behaviour in yargs (though link provided to user space workaround/patch)

Workaround is discussed in #918 by attaching your promise to the argv and then checking it in the callback.

I would say that the callback in `yargs.parse` being called before command finishes could be seen as a bug, unless documentation is written to specify that it is expected behaviour and that async handlers will _not_ halt execution of `parse`'s callback.

nyamadori:


nyamadori:
I closed this PR because this is mistake!

PR #1104, opened 2018-04-04
chore: 'false' value is no longer set when boolean flag is not passed

bcoe:
BREAKING CHANGE: yargs-parser does not populate 'false' when boolean flag is not passed

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "string-width": "^2.0.0",
     "which-module": "^2.0.0",
     "y18n": "^3.2.1",
-    "yargs-parser": "^9.0.2"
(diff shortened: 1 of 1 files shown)


bcoe:
@evocateur I'm going to work on having a patched version of yargs out soonish with this functionality.

evocateur:
Great!

bcoe:
@evocateur sorry this took a little while to release ... when you have a couple seconds mind trying `npm i yargs@next` and making sure this works properly with lerna, I'll promote the release soon.

evocateur:
Will do, thanks!

PR #1105, opened 2018-04-05
feat: remove `setPlaceholderKeys`

### LINKED ISSUE #1071: Make `setPlaceholderKeys` optional
The `setPlaceholderKeys` function has been added in #116. It sets all the options not set by the user in the CLI to `undefined`. This problematic the options object returned by `yargs` get merged with another one.
For example if you have an options object coming from a config file or somewhere else, and you want to merge it with `Object.assign` to the `yargs` options in a way that `yargs` options take precedence over the other object values.

Typically that would be a situation in which a CLI allows users to override config parameter via CLI arguments. It seems to a very common pattern.

In such situation all the flags not passed by the user (so the user does not want to override the config file) end up being set to `undefined` and override the config file value.

The workaround is to filter out the `yargs` options properties with a value equal to`undefined`.

I'm not really sure what was the need behind the implementation of `setPlaceholderKeys`, but as a result it forces to post-process `yargs` options (removing `undefined` values) to implement the very common pattern of "setting CLI arg to override config options".
 
Maybe the call to `setPlaceholderKeys` could be enable/disable with a `yargs` option?

Even better if this option, disable it by default, unless there is a good use case for setting placeholder to `undefined` that I don't see.

Would you be open to PR ?

See also yargs/yargs-parser#116 which is also related to implementation of the pattern "setting CLI arg to override config options"

Not that it matters much but this is the way `optimist` works, it doesn't create `undefined` placeholders.
```js
optimist.option('foo', {type: 'string'}).parse([])
{ _: [], '$0': '...' } // => no foo property

yargs.option('foo', {type: 'string'}).parse([])
{ _: [], '$0': '...', foo: undefined } // => foo set to undefined
```

pvdlg:
Fix #1071

Do no add `undefined` properties in parsed result object for options absent from provided `argv`.

BREAKING CHANGE: Options absent from `argv` (not set via CLI argument) are now absent from the parsed result object rather tahn being set with `undefined`

### DIFF
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -40,12 +40,12 @@ describe('yargs dsl tests', () => {
     argv.cat.should.eql(33)
   })
 
-  it('populates argv with placeholder keys for all options', () => {
+  it('do not populates argv with placeholder keys for unset options', () => {
     const argv = yargs([])
       .option('cool', {})
       .argv
 
-    Object.keys(argv).should.include('cool')
+    Object.keys(argv).should.not.include('cool')
   })
 
   it('accepts an object for implies', () => {
@@ -70,7 +70,6 @@ describe('yargs dsl tests', () => {
       )
 
     r.errors[0].should.match(/really cool key/)
-    r.result.should.have.property('x')
     r.result.should.not.have.property('[object Object]')
   })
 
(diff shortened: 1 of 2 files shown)


bcoe:
@pvdlg mind trying `npm i yargs@next` and letting me know if the parsing behavior is working the way you'd expect?

pvdlg:
Yes, it works as expected! Thanks!

pvdlg:
Don't you want to include a version of `yargs-parser` that includes yargs/yargs-parser/pull/123 though?

bcoe:
@pvdlg done, and published to `latest` no.

pvdlg:
Awesome, thanks!

PR #1118, opened 2018-04-27
PR-AN-06: Refactor yargs.js into files by method.

aorinevo:
## Summary of Changes
- Moved the following methods into their own files under /lib/api
  - Phase 1
    - boolean
    - array
    - number
    - normalize
    - count
    - string
    - skipValidation
    - requiresArg
    - nargs
    - choices
    - alias
    - demandOption
    - coerce
  - Phase 2
    - command
    - commandDir
    - usage
    - config
    - require (demand, required)

- Moved the following test suites from /test/yargs.js to separate files under /test/api
  - Phase 1
    - normalize
    - number
    - skipValidation
    - nargs
    - choices
    - coerce
  - Phase 2
    - command
    - config

- Updated test script to use glob to match against any `.spec.js` extension inside test directory
- Update package.json to run specs matching `./test/*.spec.js` and `./test/**/*.spec.js`

## Impact
- Decreased the size of yargs.js by 222 lines
- Decreased the size of test/yargs.js by 669 lines
- Increased separation of concerns


aorinevo:
Refactor corresponding tests into their own files (i.e. /test/api/...)

aorinevo:
AppVeyor is failing but on `npm run test` but on my local it passes (see screenshot).

@bcoe Any idea why this is happening?

screen shot 2018-04-29 at 11 12 17 am

aorinevo:
The main objective behind this PR was to reduce the size of `yargs.js` and `yargs.spec.js`, by breaking-up `yargs` and `yargs.spec.js` into separate files; one file per api method.  Yargs would then be the orchestrator of the different apis, with all the code abstracted away from yargs.js.

This PR is a partial solution.  It would be challenging to do this for all of yargs' apis.  Some rely on variables - primitives that are not passed by reference - higher up in the scope chain, making them nearly impossible to refactor in the same way without taking on much larger refactoring steps.

I do like the idea of using factories to generate the different yargs methods; that's what I ended up doing for the middleware api in PR https://github.com/yargs/yargs/pull/1119.

That being said, I agree that there is a performance hit here with the extra require calls.

package.json
```diff
@@ -16,6 +16,7 @@
     "decamelize": "^1.1.1",
     "find-up": "^2.1.0",
     "get-caller-file": "^1.0.1",
+    "npm": "^2.15.12",
```
evocateur: I don't think yargs needs a dependency on an outdated version of npm?

bcoe:
closing this in favor of the refactor we discussed during our monthly triage meeting.

PR #1126, opened 2018-05-10
fix: .argv and .parse() now invoke identical code path

bcoe:
* library updates to address `npm audit` issues.
* fix inconsistency between `yargs.argv` and `yargs.parse()`, which was causing issues with tests.
* fix 1,000,000 linting issues for new standard.


kaseyshue3-hash:
#2509

PR #1129, opened 2018-05-13
allow access origin input args via `yargs.processArgs`

bluelovers:
```ts
yargs.processArgs
```

this is help if we wanna do somthing follow origin argv order

### DIFF
--- a/yargs.js
+++ b/yargs.js
@@ -33,6 +33,8 @@ function Yargs (processArgs, cwd, parentRequire) {
     updateFiles: false
   })
 
+  self.processArgs = processArgs.slice()
+
   self.middleware = middlewareFactory(globalMiddleware, self)
 
   if (!cwd) cwd = process.cwd()


bluelovers:
if anyone wanna use this, before merge into master

can see here

https://github.com/bluelovers/yargs2

yargs.js
```diff
@@ -33,6 +33,8 @@ function Yargs (processArgs, cwd, parentRequire) {
     updateFiles: false
   })
 
+  self.processArgs = processArgs.slice()
```
evocateur: What problem does this solve? You don't have access to `process.argv`?

bluelovers: processArgs is from args pass by Yargs (processArgs, cwd, parentRequire)

this is not process.argv

bcoe:
hey 👋@bluelovers, echoing @evocateur's request, mind sharing an example use-case for this feature.

bluelovers:
@bcoe https://github.com/bluelovers/yargs2/blob/master/README.md

evocateur:
Literally nowhere in that scattershot README does it explain _why_ this is a useful feature.

bcoe:
@bluelovers 👋 sorry for the slow turnaround on this thread, have been working on a few other open-source projects and not concentrating on yargs recently.

I see partially where you're coming from, but if you're passing a value other than `process.argv` into yargs, couldn't you take this approach:

```
const args = ['--apple', '99']
const parsed = yargs.parse(args)
```

rather than exposing an additional parameter on the yargs object? I'm going to close this for now, _but_ if you come back with a strong argument for why this is an important use-case to solve, will definitely consider landing this.

I'm just a curmudgeon these days about landing many near features in yargs.

PR #1138, opened 2018-05-24
chore: Allow v4 of y18n

iarna:
All v4 of y18n does is drop support for 0.10 and 0.12.  Kept a broad range (3 & 4) to make it easier to dedupe trees.


evocateur:
Would `||` ranges specified in `dependencies` _usually_ be preferable to the same ranges specified in `peerDependencies`? (allowing for some cases, like plugins and singletons, where `peerDependencies` seems to have an edge)

PR #1139, opened 2018-05-26
docs: Use correct completion callback signature

matthewdias:
The completion function checks that the number of arguments is less than 3 before performing the logic that resolves a promise, so passing `done` here will cause the promise to not resolve.


evocateur:
Thanks!

PR #1151, opened 2018-06-13
I-1150: Add Redux for State Management

aorinevo:
- Add Redux for State Management


aorinevo:
Bumped major version to v12.0.0

package.json
```diff
@@ -43,7 +44,7 @@
   },
   "scripts": {
     "pretest": "standard",
-    "test": "nyc --cache mocha --require ./test/before.js --timeout=8000 --check-leaks",
+    "test": "nyc --cache mocha --require ./test/before.spec.js ./test/*.spec.js ./test/**/*.spec.js --timeout=8000 --check-leaks",
```
evocateur: `test/before.js` isn't a test, it's a setup file. Why were all the test files renamed, anyway? I really don't think that's appropriate for this PR.

reducers/context.js
```diff
@@ -0,0 +1,29 @@
+const {GET_CONTEXT, INCREMENT_CONTEXT_RESETS, ADD_FILE_TO_CONTEXT} = require('../actions/context')
+const initialState = { resets: -1, commands: [], fullCommands: [], files: [] }
+
+function getContext (state = initialState) {
+  return state
```
evocateur: Again, not clear why `getContext` exists here, seems like a leftover from previous refactoring.

test/reducers/context.spec.js
```diff
@@ -0,0 +1,17 @@
+'use strict'
+/* global describe, it */
+
+require('chai').should()
+const expect = require('chai').expect
+
+describe('context reducer', () => {
+  const { getContext, incrementContextResets } = require('../../actions/context')
+  const contextReducer = require('../../reducers/context')
+  it('should get context', () => {
```
evocateur: Given that the `default:` case of reducers just returns the state, seems like we really don't need the "get" actions at all.

package.json
aorinevo: I added folders to the test directory to organize specs for actions and reducers into their own folders.  Adding .spec was needed to target spec files and excluding test helper files.

test/reducers/context.spec.js
aorinevo: Yep, agreed, there's no need.

reducers/context.js
aorinevo: I'll split context related changes into a separate branch.

package.json
aorinevo: I can break this out into a separate PR but I need it here for now because of the change in folder structure under the test directory.

actions/context.js
```diff
@@ -0,0 +1,31 @@
+const GET_CONTEXT = 'GET_CONTEXT'
+const INCREMENT_CONTEXT_RESETS = 'INCREMENT_CONTEXT_RESETS'
+const ADD_FILE_TO_CONTEXT = 'ADD_FILE_TO_CONTEXT'
+
+// action creators
+function getContext () {
+  return {type: GET_CONTEXT}
```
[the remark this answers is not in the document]
aorinevo: Neither is needed and can be removed.  If user needs to get either, they should get them using `getState().context` or `getState().middleware`.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "yargs",
-  "version": "11.1.0",
+  "version": "12.0.0",
```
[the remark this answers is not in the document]
aorinevo: I'll remove in my the next commit.

lib/usage.js
```diff
@@ -6,6 +6,8 @@ const objFilter = require('./obj-filter')
 const path = require('path')
 const setBlocking = require('set-blocking')
 const YError = require('./yerror')
+const {setShowHelpOnFail, setFailureOutput, setUsageEpilog, setUsageDisabled, resetUsage, freezeUsage, unfreezeUsage} = require('../actions/usage')
```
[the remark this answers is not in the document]
aorinevo: Yep.

lib/usage.js
```diff
@@ -17,21 +19,12 @@ module.exports = function usage (yargs, y18n) {
...
-      enabled = true
-    }
-    failMessage = message
-    showHelpOnFail = enabled
+    store.dispatch(setShowHelpOnFail(enabled, message))
     return self
   }
 
-  let failureOutput = false
+  store.dispatch(setFailureOutput(false))
```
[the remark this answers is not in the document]
aorinevo: Yep, but the value needs to be explicitly set to false here.

package.json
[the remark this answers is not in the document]
aorinevo: Removed.

reducers/context.js
aorinevo: Done (see `aorinevo/I-1150_master_AN_redux_context`)

test/reducers/context.spec.js
aorinevo: Done

actions/middleware.js
```diff
@@ -0,0 +1,31 @@
...
+
+function resetMiddleware () {
+  return {
+    type: RESET_MIDDLEWARE
+  }
+}
+
+// action types
+module.exports = {
+  getMiddleware,
```
aorinevo: No need for `getMiddleware` since this is equivalent to `store.getState().middleware`

actions/middleware.js
```diff
@@ -0,0 +1,31 @@
+const GET_MIDDLEWARE = 'GET_MIDDLEWARE'
```
aorinevo: Remove

package.json
aorinevo: Reverted filename for before.spec.js to before.js.  Kept .spec.js for all spec files.

actions/middleware.js
aorinevo: Done

actions/middleware.js
aorinevo: Done

PR #1153, opened 2018-06-17
I-1120: Add webpack with babel

aorinevo:
- Add webpack with babel


dist/bundle.js
```diff
@@ -0,0 +1,15 @@
...
+/*!
+ * The buffer module from node.js, for the browser.
```
evocateur: Why are we versioning the build output? Are we trying to support a browser, or Node?

aorinevo: Added `target: 'node'` to `webpack.config.js` to prevent buffer from getting bundled.

aorinevo:
Adding a transpiler makes sense for apps targeting multiple browsers but since yargs is able to target specific versions of Node.js, we can simply drop Node versions over time.  We would also end up adding too much complexity for small gains.

Closing this PR.

PR #1163, opened 2018-06-23
Corrected version specifier in help example

kunagpal:
The name of the flag: `version` and the actual version value: `1.0.1` were switched.


bcoe:
@kunagpal thank you!

PR #1164, opened 2018-06-24
I-1120: Add support for ES6 using Rollup.

aorinevo:
- Add support for ES6 using Rollup.
- export bundle to `dist/bundle.js`

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,8 @@
     "os-locale": "^2.0.0",
     "require-directory": "^2.1.1",
     "require-main-filename": "^1.0.1",
+    "rollup": "^0.61.2",
+    "rollup-plugin-babel": "^3.0.5",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -19,13 +19,18 @@
     "os-locale": "^2.0.0",
     "require-directory": "^2.1.1",
     "require-main-filename": "^1.0.1",
+    "rollup": "^0.61.2",
```
aorinevo: Move rollup and rollup-plugin-babel to devDependencies.

aorinevo:
Adding a transpiler makes sense for apps targeting multiple browsers but since yargs is able to target specific versions of Node.js, we can simply drop Node versions over time.  We would also end up adding too much complexity for small gains.

Closing this PR.

Issue #1134, opened 2018-05-21
middleware does not work


vitalets:
Example from the [docs](https://github.com/yargs/yargs/blob/master/docs/api.md#middlewarecallbacks):
```js
const mwFunc1 = argv => console.log('I\'m a middleware function');
const mwFunc2 = argv => console.log('I\'m another middleware function');

require('yargs')
  .command('myCommand', 'some command', {}, function(argv){
    console.log('Running myCommand!');
  })
  .middleware([mwFunc1, mwFunc2]).argv;
```
Output:
```
  .middleware([mwFunc1, mwFunc2]).argv;
   ^
TypeError: require(...).command(...).middleware is not a function
```

Env:
```
npm info yargs version
11.0.0

node -v
v8.5.0
```

aorinevo:
Hi vitalets,

This code isn’t in v11.0.0.  Try v11.1.0.

vitalets:
Hi @aorinevo 

The same thing with 11.1. To be sure that correct version is used:
```
> npm i yargs@11.1.0
+ yargs@11.1.0
added 19 packages from 5 contributors in 5.103s

> npm list | grep yargs
└─┬ yargs@11.1.0
  └─┬ yargs-parser@9.0.2
```

Also npm is showing the latest available version is 11.0.0 not 11.1, that is strange:
![image](https://user-images.githubusercontent.com/1473072/40319313-69423674-5d30-11e8-87ba-68d327afaf93.png)

aorinevo:
Looks like the code is in master but not in any of the tagged versions.  This should be available in the next release.  An option to start using the latest features in master is to namespace yargs in npm off of the master branch and pull that into your project.  That’s what I am doing (see @aorinevo/yargs on npm) until the next release of yargs.

styfle:
Related to #1128

jaguilar87:
Any update on this? I'd really like to see middlewares in a release 🙏

aorinevo:
@jaguilar87 if you need to use the feature now, you can publish and install a namespaced version. 

Here are the steps:
  - Fork and clone the project from Github
  - In package.json:
      - change the `name` property from yargs to `{{namespace}}/yargs` (i.e. `@aorinevo/yargs` in my case)
      - change `repository.url` property to the git address of the forked project
  - Publish the project to npm (see https://docs.npmjs.com/getting-started/publishing-npm-packages)
  - Within a few minutes, the package should be available to install using `npm install {{namespace}}/yargs`

[...]

PR #1165, opened 2018-06-26
Readme: add info about installing @next version, see #1134

vitalets:


bcoe:
@vitalets awesome, thank you for the contribution; `next` worked for you as expected?

vitalets:
> @vitalets awesome, thank you for the contribution; next worked for you as expected?

Yes. Example from #1134 works correct on `yargs@next` and doesn't work on `yargs`.

PR #1166, opened 2018-06-26
Fix breaking change note in CHANGELOG.md

rwjblue:
Earlier in the changelog for 12.0.0 it mentions that Node 6 was _added_ to the testing matrix, then under breaking changes it says the opposite. I _believe_ that the breaking change section intended to say that Node 4 was dropped (it is EOL after all) not Node 6.

This will also need to be updated in the GitHub release (which is used by tolls like greenkeeper / renovate / dependabot/ etc).


bcoe:
@rwjblue great catch, thank you.

PR #1168, opened 2018-06-27
Bump decamelize & find-up

LinusU:
The only breaking changes is that support for Node.js 4.x is dropped. Since Yargs currently only supports Node >= 6 this shouldn't be a problem.


LinusU:
This will make yargs definitievly not work on Node.js 4, see #1169 as well. I'm guessing that that was the intention with 12.x though? @bcoe

PR #1169, opened 2018-06-27
Bump engines to Node.js 6.x

LinusU:
This change was mentioned in the release notes for 12.0.0, so it shouldn't be a breaking change...


bcoe:
@LinusU thank you 👍

PR #1182, opened 2018-07-15
I 1150 master an redux command

aorinevo:
- Migrate command.js state management to Redux: handlers, aliasMap, and defaultCommand
- Decoupled usage module from command.js; same could be done with validation.js

### DIFF
--- a/actions/command.js
+++ b/actions/command.js
@@ -1,12 +1,9 @@
+const RESET_COMMAND = 'RESET_COMMAND'
+const FREEZE_COMMAND = 'FREEZE_COMMAND'
+const UNFREEZE_COMMAND = 'UNFREEZE_COMMAND'
 const ADD_COMMAND_HANDLERS = 'ADD_COMMAND_HANDLERS'
-const ADD_COMMAND = 'ADD_COMMAND'
-
-function addCommand (value) {
-  return {
-    type: ADD_COMMAND,
-    value
-  }
-}
+const SET_DEFAULT_COMMAND = 'SET_DEFAULT_COMMAND'
+const SET_ALIAS_MAP = 'SET_ALIAS_MAP'
 
 function addCommandHandlers (parsedCommandCmd, cmd, description, handler, builder = {}, middlewares = [], demanded, optional) {
   const {spawnSync} = require('child_process')
@@ -18,7 +15,7 @@ function addCommandHandlers (parsedCommandCmd, cmd, description, handler, builde
     type: ADD_COMMAND_HANDLERS,
     key: parsedCommandCmd,
     value: {
-      cmd,
+      original: cmd,
       description,
       handler,
       builder,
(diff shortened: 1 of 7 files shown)


bcoe:
@timdorr sorry for reaching out on Twitter, and then taking _forever_ to actually loop you in.

This is the refactor that @aorinevo has been working on, the goal being to potentially simplify state management in yargs.

Would love the advice from a few folks in the Redux community:

* for a problem domain like yargs, is Redux the right tool to simplify our state management? some problems we have:
  * commands are a breadth-first traversal; so, sometimes if a failure occurs we need to roll back to the prior parsing state.
  * there's some coupling in our class hierarchy, e.g., sometimes validation requires output to the terminal, or rolls back the parsing state.

My concerns around Redux are:

* perhaps we've picked the wrong tool for the job, and it doesn't ultimately simplify working with yargs significantly.
* perhaps we pick the right tool, but don't use it in the best way possible to reduce the complexity of yargs' codebase.
* does the size of the yargs library increase significantly, as a side-effect of pulling in Redux?

timdorr:
Sorry for taking so long to get back to you.

I'm also unsure if Redux is the right choice for yargs. It's normally used for longer-lived applications, particularly SPAs on the web. To be fair, I'm not super-knowledgeable on the challenges that a CLI library faces, but it would seem to be a lexing/parsing problem, not a state management issue.

Redux itself is less of a library, and more a code pattern that has been implemented via a library. It's a fancy event emitter. That wouldn't seem to align well with your problem space. 

An interesting idea (and this is veering straight into Computer Science Land) is algebraic effects. I know this is super-abstract, but [this MS paper on them (page 5, sec 2.6)](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/08/algeff-tr-2016-v2.pdf) goes over an example of using them for parsing. The idea is that your lexing and parsing code remains super-simple, but emits effects as it goes that a handler can interpret for sanity checks (duplicate/conflicting args, wrong types, etc). There isn't a library for this stuff, but you can use generators to achieve it. Basically, you would build the parser function as a generator and `yield` "effects" as it parses to a handler function that would check for problems. The parser doesn't need to know about context and the handler doesn't need to know about parsing. Nice and simple.

Anyways, that's just what was on my mind. Again, I don't get the sense that Redux is the right tool for the job here. I don't have a good level of insight into your code or problem domain, so I could definitely be wrong 🤷‍♂️

bcoe:
let's reopen this issue when/if we pick up this work; currently I think it's good to concentrate more on maintenance, and this has gotten a little bit behind.

aorinevo:
Using algebraic effects can help with things like middleware - both at the command and global level - but the underlying issue around state management still persists.  Perhaps a combination of a "redux-lite" state management and algebraic effects is the right solution.  In which case, I'd still advocate for using redux.

@bcoe you have a better understanding of how state is maintained and leveraged throughout yargs.  Can you add some details on this?

@timdorr, great input!  Using algebraic effects is an interesting idea.  Super simple to implement too.

bcoe:
If you don't mind, rather than an open pull request at this point, let's move this to discussion on an issue .. I'm thinking a good start might be thinking about how we split out validation into its own re-usable module; folks have asked to use this in isolation, similarly to how we have `yargs-parser` isolated, might be a good place to start with refactoring...

Don't have super strong opinions at this point, but I do agree with @timdorr that redux might not be the right hammer -- I find the paper he pointed to interesting, almost sounds similar to a SAX XML parser.

aorinevo:
@bcoe: Will do; closing this PR.

Issue #1101, opened 2018-04-03
Locale not work in conflicts error message


FlandreDaisuki:
Here is my code:
```js
#!/usr/bin/env node
const yargs = require('yargs');

const argv = yargs
  .command('add [items...]', '',
  (yargs) => {
    yargs
      .options({
        'i': {
          alias: 'interactive',
          type: 'boolean',
        },
        'y': {
          alias: 'yes',
          type: 'boolean',
        },
        'n': {
          alias: 'no',
          type: 'boolean',
        },
      })
      .conflicts('y', 'n');
  }, (argv) => {
    console.log(argv);
  })
  .argv;
```
and output:
```
$ locale
LANG=zh_TW.utf8
LANGUAGE=
LC_CTYPE="zh_TW.utf8"
LC_NUMERIC="zh_TW.utf8"
LC_TIME="zh_TW.utf8"
LC_COLLATE="zh_TW.utf8"
LC_MONETARY="zh_TW.utf8"
LC_MESSAGES="zh_TW.utf8"
LC_PAPER="zh_TW.utf8"
LC_NAME="zh_TW.utf8"
LC_ADDRESS="zh_TW.utf8"
LC_TELEPHONE="zh_TW.utf8"
LC_MEASUREMENT="zh_TW.utf8"
LC_IDENTIFICATION="zh_TW.utf8"
LC_ALL=

$ ./cli.js add -yn


選項：
  --help             顯示說明                                             [布林]
  --version          顯示版本                                             [布林]
  -i, --interactive                                                      [布林]
  -y, --yes                                                              [布林]
  -n, --no                                                               [布林]

Arguments y and n are mutually exclusive
```

bcoe:
@FlandreDaisuki thanks for pointing this out! my guess would be that this translation might be missing from the `zh_TW` translation file. Any interest in adding the translation?

https://github.com/yargs/yargs/blob/master/locales/zh_TW.json

would love your help.

FlandreDaisuki:
The translation of exclusive arguments existed.
https://github.com/yargs/yargs/blob/6ab6a95de6358f8e42466c424567a28ed0bd45cf/locales/zh_TW.json#L39
https://github.com/yargs/yargs/blob/fc13dcd803375f3acd3df79380d8d3ecb20a881b/locales/en.json#L39

jjoselv:
This may be that is declared like 
```javascript
usage.fail(__(`Arguments ${key} and ${value} are mutually exclusive`))
```

[...]

PR #1183, opened 2018-07-17
fix: translation not working when using __ with a single parameter

jjoselv:
Closes yargs/yargs#1101

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules/
 test.js
 coverage
 package-lock.json
+.npmrc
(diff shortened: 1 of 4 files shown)


nexdrew:
@jjoselv Good find! I verified locally that this one i18n message was not being properly translated and that your change fixes it. I then wrote a test for it and pushed it to your fork, which added it to this PR.

On top of this, would you be able to provide any proper translations of the i18n string in other languages? If so, you could add those translations to this PR, if you want. Otherwise, we can worry about that later.

Thanks again!

jjoselv:
If all is well, can this PR be merged by someone please?

bcoe:
@jjoselv this should now be fixed \o/ please try `npm i yargs@next`; if the release is looking good, I will promote it next week.

jjoselv:
It works. Good to go. I'll update the i18n files at some point in time...

PR #1206, opened 2018-08-27
fix: $0 contains first arg on bundled electron app; fixes #685

### LINKED ISSUE #685: Wrong $0 in bundled electron app
We have an election app that is bundled to an .exe file.

A command line like
`bundled.exe -f filename`
will result in:
`
{ _: ['filename'],
  '$0': 'bundled.exe  -f' }
`

The following workaround:
`
const argv = yargs.parse(process.argv.slice(1));
`
will result in 
`
{ _: [],
 f: 'filename',
  '$0': 'bundled.exe  -f' }
`

Please notice that the $0 ist still wrong because it contains the "-f".

If this an expected behavior? Is there a way to work around it?

Linard11:
Default behavior of yargs is to take the first two args of process.argv
to put into $0 as this is usually the program to execute a script
e.g. 'node app.js' but for bundled electron apps this results in $0
being populated with not only the executing program but also the first
parameter. This fix modifies the check for node/iojs (and adds electron)
so in case of those 3, the executing program is skipped in $0 as it was
before, but in case where a bundled app is executed, the first parameter
is skipped.

This is only a cosmetic change for proper help printouts which use $0
A manual `yargs.parse(process.argv.slice(1))` is still necessary for
electron not to skip the first parameter, during parsing

### DIFF
--- a/yargs.js
+++ b/yargs.js
@@ -45,9 +45,9 @@ function Yargs (processArgs, cwd, parentRequire) {
   // ignore the node bin, specify this in your
   // bin file with #!/usr/bin/env node
   if (/\b(node|iojs|electron)(\.exe)?$/.test(process.argv[0])) {
-    self.$0 = process.argv.slice(1, 2);
+    self.$0 = process.argv.slice(1, 2)
   } else {
-    self.$0 = process.argv.slice(0, 1);
+    self.$0 = process.argv.slice(0, 1)
   }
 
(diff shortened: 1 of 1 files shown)


bcoe:
@Linard11 thank you for the contribution.

bcoe:
@Linard11 please try `npm i yargs@next` which has your patch; let me know if there are any issues.

PR #1207, opened 2018-08-30
Improve Norwegian Nynorsk translations

severen:
I've fixed a typo and made "argument sjekk" one word, which is more grammatically correct in Norwegian, as compounds of two nouns should be placed together _without_ a space.

Also, obligatory facepalm as I was the one who wrote these translations (although it is my second language!).


bcoe:
@severen very much appreciate this work, excited that the community has provided so many translations for yargs.

PR #1208, opened 2018-08-30
Improve Norwegian Bokmål translations

severen:
This is a continuation of #1207, as I noticed it had the same grammatical mistake.


bcoe:
@severen thank you very much; sorry for the slow turn around.

bcoe:
@severen please try `npm i yargs@next`, which should have your corrections (let me know if there are any issues).

Issue #1214, opened 2018-09-08
Middleware does not accept a single function


JamesMessinger:
According to [the docs for the `.middleware()` method](http://yargs.js.org/docs/#api-middlewarecallbacks):

> The `callbacks` parameter can be a function or a list of functions

However, it currently (as of v12.0.2) only works with an array of functions.  Passing a single function does nothing.  

```javascript
yargs
  .command('myCommand', 'some command', {}, function(argv){
    console.log('Running myCommand!');
  })
  .middleware(() => console.log('Running middleware 1'))   // <--- this never runs
  .middleware(() => console.log('Running middleware 2'))   // <--- neither does this
  .parse(['myCommand']);
```

[Here is a reproduction of the issue on RunKit.](https://runkit.com/bigstickcarpet/yargs-middleware-bug)

The problem is caused by this block of code:
https://github.com/yargs/yargs/blob/6c6c27d8c87c25e052df23abcf5875253b6b20cd/lib/middleware.js#L3-L7

I believe that `if (typeof callback === 'object')` should be `if (typeof callback === 'function')`.

I will submit a PR for this momentarily

JimiC:
Confirmed. Currently, if you need to pass a single function, it's mandatory to add it to an `Array` and pass that.

What I don't get either, is why and when a `callback`  is an `object` (See also https://github.com/yargs/yargs/blob/v12.0.2/test/middleware.js#L24).
From what I know a `callback` is **always** a `function`.

BigstickCarpet:
@JimiC - Agreed.  I think the `'object'` in the type check was just a typo or misunderstanding.  It probably should have always been `'function'` instead

BigstickCarpet:
@aorinevo - Here's the PR: https://github.com/yargs/yargs/pull/1215

JimiC:
And the related test is also adjusted.

bcoe:
@JamesMessinger @JimiC please try `npm i yargs@next`, which should have a patch that fixes this; if the release is looking good, I'll promote it next week.

JimiC:
LGTM :+1:

arkon:
Seems like this should be closed?

bcoe:
@arkon thanks for helping us groom some of our old issues 👍

PR #1215, opened 2018-09-08
Fix type check in the .middleware() method

BigstickCarpet:
Resolves Issue #1214

### DIFF
--- a/test/middleware.js
+++ b/test/middleware.js
@@ -29,6 +29,25 @@ describe('middleware', () => {
     globalMiddleware.should.have.lengthOf(1)
   })
 
+  it('runs the middleware before reaching the handler', function (done) {
+    yargs(['mw'])
+      .middleware(function (argv) {
+        argv.mw = 'mw'
+      })
+      .command(
+        'mw',
+        'adds func to middleware',
+        function () {},
+        function (argv) {
+          // we should get the argv filled with data from the middleware
+          argv.mw.should.equal('mw')
+          return done()
+        }
+      )
+      .exitProcess(false) // defaults to true.
+      .parse()
+  })
+
   it('runs all middleware before reaching the handler', function (done) {
     yargs(['mw'])
       .middleware([
@@ -56,11 +75,9 @@ describe('middleware', () => {
 
   it('should be able to register middleware regardless of when middleware is called', function (done) {
     yargs(['mw'])
-      .middleware([
-        function (argv) {
-          argv.mw1 = 'mw1'
-        }
-      ])
+      .middleware(function (argv) {
+        argv.mw1 = 'mw1'
+      })
       .command(
         'mw',
         'adds func list to middleware',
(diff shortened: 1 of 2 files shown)


test/middleware.js
```diff
@@ -24,7 +24,7 @@ describe('middleware', () => {
   it('should add a single callback to global middleware', () => {
     const globalMiddleware = []
 
-    middlewareFactory(globalMiddleware)({})
+    middlewareFactory(globalMiddleware)(function () {})
```
aorinevo: This unit test just checks to see the length of the globalMiddleware array which would equal 1 regardless of whether we add an `object` or `function`.  

We should instead check that the function gets called, similar to how it was done for passing in an array (see L33-L54).

BigstickCarpet: The test would fail if an object was passed in, because `typeof` would be `'object'`, not `'function'`, so the object wouldn't get added to the middleware array.

That said, I agree that it's worthwhile to create an additional test that verifies that a single function gets called.   I'll add that now...

aorinevo: You're right! Checking the length of the array suffices.

BigstickCarpet: I went ahead and added a new test.  Also enhanced an existing test to cover this use-case.

https://github.com/yargs/yargs/pull/1215/commits/f93cb61a25f61d240561799c0a8f92e6d904127f

test/middleware.js
```diff
@@ -56,11 +75,9 @@ describe('middleware', () => {
 
   it('should be able to register middleware regardless of when middleware is called', function (done) {
     yargs(['mw'])
-      .middleware([
-        function (argv) {
-          argv.mw1 = 'mw1'
-        }
-      ])
+      .middleware(function (argv) {
```
aorinevo: Nice!

aorinevo:
Looks good!

Reassigning to @bcoe.

BigstickCarpet:
@bcoe - any estimate as to when this PR might be merged & published?

BigstickCarpet:
btw... I also submitted [a PR on DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/28725) to add a type definition for the `.middleware()` method

aorinevo:
> @bcoe - any estimate as to when this PR might be merged & published?

Hi @BigstickCarpet, 

Can you squash the commits and update the commit message to follow conventionalcommits.org guidelines?

Once that's done, I can merge this to master. 

Also, I see that AppVeyor build step is stuck.  I can't tell why, but hopefully this will be resolved with the above changes.

I can't estimate when the next release of yargs is going to get published, but there are ways to install non-release versions of yargs, see https://github.com/yargs/yargs/issues/1134.

BigstickCarpet:
@aorinevo - It looks like AppVeyor just failed to update GitHub (maybe the webhook failed), but if you click the "Details" link, [the AppVeyor build passed successfully](https://ci.appveyor.com/project/bcoe/yargs-ljwvf/build/1483).

As for squashing the commits and editing the comments... I can do that myself if you want, but GitHub recently made it possible for repo owners to do this automatically when merging the PR.  Click the drop-down next to the "Merge" button and you'll see an option to squash the commits and replace the commit message.  It'll even automatically populate the squashed commit message with the concatenated messages of all the commits.  Really handy feature.  I use it on all of my personsal repos.

squash

aorinevo:
That works too, as long as the resulting commit message is inline conventionalcommits.org.

bcoe:
@aorinevo 👋 we should add docs about landing work! since it's all in my brain. You can just rewrite the commit message when calling "Squash and Merge", I'm fine with folks not following the convention until we land 👍

PR #1217, opened 2018-09-25
Fix NPM registry data

danmaas:
Hello - I noticed that the NPM registry entry for 'yargs' lists your homepage and repo with an HTTP URL rather than HTTPS, and therefore it shows as "insecure" in modern browsers. This PR fixes it to use HTTPS. Also, this adds the missing "author" field for NPM.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "author": "James Halliday",
   "main": "./index.js",
   "contributors": [
-    {"name": "Yargs Contributors", "url": "https://yargs.js.org/"}
+    {"name": "Yargs Contributors", "url": "https://github.com/yargs/yargs/graphs/contributors"}
   ],
   "files": [
     "index.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -2,7 +2,11 @@
   "name": "yargs",
   "version": "12.0.2",
   "description": "yargs the modern, pirate-themed, successor to optimist.",
+  "author": "James Halliday",
   "main": "./index.js",
+  "contributors": [
+    {"name": "Yargs Contributors", "url": "https://yargs.js.org/"}
```
bcoe: rather thank linking to the website here, let's link to contributors on GitHub:

https://github.com/yargs/yargs/graphs/contributors

danmaas: OK, link updated!

bcoe:
see: https://github.com/yargs/yargs/pull/1224

PR #1221, opened 2018-10-02
fix: hide `hidden` options from help output even if they are in a group

pvdlg:
```js
require('yargs')
.options({
  foo: {
    describe: 'FOO',
    group: 'Hidden',
    hidden: true
  },
  baz: {
    describe: 'BAZ',
    group: 'Hidden',
    hidden: true
  }
})
  .help()
  .argv
```

Before:
```bash
Hidden
  --foo  FOO
  --baz  BAZ

Options:
```

After:
```bash
Options:
```

### DIFF
--- a/test/usage.js
+++ b/test/usage.js
@@ -2853,6 +2853,61 @@ describe('usage tests', () => {
         '  --bar'
       ])
     })
+    it('--help should display all options except for hidden ones even in a group', () => {
+      const r = checkUsage(() => yargs('--help')
+        .options({
+          foo: {
+            describe: 'FOO',
+            hidden: true
+          },
+          bar: {},
+          baz: {
+            describe: 'BAZ',
+            group: 'Hidden:',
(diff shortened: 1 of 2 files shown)


pvdlg:
For some reason `standard` doesn't return errors locally but it returns a bunch of indentation errors on the CI.... No idea why....So I have to trigger multiple CI runs to fix those linting errors...

test/usage.js
```diff
@@ -2853,6 +2853,61 @@ describe('usage tests', () => {
         '  --bar'
       ])
     })
+    it('--help should display all options except for hidden ones even in a group', () => {
```
bcoe: \o/ tests, thank you very much.

bcoe:
@pvdlg please try `npm i yargs@next` which should have your patch; let me know if there are any issues.

pvdlg:
It works as expected, thanks

PR #1222, opened 2018-10-02
Downgrade decamelize dependency to 1.2.0 to decrease package size

Siilwyn:
I noticed the package install size creeping up quite a bit in the last versions: https://packagephobia.now.sh/result?p=yargs

After taking a look at it a big cause is the `decamelize` dependency, replacing it with `kebab-case` decreases the package install size from around 1400 kB to 670 kB.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   ],
   "dependencies": {
     "cliui": "^4.0.0",
-    "decamelize": "^2.0.0",
+    "decamelize": "^1.2.0",
     "find-up": "^3.0.0",
     "get-caller-file": "^1.0.1",
     "os-locale": "^3.0.0",


bcoe:
@Siilwyn I'd have a slight preference for figuring out why decamilize is so large, and reducing the package size, vs., switching to `kebab-case`. reason being their are advantages to using the more popular library (more eyeballs on it, if there's ever a security vulnerability, I know @sindresorhus' pretty well as a maintainer).

sindresorhus:
It's because of Unicode support. When we can target Node.js 10, it can be removed in favor of native syntax: https://github.com/sindresorhus/decamelize/blob/7e6b97deee705b5658b4efa576b0df84608f1000/index.js#L15-L17

If Yargs doesn't need Unicode support, it could downgrade to version 1.2.0 which is the same just without Unicode support.

Siilwyn:
@bcoe okay understandable, and thanks for chiming in @sindresorhus!
So then the question is if we need Unicode support, I've never come across such characters in CLI flags so my initial response would be that it's not needed. What do you think @bcoe?

pvdlg:
What is the benefit of reducing the install size from 1400 kB to 670 kB vs the risk of regression for some users due to lack of unicode support and the risk of regression due to switching to a less popular package?
The current [cost of 1GB is $0.050](https://www.backblaze.com/blog/hard-drive-cost-per-gigabyte) or less. Even on slowest internet connection 730kB is downloaded pretty fast and npm would cache the package anyway. So how 730kB makes any difference?

Siilwyn:
@pvdlg it would save a lot of bandwidth and disk space in total with 20,307,650 downloads per week!

As said above we won't switch to a different package, just an older version of the same package.

The unicode support is a new addition to the dependency and yargs doesn't claim support for this afaik, so it is not something that users expect to be there.

pvdlg:
> it would save a lot of bandwidth and disk space in total with 20,307,650 downloads per week!

The packages are downloaded over http and the content is gzipped. So that would be minimal.

How saving a few 100 megabytes for the npm CDN provider (that handles millions of GB anyway) versus loosing a functionality (yes and undocumented one) and using an outdated package that won't receive new fixes, is a worthwhile trade-off?

Siilwyn:
@pvdlg we're talking about a straightforward package here, looking at its history and the implementation I don't think missing "new fixes" is applicable. Anyway I think this is where opinions come into play, there is still the disk space which you didn't mention and for the bandwidth I'm thinking of the users with slow internet. I think it is worthwhile as every small effort helps the ecosystem as a whole.

bcoe:
@Siilwyn @pvdlg it's been an ongoing goal of yargs to support multiple locales (It's been translated into many languages, some of which don't fit the latin character set); so I like the spirit of using a decamilize/camecase library that takes into account Unicode...

at the same time, I don't love that xregexp introduces a fairly large increase in package size ... 

for the time being, rather than swapping in a new dependency, perhaps we can pin to an older version of camelize and decamelize -- with the intention of upgrading as soon as Node 10 becomes LTS.

Siilwyn:
@bcoe roger that!

Siilwyn:
@pvdlg do you know any yargs CLI apps using unicode characters in their flags?

You already brought up the trade-offs regarding fixes and size up previously, see my previous reply.

> yargs is for command line argument parsing, which is not used on embedded devices with limited disk size, on mobile or even on browsers.

That's a dangerous assumption to make, Node.js is used in very different circumstances.

bcoe:
@pvdlg if we were talking about 100kb, I'd be fully on your side of the argument; I care about supporting our foreign users. 730kb is huge though, and I dislike that decamelize pulls in a library that introduces quite a few modifications to V8's native regex parser -- my gut is that down this path are potential security concerns, such as infinite recursion.

I look forward to `v10` being LTS, so that we can better support unicode flags -- but I'm hesitant to increase the package size by nearly 700kb, for this feature ... and am equally hesitant to pull in an extended regex library.

I'm open to compromises, @sindresorhus is there any compromise that gets us better Unicode support prior to Node 10, without pulling in a third-party regex parser?

pvdlg:
> do you know any yargs CLI apps using unicode characters in their flags?

No I don't have an example handy. However that seems a lot more likely than someone installing a Node application on a floppy disk where 730kB makes a difference.

I disagree with your comment in https://github.com/yargs/yargs/pull/1222#issuecomment-426722047:
- Missing new fixes is applicable. Bugs happens. I'm pretty sure the author of `decamelize` will not backport  fixes in `1.x` because there is no reason for the extra effort
- Even on slow connection 730kB, which is 200kB gzipped, is not going to make a difference, especially that it will be cached after the first install
- The disk space is not a problem here. The cost of 700k on hard drive is $0.00000005. And every computer made in the last 20 years comes with way more disk space that 700k anyway. Unless there is a compelling demand by a user to run `yargs` on a device with extremely limited disk space, that's a case of premature optimization.

iarna:
The reason I want yargs smaller is `npx`. I write a lot of cli tools. They mostly use yargs. In fact, yargs tends to account for most of their download size. I want to run them with `npx` instead of installing them globally or as dev deps. The larger yargs and its deps, the higher the latency to running `npx` commands with it. Shrinking yargs and its deps means faster downloads means lower latency.

pvdlg:
@bcoe one way or they other it doesn't impact me, so I don't care that much from the point of view of using `yargs` in my projects.
It's just that making any sacrifice for 700kB is clearly a case of premature optimization, unless there is a compelling reasons like running in a web browser.
I see this argument of saving kilobytes at any cost in `node_modules` made quite often on many repo, and in addition of the logic flaw it forces drawback and compromises in modules that are not necessary.
Time would be better spend fixing bugs, creating features and new software, rather than chasing kilobytes in a world where the oldest computer still running probably has at least a few gigabytes of disk space.
Even if the size of `node_modules` was that important, a better way to reduce it without negative impact would be for example to use a `.gitignore` file to exclude test files. A lot of projects don't do that. Dropping features, no matter how small, is not really a good solution.

pvdlg:
@iarna I understand your point. Do we have metrics of what would be the time saved on a slow connection when running `npx` if we cut a dependency that is 200kB gzipped?

iarna:
@pvdlg It's not just download time its also extract time, but yeah, that'd be good to measure. I'd also like to see practical examples of command line option parsing will fail without proper unicode support. (It's been my experience that for most use cases, it actually still works despite not determining character boundaries correctly. It specifically fails when you put an astral character in a character class, but is it possible to make Yargs do that? The most obvious failure scenario I can think of that might impact yargs is around unicode normalization, but I'm not 100% sure that's a practical problem for it either.)

sindresorhus:
I would just depend on decamelize@1.2.0 for now and be done with it. It had gone unchanged for years before the unicode support was added. [It's 13 LOC.](https://github.com/sindresorhus/decamelize/blob/95980ab6fb44c40eaca7792bdf93aff7c210c805/index.js) It's not like it will suddenly need lots of bug fixes. I'm willing to backport potential bug fixes to v1 if that will end this discussion. Alternatively, just copy-paste the code here and bundle it for now. It's a very simple module.

pvdlg:
btw, `xregexp` includes both `src`, `lib` plus a bundled version (`xregexp-all.js`) in the npm package.
If it turns out saving a few 100 kb makes that much of a difference it's probably better to open a PR there to remove the `src` directory from the npm package, and possibly the browser bundle which cumulated represent ~70% of the package size.

bcoe:
@pvdlg dug into this a bit, I bet we could reduce the package size considerably -- but one additional concern that came up for me, is that xregexp seems to have some bugs on Node 10:

```bash
1) XRegExp() supports new regex syntax: named capture should throw an exception if characters other than A-Z, a-z, 0-9, $, and _ are used in capture names
  Message:
    Expected function to throw an Error.
  Stack:
    Error: Expected function to throw an Error.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/benjamincoe/oss/xregexp/tests/spec/s-xregexp.js:407:61)
        at <Jasmine>
  Message:
    Expected function to throw an Error.
  Stack:
    Error: Expected function to throw an Error.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/benjamincoe/oss/xregexp/tests/spec/s-xregexp.js:408:63)
        at <Jasmine>
  Message:
    Expected function to throw an Error.
  Stack:
    Error: Expected function to throw an Error.
        at <Jasmine>
        at UserContext.<anonymous> (/Users/benjamincoe/oss/xregexp/tests/spec/s-xregexp.js:409:59)
        at <Jasmine>
```

☝️ this does not occur on Node 8:

* for now let's wait until someone makes a specific request for a unicode flag, at which point we can consider upgrading decamelize, and perhaps shrinking the size of xregexp.
* otherwise, let's wait until Node 10 is LTS.

@Siilwyn do you want to create a patch for yargs-parser too? it also has the xregexp dependency.

Siilwyn:
Hey @bcoe I could not find the xregexp dependency in yargs-parser, cloned the repository and ran `npm install; npm ls xregexp`.

Issue #1130, opened 2018-05-14
Option with hidden appears if it contains the default value


klemenoslaj:
# Description
Hidden flag in an option configuration get's overruled by some properties.
One of such properties is `defualt`, defining the default value for the option.

I would expect that `hidden` flag does not get overruled by any other property.

Is that by any change bythe design?
If so then it makes sense for such information to land in the documentation.

# Working example

## Code
```
import { options, parse } from 'yargs';

options({
    test: {
        type: 'number',
        describe: 'Description of test option',
        hidden: true
    }
});
```

## Help output
```
Options:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]
```

# Broken example
```
import { options, parse } from 'yargs';

options({
    test: {
        type: 'number',
        describe: 'Description of test option',
        default: 1,
        hidden: true
    }
});
```

## Help output
```
Options:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]
  --test                                                   [number] [default: 1]
```

bcoe:
@klemenoslaj thank you for the bug report, this definitely seems like a bug. If you feel like taking a stab at fixing the issue let me know; and feel free to join the community chat here for coordinating work:

http://devtoolscommunity.herokuapp.com/

andrewcpacifico:
@bcoe this issue was fixed?

I thought to make a PR to fix it, but I cannot reproduce it.

My test script:

```javascript
require('yargs')
  .options({
    test: {
      type: 'number',
      describe: 'Description of test option',
      default: 1,
      hidden: true,
    }
  })
  .help()
  .argv
```

My output:

```
$ node . --help
Options:
  --version  Show version number                                       [boolean]
  --help     Show help                                                 [boolean]
```

pvdlg:
I couldn't reproduce either. However I found a very similar bug when an option is in a group.
See fix in #1221

bcoe:
@pvdlg @andrewcpacifico @klemenoslaj this should be fixed 👍

PR #1233, opened 2018-10-25
Add regression test for hidden options with default values

C45tr0:
Issue #1130 seemed to already be fixed. This is just to add a regression test to make sure it stays working.


bcoe:
@C45tr0 thank you for the regression test, awesome contribution.

Issue #1246, opened 2018-11-19
camelCase argument in command requires inferred alias


vasco-santos:
After [yargs/yargs-parser/commit/08c011796f83e84fe50ffb066a87ae2effadfc75](https://github.com/yargs/yargs-parser/commit/08c011796f83e84fe50ffb066a87ae2effadfc75) got merged, commands using camelCase parameters were automatically inferred to have a dash case alias.

When we use this feature combined with using `...` to optionally accept a list, throws an error `Unknow argument`, for instance in  [ipfs/js-ipfs/src/cli/commands/pin/ls.js#L7](https://github.com/ipfs/js-ipfs/blob/master/src/cli/commands/pin/ls.js#L7)

In order to fix this, we have to set a `dash-case` option. This seems a bug with `yargs-parser`, as if it inferres other alias, it should be transparent to the users.

PR #1247, opened 2018-11-19
fix: allows camel-case, variadic arguments, and strict mode to be combined

bcoe:
fixes: #1246 

CC: @vasco-santos, @boneskull


bcoe:
@vasco-santos when you have a moment please try `npm i yargs@next`, should address your issues.

vasco-santos:
yes it does! Thanks for solving this @bcoe

PR #1249, opened 2018-11-21
add an equal sign to argument

jschiarizzi:
Not sure if that is required, but for consistency i think it makes sense


evocateur:
This example is demonstrating that both styles are possible, note that the previous block uses the `=` delimiters.

jschiarizzi:
ah, that works too :)

PR #1251, opened 2018-11-26
docs(advanced): document non-singleton use, .exit() and parsed

forivall:
This adds the documentation for the non-singleton interface (as described at https://github.com/yargs/yargs/blob/a40cbc9/index.js#L4) to the markdown docs, and the `.parsed` and `.exit()` interfaces, which are useful in non-singleton usage. In particular, I noticed that they are used in [lerna](https://lernajs.io/)'s command line parsing.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -472,3 +472,12 @@ var argv = require('yargs')
      )
   .argv;
 ```
+
+### Using the non-singleton interface
+
+To use yargs without running as a singleton, do:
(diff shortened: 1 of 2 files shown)


bcoe:
@forivall thank you for adding this documentation 👏 apologize for my slow turn around on open source these days, I've been working on addressing this.

rgbkrk:
Thank you both!

Issue #1254, opened 2018-12-10
middleware may need to run before required check


trevorlinton:
One use case that's valuable to use a middleware is to lookup a value that's optional or required on an argument if there's a separate mechanism for finding it. (Think about the heroku cli and how they default the app argument by looking at the .git repo you're currently in).  This could be implemented with middleware, however the middleware runs after required checks are done.  In this circumstance if a required field could be derived using the middleware it never gets a chance to resolve those values. 

A suggested fix would be to complete parsing then execute the middleware then do a required/sanity check on the arguments passed in. 

Simple code:

```
let yargs = require('yargs')
yargs.command('foobar', 'foo then bar', {"an-option":{"demand":true, "string":true, (args) => console.log('done', args)}}
yargs.middleware([(args) => { 
  args['an-option'] = 'this option'
  console.log('middleware ran',args) 
});
yargs.argv
```

If you ran `node example.js foobar` it will simply return with the option `an-option` is required. Where what I'm proposing it would execute the command with `an-option` set to `this option`.

Thoughts? Willing to submit a pull request if one would be accepted.

aorinevo:
Hi @trevorlinton, 

I can see the value of having middleware called before required/sanity checks and would be happy to review a PR for this.  One thought is maybe to differentiate the middleware functions called before required/sanity checks from the ones that run after the checks are complete.  Maybe call them preChecksMiddleware.

trevorlinton:
I created a PR here: https://github.com/yargs/yargs/pull/1255/files I just named it `preChecksMiddleware`, at the moment only globally available, as I'm not sure the utility of having it on a per-command basis (and overloading `command` didn't seem ideal).

trevorlinton:
I should mention, the `preChecksMiddleware` gets passed two objects, the first is the parsed argv as the middleware, but the second is an object containing the array of commands as the property `commands` that could be executed, and a map of the available options and their aliases as `availableOptions`.  This is important for most of the purposes the `preChecksMiddleware` would be performing.

boneskull:
this can be closed?

PR #1255, opened 2018-12-11
Added pre-checks-middlware that runs prior to validation checks

trevorlinton:
See feature enhancement at https://github.com/yargs/yargs/issues/1254


aorinevo:
@trevorlinton I would like to use a shared branch for these changes @amurdock's changes (https://github.com/yargs/yargs/pull/1257) since they are related.  This means you should do the following:
  - Set your PR to target merge into `FEATURE/middleware-enhancements_master_AN` instead of master.
  - In your local, set your upstream to track `FEATURE/middleware-enhancements_master_AN`
     - `git fetch origin -p` (assuming origin points to yargs repo on github)
     - `git branch --set-upstream-to `origin/FEATURE/middleware-enhancements_master_AN`

I will make sure `FEATURE/middleware-enhancements_master_AN` stays up-to-date with master.

aorinevo:
@trevorlinton I would like to use a shared branch for these changes @amurdock's changes (#1257) since they are related. This means you should do the following:

Set your PR to target merge into FEATURE/middleware-enhancements_master_AN instead of master.
In your local, set your upstream to track FEATURE/middleware-enhancements_master_AN
git fetch origin -p (assuming origin points to yargs repo on github)
git branch --set-upstream-to origin/FEATURE/middleware-enhancements_master_AN`
I will make sure FEATURE/middleware-enhancements_master_AN stays up-to-date with master.

yargs.js
```diff
@@ -33,7 +33,8 @@ function Yargs (processArgs, cwd, parentRequire) {
     updateFiles: false
   })
 
-  self.middleware = middlewareFactory(globalMiddleware, self)
+  self.preChecksMiddleware = middlewareFactory(globalMiddleware, self, true)
+  self.middleware = middlewareFactory(globalMiddleware, self, false)
```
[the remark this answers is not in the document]
trevorlinton: I added the default `performPreCheck = false` to the `middlewareFactory`

trevorlinton:
@aorinevo Just checking in, was there anything else you needed on this PR?

aorinevo:
> @aorinevo Just checking in, was there anything else you needed on this PR?

Hi @trevorlinton,  I'm waiting for comments in this [PR](https://github.com/yargs/yargs/pull/1257) to be resolved before merging this in.

docs/advanced.md
```diff
@@ -473,6 +473,28 @@ var argv = require('yargs')
   .argv;
 ```
 
+#### Global Pre-check Middleware
+
+Middleware can be added (and processed) prior to any validation checks but after parsing so required fields maybe implicitly obtained from other methods (such as searching config files, looking for specific files in the current working directory), etc. These can only be set globally at the moment.
```
bcoe: maybe -> may be

trevorlinton: I rebased this and resolved the merge conflicts, tests are passing, I also changed the text as requested.  

To @bcoe's point the pre-checks-middleware does not support asynchronous functions, where as the middleware can. This is because the pre-checks-middleware must change the innerArgv object in the flow of the command run, and before the middleware is ran. If there's ever an intent to get the pre-check middleware to support async it would require drastic changes to the flow of runCommand and my bet is the rest of yargs.  

Maybe it would be useful to rename pre-checks-middleware to something else to prevent confusion, and for now, disallow asynchronous functions for pre-checks-middleware?

trevorlinton:
@bcoe @aorinevo @amurdock Is there any additional changes you'd like done on this PR? 

The only thing I can see left to perhaps do is point out that pre-checks-middleware does not support asynchronous functions while middleware does support them.  To reduce confusion we could consider renaming the pre-checks-middleware to something more appropriate (so the assumption of "async supported in middleware" isn't assumed of the pre-checks).

evocateur:
I'm in favor of the `.middleware(callbacks, applyBeforeValidation)` signature, too.

I don't think per-command `middleware()` can support the `[middleware, applyBeforeValidation]` signature, because `callbacks` can be an array? It would be confusing, even if it is possible to support.

bcoe:
@evocateur @trevorlinton do we start with pre-checks being global middleware only? my only concern is it might be a bit confusing that it's only supported in one context but not the other.

trevorlinton:
Hi @bcoe @evocateur , 

1. I'd refer to https://github.com/yargs/yargs/issues/1254 where @aorinevo asked it to be called `preChecksMiddleware` vs. `.middleware(callbacks, applyBeforeValidation)`. I can move it to using `.middleware` (as originally I wrote it, it infact did just that). I changed it `preChecksMiddleware` on @aorinevo's request. The reasoning why @aorinevo wanted it to be `preChecksMiddleware` may or may not need to be addressed -- I'm not sure.  

2. I should note having the interface named `preChecksMiddleware` vs `.middleware(callbacks, applyBeforeValidation)` doesn't address the issue that a pre-check middleware, one that is running prior to validation **cannot** be asynchronous, so calling this `preChecksMiddleware` or `middleware` might lead to confusion from users.

3. The reason it doesn't support async functions in the prechecks is it must modify the innerArgv tp add or populate options within the `runCommand` prior to the validation checks. It must also return out of `runCommand` the modified innerArgv, which, since `self.runCommand` is not asynchronous by nature, and it returns the `innerArgv`, would require me to make runCommand return a promise (or in other words, become an asynchronous function itself) and I'm sure at that point would begin one enormous yack shave as I refactored and changed interfaces for quite a few things...

I can to proceed with your (@bcoe and @evocateur) feedback and rename it to `.middleware(callbacks, applyBeforeValidation)` but I'll need to add a note in the docs that if `applyBeforeValidation` is set to true, the callbacks MUST be synchronous.

Should I proceed with the feedback from the both of you after what I've mentioned (1,2,3) above?

bcoe:
@aorinevo 👋 I'd like to vote that `applyBeforeValidation` should be a configuration option rather than adding a `preChecksMiddleware` method -- I find the addition of a method a bit confusing, but as a user I think if I read the docs, I'd understand what the config option does.

@trevorlinton let's cross the async problem later; I'm fine with it not blocking this pull request. I do think we should throw an exception if someone tries to use an async pre-check though, otherwise we'll generate support requests of folks wondering why their promises aren't working in that context.

trevorlinton:
@bcoe Sounds good, I've added a couple of tests to ensure it throws an error if async functions are passed, and to make sure pre-check middleware doesn't execute middleware and vice versa, merged the pre-check middleware documentation with the existing middleware docs and pushed up the changes.  We should be good to go unless there's anything else you'd suggest.

docs/api.md
```diff
@@ -969,6 +969,37 @@ I'm another middleware function
 Running myCommand!
 ```
 
+Middleware can be applied before validation by setting the second parameter to `true`.  This will execute the middleware prior to validation checks, but after parsing.
+
+Each callback is passed a reference to argv and context. The argv can be modified to affect the behavior of the validation and command execution.  The context contains the commands and availableOptions that matched the argv.  This feature is useful if you wish to populate a required option dynamically or based on the context prior to validation checks.
```
bcoe: I'm not opposed to adding `context` to middleware, but it feels like this feature should be a separate pull request? Sorry to be a pain, but let's keep this pull request focused on the ability to change the execution order of middleware.

docs/api.md
```diff
@@ -969,6 +969,37 @@ I'm another middleware function
...
+
+For example, an environment variable could potentially populate a required option:
+
+```js
+var argv = require('yargs')
+  .middleware(function (argv, context) {
+    argv.username = process.env.USERNAME
```
bcoe: this seems like a reasonable example of why someone might want to apply middleware earlier in execution 👍

lib/command.js
```diff
@@ -223,6 +223,24 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
     if (!yargs._hasOutput()) {
       positionalMap = populatePositionals(commandHandler, innerArgv, currentContext, yargs)
     }
+    let preCheckMiddlewares = globalMiddleware.slice(0).concat(commandHandler.middlewares || []).filter((x) => x.applyBeforeValidation === true)
```
bcoe: similarly to how we pulled `applyMiddleware` into the `middleware` module, I think for consistency this should be pulled into `middleware.js` as `applyPreCheckMiddleware`; Let's try to encapsulate as much logic specific to middleware as possible in its own helper file (`command.js` is pretty unwieldy).

lib/command.js
```diff
@@ -223,6 +223,24 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
...
+    if (preCheckMiddlewares.length > 0 && !yargs._hasOutput()) {
+      let errors = false
+      const middlewareArgs = preCheckMiddlewares.reduce(function (initialObj, middleware) {
+        let preChecksContext = {
+          'commands': JSON.parse(JSON.stringify(currentContext.commands)),
```
bcoe: I don't understand why you'd need to run `JSON.parse(JSON.stringify)` but, let's save this for a conversation related to the `context` feature 👍

test/middleware.js
```diff
@@ -49,6 +49,151 @@ describe('middleware', () => {
...
+        }
+      )
+      .exitProcess(false) // defaults to true.
+      .parse()
+  })
+
+  it('runs the pre-check middlware ensures if an async function is ran it throws an error', function (done) {
+    try {
+      yargs(['mw'])
+        .middleware([async function (argv, context) {
```
bcoe: tests are failing because Node@6 doesn't support the `async` keyword out of the box, I'd just make this a `new Promise` and you should go green 👍

test/middleware.js
```diff
@@ -49,6 +49,151 @@ describe('middleware', () => {
...
+        )
+        .exitProcess(false) // defaults to true.
+        .parse()
+    } catch (err) {
+      expect(err.message).to.equal('The passed in middleware with applyBeforeValidation set to true may not be used with async functions.')
+      done()
+    }
+  })
+
+  it('Ensure precheck middleware does not run non-precheck middleware, and vice versa', function (done) {
```
bcoe: I'd probably update the terminology in tests to use the term `applyBeforeValidation` rather than `pre-check;

`it should run middleware with applyBeforeValidation set to true before validation`

`it should run middleware with applyBeforeValidation set to false after validation`

I'm happy to do some tweaking like this myself before landing.

bcoe:
see: https://github.com/yargs/yargs/pull/1287, no need to drop the use of `async` in your tests.

docs/api.md
trevorlinton: So, the need we had downstream was to take a required option that's on quite a few commands and populate it if its not provided and some other conditions are met.  Think of it like Heroku's CLI where they populate the value of the apps `-a` option on any command that requires it should the user be in the apps source folder.  

The only way for this to work is to be able to scan what command and options are possible, and compare it with what was passed in. The `innerArgv` alone doesn't give enough information to know if the option should be populated, or even if its available for the command.

Without the context pre-validation middleware doesn't really work for our circumstances.

lib/command.js
trevorlinton: I've moved the logic from command.js into middleware.jjs

lib/command.js
trevorlinton: It's purely so that when passed into the middleware function the user cannot modify it and potentially cause some unintended behaviors.

test/middleware.js
trevorlinton: I renamed things to before validation from pre-check

trevorlinton:
@bcoe For us, the context is necessary, so I'm leaving that in. If it's a blocker to get this merge landed let me know what the concerns you may have are.  The rest of the changes are implemented.

docs/api.md
bcoe: Don't you know this in the context of your middleware though?

If you know that you need a username/password, and your middleware receives an `argv` that does not have `argv.username` and `argv.password` you can take the step of populating the argument.

-----

The reason I'm pushing back is there are various places in the flow of yargs where it would be nice to have more context about the options that have been configured on the `yargs` object:

* what options are required?
* what top-level commands have been configured?
* etc.

I would like to put more thought into this feature rather than just serializing a couple parameters specific to your use-case.

I'd rather that the second argument to the middleware be the `yargs` object itself, and that we expand the API in yargs to better expose certain contextual information about the parsing. It's my opinion that this is a distinct feature from `pre-check-middleware`, however, and we should plan it out separately.

bcoe: part of the context you seem to need is what command the middleware is executing for?

`'commands': JSON.parse(JSON.stringify(context.commands))`

I think this would be better served by us figuring out an API surface that would let us provide pre-validation middleware to a specific command.

@evocateur @aorinevo any thoughts on this one; I was thinking maybe:

`[{handler: handler, applyBeforeValidation: true}]` <-- we could optionally make the handler passed for a command specific middleware be an object like this.

lib/middleware.js
```diff
@@ -31,3 +35,16 @@ module.exports.applyMiddleware = function (argv, middlewares) {
       }
     }, argv)
 }
+
+module.exports.applyPreCheckMiddlware = function (argv, middlewares, context, aliases) {
+  for (let i = 0; i < middlewares.length; i++) {
+    let preChecksContext = {
+      'commands': JSON.parse(JSON.stringify(context.commands)),
```
bcoe: oh, I see what you're doing here, a deep copy so that you can't modify the original object? clever.

lib/middleware.js
```diff
@@ -31,3 +35,16 @@ module.exports.applyMiddleware = function (argv, middlewares) {
...
+      'availableOptions': JSON.parse(JSON.stringify(aliases))
```
bcoe: do we apply aliases after _middleware_ is run? I can understand why this makes writing the middleware annoying for your use-case @trevorlinton.

What if the second argument to middleware was `yargs`, and we expose a `getAliases` method on `yargs`. _then_, if we can figure out an API for passing `applyBeforeValidation` to command specific middleware, I think we'd have addressed your use-cases without the need for this `context` object.

CC: @aorinevo,  @Khaledgarbaya

trevorlinton:
@bcoe I understand the concerns you have, its probably best to think through how the api interfaces or what information should be passed in a bit further, I've removed the context object.  I'll open PR with perhaps a way of getting the current target command or aliases being executed? we can chat more on that PR.

bcoe:
@trevorlinton thanks for being understanding; I'm leaning towards perhaps we make the second argument `yargs` itself (like with the command `builder`), and figure out a way to pass pre-validation middleware to commands?

☝️ this would let you solve the problem you're aiming to solve with the context object right? as long as we make sure aliases are exposed on the yargs object?

trevorlinton:
@bcoe I think I can get access to the actual yargs object through hoisting, so it may not even be necessary to pass it in.  The problem comes down to knowing what the proposed vs passed in options are for a command, without the ability to figure out if the command even takes in a specific option the use case becomes quite a bit trickier.  We can start with just getting the middle applied before validation and I'll explore some options then put in a PR we can review and come up with a few solutions/implementation options.

Khaledgarbaya:
Hey Folks,

Sorry, I am very late to the party, I might be wrong, which most of the time I am, 🙈. I still think that middlewares should be run after yargs checks. I think we are adding some more unnecessary logic to them, they should be simple transformers that get some data, transform it and send it back. For the case of setting a default argument, this can be achieved by the middleware already if you set the option to optional and in your middleware do the check and fill it up. I use this in some project to check for an auth token and if no provided I read it from an `rc` file.

I'll be happy to chat more about this, and if you folks see this feature make sense go ahead and merge it, I just wanted to jump in and chat about it

Cheers,
Khaled

trevorlinton:
@bcoe I added yargs as the second argument on this.  Was there anything else you wanted added?

trevorlinton:
@Khaledgarbaya That's a valid point. I didn't like the help screen didn't show it as required however, and if the middleware was genuinely unable to derive the value I didn't want to accidentally pass/execute the downstream code.  The CLI use case I have has "plugins" which may rely on yargs to never pass in a value marked as required.

bcoe:
closing in favor of https://github.com/yargs/yargs/pull/1289, which does this _and_ a bit more.

Issue #1259, opened 2018-12-17
Specify yargs-parser options via `yargs` JS API


cspotcode:
Use-case: I want to write a command-line tool that uses yargs to parse command-line options.  I have a `--foo` boolean flag.  I do *not* want yargs to allow `--no-foo`.  So I want to pass `{'boolean-negation': false}` to yargs-parser.  (although you can certainly think of situations where other people might want to write CLI tools and pass other flags to yargs-parser; I'm just using this as an example)

It is important that users of my CLI tool do not need to be aware of nor think about yargs.  I will configure yargs to behave predictably and I'll write documentation for my tool that matches the configured behavior.  It's important that my CLI tool behaves exactly as I intend, without being accidentally influenced by external factors such as a consumer's package.json or other yargs-based CLI tools.

I'm imagining that the API looks something like this:

```
yargs.setParserConfiguration({'boolean-negation': false}).option(//... and so on
```

EDIT per [Slack conversation](https://node-tooling.slack.com/archives/C6GD3ELSJ/p1545076446048800), the preferred name for this method is either `parserConfiguration` or `yargsConfiguration`

cspotcode:
Might be related to #1248

boneskull:
While I think such a thing would be useful, and considering the paucity of information in #1248, yargs is careful to load its config from the `package.json` of your CLI app.  It won't look elsewhere unless `require.package.main` *isn't* your app.

Untested, but that's what I can deduce from reading the code.

boneskull:
this is important to mention:

`boolean-negation: true` means `--no-foo` will *expand* that to `foo: false`.  It *does not mean* `--no-foo` is not allowed.  You need to enable "strict" mode via `yargs.strict()` if you want this behavior.

bcoe:
@cspotcode please take this for a spin:

`npm i yargs@next`

There's now a method `parserConfiguration`.

bcoe:
@jean-emmanuel mind trying `npm i yargs@next` and letting me know if I've shipped your work appropriately?

shadowspawn:
#1262 has shipped, so closing this.

Issue #1248, opened 2018-11-20
« "boolean-negation": false » only works when running from the package directory


jean-emmanuel:
When running a node app from a directory that doesn't contain the appropriate package.json, boolean negation is always enabled. Any chance this could be disabled programmatically ?
Thanks a lot for the lib :)

jean-emmanuel:
Note: packaged node applications (eg electron apps) are likely to be run from an arbitrary directory that may not contain the app's `package.json`.

jean-emmanuel:
Yargs relies on `process.cwd()` to find the package.json file, this method doesn't seem to work when the scripts are bundled with browserify. Here's my workaround:

```js
var _cwd = process.cwd
process.cwd = () => { return __dirname }

// require yargs, etc
process.cwd = _cwd
```

bcoe:
@jean-emmanuel YES, we've been having an ongoing conversation in the yargs slack channel about making this something we can configure programmatically -- I think I was trying to be overly clever when I implemented this feature originally.

Want to rephrase this as a feature request, and let's implement this?

CC: @cspotcode

jean-emmanuel:
Alright, PR on the way :)

jean-emmanuel:
Yes, sorry for the delay, it works perfectly !

PR #1262, opened 2018-12-30
feat: Add `.parserConfiguration()` method, deprecating package.json config

jean-emmanuel:
A non-breaking proposal to address #1259 and #1248, it might be incomplete so please amend if needed.

### DIFF
--- a/yargs.js
+++ b/yargs.js
@@ -765,6 +765,14 @@ function Yargs (processArgs, cwd, parentRequire) {
   }
   self.getStrict = () => strict
 
+  let parserConfig = {}
+  self.parserConfiguration = function parserConfiguration (config) {
+    argsert('<object>', [config], arguments.length)
+    parserConfig = config
+    return self
+  }
+  self.getParserConfiguration = () => parserConfig
+
   self.showHelp = function (level) {
     argsert('[string|function]', [level], arguments.length)
     if (!self.parsed) self._parseArgs(processArgs) // run parser, if it has not already been executed.
@@ -1010,7 +1018,14 @@ function Yargs (processArgs, cwd, parentRequire) {
     args = args || processArgs
 
     options.__ = y18n.__
-    options.configuration = pkgUp()['yargs'] || {}
+    options.configuration = self.getParserConfiguration()
+
+    // Deprecated
+    let pkgConfig = pkgUp()['yargs']
+    if (pkgConfig) {
+      console.warn('Configuring yargs through package.json is deprecated and will be removed in the next major release, please use the JS API instead.')
+      Object.assign(options.configuration, pkgConfig)
+    }
 
     const parsed = Parser.detailed(args, options)
     let argv = parsed.argv
--- a/yargs.js
+++ b/yargs.js
@@ -1024,7 +1024,7 @@ function Yargs (processArgs, cwd, parentRequire) {
     let pkgConfig = pkgUp()['yargs']
     if (pkgConfig) {
       console.warn('Configuring yargs through package.json is deprecated and will be removed in the next major release, please use the JS API instead.')
-      Object.assign(options.configuration, pkgConfig)
+      options.configuration = Object.assign({}, pkgConfig, options.configuration)
     }
 
     const parsed = Parser.detailed(args, options)
--- a/yargs.js
+++ b/yargs.js
@@ -1023,7 +1023,7 @@ function Yargs (processArgs, cwd, parentRequire) {
     // Deprecated
     let pkgConfig = pkgUp()['yargs']
     if (pkgConfig) {
-      _logger.error('Configuring yargs through package.json is deprecated and will be removed in the next major release, please use the JS API instead.')
+      console.warn('Configuring yargs through package.json is deprecated and will be removed in the next major release, please use the JS API instead.')
       options.configuration = Object.assign({}, pkgConfig, options.configuration)
     }
 
--- a/yargs.js
+++ b/yargs.js
@@ -1023,7 +1023,7 @@ function Yargs (processArgs, cwd, parentRequire) {
     // Deprecated
     let pkgConfig = pkgUp()['yargs']
     if (pkgConfig) {
-      console.warn('Configuring yargs through package.json is deprecated and will be removed in the next major release, please use the JS API instead.')
+      _logger.error('Configuring yargs through package.json is deprecated and will be removed in the next major release, please use the JS API instead.')
       options.configuration = Object.assign({}, pkgConfig, options.configuration)
     }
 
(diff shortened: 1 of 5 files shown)


bcoe:
@jean-emmanuel this looks great, we should add a test though 👍

jean-emmanuel:
I just pushed 2 tests, what do you think ?

jean-emmanuel:
Travis' fail seems unrelated by the way.

yargs.js
```diff
@@ -1010,7 +1018,14 @@ function Yargs (processArgs, cwd, parentRequire) {
     args = args || processArgs
 
     options.__ = y18n.__
-    options.configuration = pkgUp()['yargs'] || {}
+    options.configuration = self.getParserConfiguration()
+
+    // Deprecated
+    let pkgConfig = pkgUp()['yargs']
+    if (pkgConfig) {
+      console.warn('Configuring yargs through package.json is deprecated and will be removed in the next major release, please use the JS API instead.')
```
evocateur: There's no other "raw" `console.*` call in this file, we should probably use the instance-scoped [`_logger.error()`](https://github.com/yargs/yargs/blob/27bf73923423dbe84dd2fd282fdd31d26bdb6cee/yargs.js#L956-L975) here instead.

Parenthetically, `console.warn` vs `console.error` is just semantics, it doesn't call `process.exit()` and both console methods log to `process.stderr` anyway.

jean-emmanuel: Alright, thanks for reviewing the PR ! 41fe766

jean-emmanuel: Using the scoped logger happens to break the tests as this deprecation warning is unexpected. I've reverted the commit, I'm not sure it's relevant to pass this warning for non-cli print uses.

bcoe:
@evocateur @jean-emmanuel where does this stand? would love to land the work; I'm thinking perhaps we land this as a major, and then in the next major actually retire it?

yargs.js
evocateur: Okay, fair enough. A pity about the brittle tests.

evocateur:
This strikes me as semver-minor, with removal in the next major. If there are other semver-majors waiting in the wings, it could certainly go along with that.

bcoe:
@evocateur @jean-emmanuel, @isaacs just pointed this out to me:

https://nodejs.org/api/util.html#util_util_deprecate_fn_msg_code

☝️ Node.js has a built in deprecate message that debounces.

bcoe:
let's use this -- @evocateur there's some somewhat finicky breaking changes landing in `yargs-parser`, so I'm a bit tempted to land it in a major (the ones you reviewed around handling of missing right-hand values).

yargs.js
```diff
@@ -1010,7 +1018,14 @@ function Yargs (processArgs, cwd, parentRequire) {
     args = args || processArgs
 
     options.__ = y18n.__
-    options.configuration = pkgUp()['yargs'] || {}
+    options.configuration = self.getParserConfiguration()
+
+    // Deprecated
+    let pkgConfig = pkgUp()['yargs']
```
evocateur: I like @bcoe's suggestion to use [`util.deprecate()`](https://nodejs.org/api/util.html#util_util_deprecate_fn_msg_code) here. It'll require a minor extraction of the logic, but that's arguably a good thing anyway.

```js
const util = require('util');

const getPkgConfig = util.deprecate(() => pkgUp()['yargs'], 'Configuring yargs through package.json is deprecated...');

// etc etc
```

We can test it by listening to the [warning](https://nodejs.org/api/process.html#process_event_warning) event on `process`:
```js
let called = false
process.once('warning', () => {
  called = true
})
// do the thing
// assert called === true ?
```

jean-emmanuel: Wouldn't this emit the warning whether package.json contains yargs configuration tokens or not ? This deprecates the function that checks for package.json configuration, I think it makes more sense to only warn when the deprecated feature is actually used.

evocateur: Oh, yeah, you're right, nevermind `util.deprecate`

evocateur:
@bcoe As @jean-emmanuel rightly pointed out, `util.deprecate()` isn't appropriate for something that is called _every_ time yargs parses arguments.

evocateur:
Thanks @jean-emmanuel!

jean-emmanuel:
woohooo !

bcoe:
this is now available in `yargs@13.1.0`.

plroebuck:
Uh, yeah.
* yargs-13.0.0 mentions in ["breaking changes"](https://github.com/yargs/yargs/blob/master/CHANGELOG.md#1300-2019-02-02) that Node-6 was being dropped (though "package.json" still shows `"node": ">=6"`). **Did you or didn't you?**
* Mocha _will_ support Node-6 _at least_ until end-of-life, so cannot upgrade to yargs-13 yet if support dropped (or will be without another semver-major).
* Yargs-13 introduced `yargs.parserConfiguration`, but no backport to yargs-12 available.
* Users using yargs-13 now get deprecation warning, with no visible means to address the problem from Mocha via yargs-12.

Recommendations?

plroebuck:
BTW, your documentation shows:
```js
yargs.parserConfiguration({
  "yargs": {
    "short-option-groups": true,
    "camel-case-expansion": true,
    "dot-notation": true,
    "parse-numbers": true,
    "boolean-negation": true
  }
})
```
but test fixture ["test/fixtures/configured-bin.js"](https://github.com/yargs/yargs/blob/3c6869aae7b488d2416f66c19a801c06243c075c/test/fixtures/configured-bin.js#L3) shows:
```js
var argv = require('./yargs/index.js')
  .parserConfiguration({'dot-notation': true})
  .help('help')
  .version()
  .argv
```

bcoe:
> * yargs-13.0.0 mentions in ["breaking changes"](https://github.com/yargs/yargs/blob/master/CHANGELOG.md#1300-2019-02-02) that Node-6 was being dropped (though "package.json" still shows `"node": ">=6"`). **Did you or didn't you?**

@plroebuck we can no longer promise that `Node >=6` will continue to be supported in yargs 13; one reason being flakiness we were seeing in CI/CD with the current combination of npm and Node.js shipped on 6.x. The engine field was just missed being updated.

> * Mocha _will_ support Node-6 _at least_ until end-of-life, so cannot upgrade to yargs-13 yet if support dropped (or will be without another semver-major).

We would be open to back-porting this change-request, and security patches back to `12.x`.

> BTW, your documentation shows...

☝️ I believe the documentation is wrong.

@plroebuck could you open a few issues rather than commenting on a closed issue, the odds of us losing track of your comment are high.

@jean-emmanuel interested in back-porting this to `yargs@12.x` without the deprecation warning. To do so we would create a new `12.x` branch, from the point at which we released https://github.com/yargs/yargs/commit/8789bf4f56940248316c58997b107b029dbdb297, and cherry-pick create a pull request against this branch.

PR #1270, opened 2019-01-08
update os-locale because of security vulnerability

FabricioFFC:
`os-locale` 3.1.0 uses mem `4.0.0` that fix a vulnerability: https://snyk.io/vuln/npm:mem:20180117


evocateur:
This is an in-range update, so all anyone has to do is re-install `yargs`? Or auto-fix with `npm audit fix`.

bcoe:
@FabricioFFC please give this a try:

`npm i yargs@next`

should have the security patch in question.

bcoe:
fix is now in `13.1.0`.

PR #1271, opened 2019-01-08
Edited help example to align with actual output

weswinder:
The help example using line_count.js was no longer accurate according to actual results. From what I can tell this was changed in #983 according to #1010.


bcoe:
thank you @weswinder, appreciate the contribution.

PR #1272, opened 2019-01-09
Consistent bash-completion if a command is not found

antonstefanov:
Hey,
it seems that if a command is not found, the fallback file autocompletion does not work the same as the standard bash autocompletion:

**CURRENT**
`some-cmd folder-na(tab)` -> `some-cmd folder-name `
currently the folder name is inserted, including a space at the end

**PROPOSED**
`some-cmd folder-na(tab)` -> `some-cmd folder-name/`
In the current bash versions the folder name is inserted with a slash at the end, allowing you to chain folder paths

With this change bash will decide how to handle the autocompletion.

### tested on
MacOS 10.14

### references
https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion-Builtins.html
https://stackoverflow.com/a/22827323

### DIFF
--- a/completion.sh.hbs
+++ b/completion.sh.hbs
@@ -19,10 +19,10 @@ _yargs_completions()
 
     # if no match was found, fall back to filename completion
     if [ ${#COMPREPLY[@]} -eq 0 ]; then
-      COMPREPLY=( $(compgen -f -- "${cur_word}" ) )
+      COMPREPLY=()
     fi
 
     return 0
 }
-complete -F _yargs_completions {{app_name}}
+complete -o default -F _yargs_completions {{app_name}}
 ###-end-{{app_name}}-completions-###


antonstefanov:
nvm, this can result in `No such file or directory` errors if a folder is not found, will reopen in the future

completion.sh.hbs
```diff
@@ -19,10 +19,10 @@ _yargs_completions()
 
     # if no match was found, fall back to filename completion
     if [ ${#COMPREPLY[@]} -eq 0 ]; then
-      COMPREPLY=( $(compgen -f -- "${cur_word}" ) )
+      COMPREPLY=()
```
evocateur: Ooh, I like the result of this PR. @tschaub, can you provide any context on the initial draft of this conditional?

tschaub: Hmm.  Having trouble recalling the motivation.  If `()` is a better way to fall back to filename completion, then this looks like the right change.

evocateur: @tschaub Thanks for taking a look!

bcoe:
@antonstefanov thank you for the contribution -- next we just need someone to finally implement `Zsh` completion 😛

bcoe:
@antonstefanov please try `npm i yargs@next` it should have your modifications to bash completions.

antonstefanov:
hey @bcoe nice, it works, I'm not using `Zsh` though (probably should), and can't help with the implementation

PR #1284, opened 2019-02-02
chore: update dependencies

bcoe:
skipped updating `decamelize` because it adds `75kb` to the package size (due to unicode additions yargs doesn't use).

skipped updating `standard` because it would make @trevorlinton's work landing #1255 much more annoying.

BREAKING CHANGE: see [yargs-parser@12.0.0 CHANGELOG](https://github.com/yargs/yargs-parser/blob/master/CHANGELOG.md#breaking-changes)


.travis.yml
```diff
@@ -1,5 +1,8 @@
 language: node_js
 
+before_install:
+  - if [[ "$TRAVIS_NODE_VERSION" =~ ^v6 ]]; then npm i npm -g; fi
```
coreyfarrell: This might mean that `yargs` requires npm v6.4.1 (that's what node.js 8 currently installs).  On the other hand it might be that this requirement only applies to installs which include the development dependencies.

PR #1292, opened 2019-02-12
zsh auto completion

trevorlinton:


lib/completion.js
```diff
@@ -79,8 +79,10 @@ module.exports = function completion (yargs, usage, command) {
 
   // generate the completion script to add to your .bashrc.
   self.generateCompletionScript = function generateCompletionScript ($0, cmd) {
+    let zshShell = process.env.SHELL && process.env.SHELL.indexOf('zsh') !== -1
```
bcoe: I believe this could be `const` and not `let`.

completion.zsh.hbs
```diff
@@ -0,0 +1,17 @@
...
+# yargs command completion script
+#
+# Installation: {{app_path}} {{completion_command}} >> ~/.zshrc
+#    or {{app_path}} {{completion_command}} >> ~/.zsh_profile on OSX.
+#
+_{{app_name}}_yargs_completions()
+{
+  local reply
+  local si=$IFS
+  IFS=$'\n' reply=($(COMP_CWORD="$((CURRENT-1))" COMP_LINE="$BUFFER" COMP_POINT="$CURSOR" {{app_path}} --get-yargs-completions "$\{words[@]\}"))
```
bcoe: interesting, at the end of the day this isn't too far off bash completion.

test/completion.js
```diff
@@ -196,10 +196,21 @@ describe('Completion', () => {
   })
 
   describe('generateCompletionScript()', () => {
-    it('replaces application variable with $0 in script', () => {
+    it('replaces application variable with $0 in script (bash)', () => {
+      process.env.SHELL = '/bin/bash'
```
bcoe: I like this approach 👍

completion.zsh.hbs
trevorlinton: yeah, not to much of a difference really.

lib/completion.js
trevorlinton: fixed.

Issue #1302, opened 2019-02-20
[Q] Node-6 support in yargs


plroebuck:
yargs-13.0.0 mentions in ["breaking changes"](https://github.com/yargs/yargs/blob/master/CHANGELOG.md#1300-2019-02-02) that Node-6 was being dropped (though "package.json" still shows `"node": ">=6"`).

plroebuck:
@bcoe [wrote](https://github.com/yargs/yargs/pull/1262#issuecomment-465218465):
> * yargs-13.0.0 mentions in ["breaking changes"](https://github.com/yargs/yargs/blob/master/CHANGELOG.md#1300-2019-02-02) that Node-6 was being dropped (though "package.json" still shows `"node": ">=6"`). **Did you or didn't you?**

@plroebuck we can no longer promise that `Node >=6` will continue to be supported in yargs 13; one reason being flakiness we were seeing in CI/CD with the current combination of npm and Node.js shipped on 6.x. The engine field was just missed being updated.

plroebuck:
As my other issue involves upgrading assuming Node-6 support, would I be safe stating?
* Regardless of release note contents, **another _major_ release would now be required to drop Node-6 support in yargs** (i.e., yargs-13.x _must_ retain Node-6 compatibility).

plroebuck:
Node-6 support will continue until yargs-14, per #1308.

PR #1308, opened 2019-03-01
chore: forgoing decision to drop Node 6 until April

bcoe:
for the benefit of mocha, holding off on decision to drop Node 6 support @boneskull @plroebuck.

### DIFF
--- a/test/middleware.js
+++ b/test/middleware.js
@@ -261,9 +261,10 @@ describe('middleware', () => {
     it('throws an error if async function provided', function () {
       expect(() => {
         yargs(['mw'])
-          .middleware([async function (argv) {
+          .middleware([function (argv) {
             argv.mw = 'mw'
             argv.other = true
+            return Promise.resolve(argv)
           }], true)
           .command(
             'mw',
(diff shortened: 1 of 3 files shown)


bcoe:
@boneskull @plroebuck, does this seem like a reasonable solution for the warning you're getting in mocha?

test/middleware.js
```diff
@@ -261,9 +261,10 @@ describe('middleware', () => {
     it('throws an error if async function provided', function () {
```
plroebuck: test title no longer applies...

plroebuck:
We need two things to allow us to upgrade our `yargs` dependency:
1. Documenting the decision to retain Node-6 support in current release.
1. Assurance that Node-6 support will not be dropped unless another major release of `yargs` occurs.

plroebuck:
Reference #1302

bcoe:
@plroebuck the decision not to drop Node 6 support is document here:

https://github.com/yargs/yargs/pull/1308/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1e

once we rollback this decision, we will not drop Node 6 support until `yargs@14`.

Issue #1303, opened 2019-02-20
yargs.parserConfig documentation incorrect?


plroebuck:
Your documentation shows:
```js
yargs.parserConfiguration({
  "yargs": {
    "short-option-groups": true,
    "camel-case-expansion": true,
    "dot-notation": true,
    "parse-numbers": true,
    "boolean-negation": true
  }
})
```
but ["test/fixtures/configured-bin.js"](https://github.com/yargs/yargs/blob/3c6869aae7b488d2416f66c19a801c06243c075c/test/fixtures/configured-bin.js#L3) shows:
```js
var argv = require('./yargs/index.js')
  .parserConfiguration({'dot-notation': true})
  .help('help')
  .version()
  .argv
```

plroebuck:
@bcoe [wrote](https://github.com/yargs/yargs/pull/1262#issuecomment-465218465):
> > BTW, your documentation shows...
> 
> ☝️ I believe the documentation is wrong.

shadowspawn:
The documentation has been updated (sometime in the intervening period 😄 ), thanks for the report.

https://yargs.js.org/docs/#api-reference-parserconfigurationobj

PR #1309, opened 2019-03-01
doc: clarify parserConfiguration object structure

bcoe:
CC: @plroebuck.


bcoe:
@plroebuck would this note have helped clarify configuration settings for you?

plroebuck:
Not as much as a codefence example would have....

plroebuck:
Reference #1303

bcoe:
@plroebuck I'm not quite sure what you mean by a code-fence example, mind submitting something?

My concern is that I don't want yargs to have to duplicate 100% of yargs-parser's documentation; we have quite a few configuration settings available to users.

PR #1316, opened 2019-03-10
Make Dutch translation consistent, fix Dutch language errors, and make it real Dutch

jandockx:
- "Commands:" and "command" should be translated consistently (it was "Opdrachten:" and "commando"); "commando" is much more appropriate in this context than "opdracht"
- In Dutch, there is a difference between "nummer" and "getal". A "nummer" is used for the English "digit" ([0-9]). A "getal" is what is meant here (the type of all numbers)
- the translation of "boolean" is "booleaans"
- "string" was translated as "text". "text" is not a Dutch word. In Dutch, "text" becomes "tekst". But a string is not a text.
https://www.microsoft.com/en-us/language/Search?&searchTerm=number&langID=Dutch&Source=true&productid=All%20Products translates "string" to "tekenreeks", but that is
hypercorrect. Nobody uses that term. In Dutch, we just use the English term "string" colloquially.
- "Non-option argument" should be translated as "niet-optie-argument" ("niet" instead of "non")
- The sentence aspect of "got %s, need at least %s" and "got %s, maximum of %s" was not translated sensibly
- "Missing argument value: %s" was not translated at all
- In the singular, there is no ending-'e' for an adjective if there is no article ("verplicht")
- The plural form "waarden" of "waarde" (value) seems more appropriate here than the alternative "waardes"
- The English word "check" can be used in Dutch to mean "validation", in the 1970's. In this century, we use "controle" colloquially. "Validatie", "verificatie", etc. are possible, but technical jargon
- "Implications failed:" is difficult to translate, because the English form is bad to start with. A literal translation is very awkward in Dutch. This string is used in the code as the
prelude to a list of "implications" that have failed. A better English form would be "Failed implications:", and the new Dutch translation expresses that. Furthermore, the wordt "implications",
in English, as well as in Dutch, is extremely confusing here. It is only people who know the Yargs API that understand that these are the dependencies created by the `implies()` method.
Users of the CLI have no idea what we are talking about. Yet, until a better form is found in English, it is also kept in Dutch.
- In Dutch, we only know compound substantives, not concatenated substantives ("non-option argument" -> "niet-optie-argument", "Argument check" -> "Argumentcontrole", "JSON config file" -> "JSON-config-bestand", "version number" -> "versienummer")
- Since the orginal text talks about the "config file", referring to the name of the file (`config`), the translation should do so too. It was "configuratiebestand", which is Dutch for "configuration file".
- There is no direct translation in Dutch for the idiom "mutually exclusive". The unclear attempt at doing a direct translation was replaced by an idiom that is used in Dutch, that is more descriptive.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -929,6 +929,7 @@ Locales currently supported:
 * **ja:** Japanese.
 * **ko:** Korean.
 * **nb:** Norwegian Bokmål.
+* **nl:** Dutch.
 * **pirate:** American Pirate.
 * **pl:** Polish.
 * **pt:** Portuguese.
--- a/locales/nl.json
+++ b/locales/nl.json
@@ -1,42 +1,42 @@
 {
-  "Commands:": "Opdrachten:",
+  "Commands:": "Commando's:",
   "Options:": "Opties:",
   "Examples:": "Voorbeelden:",
-  "boolean": "boolean",
+  "boolean": "booleaans",
   "count": "aantal",
-  "string": "text",
-  "number": "nummer",
+  "string": "string",
+  "number": "getal",
   "array": "lijst",
   "required": "verplicht",
   "default:": "standaard:",
   "choices:": "keuzes:",
   "aliases:": "aliassen:",
   "generated-value": "gegenereerde waarde",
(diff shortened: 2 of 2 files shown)


bcoe:
@jandockx thank you for improving our dutch translation 🎉

PR #1317, opened 2019-03-10
Change the label for "Implications failed:" in English

jandockx:
The word "implications" is extremely confusing here. It is only people who know the Yargs API that understand that these are the dependencies created by the `implies()` method.
Users of the CLI have no idea what we are talking about.

This changes the label of the concept to "required argument dependencies".

This string is used in the code as the prelude to a list of "implications" that have failed, not a warning. The earlier form was more an expression of a warning. The new form is more an expression of a title of a list.

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -29,7 +29,7 @@
   "Invalid values:": "Invalid values:",
   "Argument: %s, Given: %s, Choices: %s": "Argument: %s, Given: %s, Choices: %s",
   "Argument check failed: %s": "Argument check failed: %s",
-  "Implications failed:": "Implications failed:",
+  "Implications failed:": "Missing required argument dependencies:",
   "Not enough arguments following: %s": "Not enough arguments following: %s",
   "Invalid JSON config file: %s": "Invalid JSON config file: %s",
   "Path to JSON config file": "Path to JSON config file",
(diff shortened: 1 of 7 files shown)


locales/en.json
```diff
@@ -29,7 +29,7 @@
   "Invalid values:": "Invalid values:",
   "Argument: %s, Given: %s, Choices: %s": "Argument: %s, Given: %s, Choices: %s",
   "Argument check failed: %s": "Argument check failed: %s",
-  "Implications failed:": "Implications failed:",
+  "Implications failed:": "Missing required argument dependencies:",
```
bcoe: perhaps "missing implied arguments", or "missing dependent arguments"?

bcoe:
@jandockx 👋 just putting this back on your radar; still an edit you'd like to make?

jandockx:
Hello, @bcoe. 

I still think the current English message is confusing. 

'Missing dependent arguments' sounds fine to me, but the point is I am not a native English speaker. Advise on this topic from native English speakers (like you, I presume) might be more relevant, wouldn't it?

If your intention is for me to change the PR to 'Missing dependent arguments', pls confirm, and I'll do so.

bcoe:
@jandockx 👍 yes, please change the language to `Missing dependent arguments`.

I appreciate your help reviewing the language.

PR #1320, opened 2019-03-18
Move .hbs templates into .js to facilitate webpacking

### LINKED ISSUE #1319: Include completion .hbs templates inline to facilitate easier webpacking
Often we want our CLI tools to start really fast, and webpacking / bundling them into a single file makes them faster.  Unfortunately, webpacking will only look at `require()` / `import` and not `fs.readFileSync`, so it won't bundle yargs' `.hbs` templates.

If those templates are inlined in `.js` -- perhaps as multiline template literals -- then bundling will work automatically with less configuration.

cspotcode:
Fixes #1319


bcoe:
@cspotcode once we fully support webpack, perhaps you could add an advanced topic to yargs' docs on the topic of web-packing?

cspotcode:
Thanks, I'm not sure I'll have the time, but I *do* have a WIP project to provide common webpack prefabs: factory ions that create webpack configurations for common scenarios.  The first I'm is for node, so it fits this situation.

https://github.com/cspotcode/personal-monorepo/tree/master/packages/webpack-config-prefabs

If anyone encounters this in the future and uses the config, and finds it's missing features, they can file an issue.  It seems like everyone creating their own configs is a big waste of effort.

PR #1327, opened 2019-04-07
test: accept differently formatted output

BridgeAR:
The default for Node.js's `util.inspect` `compact` option might soon
change to `3` which slightly changes the default output for objects.
This change accepts both formats so that the test passes in all Node.js
versions.

Refs: https://github.com/nodejs/node/pull/27109


bcoe:
@BridgeAR thanks for the patch.

BridgeAR:
@bcoe is there a plan for a new version? `yargs` keeps on failing in CITGM otherwise.

PR #1337, opened 2019-04-29
docs: restore removed changelog of v13.2.0

teppeis:
In https://github.com/yargs/yargs/commit/4375680c83e76cbf5cfedbfd81c138ce45b90b54, changelog of v13.2.0 was replaced with v13.2.1.


bcoe:
thank you @teppeis, that must have been an accident.

Issue #1307, opened 2019-02-28
Add doc for `--no-` boolean prefix


AndrewFinlay:
I bumped into this the other day and it had me scratching my head.  I found the answer in some older issues but it should really be on the API doc page.

AndrewFinlay:
Just found mention of it on the 'tricks' page, maybe not such an issue after all.

willstott101:
Here is a link dump of some various bits of scattered docs about this:

https://github.com/yargs/yargs-parser#boolean-negation
#122
https://github.com/yargs/yargs/blob/master/docs/advanced.md#customizing-yargs-parser
https://yargs.js.org/docs/#api-configobject

AndrewFinlay:
@willstott101 this change has been merged into the docs, if you're happy with this solution then I'll close the issue.  Thanks.

willstott101:
LGTM I'm guessing that comes up here? http://yargs.js.org/docs/#api-booleankey

Can Ctrl-F for `--no-` and it'll be obvious straight away.

Thanks!

PR #1338, opened 2019-05-02
doc: make `--no-` boolean prefix easier to find in the docs

AndrewFinlay:
Hi all,

I bumped into #1307 while working on another yargs based project and found myself wondering what the `--no-`prefix did.  I went to the yargs docs, found the api reference and tried searching in there but couldn't find anything.  I ended up searching with github's project search to find what I was after.

So this change is quite simple, make mention of the `--no` boolean prefix on the api doc page and link to the existing doc under parsing tricks.  Hopefully that should make it a little quicker to find.


bcoe:
@AndrewFinlay thank you 👍 very much appreciate the docs update.

PR #1356, opened 2019-06-08
feat: Inline code to detect current locale from environment

### LINKED ISSUE #1354: Replace os-locale with inline code
Due to yargs passing `spawn: false` to os-locale most of the functionality of that module is unused.  The following could be used as a drop-in replacement for os-locale:

```js
function osLocale() {
  const locale = env.LC_ALL || env.LC_MESSAGES || env.LANG || env.LANGUAGE || 'en_US';
  return locale.replace(/[.:].*/, '');
}
```

The benefit is a large drop of production dependencies.  Currently a project with `yargs` as the only dependency installs 53 modules / 1.8MB to node_modules.  Without os-locale this becomes 26 modules / 1.1MB.  The size reduction is not my primary interest of this issue, the goal is cutting the number of modules that have to be audited in half.

coreyfarrell:
This eliminates half the dependencies of yargs without any change in
functionality.  The extra dependencies were used by the os-locale module
to execute detection utilities, this execution functionality was always
disabled by yargs.

Fixes #1354

### DIFF
--- a/yargs.js
+++ b/yargs.js
@@ -1181,8 +1181,9 @@ function Yargs (processArgs, cwd, parentRequire) {
     if (!detectLocale) return
 
     try {
-      const osLocale = require('os-locale')
-      self.locale(osLocale.sync({ spawn: false }))
+      const { env } = process
+      const locale = env.LC_ALL || env.LC_MESSAGES || env.LANG || env.LANGUAGE || 'en_US'
+      self.locale(locale.replace(/[.:].*/, ''))
     } catch (err) {
       // if we explode looking up locale just noop
       // we'll keep using the default language 'en'.
--- a/package.json
+++ b/package.json
@@ -22,7 +22,6 @@
     "cliui": "^5.0.0",
     "find-up": "^3.0.0",
     "get-caller-file": "^2.0.1",
-    "os-locale": "^3.1.0",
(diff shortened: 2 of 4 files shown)


yargs.js
```diff
@@ -1181,8 +1181,9 @@ function Yargs (processArgs, cwd, parentRequire) {
     if (!detectLocale) return
 
     try {
-      const osLocale = require('os-locale')
-      self.locale(osLocale.sync({ spawn: false }))
+      const { env } = process
+      const locale = env.LC_ALL || env.LC_MESSAGES || env.LANG || env.LANGUAGE || 'en_US'
+      self.locale(locale.replace(/[.:].*/, ''))
```
coreyfarrell: In theory we the try / catch block is no longer needed but the was `os-locale` was used I think it was impossible for it to throw, therefore I assume that it's possible for `self.locale()` to cause an exception.

bcoe:
@coreyfarrell incredible work, thanks for doing this.

bcoe:
@coreyfarrell when you have a sec try `npm i yargs@next`, has your refactor landed.

coreyfarrell:
It works for my (admittedly simple) use case, `npm i yargs@next` installs 26 packages (including yargs) where yargs@latest installs 53 packages.

PR #1364, opened 2019-06-12
fix!: do not allow additional positionals in strict mode

### LINKED ISSUE #1313: .strict() allows extra args
Documentation states that in strict mode, extra arguments are reported as errors. However, none of the constructs allowed by yargs seems to demonstrate that behaviour. For example, the following configs:

```js
require("yargs")
  .command("foo <bar>")
  .strict()
  .argv
```
```js
require("yargs")
  .command("foo <bar>", "foo cmd", yargs => yargs.strict())
  .strict()
  .argv
```
```js
require("yargs")
  .command({
    command: "foo <bar>",
    builder: yargs => yargs.strict()
  })
  .command({
    command: "baz <qux>",
    builder: yargs => yargs.strict()
  })
  .demandCommand()
  .strict()
  .argv;
```
all result in commands like

```
$ ./cmd foo 123 extra
$ ./cmd bar 456 extra-1 extra-2
```

being accepted without any error.

mleguen:
The test was disabled for commandless yargs, but as the validation is done on the innermost yargs, it was always disabled.

Closes #1313

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -102,7 +102,7 @@ module.exports = function validation (yargs, usage, y18n) {
       }
     })
 
-    if (commandKeys.length > 0) {
+    if ((currentContext.commands.length > 0) || (commandKeys.length > 0)) {
       argv._.slice(currentContext.commands.length).forEach((key) => {
         if (commandKeys.indexOf(key) === -1) {
           unknown.push(key)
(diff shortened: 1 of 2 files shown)


bcoe:
@mleguen thank you for the contribution.

bcoe:
@mleguen mind trying out `npm i yargs@next` and letting me know if this fix works for you?

mleguen:
I confirm the example from #1313 is now working as expected.

PR #1366, opened 2019-06-19
fix: properties accessed on singleton now reflect current state of instance

### LINKED ISSUE #1144: exception thrown when yargs.parsed.newAliases is undefined
Found something of an issue when using `yargs.showHelp()`
[L761](https://github.com/yargs/yargs/blob/master/yargs.js#L761)

it seems that when `self._parseArgs(processArgs)` is called, it is not saving the results to `self.parsed`

if you put a breakpoint on this line: [L1012](https://github.com/yargs/yargs/blob/master/yargs.js#L1012), you will see that the `self.parsed` property contains the appropriate items.

however, after this function is done, on [L762](https://github.com/yargs/yargs/blob/master/yargs.js#L762),
`self.parsed` is set back to `false` instead of an object. this leads to their being a parse error:

`TypeError: Cannot convert undefined or null to object`
`.concat(Object.keys(yargs.parsed.newAliases) || [])`

under function: `usage.showHelp(` [L240](https://github.com/yargs/yargs/blob/master/lib/usage.js#L240)

Thanks! 👍

mleguen:
`singletonify()` used to set `Argv`'s properties to `inst`'s properties values _at the time of `Argv`'s last call_.

So `Argv`'s properties and the `inst`'s ones could diverge afterwards, such as `parsed` after a call to `parse()`, still being `false` in `Argv.parsed`, not in `inst.parse`.

`singletonify()` now adds setters to `Argv` instead, looking for the value of the corresponding property in `inst` _at the time the property is used_.

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,7 +25,7 @@ function singletonify (inst) {
   Object.keys(inst).forEach((key) => {
     if (key === 'argv') {
       Argv.__defineGetter__(key, inst.__lookupGetter__(key))
-    } else if (inst[key] === 'function') {
+    } else if (typeof inst[key] === 'function') {
       Argv[key] = inst[key].bind(inst)
     } else {
       Argv.__defineGetter__(key, () => inst[key])
--- a/index.js
+++ b/index.js
@@ -25,8 +25,10 @@ function singletonify (inst) {
   Object.keys(inst).forEach((key) => {
     if (key === 'argv') {
       Argv.__defineGetter__(key, inst.__lookupGetter__(key))
+    } else if (inst[key] === 'function') {
+      Argv[key] = inst[key].bind(inst)
     } else {
-      Argv[key] = typeof inst[key] === 'function' ? inst[key].bind(inst) : inst[key]
+      Argv.__defineGetter__(key, () => inst[key])
     }
   })
 }
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -837,6 +837,17 @@ describe('yargs dsl tests', () => {
     })
   })
 
+  describe('parsed', () => {
+    it('should be false before parsing', () => {
+      yargs.parsed.should.equal(false)
+    })
+
+    it('should not be false after parsing', () => {
+      yargs.parse()
+      yargs.parsed.should.not.equal(false)
+    })
+  })
+
   // yargs.parse(['foo', '--bar'], function (err, argv, output) {}
   context('function passed as second argument to parse', () => {
     it('does not print to stdout', () => {


index.js
```diff
@@ -25,8 +25,10 @@ function singletonify (inst) {
   Object.keys(inst).forEach((key) => {
     if (key === 'argv') {
       Argv.__defineGetter__(key, inst.__lookupGetter__(key))
+    } else if (typeof inst[key] === 'function') {
+      Argv[key] = inst[key].bind(inst)
     } else {
```
bcoe: why don't we make this slightly more explicit:

`if (['$0', 'parsed'].indexOf(key) !== -1) {`

and we should be mindful about leaking any more stateful variables on the yargs object going forward.

test/yargs.js
```diff
@@ -837,6 +837,17 @@ describe('yargs dsl tests', () => {
     })
   })
 
+  describe('parsed', () => {
+    it('should be false before parsing', () => {
```
bcoe: I wonder if we could figure out the specific exception that was happening with `showHelp` and put a test around this?

mleguen: I agree. A simple "showHelp should not fail" should do the job

index.js
mleguen: I don't get the point. If we support singleton mode, shouldn't every property/method call on Argv  have the same effect as if directly called on the singleton?

bcoe: fair 👍 I just noticed that, I think, `argv`, `$0`, and `parsed` are the only properties that leak ... and honestly, I don't think we meant to leak `parsed` as part of the API.

bcoe:
@mleguen what are your feelings regarding us eventually fixing yargs, such that you can use it in singelton mode, but call `parse()` multiple times without bumping into edge-cases?

I'm happy to land this fix, but think we might want to have the above goal eventually in mind?

index.js
mleguen: Oops! I did not realize this... However, `parsed` is already leaking from the API when not using the singleton. So, does it have a sense to filter properties we accept to leak from the singleton, and not to do it when not using the singleton?

mleguen:
@bcoe We are trying to deal here with 3 different issues:

1. fixing the singleton for its behavior not to be different from the "regular" API (even if it allows accessing properties not meant to be accessed, as the regular API does) : this PR answers this point in its current state
2. fixing #1144 which this PR, as you pointed, does not yet prove to do (at least one test is missing... and maybe more as I may have wrongly related this issue to the previous one)
3. fixing parse() to be callable several times, which is yet another problem, probably the more preoccupying.

My opinion is: we should do 1 and 2 in this PR, and 3 in another one (I should be able to spend some time on it next Wednesday).

bcoe:
@mleguen sounds good to me, I'm comfortable landing this as soon as we have the test for #1144.

mleguen:
I am unable to reproduce #1144 and was indeed wrong when thinking it linked to the singletonify issue.

So this PR no longer closes #1144 (I renamed it) and I did not add tests, as we already extensively test `showHelp()`.

Issue #873, opened 2017-05-02
Environment variables trigger errors for irrelevant commands (in strict mode)


kumar303:
I find strict mode really useful because it reports errors for unknown options (this is what I want).

However, when you have a program with multiple commands it will report an error for environment variables that don't apply to the command you are running. It is common practice to define environment variables once and for all in your shell profile so this gets annoying quickly.

Here is an example `script.js` to illustrate the issue:

````js
const argv = require('yargs')
  .help('help')
  .env('MY_SCRIPT')
  .strict()
  .command('build')
  .command('run', 'runs something', (yargs) => {
    return yargs
      .options({
        'source-dir': {
          demand: false,
          default: '/some/dir',
          type: 'string',
        },
      });
  })
  .argv;

console.log(argv);
````

This works as expected (without errors):
````
$ MY_SCRIPT_SOURCE_DIR="/a/path" node script.js run
````

But this triggers an error which I think is a bug:
````
$ MY_SCRIPT_SOURCE_DIR="/a/path" node script.js build
script.js build

Options:
  --help  Show help                                                    [boolean]

Unknown argument: sourceDir
````
If it's not easily fixable, is there a workaround you can suggest?

**Comment by bcoe** [comment]
@kumar303 I'm curious, does the problem occur if you move the `.env` setting to the inner yargs:

```js
const argv = require('yargs')
  .help('help')
  .strict()
  .command('build')
  .command('run', 'runs something', (yargs) => {
    return yargs
      .env('MY_SCRIPT')
      .options({
        'source-dir': {
          demand: false,
          default: '/some/dir',
          type: 'string',
        },
      });
  })
  .argv;

console.log(argv);
```

kumar303:
No, the environment var error with `build` doesn't happen like that (since `env()` is never invoked). However, such a configuration doesn't seem realistic to me. We are depending on yargs to help us parse environment variables for options that each command, as well as default options, so I don't see how that approach would help us.

gajus:
Just came to raise a similar issue.

I've ended up completely stop using `.strict()` because it is too strict.

I don't like the current `.strict()` behaviour but the solution is not clear.

I guess one option would be to have an option that'd make `.strict()` warn (in stderr) instead of crashing the application in case an unknown property is encountered. The warn behaviour would limited to unknown properties passed via ENV; this does not effect the args passed directly to the program.

[...]

PR #1367, opened 2019-06-19
(WIP) strict: specify where to look for unknown arguments (config, env and/or argv)

mleguen:
As suggested by @cspotcode in #873 

This is a work in progress for now, do not merge yet.

### DIFF
--- a/test/validation.js
+++ b/test/validation.js
@@ -848,6 +848,47 @@ describe('validation tests', () => {
       args.foo.should.equal(true)
       args.bar.should.equal(true)
     })
+
+    it('does not fail on unknown config argument if told not to', () => {
+      yargs('build')
+        .config({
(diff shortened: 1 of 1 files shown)


bcoe:
@mleguen thanks for your contributions recently 😄

I haven't thought too hard on this topic yet; is it the right thing for us to do, to not apply strict mode to `config` or `environments`?

If it is the right thing for us to do, I think we should potentially make it something that's configurable with `.parserConfig`.

mleguen:
@bcoe you're welcome, happy to help!

I am no longer sure it is the right thing to do, and I think we could work around the original issue by allowing more selective definition of env/config to parse, in order not to have anything to ignore after parsing.

I close this PR for now, and will post another suggestion in #873 in the coming days.

PR #1369, opened 2019-06-26
fix: calling parse multiple times now appropriately maintains state (#1137)

### LINKED ISSUE #1137: when in singleton mode, calling parse() multiple times results in undefined behavior
So my situation is that I want to use a required positional argument with some non-required options. To my surprise the positional argument is not picked up as I expected. Here is what the command look like
```bash
$ > node index.js stress-test restock
index.js stress-test <notification>

Some description

Positionals:
  notification  notification type to stress test
                                        [string] [required] [choices: "restock"]

Options:
  --version             Show version number                            [boolean]
  --help                Show help                                      [boolean]
  --amount, -n          amount of notifications to create  [number] [default: 1]
  --environment, --env  environment to conduct stress test in
                                 [choices: "dev", "stg", "pro"] [default: "stg"]

Missing required argument: notification
```

If I try to give an argument that is not part of the argument choice this is what I get:
```bash
$ > node index.js stress-test test
index.js stress-test <notification>

Some description

Positionals:
  notification  notification type to stress test
                                        [string] [required] [choices: "restock"]

Options:
  --version             Show version number                            [boolean]
  --help                Show help                                      [boolean]
  --amount, -n          amount of notifications to create  [number] [default: 1]
  --environment, --env  environment to conduct stress test in
                                 [choices: "dev", "stg", "pro"] [default: "stg"]

Invalid values:
  Argument: notification, Given: "test", Choices: "restock"
```

So obviously the argument is properly picked up. Not sure what is going on here. This is what the command option looks like:

```javascript
   {
      command: 'stress-test <notification>',
      describe: 'Some description',
      builder: (yargs) => {
        return yargs.positional('notification', {
          describe: 'notification type to stress test',
          choices: ['restock'],
          type: 'string'
        }).options({
          amount: {
            default: 1,
            requiresArg: true,
            describe: 'amount of notifications to create',
            alias: 'n',
            number: true
          },
          environment: {
            default: 'stg',
            requresArg: true,
            choices: ['dev','stg','pro'],
            describe: 'environment to conduct stress test in',
            alias: 'env'
          }
        });
      }
   }
```

mleguen:
Extends the freezing/unfreezing mechanism already in place for `parse(args, cb)` to `parse()` (without arguments) to be able to call it several times with the same behavior.

Also corrects the builder function return value handler for it no longer to depend on `yargs.parsed` being set after a call to `parse()` (which is no longer the case after unfreezing).

Closes #1137

### DIFF
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -835,6 +835,29 @@ describe('yargs dsl tests', () => {
 
       r.logs[0].should.match(/Commands:[\s\S]*blerg command/)
     })
+
+    it('can be called multiple times with the same behavior', () => {
+      let counter = { foobar: 0 }
+      yargs(['test', 'foobar'])
+        .command(
+          'test <name>',
+          'increases counter',
+          yargs => yargs.positional('name', {
+            aliases: 'n',
+            describe: 'a name',
+            choices: ['foobar'],
+            type: 'string'
+          }),
+          argv => { counter[argv.name]++ }
+        )
+        .fail((msg) => {
+          expect.fail(undefined, undefined, msg)
+        })
+      yargs.parse()
+      yargs.parse()
+      yargs.parse()
+      expect(counter.foobar).to.equal(3)
+    })
   })
 
   // yargs.parse(['foo', '--bar'], function (err, argv, output) {}
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -2058,17 +2058,23 @@ describe('yargs dsl tests', () => {
     })
 
     it('allows a defaultDescription to be set', () => {
-      yargs('cmd')
+      const r = checkOutput(() => yargs('cmd --help').wrap(null)
         .command('cmd [heroes...]', 'a command', (yargs) => {
           yargs.positional('heroes', {
             default: ['batman', 'Iron Man'],
             defaultDescription: 'batman and Iron Man'
           })
         }).parse()
-
-      yargs.getOptions().defaultDescription.should.deep.equal({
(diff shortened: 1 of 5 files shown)


bcoe:
@mleguen I'll definitely have to look at this work closely, years since I touched these bits :smile: 

but this sounds like awesome progress.

test/yargs.js
```diff
@@ -2035,17 +2058,23 @@ describe('yargs dsl tests', () => {
     })
 
     it('allows a defaultDescription to be set', () => {
-      yargs('cmd')
+      const r = checkOutput(() => yargs('cmd --help').wrap(null)
         .command('cmd [heroes...]', 'a command', (yargs) => {
```
bcoe: I'm a tiny bit afraid of this change, as it seems to potentially be breaking (not a blocker, but might mean this needs to be released in a major).

Is the issue that `yargs.getOptions().defaultDescription` used to maintain state from the last command to execute, and now with this fix the freezing behavior reverts it to an initial state?

mleguen: That's exactly the point.

And the test I corrected was using the command state leaking to the parent yargs to perform its checks.

So the fix does not break any documented use case... but it could break some programs relying on that leaking command state, as our own test did.

So... major release or not?

bcoe: I say major, not a big deal, we're at 13.x already :stuck_out_tongue:

With tools like renovate bot and greenkeeper, I'd rather stay on the side of caution ... it lets us catch major bugs sooner.

PR #1371, opened 2019-07-01
docs: add section on dashes in string arguments

### LINKED ISSUE #1347: add section to docs explaining common hiccups people bump into, like bash quote handling
I'm trying to have an option that provides a string, which may start with `--`. I'm testing it now and it looks like when I pass `--` just after an option it is seen as a totally new option. See below the command I'm using and the code. I've also tried using double quotes to surround the argument, but with the same result. I'm just getting to know yargs, so maybe I'm missing something here. 

What I'd expect/want to see is that when I run `node ./dist/cli.js --thing '--hello'`, the `thing` option in `argv` would be `--hello`. When the first character is anything else than `-` it also works fine, so I can do `node ./dist/cli.js --thing '.--hello'` to result in `{ _: [], thing: '.--hello you', '$0': 'dist/cli.js' }`.

```shell
$ node ./dist/cli.js --thing '--hello'
Usage: cli.js <cmd> [options]

Options:
  --thing    Add this option!                                           [string]

Unknown argument: hello
```

```ts
import yargs from 'yargs';

const argv = yargs
  .usage('Usage: $0 <cmd> [options]')
  .strict(true)
  .options({
    thing: {
      describe: 'Add this option!',
      type: 'string',
    },
  }).argv;

console.log(argv);
```

vidavidorra:
Closes #1347

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 <p align="center">
   <b >Yargs be a node.js library fer hearties tryin' ter parse optstrings</b>
 </p>
+
 <br>
 
 [![Build Status][travis-image]][travis-url]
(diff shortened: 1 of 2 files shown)


vidavidorra:
@bcoe Please let me know if this is what you had in mind for these docs.

bcoe:
@vidavidorra thank you for the contribution 👍 sorry I've been a bit slow on OSS these days.

PR #1376, opened 2019-07-16
fix: tolerate null prototype for config objects with `extends`

### LINKED ISSUE #1372: Using `hasOwnProperty` breaks for config objects with null prototype
I'm loading a config object from a third-party library, then passing it to yargs.  This library happens to return an object with a null prototype.  yargs breaks because it's trying to call `config.hasOwnProperty`.

https://github.com/yargs/yargs/blob/master/lib/apply-extends.js#L22

coreyfarrell:
Fixes #1372

### DIFF
--- a/lib/apply-extends.js
+++ b/lib/apply-extends.js
@@ -19,7 +19,7 @@ function getPathToDefaultConfig (cwd, pathToExtend) {
 function applyExtends (config, cwd) {
   let defaultConfig = {}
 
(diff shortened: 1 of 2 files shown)


bcoe:
@coreyfarrell 👍 thank you.

bcoe:
@coreyfarrell @cspotcode please try `npm i yargs@next`, it should address this bug 👍

PR #1377, opened 2019-07-17
fix: Use decamelize from npm instead of vendored copy

coreyfarrell:
@bcoe yargs-parser already depends on decamelize@1.2.0 (and it's always `require()`'ed).  So this doesn't add anything to the node_modules directory and it avoids loading two copies of decamelize into memory by making yargs and yargs-parser use the same module.

Note that `decamelize@3` is available (requires node.js 6).  This is not used because it pulls in `xregexp` for unicode regex support (9.5MB!!).  xregexp will be dropped in a future version when node.js 10 is required.


bcoe:
@coreyfarrell this is also landed in `yargs@next`.

PR #1382, opened 2019-07-29
fix! yargs.parsed now populated before returning, when yargs.parse() called with no args

bcoe:
BREAKING CHANGE: we now only officially support yargs.$0 parameter and discourage direct access to yargs.parsed

https://github.com/yargs/yargs/pull/1369 and https://github.com/yargs/yargs/pull/1366 are incompatible, because `1369` unfreezes properties before returning (at which point yargs.parsed is the original value of `false`).

I would like to discourage users from accessing these properties on yargs, I've added a hack to make behavior the same as before, but now warn when `yargs.parsed` is accessed in singleton mode.

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,7 +28,13 @@ function singletonify (inst) {
     } else if (typeof inst[key] === 'function') {
       Argv[key] = inst[key].bind(inst)
     } else {
-      Argv.__defineGetter__(key, () => inst[key])
+      Argv.__defineGetter__('$0', () => {
+        return inst.$0
+      })
+      Argv.__defineGetter__('parsed', () => {
+        console.warn('In future major releases of yargs, "parsed" will be a private field. Use the return value of ".parse()" or ".argv" instead')
+        return inst.parsed
+      })
     }
   })
 }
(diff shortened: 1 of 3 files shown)


codemile:
>In future major releases of yargs, "parsed" will be a private field. Use the return value of ".parse()" or ".argv" instead

@bcoe the getter is triggering the warning for everyone who is using TypeScript.

For example;

```
import * as yargs from 'yargs';
```

The above will *copy* all of the exported properties from the module to a variable named `yargs`, and since the *getter* raises this warning it is triggered.

codemile:
I was able to turn off the message by changing 

```typescript
import * as yargs from 'yargs';
```

to this

```typescript
import yargs = require('yargs');
```

mleguen:
@bcoe just removed the warning in !1412 not to trigger error when importing in Typescript.

@codemile could you please confirm that with the version in the `master` branch the issue with `import *` is gone?

igloocube:
@mleguen I have just found this issue myself and can confirm when using master instead that the message is no longer displayed.

mleguen:
Thanks @igloocube

PR #1384, opened 2019-07-29
Support multiple epilogs being added

cameronhunter:
# Context
We wrap our commands in a function that adds `docs`, `maintainer`, and `slack` properties to the command module object. We use these properties as part of an automatically generated epilog that we add to each command.

# Problem
Currently subsequent calls to `epilog` will overwrite any previous epilog.

# Fix
Store epilogs in an array which is added to on each `epilog` call. Join them with new lines when building usage.

### DIFF
--- a/test/usage.js
+++ b/test/usage.js
@@ -1910,7 +1910,26 @@ describe('usage tests', () => {
     it('should display an epilog message at the end of the usage instructions', () => {
       const r = checkUsage(() => yargs('')
         .epilog('for more info view the manual at http://example.com')
+        .demand('y')
+        .wrap(null)
+        .parse()
+      )
+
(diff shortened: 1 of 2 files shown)


test/usage.js
```diff
@@ -1921,6 +1922,7 @@ describe('usage tests', () => {
         '  --version  Show version number  [boolean]',
         '  -y  [required]',
         'for more info view the manual at http://example.com',
+        'you can also find us on slack at http://devtoolscommunity.herokuapp.com',
```
bcoe: perhaps add one additional test with multiple epilogs.

cameronhunter: Sure thing. Added.

bcoe:
@cameronhunter likewise, available in `yargs@next`.

PR #1385, opened 2019-07-29
Allow completionCommand to be set via showCompletionScript

cameronhunter:
### Problem
When using `showCompletionScript` the variable `completionCommand` is `null` resulting in a bad installation instruction. For example:

`yargs.showCompletionScript('/usr/local/bin/my-app');`
```shell
###-begin-my-app-completions-###
#
#
# Installation: /usr/local/bin/my-app null >> ~/.bashrc
#    or /usr/local/bin/my-app null >> ~/.bash_profile on OSX.
#
_yargs_completions()
{
    local cur_word args type_list

    args=("${COMP_WORDS[@]}")

    type_list=$(/usr/local/bin/my-app --get-yargs-completions "${args[@]}")

      COMPREPLY=()
    fi

    return 0
}
complete -o default -F _yargs_completions my-app
###-end-my-app-completions-###
```

Notice that `null` is printed for the command.

It's possible to set the `completionCommand` variable by calling `yargs.completion()` but this will also log the completion script resulting in two scripts being logged when used in conjunction with `showCompletionScript`:

`yargs.completion().showCompletionScript('/usr/local/bin/my-app');`
```shell
###-begin-my-app-completions-###
#
#
# Installation: /usr/local/bin/my-app completion >> ~/.bashrc
#    or /usr/local/bin/my-app completion >> ~/.bash_profile on OSX.
#
_yargs_completions()
{
    local cur_word args type_list

    args=("${COMP_WORDS[@]}")

    type_list=$(/usr/local/bin/my-app --get-yargs-completions "${args[@]}")

      COMPREPLY=()
    fi

    return 0
}
complete -o default -F _yargs_completions my-app
###-end-my-app-completions-###

###-begin-my-app-completions-###
#
#
# Installation: my-app completion >> ~/.bashrc
#    or my-app completion >> ~/.bash_profile on OSX.
#
_yargs_completions()
{
    local cur_word args type_list

    args=("${COMP_WORDS[@]}")

    type_list=$(my-app --get-yargs-completions "${args[@]}")

      COMPREPLY=()
    fi

    return 0
}
complete -o default -F _yargs_completions my-app
###-end-my-app-completions-###
```

In this case the first completion script output is correct but the second one is incorrect.

### Fix
Add a second parameter to `showCompletionScript` that allows the developer to set the completion command variable.

`yargs.showCompletionScript('/usr/local/bin/my-app', 'completion');`
```shell
###-begin-my-app-completions-###
#
#
# Installation: /usr/local/bin/my-app completion >> ~/.bashrc
#    or /usr/local/bin/my-app completion >> ~/.bash_profile on OSX.
#
_yargs_completions()
{
    local cur_word args type_list

    args=("${COMP_WORDS[@]}")

    type_list=$(/usr/local/bin/my-app --get-yargs-completions "${args[@]}")

      COMPREPLY=()
    fi

    return 0
}
complete -o default -F _yargs_completions my-app
###-end-my-app-completions-###
```

It will default to "completion" in the same way that `completion()` does.

### DIFF
--- a/test/completion.js
+++ b/test/completion.js
@@ -158,6 +158,13 @@ describe('Completion', () => {
 
       r.logs[0].should.match(/\.\/test.js --get-yargs-completions/)
     })
+
+    it('allows $0 and cmd to be set', () => {
+      const r = checkUsage(() => yargs([])
+        .showCompletionScript('/path/to/my/app', 'show-completions-script'), ['test.js'])
+
+      r.logs[0].should.match(/Installation: \/path\/to\/my\/app show-completions-script/)
+    })
   })
 
   describe('completion()', () => {
--- a/test/completion.js
+++ b/test/completion.js
@@ -161,7 +161,7 @@ describe('Completion', () => {
 
     it('allows $0 and cmd to be set', () => {
       const r = checkUsage(() => yargs([])
-        .showCompletionScript('/path/to/my/app', 'show-completions-script'), ['test.js'])
+        .showCompletionScript('/path/to/my/app', 'show-completions-script'))
 
       r.logs[0].should.match(/Installation: \/path\/to\/my\/app show-completions-script/)
     })
--- a/yargs.js
+++ b/yargs.js
@@ -907,7 +907,7 @@ function Yargs (processArgs, cwd, parentRequire) {
     }
 
     // register the completion command.
-    completionCommand = cmd || 'completion'
+    completionCommand = cmd || completionCommand || 'completion'
     if (!desc && desc !== false) {
       desc = 'generate completion script'
     }
@@ -919,9 +919,10 @@ function Yargs (processArgs, cwd, parentRequire) {
     return self
   }
 
-  self.showCompletionScript = function ($0) {
-    argsert('[string]', [$0], arguments.length)
+  self.showCompletionScript = function ($0, cmd) {
+    argsert('[string] [string]', [$0, cmd], arguments.length)
     $0 = $0 || self.$0
+    completionCommand = cmd || completionCommand || 'completion'
     _logger.log(completion.generateCompletionScript($0, completionCommand))
     return self
   }


yargs.js
```diff
@@ -907,7 +907,7 @@ function Yargs (processArgs, cwd, parentRequire) {
     }
 
     // register the completion command.
-    completionCommand = cmd || 'completion'
```
bcoe: I'm slightly confused, wouldn't this have defaulted to `completion`? where was the `null` value coming from?

test/completion.js
```diff
@@ -158,6 +158,13 @@ describe('Completion', () => {
 
       r.logs[0].should.match(/\.\/test.js --get-yargs-completions/)
     })
+
+    it('allows $0 and cmd to be set', () => {
+      const r = checkUsage(() => yargs([])
+        .showCompletionScript('/path/to/my/app', 'show-completions-script'), ['test.js'])
```
bcoe: I don't think you need the `['test.js]`, as yargs is being called with empty arguments already right?

cameronhunter:
Our use-case is that we want to set `$0` and the completion command name. There are two APIs that _could_ achieve this: [`showCompletionScript`](http://yargs.js.org/docs/#api-showcompletionscript) and [`completion`](http://yargs.js.org/docs/#api-completioncmd-description-fn).

#### `showCompletionScript($0)`
Currently `showCompletionScript($0)` supports setting `$0` but not the completion command. It has a bug where it uses the variable `completionCommand` which will be `null` ([source](https://github.com/yargs/yargs/blob/07a5554c480ff3aa43cdc37b1d4ecbaf2687c779/yargs.js#L898)) since it can only be set via `completion` ([source](https://github.com/yargs/yargs/blob/07a5554c480ff3aa43cdc37b1d4ecbaf2687c779/yargs.js#L911)) resulting in the installation instructions showing `null` as the command completion:

```
###-begin-tvui-completions-###
#
#
# Installation: tvui null >> ~/.bashrc
#    or tvui null >> ~/.bash_profile on OSX.
#
```

#### `completion(cmd)`
Currently `completion` allows us to set the completion command, but not `$0`. For us this results in the completion script not working because our yargs-powered command-line app isn't on the global path.

#### Can we use `completion` and `showCompletionScript` together to resolve this issue?
I gave this a go because I was hoping that I could set the completion command using `completion('my-completion-command')` and then set `$0` when outputting with `showCompletionScript('/path/to/my/app')`. Unfortunately both of these functions output the completion script 😭

#### How I ended up where I ended up
In the end I added a second parameter to `showCompletionScript` to set the completion command name. I also fixed the `null` issue by defaulting the completion command to "completion" in the same manner as the `completion` function does.

yargs.js
cameronhunter: Yeah, I'm not adding any custom completion options, I'm wanting to leverage yargs' built-in completions. Because of this I never call the `completion()` function, just `showCompletionScript()`. The completion command will always be `null` unless `completion()` is called (source [1](https://github.com/yargs/yargs/blob/07a5554c480ff3aa43cdc37b1d4ecbaf2687c779/yargs.js#L898), [2](https://github.com/yargs/yargs/blob/07a5554c480ff3aa43cdc37b1d4ecbaf2687c779/yargs.js#L911)).

bcoe:
@cameronhunter excited to have the contributions, thank you 👍 will work on getting a release out soon.

bcoe:
@cameronhunter please try `npm i yargs@next` on a few projects, and let me know if everything is working as expected 👍 thank you for the contributions.

cameronhunter:
Thanks for merging, though I think this PR isn't acting as expected 😓 – I'm seeing the script being logged twice.

I _think_ it's due to me setting `completionCommand` in `showCompletionScript` and then this [`yargs.js#1106-1110`](https://github.com/yargs/yargs/blob/9febcbe61be4098cbb112843b899e80d2a6bb35c/yargs.js#L1106) codepath.

I'm trying to piece together a test-case but I think the correct fix is _not_ to set `completionCommand`. I'll follow-up with a PR.

PR #1386, opened 2019-07-30
feat: formalize existing callback argument to showHelp

royra:
This PR adds doc and test for an existing feature: Supplying a print callback to `usage.showHelp` instead of a `console` function name.

We use the callback in our code to explicitly specify the stdout/stderr stream dependencies of yargs - tests can use it instead of mocking the console.

Thanks for making this great library!

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -951,7 +951,7 @@ To submit a new translation for yargs:
 <a name="middleware"></a>.middleware(callbacks, [applyBeforeValidation])
 ------------------------------------
 
-Define global middleware functions to be called first, in list order, for all cli command.  
+Define global middleware functions to be called first, in list order, for all cli command.
 
 The `callbacks` parameter can be a function or a list of functions.  Each callback gets passed a reference to argv.
 
@@ -1310,10 +1310,12 @@ Generate a bash completion script. Users of your application can install this
 script in their `.bashrc`, and yargs will provide completion shortcuts for
 commands and options.
 
-.showHelp(consoleLevel='error')
+.showHelp()
 ---------------------------
 
-Print the usage data using the [`console`](https://nodejs.org/api/console.html) function `consoleLevel` for printing.
+Print the usage data to stderr.
+
+Later on, `argv` can be retrieved with `yargs.argv`.
 
 Example:
 
@@ -1323,13 +1325,27 @@ var yargs = require("yargs")
 yargs.showHelp(); //prints to stderr using console.error()
 ```
 
-Or, to print the usage data to `stdout` instead, you can specify the use of `console.log`:
+.showHelp(consoleLevel)
+---------------------------
+
+Print the usage data using the specified [`console`](https://nodejs.org/api/console.html) function for printing.
+
+Example:
 
 ```js
 yargs.showHelp("log"); //prints to stdout using console.log()
 ```
 
-Later on, `argv` can be retrieved with `yargs.argv`.
+.showHelp(printCallback)
+---------------------------
+
+Print the usage data using the specified callback, which accepts a single argument - the string to print.
+
+Example:
+
+```js
+yargs.showHelp(s => myStream.write(s)); //prints to myStream
+```
 
 .showHelpOnFail(enable, [message])
 ----------------------------------
--- a/docs/api.md
+++ b/docs/api.md
@@ -1310,14 +1310,12 @@ Generate a bash completion script. Users of your application can install this
 script in their `.bashrc`, and yargs will provide completion shortcuts for
 commands and options.
 
-.showHelp()
+.showHelp([consoleLevel | printCallback])
 ---------------------------
 
-Print the usage data to stderr.
+Print the usage data.
 
-Later on, `argv` can be retrieved with `yargs.argv`.
-
-Example:
+If no argument is provided, usage data is printed using `console.error`.
 
 ```js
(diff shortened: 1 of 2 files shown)


docs/api.md
```diff
@@ -1331,6 +1331,17 @@ yargs.showHelp("log"); //prints to stdout using console.log()
 
 Later on, `argv` can be retrieved with `yargs.argv`.
 
+.showHelp(printCallback)
+---------------------------
+
+Print the usage data using the specified callback, which accepts a single argument - the string to print.
```
bcoe: If I recall, can `showHelp` take no callback and it will result in help printing to `stdout`?

royra: @bcoe Right. Added doc section (and test) to explicitly describe that.

docs/api.md
```diff
@@ -1310,10 +1310,12 @@ Generate a bash completion script. Users of your application can install this
 script in their `.bashrc`, and yargs will provide completion shortcuts for
 commands and options.
 
-.showHelp(consoleLevel='error')
+.showHelp()
```
bcoe: To be consistent with the rest of the document I might put something like:

`.showHelp([level | printCallback])`

If no argument is provided ... your description.

If a string argument is provided ... your description

If a `printCallback` is provided ... your description.


thanks for your help fleshing out these docs :ok_hand:

royra: Fixed according to your suggestion.

Not sure the docs are very consistent in this matter.

Examples of the "multiple line" syntax: [.help](https://github.com/yargs/yargs/blob/master/docs/api.md#help), [.showHidden](https://github.com/yargs/yargs/blob/master/docs/api.md#showhidden), [.version](https://github.com/yargs/yargs/blob/master/docs/api.md#version).

I could only find a few examples of the "pipe" syntax you suggested, and they were used in scenarios with a matrix of options: `.help`, `.showHidden`, [.require/.required](https://github.com/yargs/yargs/blob/master/docs/api.md#requirekey-msg--boolean), [.usage](https://github.com/yargs/yargs/blob/master/docs/api.md#usagemessagecommand-desc-builder-handler), `.version`.

Perhaps we need a stricter syntax? We can use the [typings from @types/yargs package](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/yargs/index.d.ts) as a starting point.

docs/api.md
```diff
@@ -1310,25 +1310,31 @@ Generate a bash completion script. Users of your application can install this
 script in their `.bashrc`, and yargs will provide completion shortcuts for
 commands and options.
 
-.showHelp(consoleLevel='error')
+.showHelp([consoleLevel | printCallback])
 ---------------------------
 
-Print the usage data using the [`console`](https://nodejs.org/api/console.html) function `consoleLevel` for printing.
```
mleguen: This change sould not remove the default value of `consoleLevel` in the signature.

docs/api.md
```diff
@@ -1310,25 +1310,31 @@ Generate a bash completion script. Users of your application can install this
...
+Print the usage data.
 
-Example:
```
mleguen: Shouldn't we keep `Example:` before examples?

docs/api.md
```diff
@@ -1310,25 +1310,31 @@ Generate a bash completion script. Users of your application can install this
...
+If no argument is provided, usage data is printed using `console.log`.
```
mleguen: Isn't it using `console.error` by default?

test/usage.js
```diff
@@ -2345,7 +2345,27 @@ describe('usage tests', () => {
       ])
     })
 
-    it('should call the correct console.log method', () => {
+    it('should print the help to stdout when no arguments were specified', () => {
```
mleguen: Isn't it `stderr` rather than `stdout` by default?

And is the wording `stdout`/`stderr` pertinent when we use `console.log`/`.error` which may have been redefined by the calling code?

docs/api.md
royra: It's not removing the default value - see the second paragraph under "if no argument is provided". I fixed it according to the [comment](https://github.com/yargs/yargs/pull/1386#discussion_r309314320) by @bcoe. LMK if you have another specific change in mind.

docs/api.md
royra: Right now there is an example for every option - no args, string arg, and function arg. Would you like me to move the examples to separate section after all the text?

docs/api.md
royra: Right, fixed.

test/usage.js
royra: Fixed to "should print the help using console.error".

docs/api.md
mleguen: You're right, sorry.

docs/api.md
mleguen: No, that's okay. Most examples in the page does seem to have this anyway.

docs/api.md
mleguen: I agree with @royra about the need of not reinventing a syntax of our own, and using widely used syntax as typescript or [jsdoc](https://jsdoc.app/tags-type.html).

I would suggest merging this PR as is, and opening a separate issue to review all docs to apply a standard typing syntax.

@bcoe are you OK with this?

bcoe:
@royra your changes are in `npm i yargs@next`.

PR #1388, opened 2019-07-31
Don't set stateful completionCommand in showCompletionScript

cameronhunter:
I tried to write a test case for this but couldn't find one that reproduced the issue. I _think_ the issue only exists when using `yargs` as a singleton. The issue is that two completion scripts are logged, one from the explicit call to `showCompletionScript` and the second from [`yargs.js#1108`](https://github.com/yargs/yargs/blob/9febcbe61be4098cbb112843b899e80d2a6bb35c/yargs.js#L1108).

### DIFF
--- a/yargs.js
+++ b/yargs.js
@@ -923,8 +923,7 @@ function Yargs (processArgs, cwd, parentRequire) {
   self.showCompletionScript = function ($0, cmd) {
     argsert('[string] [string]', [$0, cmd], arguments.length)
     $0 = $0 || self.$0
-    completionCommand = cmd || completionCommand || 'completion'
-    _logger.log(completion.generateCompletionScript($0, completionCommand))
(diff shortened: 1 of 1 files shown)


bcoe:
@cameronhunter agreed that it seems weird that `completionCommand` is set when call the `show` method ... what was the side effect this was causing for you?

cameronhunter:
The side-effect was two completion scripts being logged.

bcoe:
@cameronhunter I'm definitely a bit confused by this emergent bug :stuck_out_tongue_closed_eyes:, but I've seen stranger things.

bcoe:
@cameronhunter could you please try:

```
npm i yargs@next
```

It should have your fix 👍

bcoe:
@cameronhunter this is now released in `yargs@14.2.0`.

PR #1394, opened 2019-08-03
ES6 class as a command

hnw:
Currently, we can use Node.js module as a yargs command. However, ES6 class cannot be applied to `.command()` because yargs judges whether module or not with `typeof cmd === 'object'` but ES6 class is `'function'`.

Many Node.js modules return `object` as a top level module, so they have no problem when someone want to use them as yargs commands. However some modules return ES6 class, and we cannot make them as yargs commands.

In this Pull Request, I propose that not only plain objects but also ES6 classes are accepted as yargs commands.

For example, suppose XyzBank class.

```javascript
class XyzBank {
  constructor(spec) {
  }
  withdraw(amount) {
  }
  deposit(amount) {
  }

  /* helper methods for yargs */
  static get command() {
    return 'xyz <command>'
  }
  static get description() {
    return 'banking operation for Xyz bank'
  }
  static builder(yargs) {
    return yargs
      .middleware(argv => argv._obj = new XyzBank(argv))
      .command({
        command:  'withdraw <amount>',
        describe: 'withdraw money',
        handler:  async (argv) => await argv._obj.withdraw(argv.amount)
      })
      .command({
        command:  'deposit <amount>',
        describe: 'deposit money',
        handler:  async (argv) => await argv._obj.desposit(argv.amount)
      })
  }
}

module.exports = XyzBank;
```

Then, we can instantiate this class as usual, and we also use it as a yargs command.

```javascript
yargs.command('$0', 'default command', require('./banks/xyzbank');
```

or

```javascript
yargs.commandDir('./banks')
```

Such capability should be convenient for many yargs users.

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -26,7 +26,7 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
     if (Array.isArray(cmd)) {
       aliases = cmd.slice(1)
       cmd = cmd[0]
-    } else if (typeof cmd === 'object') {
+    } else if (typeof cmd === 'object' || typeof cmd === 'function') {
       let command = (Array.isArray(cmd.command) || typeof cmd.command === 'string') ? cmd.command : moduleName(cmd)
       if (cmd.aliases) command = [].concat(command).concat(cmd.aliases)
       self.addHandler(command, extractDesc(cmd), cmd.builder, cmd.handler, cmd.middlewares)
--- a/lib/command.js
+++ b/lib/command.js
@@ -34,7 +34,7 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
     }
 
     // allow a module to be provided instead of separate builder and handler
-    if (typeof builder === 'object' && builder.builder && typeof builder.handler === 'function') {
+    if ((typeof builder === 'object' || typeof builder === 'function') && builder.builder && typeof builder.handler === 'function') {
       self.addHandler([cmd].concat(aliases), description, builder.builder, builder.handler, builder.middlewares)
       return
     }


mleguen:
In the example you gave, couldn't you achieve the same goal by splitting your module in two, one for the `XyzBank` class and one for the yargs command object?

Wouldn't that be as concise and more readable (by separating business and technical stuff)?

craigphicks:
Couldn't you build an ES wrapper around `yargs` where with interface you want? 
As it happens, I was wondering about separate `yargs` with separate states and it seems to overlap with your conversation.  See [post](https://github.com/yargs/yargs/issues/1402#issuecomment-521767817).  

(NOTE: the following was run with the node modules flag
```
node --experimental-modules ...
```
)

The conversion to an own-state ES6 class is - 
```
'use strict'
class YargsClass {
    constructor(yargs){
        this.yargs=yargs
    }
}
export default async function createYargsClass(){
    var yargs = await import('../node_modules/yargs/index.js')
    return new YargsClass(yargs)
}
```

Because `yargs` is imported inside `createYargsClass` the `yargs` state is individualized for the returned class.  Perhaps this might work as a way to pass the `argv` as an argument in - 
```
export default async function createYargsClass(argvIn){
     var process = { argv:argvIn }
    var yargs = await import('../node_modules/yargs/index.js')
    return new YargsClass(yargs)
}
```
(but I haven't tested it).

Anyway, you write all interfaces you want  into the ES6 `YargsClass`.

craigphicks:
@hnw 

Now applying this to your bank project, with a slight change in perspective:

```
class XyzBankTransaction extends YargsClass {
  constructor(commandLine) {
    super(commandLine)
  }
  withdraw(amount) {
  }
  deposit(amount) {
  }

  /* helper methods for yargs */
  static get command() {
    return 'xyz <command>'
  }
  static get description() {
    return 'banking operation for Xyz bank'
  }
  static builder() {
    return this.yargs
      .middleware(argv => argv._obj = new XyzBank(argv))
      .command({
        command:  'withdraw <amount>',
        describe: 'withdraw money',
        handler:  async (argv) => await argv._obj.withdraw(argv.amount)
      })
      .command({
        command:  'deposit <amount>',
        describe: 'deposit money',
        handler:  async (argv) => await argv._obj.desposit(argv.amount)
      })
  }
}

module.exports = XyzBankTransaction;
```

Suppose your command lines are arriving in a stream, and they will processed asynchronously, which appears to be the case as your `handler`s are asynchronous.  Then the processing of transaction may overlap and `yargs` singleton model will cause errors.  It is necessary to have a separate `yargs` object for each transaction - and that is what the `YargsClass` provides.
 
Maybe it should be called `YargsNonSingletonClass`.

Thanks for your ideas and yoroshiku onegaishimasu.

bcoe:
@mleguen I'm not as opposed to supporting classes, and eventually ESM modules, build I'd like the loading logic to be refactored a bit so that it actually checks for the appropriate keys, and reports an error appropriately when they're missing.

I think this would be good for objects as well.

lib/command.js
```diff
@@ -26,15 +26,15 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
     if (Array.isArray(cmd)) {
       aliases = cmd.slice(1)
       cmd = cmd[0]
-    } else if (typeof cmd === 'object') {
+    } else if (typeof cmd === 'object' || typeof cmd === 'function') {
```
bcoe: let's pull this into a helper that insures, if `cmd` is a `function`, it has appropriate fields, e.g.,

It should have a `description`, or `builder`, or `handler` or `middlewares`.

Perhaps we could apply the same check for when `typeof cmd === 'object'`, and simply do:

```
else if (isCommandDescription(cmd)) {
      let command = (Array.isArray(cmd.command) || typeof cmd.command === 'string') ? cmd.command : moduleName(cmd)
      if (cmd.aliases) command = [].concat(command).concat(cmd.aliases)
      self.addHandler(command, extractDesc(cmd), cmd.builder, cmd.handler, cmd.middlewares)
}
```

lib/command.js
```diff
@@ -26,15 +26,15 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
...
       let command = (Array.isArray(cmd.command) || typeof cmd.command === 'string') ? cmd.command : moduleName(cmd)
       if (cmd.aliases) command = [].concat(command).concat(cmd.aliases)
       self.addHandler(command, extractDesc(cmd), cmd.builder, cmd.handler, cmd.middlewares)
       return
     }
 
     // allow a module to be provided instead of separate builder and handler
-    if (typeof builder === 'object' && builder.builder && typeof builder.handler === 'function') {
+    if ((typeof builder === 'object' || typeof builder === 'function') && builder.builder && typeof builder.handler === 'function') {
```
bcoe: Similarly here, we can create a similar helper that's `isBuilderAndHandler`.

bcoe:
@hnw sorry for the slow feedback, I think support for `classes` is a good idea, I added a couple suggestions that address my concerns.

bcoe:
@hnw I'm still open to adding this feature, please feel free to reopen if you have time to see the work over the finish line 👍

PR #1395, opened 2019-08-10
fix: detect zsh when zsh isnt run as a login prompt

forivall:
I don't run zsh as a login prompt, and I haven't `chsh`'d from bash to zsh, and so yargs was still giving me the bash completion when running it inside of zsh.

### DIFF
--- a/lib/completion.js
+++ b/lib/completion.js
@@ -8,7 +8,8 @@ module.exports = function completion (yargs, usage, command) {
     completionKey: 'get-yargs-completions'
   }
 
-  const zshShell = process.env.SHELL && process.env.SHELL.indexOf('zsh') !== -1
+  const zshShell = (process.env.SHELL && process.env.SHELL.indexOf('zsh') !== -1) ||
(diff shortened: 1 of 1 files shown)


bcoe:
@forivall thanks for the contribution :+1: I don't use `zsh` personally, so it's great to have someone kick the tires of the completion logic.

Like @mleguen says, tests are failing because of linting, you might be able to simply run:

```
./node_modules/.bin/standard --fix
```

If you don't get around to fixing the linting, I'll work on landing this soon myself.

forivall:
Whoops, yeah, I forgot about that comment. I'll get to it tomorrow. Thanks!!

~ Emily Klassen

forivall:
the `--fix` flag with standard didn't auto-fix. So i added the parens manually, and now it passes the standard check.

bcoe:
@forivall thank you for the contribution :smile:

bcoe:
@forivall this is now released in `yargs@14.2.0`.

PR #1396, opened 2019-08-10
chore: Minor refactor

jamesgeorge007:
refactor code to a more concise version.

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -209,6 +209,24 @@ module.exports = function validation (yargs, usage, y18n) {
     return implied
   }
 
+  function manipulateImplications (num, val) {
+  	// convert string '1' to number 1
+  	num = Number(val)
+    val = isNaN(num) ? val : num
+
+    if (typeof val === 'number') {
+      // check length of argv._
+      val = argv._.length >= val
+    } else if (val.match(/^--no-.+/)) {
+      // check if key/value doesn't exist
+      val = val.match(/^--no-(.+)/)[1]
+      val = !argv[val]
+      } else {
+        // check if key/value exists
+        val = argv[val]
+      }
+  }
+
   self.implications = function implications (argv) {
     const implyFail = []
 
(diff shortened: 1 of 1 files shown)


jamesgeorge007:
cc @bcoe

bcoe:
@jamesgeorge007 looks like a reasonable refactor, but tests are failing because `argv` does not exist in the context of the new helper; I might also be tempted to call the helper something closer to what it does.

As an example, `coerceValueType`, or something like that.

jamesgeorge007:
@bcoe CI is still failing :thinking:

jamesgeorge007:
@mleguen still the build doesn't turn green :thinking:

bcoe:
@jamesgeorge007 if you clone this repository, and run:

```
npm i
npm t
```

Inside the folder, I think you should be able to see these exceptions happening on your local machine (without forcing CI/CD to run).

bcoe:
👋 @jamesgeorge007 is this still something you'd like to see over the finish line?

jamesgeorge007:
yeah

lib/validation.js
```diff
@@ -209,43 +209,36 @@ module.exports = function validation (yargs, usage, y18n) {
...
     const implyFail = []
 
     Object.keys(implied).forEach((key) => {
       const origKey = key
       ;(implied[key] || []).forEach((value) => {
-        let num
         let key = origKey
         const origValue = value
+        // eslint-disable-next-line
+        [key, value] = [key, value].map((item) => coerceValueType(argv, item))
```
jamesgeorge007: @bcoe It seems we aren't getting the intended values while destructuring `key` and `value`.

mleguen: The issue disappears if you do not use destructuring, but neither do I understand why, nor am I able to reproduce such a destructuring issue in a simpler code.

```suggestion
        key = coerceValueType(argv, key)
        value = coerceValueType(argv, value)
```

But in the end, IMHO, having 2 lines is more readable than using a map here...

lib/validation.js
```diff
@@ -209,43 +209,36 @@ module.exports = function validation (yargs, usage, y18n) {
     return implied
   }
 
+  function coerceValueType (argv, val) {
```
mleguen: I suggest renaming this function, as it is not coercing the value type, but checking its presence.

```suggestion
  function keyExists (argv, key) {
```

lib/validation.js
mleguen: And this way you can also drop the "please don't lint my hack" hint above ;-)

bcoe:
@mleguen feel free to land this when you're happy?

lib/validation.js
```diff
@@ -209,43 +209,37 @@ module.exports = function validation (yargs, usage, y18n) {
...
   self.implications = function implications (argv) {
     const implyFail = []
 
     Object.keys(implied).forEach((key) => {
       const origKey = key
       ;(implied[key] || []).forEach((value) => {
-        let num
         let key = origKey
         const origValue = value
+        // eslint-disable-next-line
```
mleguen: This hint should no longer be useful

mleguen:
Thanks @jamesgeorge007 

Could you please remove the hint which is no longer useful before I merge this?

jamesgeorge007:
@mleguen done :+1:

mleguen:
Thanks @jamesgeorge007

PR #1400, opened 2019-08-13
fix(docs): fix incorrect parserConfiguration documentation

rvolgers:
Looks like somebody copied the parser options from `package.json` and accidentally copied a bit too much. Hope I can save the next guy the 15 minutes it took me to figure this out...


bcoe:
@rvolgers thank you, good catch.

Issue #1323, opened 2019-03-22
When positional argument is named as some option conflict is not detected


AuHau:
I named a positional argument `version` which resolved in very weird behaviors. After quite some debugging I realized that during parsing you convert positional arguments into options yet you don't verify if there is already an option with the same name. In my case, there is a conflict with the built-in `--version` option.

I guess this is quite an edge-case, because having option and argument with same name does not really make sense, but there should be check for that.

terrarum:
I've run into this problem as well with yargs 14.0.0.

I have a subcommand that has the positional `<version>`.

Running `ds box set 2019` gives me 

```
{ _: [ 'box', 'set' ], '$0': 'devstack', version: true }
```

If I rename version to `verfsion` I get

```
{ _: [ 'box', 'set' ], '$0': 'devstack', verfsion: 2019 }
```

Same if I call the positional `<help>` vs `<helpp>`, I get true instead of the expected value.

bcoe:
@AuHau or @terrarum mind providing a minimal example of the yargs configuration that causes the bug your seeing?

terrarum:
Using yargs 14:

`index.js`:
```js
const yargs = require('yargs');

yargs.command(
  'set1 <version>',
  'Change version of some other thing',
  () => {},
  (argv) => {
    console.log(argv);
  }
);

yargs.command(
  'set2 <blursion>',
  'Change blursion of some other thing',
  () => {},
  (argv) => {
    console.log(argv);
  }
);

yargs.argv;
```

Running `node index.js set1 1.0` gives me 
```
{ _: [ 'set1' ], '$0': 'index.js', version: true }
```

Running `node index.js set2 1.0` gives me 
```
{ _: [ 'set2' ], '$0': 'index.js', blursion: 100 }
```

bcoe:
:+1: what's the proper behavior? I'm wondering if explicitly setting `version` or `help`, or a handful of other settings that are defaulted to being enabled, should just take precedence and cancel out their prior configuration.

I'm concerned about throwing an exception, as I believe it would break a lot of apps that are already in the wild.

terrarum:
In my scenario I'm writing a command that lets me set the version of an AMI that another tool should use, so `version` felt like the appropriate name for that positional.

[...]

PR #1404, opened 2019-08-18
Throw on help or version key reuse attempt

mleguen:
As stated in #1323 , one can be tempted to name a positional `help` or `version`, without knowing `yargs` already uses this keys by default. The resulting weird behavior is not easy to trace back to this.

This PR adds checks to throw an exception when trying to define an option or a positional with a key or an alias already used by the help and version options, and corresponding tests.

### DIFF
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -1959,6 +1959,122 @@ describe('yargs dsl tests', () => {
     })
   })
 
+  describe('.version() and .option()', () => {
+    it('prevents .option() from reusing the default version key', () => {
+      expect(() => {
+        yargs(['--version', 'test'])
+          .option('version', {
+            type: 'string'
+          })
+          .parse()
+      }).to.throw(/'version' key already used by version option/)
+    })
+
+    it('prevents .option() from reusing the default version key as an alias', () => {
+      expect(() => {
+        yargs(['--version', 'test'])
+          .option('someoption', {
+            alias: ['version'],
+            type: 'string'
+          })
+          .parse()
+      }).to.throw(/'version' key already used by version option/)
+    })
+
+    it('prevents .option() from reusing a custom version key', () => {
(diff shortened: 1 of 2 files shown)


bcoe:
@mleguen I'm so so on this change. For starters, help can be disabled by calling `.help(false)` or `.version(false)` which we would need to continue supporting.

My bigger concern is that there are probably thousands of programs using yargs in the wild that explicitly set `help`, and would now start throwing. A warning message would be lower impact, and would still draw attention to the issue.

bcoe:
I wonder if this is a case where doing nothing would be the best thing though :stuck_out_tongue_closed_eyes:

mleguen:
> For starters, help can be disabled by calling .help(false) or .version(false) which we would need to continue supporting

Sure, it can be disabled that way, but only once you understand your code is failing because of `showHelp` is called by default internally by yargs, and is already using the `help` key.

This MR does not change anything about supporting this. A test already check this for `.version(false)`, but I see now there is none for `.help(false)` so I am going to add it.

mleguen:
Dedicated tests added to prove we continue supporting normal uses of `.help` and `.version`.

I also removed some this PR added which I think unuseful (using an alias with positional).

bcoe:
I think throwing is probably too aggressive, what if we are just smart about not setting `.version` or `.help`, etc., if a user has configured them separately.

bcoe:
yargs@4.x is a bit of a mess for commands right now unfortunately, due to our attempts to better handle state between multiple calls to `parse()`, I'd like to avoid any more breaking changes like this until we can get things a bit more stable.

bcoe:
I'm definitely open to us figuring out a patch that protects folks from this bug, but probably the right thing to do will be making sure we don't stomp on user defined settings with **magic** settings.

Issue #1363, opened 2019-06-12
Question about extends vs cli params


soulgalore:
Hi! First I've been a using yargs for many years in sitspeed.io and it works really great. Thanks a lot!

Lately I've started using `extends` to extend other configurations and I have a question. Let me show an example:
**parent.json**
```
{
  "a": {
    "b": 1
  }
}
```

**child.json**
```
{
  "extends": "/path/to/parent.json",
  "a": {
    "c": 2
  }
}
```

In my case the the child overrides all properties on the parent, so c exists but not b. But if I use the child and add use a cli option like `--a.b 2` they are merged and both c and b exists on the object. Is there any reason that it behaves differently? 

The way we use extends is that on the server we keep some secrets in the one config file and then we have the rest of the config open in a git repo, so it's easy for users to see our full setup. But we have a couple of clashing namespaces that get overruled. 

Or is it a better way to do that?

Best
Peter

bcoe:
Hey Peter :wave: thanks for the kind words. I honestly haven't used the `extends` functionality that much personally :laughing: 

In general we could use a bit of an overall of how our dot notation functionality works ... it's buggy. I think that your example should be equivalent to:

```
--a.b 1 --a.c 2
```
I think I'd be open to a patch that corrects this behavior ... I don't think I'll have many cycles to add the functionality myself anytime soon unfortunately (so if anyone reads this and wants to contribute, your patch would be greatly appreciated).

bcoe:
@soulgalore just a thought, I might make this a configuration option (and default to the current behavior) ... for folks who are using this to build tools that except a configuration file (like a nycrc.json, or eslint.json) the merging _might_ be a surprise?

soulgalore:
Yes I see. It is not super important for me, I can have a go at a PR when I get back after vacation in the middle of August. At the moment we override with ENV instead, that works fine good for us.

PR #1411, opened 2019-08-23
feat: make it possible to merge configurations when extending other config.

soulgalore:
Default behavoir is as before. But you can add 'merge-extends': true
to your parser configuration that will change so that configurations
will be deeply merged instead of overwritten per key.

See https://github.com/yargs/yargs/issues/1363


yargs.js
```diff
@@ -330,7 +330,7 @@ function Yargs (processArgs, cwd, parentRequire) {
     argsert('[object|string] [string|function] [function]', [key, msg, parseFn], arguments.length)
     // allow a config object to be provided directly.
     if (typeof key === 'object') {
-      key = applyExtends(key, cwd)
+      key = applyExtends(key, cwd, self.getParserConfiguration()['merge-extends'])
```
soulgalore: Is this the correct way to pass on configuration and if so, what name do you prefer :)

lib/apply-extends.js
```diff
@@ -16,7 +16,22 @@ function getPathToDefaultConfig (cwd, pathToExtend) {
   return path.resolve(cwd, pathToExtend)
 }
 
-function applyExtends (config, cwd) {
+function mergeDeep (target, config1, config2) {
```
soulgalore: Could it be a smarter way to do this and do we miss any case? When I tried the deep merging code got quite complicated fast so tried to have as simple version as possible, but I'm not sure if I miss any cases?

mleguen: I could not think of a case you missed, and as it is a new feature, there is no risk you are going to break any existing code if you missed something.

mleguen: However, apart from mimicking `Object.assign`, what is the use of having a `target` parameter which is always `{}`? I would drop it.

lib/apply-extends.js
```diff
@@ -16,7 +16,22 @@ function getPathToDefaultConfig (cwd, pathToExtend) {
...
+  const isObject = obj => obj && typeof obj === 'object' && !Array.isArray(obj)
+  for (let key of Object.keys(config1)) {
```
mleguen: I'd replace that first `for` loop by an `object.assign`

```suggestion
  Object.assign(target, config1)
```

yargs.js
mleguen: As this is a yargs options, not related to the parser, I think you should not use `getParserConfiguration()`.

I suggest instead adding a `mergeExtends()` function to set a `mergeExtends` boolean variable (as `strict()` does for example).

lib/apply-extends.js
soulgalore: Thanks you are right, I've dropped it.

lib/apply-extends.js
soulgalore: Fixed.

yargs.js
soulgalore: Yes, I did fully understood how everything work together, thanks for pointing me in the right direction!

soulgalore:
@mleguen Thanks for taking the time to review! I've pushed an updated version, please let me know if there's something else I missed!

yargs.js
```diff
@@ -772,6 +772,14 @@ function Yargs (processArgs, cwd, parentRequire) {
   }
   self.getStrict = () => strict
 
+  let mergeExtends = false
+  self.mergeExtends = function (enabled) {
```
bcoe: rather than adding a new method `mergeExtends` let's add this as a new `. parserConfiguration`, I think a good name for the configuration setting would be: `deepMergeConfig`, so you'd do this:

```
yargs.parserConfig({'deep-merge-config': true})
```

soulgalore: Let me fix that. I just added an entry to https://github.com/yargs/yargs/blob/master/docs/advanced.md#customizing-yargs-parser but let me know if I should do something more.  Really looking forward to this if it gets merge, so I can release my depending projects :)

bcoe:
@soulgalore could you please try:

`npm i yargs@next` and let me know if it works for you use-cases?

soulgalore:
Thank you @bcoe ! I've tested and it works great!

soulgalore:
One thing though @bcoe - when I tested at work I run into an issue of a three step extend described here: https://phabricator.wikimedia.org/T232467
Where the new functionality doesn't work.

I'll add a new test case first thing tomorrow and verify.

mleguen:
@bcoe @soulgalore one of the tests added by this PR breaks on master branch :-(

```bash
  559 passing (37s)
  1 failing

  1) yargs dsl tests
       config
         extends
           deep merges configs when extending when deep-merge-config=true:
     AssertionError: expected 1 to deeply equal [ 1 ]
      at Context.it (test\yargs.js:1398:25)
```

soulgalore:
@mleguen Can you share some more context? It works when I tested on my local machine with a fresh checkout just now and on Travis: https://travis-ci.org/yargs/yargs/builds/581897033

mleguen:
@soulgalore sorry, it was a problem of outdated dependencies on my side

bcoe:
published in `yargs@14.2.0` sorry for the slow cadence lately; thanks for the contribution @soulgalore, let me know if you bump into any issues.

PR #1412, opened 2019-08-26
fix: populate correct value on yargs.parsed and stop warning on access

bcoe:
This should address a couple minor issues that lead to #1144:

1. `self.parsed` was being populated with an `argv` instance rather than a `yargsParser.detailed` instance, leading to the missing fields that raised exceptions.
2. once `1.` was addressed, the help message was incorrect, because `async` handlers prior to this relied on state not being reset; to solve this, we now cache the help message that _would_ be displayed at the time of a promise's invocation.
3. there was one final issue related to our `self.parsed`, TypeScript executes all methods on a class, making our warning message display even though it wasn't explicitly invoked (I've reverted the warning message).

*TypeScript Issue:*

```js
var __importStar = (this && this.__importStar) || function (mod) {
    if (mod && mod.__esModule) return mod;
    var result = {};
    if (mod != null) for (var k in mod) if (Object.hasOwnProperty.call(mod, k)) result[k] = mod[k];
    result["default"] = mod;
    return result;
};
```

@gajus could I get you to try this out?

fixes: #1144

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -232,7 +232,6 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
 
       let handlerResult
       if (isPromise(innerArgv)) {
-        yargs.getUsageInstance().cacheHelpMessage()
         handlerResult = innerArgv.then(argv => commandHandler.handler(argv))
       } else {
         handlerResult = commandHandler.handler(innerArgv)
--- a/lib/command.js
+++ b/lib/command.js
@@ -174,7 +174,7 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
     let numFiles = currentContext.files.length
     const parentCommands = currentContext.commands.slice()
 
-    // what does yargs look like after the buidler is run?
+    // what does yargs look like after the builder is run?
     let innerArgv = parsed.argv
     let innerYargs = null
     let positionalMap = {}
@@ -230,14 +230,19 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
 
       innerArgv = applyMiddleware(innerArgv, yargs, middlewares, false)
 
-      const handlerResult = isPromise(innerArgv)
-        ? innerArgv.then(argv => commandHandler.handler(argv))
-        : commandHandler.handler(innerArgv)
+      let handlerResult
+      if (isPromise(innerArgv)) {
+        yargs.getUsageInstance().cacheHelpMessage()
+        handlerResult = innerArgv.then(argv => commandHandler.handler(argv))
+      } else {
+        handlerResult = commandHandler.handler(innerArgv)
+      }
 
       if (isPromise(handlerResult)) {
-        handlerResult.catch(error =>
+        yargs.getUsageInstance().cacheHelpMessage()
+        handlerResult.catch(error => {
           yargs.getUsageInstance().fail(null, error)
-        )
+        })
       }
     }
 
--- a/lib/command.js
+++ b/lib/command.js
@@ -241,7 +241,11 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
       if (isPromise(handlerResult)) {
         yargs.getUsageInstance().cacheHelpMessage()
         handlerResult.catch(error => {
-          yargs.getUsageInstance().fail(null, error)
+          try {
+            yargs.getUsageInstance().fail(null, error)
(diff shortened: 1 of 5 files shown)


gajus:
Can confirm that this fixes the issue I have reported originally.

lib/command.js
```diff
@@ -230,14 +230,19 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
 
       innerArgv = applyMiddleware(innerArgv, yargs, middlewares, false)
 
-      const handlerResult = isPromise(innerArgv)
-        ? innerArgv.then(argv => commandHandler.handler(argv))
-        : commandHandler.handler(innerArgv)
+      let handlerResult
+      if (isPromise(innerArgv)) {
+        yargs.getUsageInstance().cacheHelpMessage()
```
mleguen: I do not understand why you cache help message here? Shouldn't caching it before `handlerResult.catch` below be enough? Removing this line does not break any test.

test/command.js
```diff
@@ -1426,36 +1426,66 @@ describe('Command', () => {
...
-      .parse()
+    })
 
-    handler.then(called => {
-      called.should.equal(true)
-      parsed.pos.should.equal('hello')
-      return done()
+    // see: https://github.com/yargs/yargs/issues/1144
+    it('displays error and appropriate help message when handler fails', (done) => {
+      // TODO: debug why exception bubbles to unhandledRejection
```
mleguen: The exception only shows up when exitProcess is set to false because when exitProcess is true, the process is killed by exit before being able to report about the unhandled rejection.

Once corrected the unhandled rejection (see comment on `lib/command.js`), the bad news is I cannot see how to write this test...

lib/command.js
```diff
@@ -230,14 +230,19 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
...
+        handlerResult = innerArgv.then(argv => commandHandler.handler(argv))
+      } else {
+        handlerResult = commandHandler.handler(innerArgv)
+      }
 
       if (isPromise(handlerResult)) {
-        handlerResult.catch(error =>
+        yargs.getUsageInstance().cacheHelpMessage()
+        handlerResult.catch(error => {
           yargs.getUsageInstance().fail(null, error)
```
mleguen: Your unhandled rejection comes from here: when exitProcess is set to false, the last thing fail does is throwing, so your catch handler throws itself, the promise is still rejected, and this rejection is not handled.
```suggestion
          try {
            yargs.getUsageInstance().fail(null, error)
          }
          catch() {
          }
```

mleguen:
@bcoe IMO, we have a conception issue here: we should not accept async handlers in sync chained functions.

I recommend:
- either deprecating the use of async command handlers, and ignoring command handler results, without trying to display help (which would be consistent with what happen when a sync handler throws: no help is displayed)
- or having runCommand and all depending functions become async too, which would break the ability to chain calls after such methods

gajus:
> or having runCommand and all depending functions become async too, which would break the ability to chain calls after such methods

What is the impact of breaking ability to chain methods?

The last suggestion sounds reasonable. First one sounds like a quick workaround that will leave up to the individual yargs consumer to figure out their own way of handling async errors.

mleguen:
This conception issue is what leads to your last test being so difficult to write.

Any client program using async handlers and wanting to perform actions after yargs is done parsing will have the same problem: we cannot know when yargs is done parsing when using async handler.

mleguen:
> What is the impact of breaking ability to chain methods?

In fact, I think it would only prevent chaining `showHelp()`, as `parse()` is already not chainable, and I cannot see any use case for chaining `showHelp()`.

However, the major impact would be that, when using async handlers, parse would no longer return directly the parsing results, but a promise. This would be great for people wanting to do things after parsing, but could break existing code.

Next major release?

mleguen:
In order not to block this fix, here is what I suggest:
- fix the unhandled rejection issue
- remove the last test, as not being writable yet
- merge this PR
- open a new issue about `parse()` and `showHelp()` having to return a promise when a command handler does so
- write a PR for this new issue, including the missing test (I can handle this)

@bcoe what is your opinion on this?

bcoe:
@mleguen ready for review 👍

test/command.js
mleguen: OK, nice way to write this test.

PR #1418, opened 2019-08-31
Removed "default: false" on verbose.

sliftist:
Proposal to remove "default: false" on verbose in the README.

I was mildly confused when I copied the example and called it with `-v`, expecting no value to default the value to true. However it appears as if the default value overrides both the "not passed" value, and the "passed with no value" parameter. This makes me believe "default: false" should never be used, as the default of default is effectively false.

At the very least I believe it shouldn't be part of the example in the README, especially on conjunction with the verbose parameter, which makes a program that uses verbose differently than any program I have ever seen.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,9 @@ require('yargs') // eslint-disable-line
     serve(argv.port)
   })
   .option('verbose', {
-    alias: 'v'
+    alias: 'v',
+    type: 'count',
+    description: 'Run with verbose logging. Additition 'v's (-vv, -vvv, etc) to run with even more verbose logging.'
   })
   .argv
 ```
--- a/README.md
+++ b/README.md
@@ -79,8 +79,7 @@ require('yargs') // eslint-disable-line
     serve(argv.port)
   })
   .option('verbose', {
-    alias: 'v',
-    default: false
+    alias: 'v'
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -79,8 +79,7 @@ require('yargs') // eslint-disable-line
     serve(argv.port)
   })
   .option('verbose', {
-    alias: 'v',
-    default: false
+    alias: 'v'
```
bcoe: perhaps add a `describe: 'run with verbose logging'`, I'm comfortable with dropping the `false` 👍

mleguen: If we want this example to show a standard way of defining a verbose option, I suggest also adding a `type: 'count'`, as it is the way a verbose option often works (v = 0 when the option is absent, 1 when it is present, up to n when it is repeated, 4 for example with `-vvvv`).

bcoe: I was thinking of suggesting this too, but thought it might be a bit too advanced of a feature to show off near the top of the README. I'm comfortable with whichever approach @sliftist takes 👍 but think it would be good to add a `description` either way.

sliftist: Cool, I didn't know `type: 'count'` worked like that. Both changes added to the pull request.

PR #1422, opened 2019-09-11
fix: #1335: async middleware called twice

### LINKED ISSUE #1335: Async middleware called twice
I need to refresh token on every command. As a solution i've been tried to create async middleware but encountered a problem: middleware called twice per command if i return a Promise or make async function. Simplified form:
middleware
```js 
const refreshTokenMiddleware = () => {
  console.log('middleware: start')
  return new Promise(resolve => setTimeout(resolve, 10)).then(() =>
    console.log('middleware: complete')
  )
}
```
register
```js
yargs
  .commandDir('commands')
  .middleware(refreshTokenMiddleware)
```
command handler
```js
exports.handler = async () => {
  console.log('handler:start')
  await new Promise(resolve => setTimeout(resolve, 10))
  console.log('handler:complete')
}

```
Here is the output:
```
middleware: start
middleware: complete
handler:start
middleware: start
middleware: complete
handler:complete
```

I have tried to trace down this behaviour, but without any luck (insufficient time).

mleguen:
Closes #1335

Async middleware where called:
- once for the innermost subcommand (expected)
- then once for every parent of this subcommand (unexpected)

Due to an incorrect check in the first run of middleware (before validation).

CI will probably fail as #1411 introduced a broken test in master.

### DIFF
--- a/lib/middleware.js
+++ b/lib/middleware.js
@@ -40,8 +40,7 @@ function applyMiddleware (argv, yargs, middlewares, beforeValidation) {
   const beforeValidationError = new Error('middleware cannot return a promise when applyBeforeValidation is true')
   return middlewares
     .reduce((accumulation, middleware) => {
-      if (middleware.applyBeforeValidation !== beforeValidation &&
-          !isPromise(accumulation)) {
+      if (middleware.applyBeforeValidation !== beforeValidation) {
         return accumulation
       }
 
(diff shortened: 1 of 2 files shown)


mleguen:
Woops! Forgot to add a test to prove the issue is fixed, coming soon...

mleguen:
Ready for review!

mleguen:
Thanks @bcoe 

I knew it... :-)

It should be OK now.

bcoe:
@mleguen released in `yargs@14.2.0`.

PR #1423, opened 2019-09-12
fix: support merging deeply nested configuration

soulgalore:
There was a bug in #1411 where I've missed passing on the merge config status, that mean that we only merged the first step if you make multiple extends as mentioned in https://github.com/yargs/yargs/pull/1411#issuecomment-529964443

I added a new test case, let me know if you want me to change the old one instead.


bcoe:
@soulgalore 🎉 this is released in `yargs@14.2.0` thanks for the contribution.

PR #1424, opened 2019-09-14
Fix promise check to accept any spec conform object

micha149:
I came across this problem when I was creating a CLI tool which uses [NodeGit](https://www.nodegit.org/) internally. NodeGit is using its own fork of es6 promises and so yargs failed to handle my async code because it didn't return an instance of `Promise`. But yargs should accept all promises as long as they are spec conform, regardless if its an native promise, a polyfill, a bluebird promise or any custom implementation.

So, I fixed the code to look for thenable objects instead of making an instanceof check.

### DIFF
--- a/test/is-promise.js
+++ b/test/is-promise.js
@@ -1,0 +1,34 @@
+'use strict'
+/* global describe, it */
+
+const { expect } = require('chai')
+const isPromise = require('../lib/is-promise')
+
+describe('isPromise', () => {
+  it('returns `false` on non promise value', () => {
+    expect(isPromise('hello there.')).to.be.equal(false)
+  })
+
+  it('returns `true` on es6 promise', () => {
+    const esPromise = new Promise(() => {})
+    expect(isPromise(esPromise)).to.be.equal(true)
+  })
+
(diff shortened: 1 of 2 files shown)


micha149:
😞Pipeline failed because of an timeout… Can somebody rebuild it?

test/is-promise.js
```diff
@@ -0,0 +1,30 @@
+'use strict'
+/* global describe, it */
+
+const { expect } = require('chai')
+const isPromise = require('../lib/is-promise')
+
+describe('isPromise', () => {
```
mleguen: Could you please add a test to check that for a non-falsy scalar value isPromise returns false?

This is IMHO one of the main use cases.

mleguen:
Travis CI build restarted

test/is-promise.js
micha149: Sure. I added a test on a string value

mleguen:
Thanks @micha149

micha149:
My pleasure.

bcoe:
@micha149 thanks for the contribution 🎉 this should now be available in `yargs@14.2.0`.

PR #1425, opened 2019-09-16
docs: update supported locales

kidonng:


docs/api.md
```diff
@@ -895,14 +896,16 @@ Locales currently supported:
...
+* **nn:** Norwegian Nynorsk.
 * **pirate:** American Pirate.
 * **pl:** Polish.
 * **pt:** Portuguese.
 * **pt_BR:** Brazilian Portuguese.
 * **ru:** Russian.
 * **th:** Thai.
 * **tr:** Turkish.
-* **zh_CN:** Chinese.
+* **zh_CN:** Simplified Chinese.
```
mleguen: IMHO, we should use "PRC Mainland Chinese" for zh_CN and "Taiwan Chinese" for zh_TW, to keep as close as possible to [IANA names](http://www.iana.org/assignments/language-subtag-registry/language-subtag-registry).

mleguen: @bcoe what is your opinion in this matter?

kidonng: I think this is a conventional expression and it can convey the information properly. We do have `zh_HK` for the Chinese language used in Hong Kong but people usually won't distinguish them.
Nevertheless, I agree with your point and will make some changes to see if it's better.

bcoe: I recently translated an arcade game to `zh_CN` and the team I worked on used the phrasing `Simplified Chinese`, I think I'm comfortable with this, and differ to @kidonng.

bcoe: ahh, I see we opted for "Mainland", @kidonng are you fine with where we landed in this PR?

kidonng: Sure ☺️ It's up to you to choose the words that best suits you.

PR #1441, opened 2019-10-07
fix(docs): TypeScript import to prevent a future major release warning

### LINKED ISSUE #1439: Warning about `parsed` in typescript
When using this with typescript and using `import * as yargs from 'yargs';`, I get the following log when running the code. I fixed this by using `import yargs from 'yargs';`, and think it has to do with exactly what is imported. If that indeed is the case, I think the [typescript example](https://github.com/yargs/yargs/blob/master/docs/typescript.md) should be updated. If so, I'd be happy to create a PR, but it really just is removing a couple of characters.

```
In future major releases of yargs, "parsed" will be a private field. Use the return value of ".parse()" or ".argv" instead
```

vidavidorra:
Importing using `import yargs from 'yargs';` prevents a warning log to console about the `parsed` property being dropped in a future major version.

In future major releases of yargs, "parsed" will be a private field. Use the return value of ".parse()" or ".argv" instead

Closes #1439

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -7,7 +7,7 @@ The following `.options()` definition:
 
 ```typescript
 #!/usr/bin/env node
-import * as yargs from 'yargs';
+import yargs from 'yargs';
 
 const argv = yargs.options({
   a: { type: 'boolean', default: false },


bcoe:
@vidavidorra thank you for the contribution :+1:

PR #1447, opened 2019-10-12
missing array choice regression test

kellyselden:
Failing test case for https://github.com/yargs/yargs-parser/issues/205.

If you change yargs-parser to v13.1.1, the test passes.

In use [here](https://github.com/CrowdStrike/faltest/blob/5403b37f6d527764eee5ae60988b1dbdd445d065/packages/cli/src/index.js#L90-L97)

### DIFF
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -110,6 +110,18 @@ describe('yargs dsl tests', () => {
     argv.looks.should.eql('good')
   })
 
+  it('should ignore a missing array choice with an empty default', () => {
+    const argv = yargs(['--looks', '--looks', 'good'])
+      .option('looks', {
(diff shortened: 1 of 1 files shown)


bcoe:
thanks @kellyselden :+1:

test/yargs.js
```diff
@@ -110,6 +110,18 @@ describe('yargs dsl tests', () => {
     argv.looks.should.eql('good')
   })
 
+  it('should ignore a missing array choice with an empty default', () => {
```
mleguen: In yargs/yargs-parser#205, the bug seems to depend on the use of `flatten-duplicate-arrays` or not. Souldn't this test reproduce this?

PR #1453, opened 2019-10-16
Yargs async

mleguen:
See yargs/yargsa#4


bcoe:
@mleguen I'm excited about this work ... it represents a big change to yargs that I didn't expect to have the time to really ever dig into and take on.

It's my goal to set aside a few hours to open-source on Saturday this week, to really dive in and understand this work and your underlying proposal.

One thought, it might be interesting to make an effort to update `docs/` and `README.md` to reflect what this new async world will look like for our existing samples; this will be a good indicator of how disruptive we expect these changes to be.

test/command.js
```diff
@@ -61,8 +61,8 @@ describe('Command', () => {
       argv['baz-qux'].should.equal('world')
     })
 
-    it('populates argv with camel-case variants of variadic args when possible', () => {
-      const argv = yargs('foo hello world !')
+    it('populates argv with camel-case variants of variadic args when possible', async () => {
```
bcoe: adding some specific tests for chains of `async` middleware, would be good, I'd expect this to work like:

```
for (const middleware of stack) {
  await middleware;
  await command;
}
```

mleguen:
> One thought, it might be interesting to make an effort to update `docs/` and `README.md` to reflect what this new async world will look like for our existing samples; this will be a good indicator of how disruptive we expect these changes to be.

@bcoe Don't worry: updating docs is in my TODO list, and may be in my mind the most important part of this work...

mleguen:
@bcoe Regarding your suggestion of yargsa:
- pros: it would indeed temper the potentially disruptive effect of this new API
- cons: as most of the code is impacted, this would mean, in practice, either maintaining 2 branches of code (yargs and yargsa), or deprecating yargs as is and maintaining only yargsa

mleguen:
@bcoe I do not see however how we could benefit from top level await, as we have no dynamic initializing in Yargs. We would export an async API, but we would export it statically, so there would be no need for us to use top level await.

mleguen:
@bcoe about c8: If I understand well, this is a replacement for nyc, using builtin node functionality? This looks exciting!

However, the drawback would be the need for node v12, meaning we would have to support both yargs and yargsa at least for 18 months (until node v10 is out of maintenance). Would we have enough time for that?

bcoe:
@mleguen I'm empathetic to the drawbacks of having an async version of yargs in conjunction with a sync version, however, some of the biggest users of yargs require that its behavior is sync (`nyc`, `mocha`, etc).

And I think that the disruption to the existing community is (at this time) not worth the benefit we'd get from an async API. As you've mentioned yourself, the async features we have are a bit odd, and not super compelling:

* middleware.
* handler responses.
* completions.

However, as we Node.js/JavaScript continues to evolve, and we get top-level await, async-iterators, and other delightful abstractions for async code, I think that there will start to be a reasonable case for yargs being async.

I don't think it would be worth making this jump for the next few majors though, and we'd provide more value to the community continuing to address some of the backlog of bugs, oddities, etc. This makes me think if we'd like to start experimenting with an async API, it would at least be worth considering floating two versions of yargs at once.

bcoe:
@mleguen I've created https://github.com/yargs/yargsa, which you are a maintainer of, I've also published https://www.npmjs.com/package/yargsa (as a placeholder for our work).

I'd like to advocate that we pull together an async version of yargs in this repo, I think it would be a great opportunity to play with:

* async/await (and top level await as soon as it becomes available).
* ESM modules (I'd advocate we potentially try to write this library using ESM modules from the get go, as long as we're experimenting with future APIs.

Let's use this as a playground to flesh out the version of yargs we'd like to see in the future.

mleguen:
@bcoe Thank you!

And Wow! for your hard work of the two past weeks on yargs!

> I'd advocate we potentially try to write this library using ESM modules

Why not switching to typescript?
- we could output ESM modules (and other formats as well if needed)
- we would benefit from typings in our devs
- our typescript users would have accurate typings without having to rely on potentially outdated @types/yargs
- our non typescript users would not be impacted
- etc.

At the moment, Yargs is the only project I am still developping in plain JS

mleguen:
@bcoe As the yargsa repository is only a stub at the moment, I am going to complete the feature first here.

Once it is complete, I will initialize yargsa with the latest version of yargs, and merge it there.

You did not expect yargsa to be a full rewrite of yargs, from scratch?

mleguen:
By the way, I removed the limitation on async middlewares, which could only be used after validation, not before.

This limitation was due to the complexity of handling promises in the sync API.

mleguen:
@bcoe I have a strange coverall result [here](https://coveralls.io/builds/26965467/source?filename=lib%2Fvalidation.js#L198) with a closing brace considered as an uncovered line.

Could it be c8 related?

mleguen:
I also aligned the async behavior of all async allowed callbacks (command builder and handlers, middlewares and completion function).

They can now all be:
- either a sync function
- or an async function returning a promise
- or an async function calling a done callback

mleguen:
@bcoe @cspotcode, @petrgrishin, @Goodluckhf ready for review

lib/command.js
```diff
@@ -185,7 +184,15 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
     if (typeof commandHandler.builder === 'function') {
       // a function can be provided, which builds
       // up a yargs chain and possibly returns it.
-      innerYargs = commandHandler.builder(yargs.reset(parsed.aliases))
+      if (commandHandler.builder.length < 2) {
+        innerYargs = await commandHandler.builder(yargs.reset(parsed.aliases))
+      } else {
+        // builder function using a done callback
+        innerYargs = await new Promise(resolve =>
+          commandHandler.builder(yargs.reset(parsed.aliases), resolve)
```
Goodluckhf: What about error? I think it should be something like this:
```
 await new Promise(
     (resolve, reject) =>
          commandHandler.builder(yargs.reset(parsed.aliases),(err, result) => { 
             if (err) return reject(err); 
             resolve(result);
         })
)
```

But better to use `promisify` ofc

lib/command.js
```diff
@@ -232,23 +239,16 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
...
-          try {
-            yargs.getUsageInstance().fail(null, error)
-          } catch (err) {
-            // fail's throwing would cause an unhandled rejection.
-          }
-        })
+      try {
+        await applyMiddleware(innerArgv, yargs, middlewares, false)
+        if (commandHandler.handler.length < 2) {
+          await commandHandler.handler(innerArgv)
```
Goodluckhf: above in this code `innerArgv` was overridden but now you don't override value

lib/middleware.js
```diff
@@ -36,29 +35,14 @@ function commandMiddlewareFactory (commandMiddleware) {
...
-          : Object.assign(accumulation, result)
-      }
-    }, argv)
+async function applyMiddleware (argv, yargs, middlewares, beforeValidation) {
+  for (let middleware of middlewares.filter(m => m.applyBeforeValidation === beforeValidation)) {
+    Object.assign(argv,
+      middleware.length < 3
+        ? await middleware(argv, yargs)
+        // middleware using a done callback
+        : await new Promise(resolve => middleware(argv, yargs, resolve))
```
Goodluckhf: here the same. You've missed error rejection

lib/command.js
cspotcode: What is the type of `commandHandler.builder` in this codepath?  What is the type of `commandHandler`?

cspotcode:
> > I'd advocate we potentially try to write this library using ESM modules
> 
> Why not switching to typescript?

@mleguen @bcoe Yes, the library should be written in TypeScript.  As is, the code is unnecessarily annoying to review because it's not clear the type of values passing through the codebase.

test/command.js
mleguen: Oops! I forgot this point. I will add some tests.

lib/command.js
mleguen: You're right about error handling:
- we should use the standard JS callback format
- if doing so, we should use promisify instead of reinventing the wheel

I'll add some commit to change that (and not only for builder functions).

commandHandler.builder is a function in this codepath (checked a few lines above).

lib/command.js
mleguen: This was a redundant overriding, as applyMiddleware already modifies its first parameter.

So I removed it for it to be more explicit.

lib/middleware.js
mleguen: You're right. The same logic is to be applied to builders, handlers, middlewares and completion functions.

lib/middleware.js
```diff
@@ -36,29 +35,14 @@ function commandMiddlewareFactory (commandMiddleware) {
...
+    )
+    // TODO: add a test for middleware calling done
```
mleguen: It looks like I also missed a remaining TODO...

bcoe:
@cspotcode @mleguen I'm open to TypeScript.

lib/command.js
mleguen: done

lib/middleware.js
mleguen: done

lib/middleware.js
mleguen: removed

test/command.js
mleguen: done

cspotcode:
@bcoe @mleguen Awesome, feel free to ping me on anything related to a TS port.

I took a stab at it a while back, so I went ahead and pushed my branch and filed a PR.  https://github.com/yargs/yargs/pull/1478

bcoe:
@cspotcode @mleguen would you be open to closing this PR, and moving the pertinent conversation we've had to a conversation on yargsa?

cspotcode:
Yeah, works for me.

bcoe:
Conversation moved here: https://github.com/yargs/yargsa/issues/4

@mleguen do you think you'd be open to making an attempt at PR on `yargsa`? If we want to start simple, it could literally be a fork of `yargs`, but with your `async` work from this branch ...

if you, @cspotcode, and others have the will and the cycles, we could also try an approach that's slightly more from the ground up.

OJezu:
yargs/yargsa repository does not seem to be public, Github returns 404 for me.

tamj0rd2:
Did anything else happen with yargsa?

cspotcode:
If I recall correctly, nothing much happened with yargsa.  If I recall correctly, there was disagreement on the name of a function in the API surface, and the code being JS instead of TS was making it unnecessarily time-consuming to reason about any refactors.

Someone was working on porting yargs to TS, which seems to be mostly complete.  They said they no longer have time to lead that effort.  I'm not sure what still remains to be done.

tamj0rd2:
> Someone was working on porting yargs to TS, which seems to be mostly complete. They said they no longer have time to lead that effort. I'm not sure what still remains to be done.

Do you know if that PR or branch still exists somewhere? I'd be happy to contribute to it

cspotcode:
Yeah, I guess that's it.  Seems like steps 5-9 are pretty straightforward.  @types/yargs should already have tests which can be used to verify the bundled declarations.  Contributions to @types get merged within a week, so that can happen quickly.  I don't know if any work has been done converting yargs-parser to TS.  I wonder what ts-migrate would do with it?

PR #1456, opened 2019-10-21
Patch 1

dahim2006:
Add missing french translation


dahim2006:
The translation of a message was missing : I added it in the FR locale file

PR #1459, opened 2019-10-26
fix: -- (stop parse) was not being respected by commands

### LINKED ISSUE #1457: Unknown arguments cause incorrect command positional `coerce()`
As evidenced in https://github.com/lerna/lerna/issues/2315, when arguments are passed after `--` with the default `parserConfiguration` (append them to `argv._`), `yargs@14.2.0` causes an incorrect `coerce("")` call on a command's positional argument, as well as re-parsing any flags (which fails when `strict()` is enabled).

```js
require('yargs')
  .strict()
  .option('loglevel', {
    type: 'string',
  })
  .command('repro [stuff]', 'example', yargs => {
    yargs.positional('stuff', {
      describe: 'things',
      coerce: stuff => {
        if (stuff === '') {
          throw new Error('fail');
        }
        return stuff;
      },
    });
  }, argv => {
    console.log(argv);
  })
  .parse('repro -- --loglevel=woo');
```

It _also_ causes the valid `--loglevel` option in the example above to fail, once the `coerce()` method is modfied to ensure that the parameter is in fact non-empty first (or if it's an identity function `stuff => stuff`, or removed entirely). So that's fun.

The current workaround is to enable `populate--` in yargs-parser:
```js
  .parserConfiguration({
    "populate--": true,
  })
```

This seems sub-optimal.

bcoe:
@evocateur I believe this addresses at least half of the issues you described in #1457 (it sounded like you were potentially describing two bugs).

mind trying the branch out?

fixes #1457

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -227,9 +227,12 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
 
     if (commandHandler.handler && !yargs._hasOutput()) {
       yargs._setHasOutput()
+      // to simplify the parsing of positionals in commands,
+      // we temporarily populate '--' rather than _, with arguments
+      const populateDoubleDash = !!yargs.getOptions().configuration['populate--']
+      if (!populateDoubleDash) yargs._copyDoubleDash(innerArgv)
 
       innerArgv = applyMiddleware(innerArgv, yargs, middlewares, false)
-
       let handlerResult
       if (isPromise(innerArgv)) {
         handlerResult = innerArgv.then(argv => commandHandler.handler(argv))
@@ -349,7 +352,12 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
     // short-circuit parse.
     if (!unparsed.length) return
 
-    const parsed = Parser.detailed(unparsed, options)
+    const config = Object.assign({}, options.configuration, {
+      'populate--': true
+    })
+    const parsed = Parser.detailed(unparsed, Object.assign({}, options, {
+      configuration: config
+    }))
 
     if (parsed.error) {
       yargs.getUsageInstance().fail(parsed.error.message, parsed.error)
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "yargs-test-extends": "^1.0.1"
   },
   "scripts": {
-    "pretest": "standard",
+    "posttest": "standard",
     "test": "nyc --cache mocha --require ./test/before.js --timeout=12000 --check-leaks",
     "coverage": "nyc report --reporter=text-lcov | coveralls",
     "release": "standard-version"
--- a/test/command.js
+++ b/test/command.js
@@ -135,6 +135,45 @@ describe('Command', () => {
         .parse()
       parseCount.should.equal(1)
     })
+
+    // see: https://github.com/yargs/yargs/issues/1457
+    it('handles -- in conjunction with positional arguments', () => {
+      let called = false
+      const argv = yargs('foo hello world series -- apple banana')
+        .command('foo <bar> [awesome...]', 'my awesome command', noop, (argv2) => {
+          argv2.bar.should.eql('hello')
+          argv2.awesome.should.eql(['world', 'series'])
+          argv2['_'].should.eql(['foo', 'apple', 'banana'])
+          called = true
+        })
+        .parse()
+      argv.bar.should.eql('hello')
+      argv.awesome.should.eql(['world', 'series'])
+      argv['_'].should.eql(['foo', 'apple', 'banana'])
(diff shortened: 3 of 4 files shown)


evocateur:
Here's the test I wrote locally to validate the weird `strict()` + options passed after `--`:
```diff
diff --git a/test/validation.js b/test/validation.js
index bd0bd1d..e9dc0d6 100644
--- a/test/validation.js
+++ b/test/validation.js
@@ -848,6 +848,19 @@ describe('validation tests', () => {
       args.foo.should.equal(true)
       args.bar.should.equal(true)
     })
+
+    it('does not fail when unrecognized option is passed after --', () => {
+      const args = yargs('ahoy ben -- --arrr')
+        .strict()
+        .command('ahoy <matey>', 'piratical courtesy')
+        .option('arrr', { boolean: true, describe: false })
+        .fail((msg) => {
+          expect.fail(msg)
+        })
+        .parse()
+      args.matey.should.equal('ben')
+      args._.should.deep.equal(['ahoy', '--arrr'])
+    })
   })
 
   describe('demandOption', () => {
```

bcoe:
@evocateur I will try to back-port into 14, if it applies cleanly.

bcoe:
@evocateur please go ahead and test with `npm i yargs@next-14`, I will have `v15` out relatively soon.

bcoe:
@evocateur when you have a moment, could you try `npm i yargs@next` and see if your issue has been addressed?

evocateur:
@bcoe Sorry for the delay. I think we have a problem :(

_Both_ v14.2.1 and v15.0.0 fail across all platforms in `lerna@latest`:
- [yargs-14-next](https://github.com/lerna/lerna/compare/yargs-14-next) ([travis](https://travis-ci.org/lerna/lerna/builds/610978120))
- [yargs-next](https://github.com/lerna/lerna/compare/yargs-next) ([travis](https://travis-ci.org/lerna/lerna/builds/610980333))

The most common error is `TypeError: Cannot delete property '--' of #<Object>`, stack points to [this line](https://github.com/yargs/yargs/blob/5cc2b5e15b0c54022420ecc9ee533d89a88e4115/yargs.js#L1195). The other test failures seem to be related to "known" flags _after_ an unknown positional being swallowed incorrectly.

Running the integration tests (`npm run integration`, which don't use the artificial `./helpers/command-runner`, but call lerna as a proper CLI via `execa`), we encounter the same `TypeError: Cannot delete property '--' of #<Object>`:
```
 FAIL  integration/lerna-run-since.test.js
  ● lerna run --since

    Command failed: node /Users/daniels/github/lerna/core/lerna/cli.js run test --since --concurrency=1 -- --silent
    /Users/daniels/github/lerna/node_modules/yargs/yargs.js:1183
          else throw err
               ^

    TypeError: Cannot delete property '--' of #<Object>

      42 |     .command(runCmd)
      43 |     .command(versionCmd)
    > 44 |     .parse(argv, context);
         |      ^
      45 | }
      46 |

      at Object.Yargs.self._copyDoubleDash (node_modules/yargs/yargs.js:1195:5)
      at Object.parseArgs [as _parseArgs] (node_modules/yargs/yargs.js:1096:60)
      at Object.parse (node_modules/yargs/yargs.js:583:25)
      at main (core/lerna/index.js:44:6)
      at Object.<anonymous> (core/lerna/cli.js:11:15)
      at makeError (node_modules/execa/index.js:174:9)
      at Promise.all.then.arr (node_modules/execa/index.js:278:16)
```
(this is output from the branch using `yargs@15`, but it's the same error with `yargs@14.2.1`)

The test failures are **all** in modules that _explicitly_ specify `yargs.parserConfiguration({ "populate--": true })` in their command builder (it's never set globally). When removing it, say, [from the version command](https://github.com/lerna/lerna/blob/5344820fc65da081d17f7fd2adb50ffe7101905b/commands/version/command.js#L159-L162), the tests for that module pass. This "works" for the version command, but others like `lerna run` and `lerna exec` _depend_ on `"populate--": true`.

NoahZinsmeister:
My lerna@3.18.4 package is also broken in the same way (implicit yargs@14.2.1 dependency)

bcoe:
@evocateur @NoahZinsmeister can you send a failing test? to me this seems like another dependency in lerna is mucking with object definitions:

```
> const obj = {}
undefined
> delete obj['--']
true
> 
```

Is there another dependency which is potentially mucking with object creation?

NoahZinsmeister:
possibly @bcoe ! the best i can do atm is give you the steps i'm doing to reproduce the bug locally:

- `git clone https://github.com/NoahZinsmeister/web3-react.git`
- `yarn`
- `yarn start`

PR #1460, opened 2019-10-27
refactor!: remove package.json-based parserConfiguration

bcoe:
removes deprecated feature, in prep for breaking release.

### DIFF
--- a/yargs.js
+++ b/yargs.js
@@ -1,4 +1,5 @@
 'use strict'
+
 const argsert = require('./lib/argsert')
 const fs = require('fs')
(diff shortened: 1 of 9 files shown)


yargs.js
```diff
@@ -603,7 +604,7 @@ function Yargs (processArgs, cwd, parentRequire) {
 
       const demand = opt.demand || opt.required || opt.require
 
-      // deprecated, use 'demandOption' instead
+      // A required option can be specified via "demand: true".
```
bcoe: I don't think it's worth the disruption that dropping this feature would cause.

Issue #1465, opened 2019-10-27
chore: get to 100% branch coverage


bcoe:
## Background

One of the other projects I've been working on is [c8](https://github.com/bcoe/c8), which collects coverage using V8's built in coverage (rather than using transpilation).

As of Node 13, enough bugs have been worked out in V8/c8 that we can actually hit 100% line coverage with yargs (and I've started [moving us over to dog-fooding c8](https://github.com/yargs/yargs/pull/1464)).

## Help Wanted

I would like for us to try to move towards 100% branch coverage with yargs, in the process of doing so we should file bugs against V8 for any issues we continue to find with their coverage support.

## Upstream Issues in V8

_we can file any outstanding issues against V8 here._

## Contributing

1. run the yargs test suite.
2. submit a patch adding a test for one of the untracked branches of coverage.

nicolasacquaviva:
When I try to run `npm run coverage` I get `Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}`. Am I missing something?

bcoe:
@nicolasacquaviva `npm run coverage` is used on Travis to send coverage to coveralls.io. If you simply run `npm test`, it will collect and display coverage 👍

dhuang612:
Is there anything left to work on?

Thanks

bcoe:
@dhuang612 yeah :smile: there's still missing block coverage for quite a few statements, if you clone the repo and run tests, they should jump out at you.

PR #1467, opened 2019-10-28
test: cover missing filter arg in obj-filter

RobinCsl:
This is a contribution to https://github.com/yargs/yargs/issues/1465

BEFORE
 obj-filter.js           |      100 |       75 |    66.67 |      100 |              4,10 |

AFTER
  obj-filter.js           |      100 |    88.89 |      100 |      100 |                10 |

### DIFF
--- a/test/obj-filter.js
+++ b/test/obj-filter.js
@@ -1,0 +1,16 @@
+'use strict'
+/* global describe, it */
+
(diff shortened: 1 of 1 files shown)


bcoe:
@RobinCsl thanks for the contribution :tada:

PR #1468, opened 2019-10-29
Add tests for argsert warning to display positional information

jaredgreen:
Contribution towards code coverage for #1465 

This gets `lib/argsert.js` to 100% coverage. Technically only one test increased coverage, but I think there is value in both tests as there wasn't actually coverage of any positions aside from first.

Before:
image

After:
image


bcoe:
@jaredgreen awesome \o/ thank you for the contribution.

Issue #294, opened 2015-11-08
suggestion: simplify argv object / granular control


boneskull:
Say you've gathered `argv` like so:

``` js
const argv = yargs.option('baloney', {
  alias: 'b'
})
  .demand(1)
  .argv;
```

I _always_ do something like this afterwards:

``` js
const files = argv._;
const options = argv.filter(function(value, key) {
  return key.length > 1;
});

mainFunc(files, options);
```

Use case:
- The executable using **yargs** is simply an wrapper around an API; the CLI is only one method of accessing the API.
- The API validates its input.
- The fewer keys to worry about, the easier it is to validate `options`.
- Moreover, data duplication sux.

I'd like to propose similar behavior:
- The options object supports a new property, `store`.
- The value of `store` determines in which `argv` property the value is stored.
- If `store` is an `Array` of `string` values, it will be stored in each supplied property.
- Omitting `store` retains default behavior.

I don't _think_ this use case is entirely off-the-wall.  It provides granular control over what ends up in `argv`--your CLI needn't name its flags as expected by your API.

bcoe:
One idea I've been kicking around, that addresses a similar issue, is creating a new module `alias-object`. This module would handle all the logic around alias keys (this is more annoying than you'd think, supporting transitive aliases, etc.

Perhaps rather than adding an additional options setting, we could expose this new object? It would have a method that returns all keys/values without alias keys/values.

boneskull:
This would solve my use case. Granted there are myriad workarounds, but it's save me some boilerplate.

coreyfarrell:
@boneskull recent versions of yargs-parser have an option that I think accomplishes your goal:
```js
const argv = yargs.option('baloney', {
  alias: 'b'
})
  .demand(1)
  .parserConfiguration({'strip-aliased': true})
  .argv
```

See https://github.com/yargs/yargs-parser#strip-aliased and the following option `strip-dashed`.

bcoe:
@coreyfarrell I had forgotten that this feature was added, good catch @coreyfarrell.

Issue #276, opened 2015-10-09
Accessing defaults in .check()


petehotchkiss:
Within the scope of the .check() method - do we have access to the default values?

I'd like to set up a pseudo 'demand' that is soft enough to leta task be called with no arguments, but if one given argument  - "d" - is passed then it should insist on "e" also being included

var foo = require('yargs')
                    .alias('e', ['E', 'elephant'])
                    .alias('d', ['D', 'dog'])
                    .default({ e: '12345', d:false })
                    .check( function(a,b) {
                         // Check to see if d is true AND that e isnt the default value '12345'
                    })
                    .agrv;

So I'm figuring it would be  a case of trying

if( a.e == default_array['e']) { // its default - so throw the error };

But I'm not familiar - nor can I see any documentation to - with how to get at values we've defined as being the default.

Tell me if I'm barking up the wrong tree here

thanks

nexdrew:
@petehotchkiss Good question! You should be able to access default values using something like `yargs.getOptions().default['e']` (assuming `var yargs = require('yargs')`). Give that a shot and let us know.

Also, have you seen the [`.implies()` method](https://www.npmjs.com/package/yargs#implies-x-y)? It's close to what you want, but I don't think it makes a distinction between specified vs default values.

shadowspawn:
An answer was provided, and no further activity in a month. Closing this as resolved.

Feel free to open a new issue if it comes up again, with new information and renewed interest.

PR #1472, opened 2019-11-05
feat: configuration setting for running detailed parse

### LINKED ISSUE #1334: Presence of an argument?
Say I have the arguments --abc N. If I make the default for abc be 5, then yargs will create the .abc value with 5. However, this doesn't tell me if --abc was actually seen on the command line. Is there a way to see an argument was given at all?

bcoe:
You can now set `yargs.parserConfiguration({detailed: true})` and receive detailed results from `yargs-parser`.

This provides an interface into @juergba's [defaulted functionality](https://github.com/yargs/yargs-parser/pull/211).

fixes #1334 

CC: @coreyfarrell.

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -234,10 +234,15 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
 
       innerArgv = applyMiddleware(innerArgv, yargs, middlewares, false)
       let handlerResult
+      let detailedParse = combineDetailedParses(yargs, positionalMap)
       if (isPromise(innerArgv)) {
-        handlerResult = innerArgv.then(argv => commandHandler.handler(argv))
+        handlerResult = innerArgv.then(argv => {
+          if (detailedParse) detailedParse.argv = argv
+          commandHandler.handler(detailedParse || argv)
+        })
       } else {
-        handlerResult = commandHandler.handler(innerArgv)
+        if (detailedParse) detailedParse.argv = innerArgv
+        handlerResult = commandHandler.handler(detailedParse || innerArgv)
       }
 
       if (isPromise(handlerResult)) {
@@ -260,6 +265,19 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
     if (numFiles > 0) currentContext.files.splice(numFiles * -1, numFiles)
 
     return innerArgv
+  }
+
+  function combineDetailedParses (yargs, positionalMap) {
+    let detailedParse
+    if (yargs.getParserConfiguration().detailed) {
+      detailedParse = Object.assign({}, yargs.parsed)
+      Object.keys(positionalMap).forEach(k => {
+        if (detailedParse.defaulted[k] && !positionalMap[k].defaulted) {
+          delete detailedParse.defaulted[k]
+        }
+      })
+    }
+    return detailedParse
   }
 
   function shouldUpdateUsage (yargs) {
@@ -378,6 +396,15 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
         }
       })
     }
+
+    // Populate defaulted value from parse on positionalMap. This allows us to
+    // combine the option and positional detailed parse information.
+    Object.keys(positionalMap).forEach(k => {
+      positionalMap[k] = {
+        value: positionalMap[k],
+        defaulted: !!parsed.defaulted[k]
+      }
+    })
   }
 
   self.cmdToParseOptions = function (cmdString) {
--- a/lib/command.js
+++ b/lib/command.js
@@ -177,7 +177,7 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
     // what does yargs look like after the builder is run?
     let innerArgv = parsed.argv
     let innerYargs = null
-    let positionalMap = {}
+    let detailedPositionalParse = {}
     if (command) {
       currentContext.commands.push(command)
       currentContext.fullCommands.push(commandHandler.original)
@@ -215,7 +215,7 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
     }
 
     if (!yargs._hasOutput()) {
-      positionalMap = populatePositionals(commandHandler, innerArgv, currentContext, yargs)
+      detailedPositionalParse = addPositionalsToArgv(commandHandler, innerArgv, currentContext, yargs)
     }
 
     const middlewares = globalMiddleware.slice(0).concat(commandHandler.middlewares)
@@ -223,7 +223,7 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
 
     // we apply validation post-hoc, so that custom
     // checks get passed populated positional arguments.
-    if (!yargs._hasOutput()) yargs._runValidation(innerArgv, aliases, positionalMap, yargs.parsed.error)
+    if (!yargs._hasOutput()) yargs._runValidation(innerArgv, aliases, detailedPositionalParse, yargs.parsed.error)
 
     if (commandHandler.handler && !yargs._hasOutput()) {
       yargs._setHasOutput()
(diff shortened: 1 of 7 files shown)


lib/command.js
```diff
@@ -367,17 +383,30 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
...
         if (positionalKeys.indexOf(key) !== -1) {
-          // any new aliases need to be placed in positionalMap, which
-          // is used for validation.
-          if (!positionalMap[key]) positionalMap[key] = parsed.argv[key]
+          // any new aliases need to be placed in parsedPositionals, so that
+          // the new keys pass strict validation.
+          if (!detailedPositionalParse[key]) {
+            detailedPositionalParse[key] = {
+              value: parsed.argv[key],
+              defaulted: !!parsed.defaulted[key]
```
mleguen: If detailerPositionalParse's list of fields is intended to grow, we should not have to add these fields in 2 places.

I recommand refactoring this part.

lib/command.js
```diff
@@ -348,9 +366,7 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
         unparsed.push(value)
       })
     })
-
-    // short-circuit parse.
-    if (!unparsed.length) return
+    if (!unparsed.length) return detailedPositionalParse
```
mleguen: In that case, postProcessPositionals (now runParserOnPositionals) used to return nothing and populatePositionals (now addPositionalsToArgv) used to return positionalMap.

Now, runParserOnPositionals returns `{}` and addPositionalsToArgv returns it too. Isn't something missing? Shouldn't detailedPositionalParse contain at least values from positionalMap before being returned?

bcoe:
@mleguen I agree about concerns of adding complexity, however I also feel we need some way to expose a `detailed` parse from `yargs-parser` (which is an advanced feature @coreyfarrell asked for for `nyc`, and other folks have asked for multiple times throughout the years). If we're going to add this feature, I think we need it to work in `.parse()`, `command()`, and all other locations where a parse occurs.

tldr; I think this ultimately is actually a good replacement to `yargs.parsed`, which only worked in some contexts ... open to other recommendations.

test/yargs.js
```diff
@@ -873,6 +873,127 @@ describe('yargs dsl tests', () => {
...
+          ;(!!parse.defaulted.id).should.equal(false)
+          ;(!!parse.defaulted.foo).should.equal(true)
+          ;(!!parse.defaulted.banana).should.equal(true)
+          parse.aliases.id.should.include('identifier')
+          return done()
+        }, [(argv) => {
+          return {
+            hello: 'world'
+          }
+        }])
```
coreyfarrell: What is this final callback argument?  Seems like this test could do without it - also with `detailed: true` I'd expect this callback to use `({argv})` argument (parse.argv instead of just argv).

test/yargs.js
```diff
@@ -873,6 +873,127 @@ describe('yargs dsl tests', () => {
...
+            .positional('foo', {
+              default: 'bar'
+            })
+            .alias('id', 'identifier')
+        }, (parse) => {
+          parse.argv.id.should.equal('abc123')
+          parse.argv.identifier.should.equal('abc123')
+          parse.argv.foo.should.equal('bar')
+          parse.argv.banana.should.equal('yellow')
+          parse.argv.hello = 'world'
```
coreyfarrell: Setting `parse.argv.hello` seems unneeded?

test/yargs.js
```diff
@@ -873,6 +873,127 @@ describe('yargs dsl tests', () => {
...
+            hello: 'world'
+          }
+        }])
+        .parserConfiguration({
+          detailed: true
+        })
+        .option('banana', {
+          default: 'yellow'
+        })
+        .parse('foo abc123 --apple red')
```
coreyfarrell: As with the yargs-parser tests it would be good if we could use `--apple green` to show that yargs.defaulted.apple is cleared when the default value is explicitly set to the default value by the user.  Also missing a test above to verify parse.defaulted.apple is unset.

test/yargs.js
```diff
@@ -873,6 +873,127 @@ describe('yargs dsl tests', () => {
...
+        .command('foo [id] [foo]', 'foo command', (yargs) => {
+          yargs
+            .option('apple', {
+              default: 'green'
+            })
+            .positional('id', {
+              default: 'blerg'
+            })
+            .positional('foo', {
+              default: 'bar'
```
coreyfarrell: What is the expected behavior when a default is not configured?  If you remove this `default: 'bar'` would this mean `parse.defaulted.foo == true && parse.argv.foo === undefined`?

test/yargs.js
```diff
@@ -873,6 +873,127 @@ describe('yargs dsl tests', () => {
...
+          detailed: true
+        })
+        .option('banana', {
+          default: 'yellow'
+        })
+        .parse('foo abc123 --apple red')
+      parse.argv.id.should.equal('abc123')
+      parse.argv.identifier.should.equal('abc123')
+      parse.argv.foo.should.equal('bar')
+      parse.argv.banana.should.equal('yellow')
```
coreyfarrell: Need to check argv.apple and defaulted.apple.

test/yargs.js
bcoe: this is `middleware`, which can be passed as an array as the final argument to a `command`. The idea with middleware is that it might be a third-party plugin that someone has written, so I think the API surface should continue getting `argv` (otherwise someone writing middleware would need to detect a detailed parse, vs., a regular parse).

bcoe:
@coreyfarrell is there a more minimal API surface that would work for your use-case, maybe we could just get away with introducing `.parseDetailed` and ignoring the `.command` use case until someone asks for it.

coreyfarrell:
> ignoring the .command use case until someone asks for it.

For my use case `const {argv, defaults} = yargs.parse(argsArray)` would be adequate, though ignoring the `.command` use case for now would prevent adding it without a breaking change.  Right now setting `detailed: true` has no meaning anywhere, so using it to change the API of both `yargs.parse` and `.command` handlers is not a breaking change.

bcoe:
we have this ancient pull request as well, asking us to expose `defaults`:

https://github.com/yargs/yargs/issues/276

I don't love the idea of adding a `yargs.defaults()` because it feels a bit clunky, why no `yargs.aliases()`, `yargs.etc` ... perhaps a compromise would be a method like:

```js
yargs.isDefaulted(string) // returns true or false depending on whether the optoin was defaulted.
```

☝️ I could imagine adding a handful of options like this.

bcoe:
somewhat similar: https://github.com/yargs/yargs/issues/294, a way to filter out any keys that have aliases.

coreyfarrell:
The API of the `yargs` object is difficult as many of the functions mean "add this thing" rather than "get information about this thing".  So I think `yargs.isDefaulted(string)` makes sense.  FWIW my previous message mentioned `const {argv, defaults} = yargs.parse(argsArray)`, this is incorrect.  It should have said `const {argv, defaulted} = yargs.parse(argsArray)`.  So I think `yargs.defaulted` avoids the confusion that `yargs.defaults` would create, though I say that from the prospective of a native English speaker.  I'm not sure if `defaulted` would be so clear to people who are not native English speakers so maybe `isDefaulted(string)` would be better?  If we go in that direction then should we consistently expose `isDefaulted(string)` instead of `defaulted`?

juergba:
What about the `error` object, which is populated by yargs-parser?
Does it really throw in Yargs? Otherwise it should also be accessible in Yargs to make sense.
In yargs-parser it's populated only and does not throw.

mleguen:
@coreyfarrell As a non-native English speaker, I am used to seeing (and writing) isXxxed functions for this kind of check. So I would recommend using isDefaulted. But that is only ONE non-native english speaker opinion...

bcoe:
@mleguen @coreyfarrell I'm leaning towards the `isDefaulted`, vs., the current state of this branch.

bcoe:
I need to actually sit down and work on the refactor discussed here, i.e., adding a helper method rather than a sweeping change to how parsing works based on a configuration setting.

bcoe:
let's revisit this when someone has time, it would be good to expose the `defaulted` behavior, but I'm not 100% sure of the best implementation just yet.

bcoe:
I think @coreyfarrell suggested something along the lines of a symbol on the `argv` object, which could be used by `yargs.isDefaulted(argv, key)` to check whether a given field was defaulted.

Issue #1452, opened 2019-10-16
Yargs async


mleguen:
Follow-up of #1420.

If we are going to introduce a breaking change by changing .parse() return type (and the like), I think that, now that we are planning to drop node 6 support, we should go further by making .parse() async, ie _always_ returning a promise.

This would not be a pain for our users, as they could use async/await too in their now at-least-node-8-compatible code. I am going to post some future API use examples here to illustrate this.

This would also ease the work a lot for us, and simplify already existing code (we would no longer have to deal with "maybe promises"). I am also going to post the intended approach here.

mleguen:
## Future API usage examples

### With `.exitProcess(true)` (default)

```js
async main() {
  const argv = await yargs
    .command('asyncCmdSuccess', 'description', () => {}, () => Promise(
      (resolve) => setTimeout(() => resolve(), 500)
    ))
    .command('asyncCmdError', 'description', () => {}, () => Promise(
      (_, reject) => setTimeout(() => reject(), 500)
    ))
    .command('syncCmdSuccess', 'description', () => {}, () => {})
    .command('syncCmdError', 'description', () => {}, () => {
      throw new Error()
    })
    .parse()
  // Parsing is OK and the handler returned (if sync) or its promise was resolved (if async)
  doWatheverYouWantWithParsingResults(argv)
  // No try/catch, as the process would be exited before the catch
  // and before node could emit an "unhandled rejection" warning
}

main()
```

### With `.exitProcess(false)`

```js
async main() {
  try {
    const argv = await yargs
      .command('asyncCmdSuccess', 'description', () => {}, () => Promise(
        (resolve) => setTimeout(() => resolve(), 500)
      ))
      .command('asyncCmdError', 'description', () => {}, () => Promise(
        (_, reject) => setTimeout(() => reject(), 500)
      ))
      .command('syncCmdSuccess', 'description', () => {}, () => {})
      .command('syncCmdError', 'description', () => {}, () => {
        throw new Error()
      })
      .exitProcess(false)
      .parse()
    // Parsing is OK and the handler returned (if sync) or its promise was resolved (if async)
    doWatheverYouWantWithParsingResults(argv)
  }
  // Without try/catch, node would emit an "unhandled rejection" warning if parsing is KO
  catch(err) {
    // The syntax is incorrect or the handler threw (if sync) or its promise was rejected (if async)
    dealWithParsingErrors(err)
  }  
}

main()
```

mleguen:
## General approach

[...]

PR #1473, opened 2019-11-07
feat: onFinishCommand handler

petrgrishin:
Example usage:
image

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -449,6 +449,20 @@ yargs.parserConfiguration({
 See the [yargs-parser](https://github.com/yargs/yargs-parser#configuration) module
 for detailed documentation of this feature.
 
+## Command finish hook
+### Example
+```
+yargs
+    .comand('cmd', ..., async () => {
+        await this.model.find()
+        return Promise.resolve('result value')
+    })
+    .onFinishCommand(async (resultValue) => {
+        await this.db.disconnect()
+        process.exit()
+    }).argv;
+```
+
 ## Middleware
 
 Sometimes you might want to transform arguments before they reach the command handler.
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -2359,32 +2359,36 @@ describe('yargs dsl tests', () => {
   describe('onFinishCommand', () => {
     it('use with promise', (done) => {
       const result = 'noop-result'
-      yargs(['noop']).command('noop', 'a noop command', noop, () => {
-        return Promise.resolve(result)
+      let calledTimes = 0
+      yargs(['noop']).command('noop', 'a noop command', noop, async () => {
+        return result
       }).onFinishCommand(async (commandResult) => {
         commandResult.should.eql(result)
+        calledTimes++
       })
-        .parse('noop', (err, argv) => {
-          if (err) {
-            return done(err)
-          }
-          return done()
-        })
+        .parse('noop')
+
+      setTimeout(() => {
+        calledTimes.should.eql(1)
+        done()
+      }, 5)
     })
 
     it('use without promise', (done) => {
       const result = 'noop-result'
+      let calledTimes = 0
       yargs(['noop']).command('noop', 'a noop command', noop, () => {
         return result
-      }).onFinishCommand(async (commandResult) => {
+      }).onFinishCommand((commandResult) => {
         commandResult.should.eql(result)
+        calledTimes++
       })
-        .parse('noop', (err, argv) => {
-          if (err) {
-            return done(err)
-          }
-          return done()
-        })
+        .parse('noop')
+
+      setTimeout(() => {
+        calledTimes.should.eql(1)
+        done()
+      }, 5)
     })
   })
 })
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -2355,4 +2355,36 @@ describe('yargs dsl tests', () => {
     // #1482.
     argv['--'].should.eql(['foo'])
   })
+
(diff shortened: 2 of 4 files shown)


test/yargs.js
```diff
@@ -2339,4 +2339,36 @@ describe('yargs dsl tests', () => {
         })
     })
   })
+
+  describe('onFinishCommand', () => {
+    it('use with promise', (done) => {
+      const result = 'noop-result'
+      yargs(['noop']).command('noop', 'a noop command', noop, () => {
```
Goodluckhf: you can just use `async` here

test/yargs.js
```diff
@@ -2339,4 +2339,36 @@ describe('yargs dsl tests', () => {
...
+        return Promise.resolve(result)
+      }).onFinishCommand(async (commandResult) => {
```
Goodluckhf: Why do you need this function to be `async`?

mleguen:
@bcoe Perhaps redundant with #1452?

Or perhaps not if #1452 only lands in yargsa...

Not incompatible, anyway.

docs/advanced.md
```diff
@@ -447,6 +447,20 @@ yargs.parserConfiguration({
...
 
+## Command finish hook
+### Example
+```
+yargs
+    .comand('cmd', ..., async () => {
+        await this.model.find()
+        return Promise.resolve('result value')
+    })
+    .onFinishCommand(async (resultValue) => {
```
mleguen: Instead of a general yargs.onFinishCommand(), wouldn't we cover more use cases by adding an optional parameter to yargs.command(), which would be the finishCommand handler for this command?

EDIT: After a 2nd thought, a per-command handler does not make sense, as it is already easy, when writing an async command handler, to handle the termination in the handler itself.

And a per-command handler does not cover the commandDir use case, when the main yargs file does not have the hand on how the command handlers are written. Your general yargs.onFinishCommand does.

So keep it as is.

test/yargs.js
```diff
@@ -2339,4 +2339,36 @@ describe('yargs dsl tests', () => {
...
+        commandResult.should.eql(result)
```
mleguen: If your command handler is long enough to execute (if you add a timer for example), this assertion should be ignored, as done() will have been called by parse callback before.

yargs.js
```diff
@@ -483,6 +483,12 @@ function Yargs (processArgs, cwd, parentRequire) {
     return self
   }
 
+  self.onFinishCommand = function (f) {
+    argsert('<function>', [f], arguments.length)
+    self._onFinishCommand = f
```
mleguen: Shouldn't _onFinishCommand be freezed/unfreezed?

bcoe:
rather than adding additional functionality to yargs for handling async behavior, I'd rather see us combine efforts [here](https://github.com/yargs/yargsa/issues/4) and figure out what an async-first version of yargs might look like.

bcoe:
@mleguen @petrgrishin if there was a stop gap between a fully async version of yargs, and something better than what we have today, I'd be open to this too; but would love for folks to collaborate and figure out what those two versions of the library are.

mleguen:
@bcoe @petrgrishin Sure, this was the idea behind my comments on this issue.

I feel however that yargsa and this issue are complementary:
- yargs focuses on async API (ie medling with promises)
- this issue could bring the "tell application the command handler is done" feature for sync applications

I would then recommend sticking here to a standard sync API, ie:
- adding an optional done argument to the command handler
- calling onFinishCommand when done is called (if the handler uses it) or when the handler returns (if not using it)

And applications using promise-returning command handler should be encouraged to use yargsa.

What would you think of this?

bcoe:
@mleguen your suggestion of using this work as an intermediate step towards a fully async API seems reasonable to me (my wish would be for us to get down to a healthier number of open issues on the existing yargs API, before we made the sweeping change proposed in [yargsa](https://github.com/yargs/yargsa/issues/4)).

@petrgrishin would you be open to addressing some of the code review brought up here?

yargs.js
petrgrishin: Done, thx

test/yargs.js
petrgrishin: Remove `async`, thx

test/yargs.js
petrgrishin: Ok

test/yargs.js
petrgrishin: 👌

dapplion:
Great feature! When will a version with this feature be released?

docs/advanced.md
```diff
@@ -449,6 +449,20 @@ yargs.parserConfiguration({
 See the [yargs-parser](https://github.com/yargs/yargs-parser#configuration) module
 for detailed documentation of this feature.
 
+## Command finish hook
+### Example
+```
+yargs
+    .comand('cmd', ..., async () => {
```
dapplion: Typo, should be `command` not `comand`

mleguen: Thanks @dapplion. Mind doing a small PR to fix this typo?

dapplion: > Thanks @dapplion. Mind doing a small PR to fix this typo?

Sure @mleguen: https://github.com/yargs/yargs/pull/1506

bcoe:
@petrgrishin this is released now, take it for a spin and make sure it's working for you?

One thought for the future, should we potentially make `onFinishCommand` cause yargs to return a Promise, so that you could `await` a command finishing? might be a potential API improvement in the future?

PR #1477, opened 2019-11-09
feat: Expose `Parser` from `require('yargs/yargs')`

### LINKED ISSUE #1475: Expose yargs-parser
Would you be willing to have yargs re-expose yargs-parser?  My hope is for nyc to be able to do something like `const {Parser} = require('yargs/yargs')`.  This would have the benefit of guaranteeing that when nyc uses the exact same version of `yargs-parser` as yargs itself.  It would also reduce the maintenance burden for packages like nyc, for example right now `yargs-parser@16.1.0` is available but latest yargs uses `yargs-parser@^15.0.0` so anyone who blindly upgrades (my CI passed so it's good) would cause two copies of yargs-parser to get installed.

coreyfarrell:
Fixes #1475

### DIFF
--- a/test/parser.js
+++ b/test/parser.js
@@ -1,0 +1,11 @@
+'use strict'
+/* global it */
+
+const yargs = require('../yargs')
+const Parser = require('yargs-parser')
+
+require('chai').should()
+
+it('should expose yargs-parser as Parser', () => {
+  yargs.Parser.should.equal(Parser)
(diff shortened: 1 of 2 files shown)


bcoe:
@coreyfarrell mind trying `npm i yargs@next` on a few projects when you have a moment, it has these changes for you 👍

PR #1478, opened 2019-11-15
TypeScript port from a while ago

cspotcode:
This is a WIP.  It'll need to be merged with the latest master, maybe pushed to the yargsa repo instead.  And all the type declarations will have to be integrated so that inference works the same as people expect from @types/yargs.


bcoe:
@cspotcode let's move this work to yargsa, I'm excited to see if you and @mleguen can come up with a compelling module that uses the core of yargs, but is async/typed from the ground up.

Let me know if there's anything I can do to help with this work, I don't have a ton of cycles right now, but am curious to see if this experiment ends up being worthwhile.

mleguen:
> a compelling module that uses the core of yargs, but is async/typed from the ground up

@ bcoe What are your precise expectations about yargsa using "the core of yargs"? If you mean using yargs-parse, then yes, we can. But if you mean keeping compatibility with bits of yargs javascript sync code, I am afraid it cannot see how to do this. Switching one project to async (and typescript) and keeping the other sync (and pure javascript) will lead to incompatible codebase.

bcoe:
@mleguen my thinking is that there's quite a bit of logic in yargs around validation, commands, completion, some of those bits can perhaps be used in isolation? I'm not married to pulling in the whole of yargs wholesale.

cspotcode:
I say we pull everything into one branch, iterate on it there, and see if there are any concrete reasons to split into multiple modules when we're done.  We should also keep in mind that there's nothing stopping someone from doing `require('yargs').parser(/* using only the parser here */)` and it doesn't mean the added complexity of multiple repos & multiple modules.

We'll be able to iterate more quickly if it's all in one repo and all on the table.

mleguen:
@bcoe @cspotcode Following your comments, here is what I suggest:
- keep the same repo for now
- refactorize the async PR:
  - to isolate generic code (validation, commands, completion, etc.) from the sync-specific workflow in the sync API
  - to publish the async API as require('yargs').yargsa, using the generic code and an async-specific workflow
- once fully reviewed, merge it into master for both API to be available in each version of yargs
- keep other "breaking changes" to the async API
- apply non breaking changes to both API, such as the ts port, which should have no impact on projects using yargs
- deprecate the sync API once most projects have switched to the async API

Or should we switch to ts first, and publish the async API afterwards? Writing, testing and reviewing such a change could be easier and safer in ts than in pure js.

In both cases, I would have to rewrite most of the work I already did on the async PR, but I believe this would improve the maintenability of the project.

bcoe:
@mleguen I like this idea, especially if we can refactor out more shared parts :+1:

Issue #1487, opened 2019-11-20
Async API in the same module as the sync one


mleguen:
As discussed in #1478 , the async API of yargs should be accessible as follows:

```js
const yargsa = require('yargs').yargsa
```

which would increase code reuse between the sync and async API, and the project maintenability, compared to the first approaches described in #1420 and yargs/yargsa#4.

Todo:
- [x] describe the future API (ussage examples)
- [ ] describe the intended approach (dev and tests)
- [ ] do it

mleguen:
## Future API usage examples

### With `.exitProcess(true)` (default)

```js
const yargsa = require('yargs').yargsa

async main() {
  const argv = await yargsa
    .command('asyncCmdSuccess', 'description', () => {}, () => Promise(
      (resolve) => setTimeout(() => resolve(), 500)
    ))
    .command('asyncCmdError', 'description', () => {}, () => Promise(
      (_, reject) => setTimeout(() => reject(), 500)
    ))
    .command('syncCmdSuccess', 'description', () => {}, () => {})
    .command('syncCmdError', 'description', () => {}, () => {
      throw new Error()
    })
    .parse()
  // Parsing is OK and the handler returned (if sync) or its promise was resolved (if async)
  doWatheverYouWantWithParsingResults(argv)
  // No try/catch, as the process would be exited before the catch
  // and before node could emit an "unhandled rejection" warning
}

main()
```

### With `.exitProcess(false)`

```js
const yargsa = require('yargs').yargsa

async main() {
  try {
    const argv = await yargsa
      .command('asyncCmdSuccess', 'description', () => {}, () => Promise(
        (resolve) => setTimeout(() => resolve(), 500)
      ))
      .command('asyncCmdError', 'description', () => {}, () => Promise(
        (_, reject) => setTimeout(() => reject(), 500)
      ))
      .command('syncCmdSuccess', 'description', () => {}, () => {})
      .command('syncCmdError', 'description', () => {}, () => {
        throw new Error()
      })
      .exitProcess(false)
      .parse()
    // Parsing is OK and the handler returned (if sync) or its promise was resolved (if async)
    doWatheverYouWantWithParsingResults(argv)
  }
  // Without try/catch, node would emit an "unhandled rejection" warning if parsing is KO
  catch(err) {
    // The syntax is incorrect or the handler threw (if sync) or its promise was rejected (if async)
    dealWithParsingErrors(err)
  }  
}

main()
```

bcoe:
[...]

PR #1488, opened 2019-11-20
Feature/1487 initialize async api

mleguen:
First step of #1487 

I recommend merging as soon as it is reviewed and corrected, as it will prevent master and the works on yargsa from drifting too much one from the other.


mleguen:
WIP: I still have some linux/ios tests to fix (I'm developping on a windows machine, so these tests are skipped on local runs).

I should also add the yargsa API documentation initialization here.

bcoe:
@mleguen I need to give this a through read through, but my first reaction is I don't love the term _legacy_ for the existing API. 

Yargs is a library used by millions of GitHub repos, tens of thousands of npm repositories, and I personally am not going to be racing to encourage [nyc](https://github.com/istanbuljs/nyc), [mocha](https://github.com/mochajs/mocha), or other major consumers to switch to the new API surface; nor do I think it's a good idea to suggest to the community that the API surface will soon be broken. 

The fact is _for 95% of use-cases_,  the sync API makes writing your application easier; if only because it will be years before top-level `async`/`await` is available across all supported Node versions.

I'm open to having an asynchronous export introduced to yargs (I think it's a good idea), and perhaps we could use it to remove some of the asynchronous logic from the existing API -- like you suggested initially (it's weird that that the two paradigms are mixed).

In my mind, this would mean that you need to import `yargsa` (or perhaps `yargs.async` would be more elegant since we're adding this to the library), If you want to use `async`/`await` in:

* command handlers.
* middleware.
* completions.

We could then strip the half-baked async behavior out of the `sync` version of the API. Where I think this could potentially benefit us, would be that it puts us in a position to simplify the command handling logic, and start to refactor that part of the system.

----

I think this work is a very worthwhile experiment, but when I think about the longterm health of this project, getting the open issues below 200 is more important to me (in the limited time I have to commit to the project) than a complete overhaul of the library.

I think we should keep this in mind with our goals for async functionality 👌

README.md
```diff
@@ -68,8 +68,8 @@ Retreat from the xupptumblers!
 ```javascript
 #!/usr/bin/env node
 require('yargs') // eslint-disable-line
-  .command('serve [port]', 'start the server', (yargs) => {
-    yargs
```
bcoe: What's the motivation for shortening this from `yargs` to `y`? Were you thinking because the object might be either a `yargs` or `yargsa` instance.

If we're going to have the asynchronous API surface in `yargs` itself, I think I'd prefer that we have something like:

```
const yargs = require('yargs').async
```

Similar to `require('fs').promises` in Node itself, at which point I'd just keep calling the API `yargs` in samples like this.

index.js
```diff
@@ -14,6 +14,8 @@ function Argv (processArgs, cwd) {
   return argv
 }
 
+Argv.yargsa = require('./yargsa')
```
bcoe: My thinking is, let's call this `Argv.async`, at which point, rather than thinking of `yargsa` as the successor of yargs, we frame it as the import you should use if you're using the async version of the API.

test/api-test/index.js
```diff
@@ -0,0 +1,23 @@
+'use strict'
+/* global describe */
+
+/**
```
bcoe: As a rule, I usually try to make the directory structure of the tests match the directory structure of the project.

If you would like to make a case for moving tests, I would advocate:

```
test/lib/completion.js
test/lib/integration.js
test/integration/index.js
test/yargs.js
test/yargs-async.js
```

---

However, could we hold off on moving around all the tests until a later date? It makes this pull request much harder to review because it's hard to tell what's new related to the async API, and what is just tests being moved into a new folder.

mleguen:
@bcoe I will give it a 2nd try in a few days to take your remarks into account, and answer your questions.

Some of the answers are already in the commit history (I tried to have a commit history as clean as possible to ease the review, knowing a lot of files would be changed).

About the (maybe poorly chosen) term of *legacy*, sorry for misunderstanding your intentions when you first introduced the *yargsa* name. No problem for me to keep this PR to only dealing with the async issue, as this was my initial intention.

coreyfarrell:
I have to agree this PR is too large.  If `test/*.js` needs to be moved to `test/api-test/*.js` then lets get a separate PR for that.  Preferably with two commits, first commit would be the result of `git mv` commands with no other changes or additions.  Second commit with the edits/additions needed to fix the tests (updating path references and creating new test/*.js files for example).  Structuring like this will make it easy to review / approve these renames.

mleguen:
@coreyfarrell 

>  Preferably with two commits, first commit would be the result of git mv commands with no other changes or additions. Second commit with the edits/additions needed to fix the tests (updating path references and creating new test/*.js files for example). Structuring like this will make it easy to review / approve these renames.

Well, that is exactly what I did here :-)

coreyfarrell:
What I'm suggesting is that a PR that just moves things around without anything else can get approved much easier than a PR that combines large code movement with the actual API changes.

mleguen:
@coreyfarrell OK, I'll do it that way next time.

@bcoe I close this PR for now, to let everyone involved have time to think about *if* and *how* they really want async support for yargs.

I am afraid a do-not-change-too-much approach will not lead us to a better async support than what we already have today, although I completely understand that maintaining support for the huge sync userbase has to be the top priority.

bcoe:
@mleguen I think your idea of a `yargs.yargsa` or `yargs.async` export is great, I'm wondering if perhaps we just start [fleshing out the API here](https://github.com/yargs/yargs/pull/1491); feel free to send patches against that branch with edits to the document.

Then, once we have an API we love documented, let's remove the async functionality from the existing API and build out the async module?

Issue #1420, opened 2019-09-04
Accepting sync and async callbacks while keeping yargs functions sync-only is incoherent


mleguen:
This issue is a follow-up of the discusion with @bcoe and @gajus in #1412

Example with command handlers:
- at the moment:
  - we accept async handlers for commands
  - we asynchronously log an error in `parse()` and `showHelp()` (and maybe other functions as well?) when an async handler reject
  - as `parse()` and `showHelp()` are sync, the calling program has no way to know if and when such an error will occur
- expected behavior:
  - when an async handler needs to be called, the calling Yargs function (`parse()`, `showHelp()`, etc.) should return a promise, to be resolved/rejected after the handler completes, for the calling program to know what is happening
  - when the handler is sync, the current behavior should be kept, ie the calling Yargs function should not return a promise

Before opening any PR, please detail in this issue the intended conception for review, and do not limit it to the command handler example detailed here (there may be other cases of async input).

mleguen:
Async middlewares would have to be taken into account as well in the conception.

mleguen:
Where are promises currently used in Yargs?

01. command handler callbacks in `command.runCommand()`:
    - if the callback returns a promise, it is asynchronously checked by Yargs to emit a validation error if rejected
    - the promise of this check is lost (neither stored nor returned by `runCommand()`)
    - calling applications never know if they are going to receive an asynchronous validation error, so strictly speaking, never know if the parsing is successful
    - this behavior is not documented
01. middleware callbacks in `middleware.applyMiddleware()`
    - if a middleware is to be applied before validation, asynchronous callback are forbidden
    - if a callback returns a promise, the following middleware callbacks are ran asynchronously, once the promise is fulfilled
    - the command handler is also ran asynchronously once the last middleware's promise is fulfilled
    - the promise of the last middleware is returned by `applyMiddleware()`, `runCommand()`, etc., up to `parse()`
    - this allows calling applications to know when middlewares are done, but is completely incoherent with the way async command handlers are handled
    - calling applications still do not know if they are going to receive an asynchronous validation error afterwards, as the promise of the handler execution is still lost
    - none of these behaviors are documented
01. completion functions in `completion.getCompletion()`:
    - if the completion function returns a promise, the `done()` callback is ran asynchronously
    - the promise of `done()` execution is correctly returned by `completion.getCompletion()`
    - however, it is lost in `yargs.getCompletion()` which does not return it
[...]

PR #1491, opened 2019-11-25
docs: describe yargs.async API surface

bcoe:
## Background

@mleguen and @petrgrishin, in #1473, #1488, #1420, and #1487, have kicked off a conversation about what a better thought out asynchronous API surface for yargs would look like ... After dragging my feet initially, I now officially find myself roped in 😆 ... 

and getting a little bit excited about the potential of this cleaning up some of the ugly nooks and crannies of the yargs API.

## What is This?

As [suggested in #1487](https://github.com/yargs/yargs/issues/1487#issuecomment-557979631), I think we should start by collaborating on a document that thoroughly describes yargs' async API surface, before we dive in with the implementation

### DIFF
--- a/docs/async.md
+++ b/docs/async.md
@@ -1,0 +1,36 @@
+# Async API
+
+_Note: this document describes an API that has not yet been fully implemented._
+
+yargs exposes an async API surface, making it easier to compose command driven
+applications that perform asynchronous operations.
+
+As an example, perhaps you would like to create a command line application that
+fetches a URL, like curl:
+
+```js
+const fetch = require('node-fetch')
+const yargs = require('yargs').async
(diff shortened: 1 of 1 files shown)


bcoe:
@petrgrishin I have slight concerns about the mixing of callbacks and promises in #1473, curious what your thoughts are on this PR as a potential direction for async behavior, mainly getting a `promise` back from a command if the command has a promise.

bcoe:
@mleguen, looks like we're opting to start with @petrgrishin's implementation? I still think this might be worth revisiting, would potentially be nice to have a way to opt into promises from yargs rather than using callbacks, but let's see if folks like the approach released in `yargs@15.1.0`?

Issue #1454, opened 2019-10-16
Number as option alias not working as advertised


madflow:
Hello!

When using a number as alias - it is advertised in the `help` section as `-1` - but this does not work. (The option stays undefined).

```
// run.js
const argv = require('yargs')
  .option('negativeone', {
    alias: '1',
    type: 'boolean',
    description: 'Negative one'
  }).argv;

console.log(argv);
```
```
➜  yargs-test node run.js --help
Optionen:
  --help             Hilfe anzeigen                                    [boolean]
  --version          Version anzeigen                                  [boolean]
  --negativeone, -1  Negative one                                      [boolean]
```

> ➜  yargs-test node ./run.js --negativeone
{ '1': true, _: [], negativeone: true, '$0': 'run.js' }

> ➜ yargs-test node ./run.js -1           
{ _: [ -1 ], '$0': 'run.js' }

> ➜ yargs-test node ./run.js --1
{ '1': true, _: [], negativeone: true, '$0': 'run.js' }

mleguen:
@madflow Did you try setting the `parse-numbers` option to false?

```js
yargs.parserConfiguration({
  "parse-numbers": false
})
```

mleguen:
No, it does not work either.

bcoe:
do we want to support numbers as options, is a question worth asking ... this feels a little weird.

madflow:
> do we want to support numbers as options, is a question worth asking ... this feels a little weird.

@bcoe I am not sure at whom this question is directed - but since I logged this issue: It is weird and felt nonintuitive in our "downstream" application. We removed the number alias.

mleguen:
@bcoe I also fell its weird (but maybe some people use it?)

However, if we decide we do support numbers as options, shouldn'd we **explicitely** not support it (ie throw when a number is passed as a key or an alias, and stop parsing `--1` as `{ '1': true }` in yargs-parser)?

benjie:
I don't think numeric CLI flags are weird; quite a few of the standard unixy tools use numeric CLI flags. I'm sure many of us have used `kill -9` from time to time, but here's a few others off the top of my head:

#### grep

```
$ grep --help
Usage: grep [OPTION]... PATTERN [FILE]...
Search for PATTERN in each FILE.
[...]
  -C, --context=NUM         print NUM lines of output context
  -NUM                      same as --context=NUM
```

#### gzip

```
$ gzip --help
Usage: gzip [OPTION]... [FILE]...
[...]

PR #1493, opened 2019-11-27
fix: show 2 dashes on help for single digit option key or alias

mleguen:
To be coherent with what yargs/parser considers as an option single dash key, do not display '-1' but '--1' when '1' is used as an option key or an alias.

See #1454

### DIFF
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -272,7 +272,7 @@ module.exports = function usage (yargs, y18n) {
             // for the special positional group don't
             // add '--' or '-' prefix.
             if (groupName === self.getPositionalGroupName()) return sw
-            else return (sw.length > 1 ? '--' : '-') + sw
+            else return (/^[^0-9]$/.test(sw) ? '-' : '--') + sw
           })
           .join(', ')
 
--- a/test/usage.js
+++ b/test/usage.js
@@ -1121,6 +1121,51 @@ describe('usage tests', () => {
       ])
     })
 
+    it.only('should use 2 dashes for 1-digit key usage', () => {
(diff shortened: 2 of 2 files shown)


bcoe:
fixes: #1454

ruyadorno:
@bcoe heads up this changes does not quite fits the example you gave of `--100` since it's checking for a single-digit only...

...and unfortunately it breaks the help page for one of my modules that uses the `-0` standard, similar to `xargs`:

```
XARGS(1)                  BSD General Commands Manual                 XARGS(1)

NAME
     xargs -- construct argument list(s) and execute utility

SYNOPSIS
     xargs [-0opt] [-E eofstr] [-I replstr [-R replacements]] [-J replstr] [-L number] [-n number [-x]] [-P maxprocs]
           [-s size] [utility [argument ...]]

DESCRIPTION
     The xargs utility reads space, tab, newline and end-of-file delimited strings from the standard input and exe-
     cutes utility with the strings as arguments.

     [... redacted for brevity]

     The options are as follows:

     -0      Change xargs to expect NUL (``\0'') characters as separators, instead of spaces and newlines.  This is
             expected to be used in concert with the -print0 function in find(1).
```

ruyadorno:
I ended up figuring out that **yargs-parser** itself does not have support to such pattern (which surprises me given its popularity within some of the day-to-day tools I use)

I gave it a try at actually implementing support to it in the parser but not really sure how that will go 😬 https://github.com/yargs/yargs-parser/pull/255 - if it eventually makes its way into the parser then we also should update `lib/usage.js` to match it

Issue #1434, opened 2019-10-01
getCompletion example is not working


agrakhov:
I'm trying to play with `.getCompletion` but not sure if its working correctly. If I'm following the [example in docs](https://github.com/yargs/yargs/blob/master/docs/api.md#getcompletionargs-done) and have this in `test.js`:
```
#! /usr/bin/env node
require('yargs')
	.option('foobar')
	.option('foobaz')
	.completion()
	.getCompletion(['./test.js', '--foo'], function(completions) {
		console.log(completions)
	})
// I've tried to add .argv at the end - not helping
```
when I try to run `./test.js --foo` following error is reported:
```
.../test-cli/node_modules/yargs/lib/completion.js:76
        const keyAndAliases = [key].concat(aliases[key] || [])
                                                  ^

TypeError: Cannot read property 'help' of undefined
    at Object.keys.forEach (.../test-cli/node_modules/yargs/lib/completion.js:76:51)
    at Array.forEach (<anonymous>)
    at Object.getCompletion (.../test-cli/node_modules/yargs/lib/completion.js:74:43)
    at Object.Yargs.self.getCompletion (.../test-cli/node_modules/yargs/yargs.js:933:16)
    at Object.<anonymous> (.../test-cli/test.js:6:3)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
```

I hope its just incorrect example in docs )

JakeRowsell89:
I'm able to reproduce. Out of curiosity, are you on zshell? There seems to be some zshell specific code

agrakhov:
@JakeRowsell89 yep I'm using ohmyzsh.

mleguen:
@agrakhov @JakeRowsell89 

This should be fixed now in master.

mleguen:
Fixed in 15.1.0

PR #1495, opened 2019-11-27
fix getCompletion() no longer working for options

mleguen:
getCompletion() throws when completing an option, as it needs `parsed.aliases`, which is lost (unfreezed) when calling `.parse()` with arguments.

See #1434 

This is a side effect of fixing `.parse()` to be able to call it several times, a few months ago.

This PR provides `parsed` to completion before unfreezing it.

### DIFF
--- a/lib/completion.js
+++ b/lib/completion.js
@@ -8,6 +8,11 @@ module.exports = function completion (yargs, usage, command) {
     completionKey: 'get-yargs-completions'
   }
 
+  let aliases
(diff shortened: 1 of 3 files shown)


bcoe:
@mleguen thanks for the patch :ok_hand:

bcoe:
@mleguen this should be out the door in `15.1.0` 👌 thanks for all the hard work.

PR #1497, opened 2019-11-29
fix: misspelling of package.json `engines` field

coreyfarrell:


bcoe:
@coreyfarrell thanks for the patch, mind trying this out when you have the cycles? yargs@15.1.0 was the first version published by a GitHub app, so I'm interested in making sure there weren't any oddities.

coreyfarrell:
`npm info yargs engines` gives the expected information and I just ran the nyc tests locally using latest yargs - no failures.  So I think we're good!

PR #1507, opened 2019-12-04
feat: complete short options with a single dash

### LINKED ISSUE #265: Bash completion:  Long vs short options and number of dashes
When using the bash completion feature, yargs completes short options with two
leading dashes instead of just one:

test.js:

``` javascript
var argv = require('yargs')
  .alias('l', 'long')
  .describe('l', 'an option')
  .help('h')
  .alias('h', 'help')
  .completion('completion')
  .argv;

console.log(argv);
```

```
$ node test.js --get-yargs-completions
--l
--h
```

This isn't a huge issue because using the double-dashed short options still works:

```
$ node test.js --l foo
{ _: [],
  h: false,
  help: false,
  l: 'foo',
  long: 'foo',
  '$0': 'test.js' }
```

But I think it would be much better to always use the long options for completions, as they are generally more descriptive. If a long option is not available it should fall back to the single-dashed short version.

mleguen:
When the user did not already enter two dashes, and when the option is not a single digit one.

Closes #265

### DIFF
--- a/lib/completion.js
+++ b/lib/completion.js
@@ -81,11 +81,14 @@ module.exports = function completion (yargs, usage, command) {
         const keyAndAliases = [key].concat(aliases[key] || [])
         const notInArgs = keyAndAliases.every(val => args.indexOf(`--${val}`) === -1)
         if (notInArgs) {
+          const startsByTwoDashes = s => /^--/.test(s)
+          const isShortOption = s => /^[^0-9]$/.test(s)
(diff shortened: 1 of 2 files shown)


bcoe:
@mleguen thank you for all the hard work around completions!

bcoe:
@mleguen this should be released now, if you'd like to try it out.

mleguen:
It is working as expected. Thanks.

PR #1508, opened 2019-12-04
fix: populate positionals when unknown-options-as-args is set

mleguen:
Positionals where no longer populated using `unknown-options-as-args`, as yargs turned them into options before handing them to yargs, but withoud declaring the corresponding option. So they were handled by yargs-parser as unknown options.

The "hidden" options for the positionals are now declared when using `unknown-options-as-args`.

Closes #1444.

PR #1509, opened 2019-12-04
feat(completion): takes negated flags into account when boolean-negation is set

### LINKED ISSUE #1431: Completions should suggest `--no-*` flags when appropriate
When a boolean flag defaults to `true`, I pass `--no-foo` to set it to false.

yargs tab-completions should propose the `--no-foo` flag when:

a) default for the flag is `true`
b) the yargs parser is configured to allow `--no-*` variants for boolean flags.

EDIT fixed a typo that I had backwards

mleguen:
Closes #1431 

When boolean-negation is set:
- completes --no-flag when flag defaults to true
- avoids repeating flags whose negated counterpart is already present

### DIFF
--- a/lib/completion.js
+++ b/lib/completion.js
@@ -76,21 +76,30 @@ module.exports = function completion (yargs, usage, command) {
 
     if (current.match(/^-/) || (current === '' && completions.length === 0)) {
       const descs = usage.getDescriptions()
-      Object.keys(yargs.getOptions().key).forEach((key) => {
+      const options = yargs.getOptions()
+      Object.keys(options.key).forEach((key) => {
+        const negable = !!options.configuration['boolean-negation'] && options.boolean.includes(key)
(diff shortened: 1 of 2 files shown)


mleguen:
Warning : either this PR or #1508 will need some conflict-handling before merging, as both impact the same lines in completion.js.

bcoe:
:+1: let's hold off on landing until we have a plan. Perhaps you could make a PR against this PR, for #1508 rather than us having both PRs open at the same time.

mleguen:
I wronlgy mentionned #1508, but is #1507 which should conflict with this PR.

Instead of doing something complicated, let's merge #1507 first (I will need reviewing first), and I will then rebase this PR.

bcoe:
@mleguen bother you to rebase?

mleguen:
@bcoe I should have some time to do it next week

mleguen:
@bcoe done!

PR #1512, opened 2019-12-10
add link to replit run

eankeen:
i think it would be pretty cool to let people see how this code works (and edit / preview the program) right in their browser, without any manual downloads or installation. i'm thinking we can add a 'run on repl.it' button, which redirects to something like [this](https://repl.it/@eankeen/run-yargs):

image

### DIFF
--- a/.replit
+++ b/.replit
@@ -1,0 +1,2 @@
+language = "nodejs"
+run = "yarn add yargs-parser && yarn link && yarn link yargs && cd example && node divide -x 2 -y 4"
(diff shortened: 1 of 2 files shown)


bcoe:
@eankeen I think this is a neat idea, I wonder how practical it would be to have examples of a few key features in our documentation, rather than just one repl.it link at the top of the README? I don't think that just getting dumped into a terminal with yargs installed is the most valuable.

bcoe:
👋 @eankeen closing this for now, I'm open to this work, but I don't think that just importing yargs doesn't add enough value.

I would rather add links in our docs for targeted replit examples. If you want to discuss, there's a chat room [here](http://devtoolscommunity.herokuapp.com/).

Issue #1511, opened 2019-12-10
Fix typescript documentation with the right syntax for commonjs module import


mleguen:
See the [related paragraph](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require) in the typescript documentation.

We should do :
```
import yargs = require('yargs');
```

Issue #1455, opened 2019-10-19
TypeScript examples don't compile


lf-:
Hi! I've been converting a project to use Yargs recently and was having some issues with the compiler being dissatisfied with my program, so I tried one of the examples in `docs/typescript.md` to see if I could better understand the issue.

Unfortunately, it seems to have the same problem as my code was having:
```
test.ts:4:20 - error TS2339: Property 'options' does not exist on type '{ <K extends never, V>(key: K, value: V, description?: string): Argv<Omit<{}, K> & { [key in K]: V; }>; <K extends string, V>(key: K, value: V, description?: string): Argv<{ [key in K]: V; }>; <D extends { [key: string]: any; }>(defaults: D, description?: string): Argv<...>; }'.

4 const argv = yargs.options({
                     ~~~~~~~
```

It appears that the compiler is possibly not seeing whatever object results from `import yargs from 'yargs'` as an instance of Argv, which seems very strange indeed. How should we fix the examples so they compile, or fix the definitions so they work with the examples?

Contents of `test.ts`:

```typescript
#!/usr/bin/env node
import yargs from 'yargs';

const argv = yargs.options({
  a: { type: 'boolean', default: false },
  b: { type: 'string', demandOption: true },
  c: { type: 'number', alias: 'chill' },
  d: { type: 'array' },
  e: { type: 'count' },
  f: { choices: ['1', '2', '3'] }
}).argv;
```

Versions:
```json
  "dependencies": {
    "yargs": "^14.2.0"
  },
  "devDependencies": {
    "@types/node": "^12.11.1",
    "@types/yargs": "^13.0.3",
    "typescript": "^3.6.4"
  },
```

lf-:
It is notable that using `const yargs = require('yargs')` results in this working properly. So there is something wrong with importing it like a module. Although I could submit a pull request to blindly change the docs to the older import syntax, I think this presents an opportunity to better understand how the modules system works and why it is producing this result. Does anyone know why this seems to happen only when importing as a module?

lf-:
Aha!
```typescript
import * as yargs from 'yargs';
```

results in proper types working. Is this a code smell or should I just file a pull request on the docs to change them to this syntax? (again, I don't know about the conventions around modules in TS)

RikuMantysalo:
I think it depends on your tsconfig.

If you have `esModuleInterop` set to true it works fine with `import yargs from 'yargs'`

bcoe:
[...]

PR #1513, opened 2019-12-11
fix(docs): use recommended cjs import syntax for ts examples

mleguen:
Fix typescript examples to use the right import syntax for a commonjs module such as yargs.

See explanation in typescript documentation [here](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require):

> Both CommonJS and AMD generally have the concept of an exports object which contains all exports from a module.
>
> [...]
>
> When exporting a module using export =, TypeScript-specific import module = require("module") must be used to import the module.

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -7,9 +7,7 @@ The following `.options()` definition:
 
 ```typescript
 #!/usr/bin/env node
-import * as yargs from 'yargs';
-// or with the "esModuleInterop" compiler option set to "true":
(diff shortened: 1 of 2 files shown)


mleguen:
Closes #1511

mleguen:
Closes #1455

lf-:
@mleguen you have to put the "closes" in the initial PR message for GitHub to recognize it (you'll see an underline under "closes" when it's right)

Issue #1534, opened 2020-01-29
Help is displayed twice when an async handler fails with exitProcess(false)


mleguen:
#### index.js
```js
const yargs = require('yargs')

yargs()
  .command('cmd', 'test command', {}, () =>
    new Promise((resolve, reject) => 
      setTimeout(reject, 10)
    )
  )
  .exitProcess(false)
  .parse('cmd')
```

Results:
```bash
$ node index.js
index.js cmd

test command

Options:
  --help     Affiche de l'aide                                         [booléen]
  --version  Affiche le numéro de version                              [booléen]

index.js cmd

test command

Options:
  --help     Affiche de l'aide                                         [booléen]
  --version  Affiche le numéro de version                              [booléen]
```

Expected:
```bash
$ node index.js
index.js cmd

test command

Options:
  --help     Affiche de l'aide                                         [booléen]
  --version  Affiche le numéro de version                              [booléen]
```

mleguen:
Probably due to showHelp beeing called when the handler fails, after yargs is unfreezed, so showHelp calls parse again, calling the handler a second time and failing a second time.

mleguen:
I recommend:
- leaving this bug as is
- and adding a warning in the doc not to use an async handler if parse is intended to be called several times, as an help message for an async handler failure can occur in the middle of the next call to `.parse()`

I can see no way of dealing "cleanly" with async handlers while yargs is sync.

@bcoe what is your opinion on this?

bcoe:
@mleguen I agree with adding a caveat for `async` handlers for the time being.

PR #1535, opened 2020-01-29
fix: help always displayed for the first command parsed having an async handler

### LINKED ISSUE #1533: Bug: .parse not resetting context
Sample code as `bug.js`:
```

const parser = yargs
  .command({
    command: 'foo <bar>',
    description: 'foo subcommand',
    handler: async ({ bar }) => {
      await new Promise((resolve) => {
        setTimeout(() => {
          resolve()
        }, 100)
      })
      console.log('foo command', bar)
    }
  })
  .demandCommand(1, 'You need at least one command before moving on')
  .strict(true)
  .version()
  .help()
  .example('bot [command] --help', 'get detailed help on any command')

parser.parse(process.argv.slice(2), {}, (err, argv, output) => {
  if (err) {
    console.error(output)
    process.exit(1)
  } else if (output) {
    console.log(output)
  }
})

setTimeout(() => {
  parser.parse('', {}, (err, argv, output) => {
    if (err) {
      console.error(output)
      process.exit(1)
    } else if (output) {
      console.log(output)
    }
  })
}, 1000)
```

When `yargs` is `^15.1.0` we see the output:
```
$ node bug.js foo foo-bar
foo command foo-bar
bug.js foo <bar>

foo subcommand

Options:

You need at least one command before moving on
```

When `yargs` is `13.3.0` we see the output:
```
$ node bug.js foo foo-bar
foo command foo-bar
bug.js <command>

Commands:
  bug.js foo <bar>  foo subcommand

Options:

Examples:
  bot [command] --help  get detailed help on any command

You need at least one command before moving on
```

The issue is that we are using `yargs` in a bot setting, where `.parse` is called multiple times. 

With older version of `yargs` we see that on every `.parse` call it reset the context, where every command will be parsed properly, with newer versions of `yargs` when `.parse` is called multiple times, the context is not reset. 

This happened in v14 and later

mleguen:
Close #1533 

After a call to `.parse()` for a command with an async handler, next calls to parse, if they need to display help, display help for this command.

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -256,6 +256,9 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
             // fail's throwing would cause an unhandled rejection.
             }
           })
+          .then(() => {
+            yargs.getUsageInstance().clearCachedHelpMessage()
+          })
       } else {
         if (handlerFinishCommand) {
           handlerFinishCommand(handlerResult)
--- a/lib/usage.js
+++ b/lib/usage.js
@@ -411,6 +411,12 @@ module.exports = function usage (yargs, y18n) {
     cachedHelpMessage = this.help()
   }
 
+  // however this snapshot must be cleared afterwards
+  // not to be be used by next calls to parse
+  self.clearCachedHelpMessage = function () {
+    cachedHelpMessage = undefined
+  }
+
   // given a set of keys, place any keys that are
   // ungrouped under the 'Options:' grouping.
   function addUngroupedKeys (keys, aliases, groups) {
--- a/test/helpers/utils.js
+++ b/test/helpers/utils.js
@@ -34,8 +34,7 @@ exports.checkOutput = function checkOutput (f, argv, cb) {
     }
     process.emit = function emit (ev, value) {
       if (ev === 'uncaughtException') {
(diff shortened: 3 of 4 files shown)


bcoe:
@mleguen caching in general can lead to some gnarly bugs (I've heard it called "one of the hard problems in computer science").

Another option, do we stop caching, how much do we care about the performance gain? Or did we add caching to address another bug?

mleguen:
This is not a performance cache, only something to still be able to display help for the command whose async handler is "running in the background" when yargs returned and can already have been modified.

IMHO, this is a bad idea anyway, so I would be happy to drop it.

bcoe:
@mleguen would we regress a bug that the cache was added to address? I would prefer simplifying if we think this is a feature that people don't need too often ... it might be worth keeping if we specifically added this to address problems.

mleguen:
@bcoe Here is what I suggest:
- we merge this PR first to fix #1533 
- we do not drop caching as it is still useful for applications not calling parse on the same yargs instance several times (and we have no better solution at the moment with yargs not being async)
- we deal with #1534 by stating in the docs that command handlers should **not** return a promise if parse is to be called several times

PR #1536, opened 2020-01-29
Use process.defaultApp to detect built electron app

mleguen:
If process.defaultApp is false (for built electron app), remove only the first element of process.argv (the app binary) and not the two first elements (node + the app script).

See https://github.com/electron/electron/issues/4690#issuecomment-217435222 for mor info.

Solve part of #685

### DIFF
--- a/lib/process-argv.js
+++ b/lib/process-argv.js
@@ -1,0 +1,20 @@
+function getProcessArgvBinIndex () {
+  // Built Electron app: app argv1 argv2 ... argvn
+  // (see https://github.com/electron/electron/issues/4690#issuecomment-217435222)
+  if (process.defaultApp === false) return 0
+  // Default: node app.js argv1 argv2 ... argvn
+  return 1
+}
+
+function getProcessArgvWithoutBin () {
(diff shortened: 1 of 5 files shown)


lib/process-argv.js
```diff
@@ -0,0 +1,20 @@
+function getProcessArgvBinIndex () {
+  // Built Electron app: app argv1 argv2 ... argvn
+  // (see https://github.com/electron/electron/issues/4690#issuecomment-217435222)
+  if (process.defaultApp === false) return 0
```
bcoe: is `process.defaultApp` set by electron, probably worth noting this in the comment as well.

mleguen: OK, I'll do it ASAP

mleguen: Comment added

bcoe:
@mleguen friendly reminder to make sure you use conventional commits when you merge, or else stuff doesn't show up in the next release, I went ahead and amended.

mleguen:
@bcoe Oops! Sorry and thanks. Would you have a link to some doc about the requested commit syntax (is it the same as the one used by nest.js?) for me to do it as expected next time?

PR #1540, opened 2020-02-09
feat: introduces strictCommands() subset of strict mode

bcoe:
I have met your demands @boneskull 

This seemed like a reasonable feature request, and I'm excited to see whether you get use out of it.

Code review appreciated.

---

~BREAKING CHANGE: if you use `.demandCommand()` or `.demandCommand(1)`, in conjunction
with defining commands, `.strictCommands()` is enabled automatically.~

Since the breaking changes to `yargs-parser` were for undocumented behavior (and 15.x.x was actually a regression with regards to the parsing of booleans, I'm going to try to release 15.x.x as a non-breaking change; if the candidate release seems to be breaking too many folks, we can revisit this decision).

### DIFF
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -16,13 +16,6 @@ module.exports = function validation (yargs, usage, y18n) {
     const demandedCommands = yargs.getDemandedCommands()
     // don't count currently executing commands
     const _s = argv._.length - yargs.getContext().commands.length
-    const commandKeys = yargs.getCommandInstance().getCommands()
-
-    // for the special case of yargs.demandCommand() and yargs.demandCommand(1), if
-    // yargs has been configured with commands, we automatically enable strictCommands.
-    if (commandKeys.length && demandedCommands._ && demandedCommands._.min === 1 && demandedCommands._.max === Infinity) {
-      yargs.strictCommands()
-    }
 
     if (demandedCommands._ && (_s < demandedCommands._.min || _s > demandedCommands._.max)) {
       if (_s < demandedCommands._.min) {
@@ -142,8 +135,8 @@ module.exports = function validation (yargs, usage, y18n) {
 
     if (unknown.length > 0) {
       usage.fail(__n(
-        'Unknown argument: %s',
-        'Unknown arguments: %s',
+        'Unknown command: %s',
+        'Unknown commands: %s',
         unknown.length,
         unknown.join(', ')
       ))
--- a/lib/validation.js
+++ b/lib/validation.js
@@ -16,6 +16,13 @@ module.exports = function validation (yargs, usage, y18n) {
     const demandedCommands = yargs.getDemandedCommands()
     // don't count currently executing commands
     const _s = argv._.length - yargs.getContext().commands.length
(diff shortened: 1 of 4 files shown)


lib/validation.js
```diff
@@ -120,6 +127,32 @@ module.exports = function validation (yargs, usage, y18n) {
...
+      argv._.slice(currentContext.commands.length).forEach((key) => {
+        if (commandKeys.indexOf(key) === -1) {
+          unknown.push(key)
+        }
+      })
+    }
+
+    if (unknown.length > 0) {
+      usage.fail(__n(
+        'Unknown argument: %s',
```
evocateur: Shouldn't this be `"Unknown command"`? The error as written looks too similar to the "unknown argument" output.

lib/validation.js
```diff
@@ -16,6 +16,13 @@ module.exports = function validation (yargs, usage, y18n) {
     const demandedCommands = yargs.getDemandedCommands()
     // don't count currently executing commands
     const _s = argv._.length - yargs.getContext().commands.length
+    const commandKeys = yargs.getCommandInstance().getCommands()
+
+    // for the special case of yargs.demandCommand() and yargs.demandCommand(1), if
+    // yargs has been configured with commands, we automatically enable strictCommands.
+    if (commandKeys.length && demandedCommands._ && demandedCommands._.min === 1 && demandedCommands._.max === Infinity) {
+      yargs.strictCommands()
```
evocateur: If this is the only reason the PR is considered a breaking change, I would argue in favor of postponing this particular behavior until later, and allowing a normal feature release for this. I'm not convinced the major version fatigue is worth it, here.

bcoe: @evocateur the new version of [yargs-parser](https://github.com/yargs/yargs-parser/blob/master/CHANGELOG.md#-breaking-changes) has two, I think important, breaking changes that I'd like to shield us from with a major in `yargs`.

lib/validation.js
bcoe: you are 100% correct ... the reason I was thinking of using `Unknown argument:` was to avoid having to update all the languages yargs has been translated into, alternatively we could introduce the new phrase and let folks gradually submit translations.

lib/validation.js
evocateur: Ah, fair. Might as well, then.

lib/validation.js
evocateur: I think it's more important to be precise with error messages, even if it adds another untranslated string.

bcoe:
@boneskull @mleguen I think I'm going to hold off on the breaking change until the next major, based on conversations we had in the tooling chat.

lib/validation.js
bcoe: 👍 @mleguen mind making a PR for the french translation in a separate PR, as I know we've been fixing a few issues with translations recently.

mleguen: @bcoe No problem. I will do it when I will rebase the PR for french translations, once this one is merged.

bcoe:
@boneskull when you have a moment try `npm i yargs@next` it has the strict command functionality you requested.

Issue #1561, opened 2020-02-20
Bad French translation for "default"


mleguen:
To display "default", we reuse the translation for "default:" and remove the final character of the translations, expecting ":" to be translated as ":", which is not the case in all languages.

In French, this leads "default" to be translated as "défaut " (with a final space) instead of "défaut".

"default" and "default:" should be 2 different translations.

PR #1564, opened 2020-02-20
fix(locales): remove extra space in French for 'default'

mleguen:
fixes: #1561


bcoe:
:+1: thanks for the contribution, it's great to have someone who speaks another language help with these translations.

PR #1566, opened 2020-02-26
[WIP] fix: German translation

danieldjewell:
Although I'm not a *native* German speaker (I *am* a native English
    speaker though), there were two specific
things that stood out in the translation as strange. (I will double
    check with native speakers if I have the chance -- and any native
    speakers, please share your thoughts.)

I did do some double and triple checking on word usage and other
translations, as well.

@maxrimue - I'm guessing that you're a native speaker of German, and you
contributed to the translation in the past - perhaps if you have a
minute you can review this?

* "count" - Based on the way that this would be used, I'm looking at it
as "a count of something" ... "a quantity of something" ... and for that
I think "Anzahl" would be a better choice. "Zähler" would be more of "a
counter" or "something that counts" (or even "someone who counts").

* "choices" - "Möglichkeiten" would more literally translate as
"possibilities" (to say "es ist möglich" is to say "it is
    possible/capable of happening/occurring"). But that's kinda what it
literally means - possibilities. "Auswahl" (and the verb auswählen - to
    choose, to select) would be a choice from a set of options. And
"optionen" would literally be the plural of "options"... I *suppose* one
could write "Auswahlmöglichkeiten" ... but to my ear and understanding,
      that would refer to the **availability** of **an opportunity to
      select something** not the **available selection options**
      themselves.


danieldjewell:
One quick other observation:

It *might* be necessary/a good idea to add additional options to the locales because some of the words as used in English can be *both* nouns **and** verbs simultaneously. (e.g. "count" can be "to count something" , "a total count", "hey you, go count these widgets" .... there are different words in German for "the action of counting", "the numerical quantity after counting something", and so on...) 

@mleguen I will first very quickly admit that my French is *terrible* - however, as you probably know, something like 50-60% of English vocabulary comes from French. (Which is why I can go to France and usually **read** everything just fine but I cannot speak or pronounce the words I am reading.) My point is: a translation between EN<>FR is likely more direct and easier ("choice" = "choix" - interestingly, the [FR-Wiktionary article on "choix"](https://fr.wiktionary.org/wiki/choix) even lists translations into German - "Wahlmöglichkeiten" works too, I suppose...). Just a thought :)

mleguen:
@danieldjewell Although I cannot review this myself (I am not a native speaker), my brother lives in Germany, and I am going to have him try to find a native speaker to have a look at your suggestions.

> It might be necessary/a good idea to add additional options to the locales because some of the words as used in English can be both nouns and verbs simultaneously

In a past professional experience, the home-brewed localization system we used allowed us to add an optional context string to translations, to distinguish such cases.

@bcoe Is it something that could be added to y18n?

bcoe:
@mleguen I'm certainly open to extending on things, if need be. I think it would be good to get a native speaker to double check these translations (before we add any functionality).

maxrimue:
I'm all for changing `count` to `Anzahl`, I get the reasoning and am a little unsure why I chose `Zähler` in the past. When it comes to `Auswahloptionen`, I would perhaps go with `Optionen` directly instead, as "Auswahl" and "Optionen" are very similar in their meaning and rare to hear in that combination. I would almost argue that even in the English language "options" is more common than "choices" in this context

bcoe:
@danieldjewell sorry for leaving this for so long, do @maxrimue's suggestions make sense?

Perhaps we could address those nits, and work on getting this merged?

danieldjewell:
Apologies from me too @bcoe - A little distracted with everything going on in the world right now (hope everyone is staying safe and healthy!)

Thank you @maxrimue -- I totally see what you mean about ```Optionen``` -- and you're totally right, in English one would most likely use "options" (it does get kinda weird and contextual .. like for a prompt "Please select from these options:" or "Please choose an option:" ... funny when I stop to think about it, never ceases to amaze me how complicated languages can be :)....)

Issue #1570, opened 2020-03-01
Regression: 15.2.0 ignores nargs for array


kevinoid:
Yargs 15.2.0 included an incompatible change in how `nargs` is handled for `array` options.  For example, code for a repeatable option where the option should have a single argument on each occurrence (e.g. to behave like [`curl`'s `--header` option](https://curl.haxx.se/docs/manpage.html#-H)):

```js
#!/usr/bin/env node
const yargs = require('yargs')
console.log(
  yargs
    .option('header', {
      requiresArg: true,
      array: true,
      nargs: 1
    })
    .argv
)
```
When invoked as `./example.js --header foo bar` with `yargs@15.1.0` this would print:
```js
{ _: [ 'bar' ], header: [ 'foo' ], '$0': 'example.js' }
```
with `yargs@15.2.0` this would print:
```js
{ _: [], header: [ 'foo', 'bar' ], '$0': 'example.js' }
```
I bisected the change to b9409da (#1553) then to yargs/yargs-parser@4cbc188 as a result of #1098.  I'm not sure if this is a bug in `yargs-parser` or if this is the intended behavior (in which case it might warrant a major version bump and some guidance about how to get the previous behavior).

Thanks for considering,
Kevin

bcoe:
My thinking on this regression is that using `array` and `nargs` in conjunction was never documented behavior, so much as a side effect folks found (I noticed `mocha` did this as well, so you're not alone).

I will look at adding this behavior back to `yargs-parser`, so that the next release of yargs doesn't need to be breaking, but I would rather that this behavior in the future was more explicit, with something like:

```
#!/usr/bin/env node
const yargs = require('yargs')
console.log(
  yargs
    .option('header', {
      array: true,
      demand: 1
    })
    .argv
)
```

In general I'm not happy with array parsing, and would like to move towards arrays not being "greedy" by default, see:

https://github.com/yargs/yargs-parser#greedy-arrays

Instead I would rather the default behavior is closer to what you're describing, mainly `have a single argument on each occurrence`, potentially with introducing a `demand` option which would require that an array be a certain size.

If we have this, I'm honestly tempted to drop the `nargs` feature in the next major version of yargs and yargs-parser.

kevinoid:
Thanks for the detailed explanation @bcoe!  Sounds great.

[...]

Issue #1098, opened 2018-03-27
requiresArgs: true breaks expected array parsing behavior in v11


dorellang:
I noticed that when you have an array option with `requiresArgs: true`, then only one argument is parsed, even when the docs state that `--foo foo bar` should be parsed as `['foo', 'bar']`. 

For instance, the following code:
```
yargs
  .option('i', {
    alias: 'items',
    type: 'array',
    requiresArg: true,
    description: "Some items to be parsed."
  })
  .argv;
```

Would parse ```node test.js -i item1 item2 item3```   as 

```
{ _: [],
  help: false,
  version: false,
  i: [ 'item1' ],
  items: [ 'item1'],
  '$0': 'test.js'
}
```

But according the docs I would expect `argv.i == argv.items == ['item1', 'item2', 'item3']`, which happens when requiresArg is not specified.

This was working on v10. Apparently [this PR](https://github.com/yargs/yargs/pull/1054) changed the behavior.

tristanls:
The workaround I ended up using is checking for non-zero length in `coerce`:
```
            coerce: opt =>
            {
                if (opt.length < 1)
                {
                    throw new Error("Missing required argument: foo")
                }
                return opt;
            },
```

mleguen:
If I understand well, Yargs is now lacking a way of indicating an array option should have *at least one value*.

@bcoe Shouldn't we change nargs signature from `.nargs(key, count)` to `.nargs(key, min, max = min)`?

- `.nargs(key, 1)` would mean 1 and only 1 argument (unchanged), so would `.nargs(key, 1, 1)`
- `.nargs(key, 1, 3)` would mean 1 to 3 arguments
- `.nargs(key, 1, Infinity)` would mean at least 1 argument

This is something I already do in another parsing tool I'm developping, and it works well.

mleguen:
And in that cas `requireArgs` should be an alias for `.nargs(key, 1, Infinity)`, repairing the broken behavior.

rulatir:
@tristanls, do you think it is possible to somehow reuse the localized error message that would thrown by the built-in assertion triggered by `requiresArg`?

rulatir:
Hotfix the whopping huge failure-to-implement-spec bug first, reengineer the API later.

tristanls:
[...]

PR #1572, opened 2020-03-02
fix: address ambiguity between nargs of 1 and requiresArg

bcoe:
In #1054, `requiresArg` was refactored to use a `nargs` of `1` to specify that a value was required following an option (`--x foo` is valid, `--x` is not valid).

The problem with this design decision  was that it created an ambiguity when `array` and `nargs` were used in conjunction, one could not distinguish between:

* an array that should only accept one value.
* an an array that required _at least_ one value.

This behavior lead to issues like https://github.com/yargs/yargs/issues/1570, https://github.com/yargs/yargs/issues/1098, and https://github.com/mochajs/mocha/pull/4196.

---

To solve this problem [`nargs` now accepts the special value `NaN`](https://github.com/yargs/yargs-parser/pull/251), this value is used by `requiresArg` to specify that an `array` accepts "one or more" values following it. A `nargs` of `1` used in conjunction with `array`, continues to indicate that just one value should be accepted by an array.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "string-width": "^4.2.0",
     "which-module": "^2.0.0",
     "y18n": "^4.0.0",
-    "yargs-parser": "^17.1.0"
+    "yargs-parser": "^18.0.0"
   },
   "devDependencies": {
     "c8": "^7.0.0",
--- a/test/yargs.js
+++ b/test/yargs.js
@@ -2457,6 +2457,36 @@ describe('yargs dsl tests', () => {
       argv._.should.eql(['item2', 'item4'])
     })
 
+    it('should apply nargs with higher precedence than requiresArg: true', () => {
+      const argv = yargs(['-i', 'item1', 'item2', '-i', 'item3', 'item4'])
+        .option('i', {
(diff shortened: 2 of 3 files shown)


bcoe:
@kevinoid please try `npm i yargs@next`, I believe I've come up with a way to support both types of behavior that folks wanted (by changing the behavior of `requiresArg` slightly).

kevinoid:
Thanks @bcoe!  Unfortunately it didn't appear to fix the example from #1570 for me.  Steps I ran:
```sh
cd "$(mktemp -d)"
npm init -y
npm install yargs@next
cat > example.js <<'EXAMPLE_JS'
#!/usr/bin/env node
console.log(
  yargs
    .option('header', {
      requiresArg: true,
      array: true,
      nargs: 1
    })
    .argv
)
EXAMPLE_JS
node example.js --header foo bar
```
Which printed
```js
{ _: [], header: [ 'foo', 'bar' ], '$0': 'example.js' }
```
which matches the behavior from `yargs@15.2.0` rather than `yargs@15.1.0`.

Let me know if there's anything else I can do to help.

PR #1580, opened 2020-03-07
feat: add usage for single-digit boolean aliases

ruyadorno:
Add proper usage output for the now supported single-digit boolean aliases which landed in yargs-parser.

ref: https://github.com/yargs/yargs-parser/pull/255

/cc @bcoe @mleguen 

relates to: #1493


ruyadorno:
🎉

Issue #1518, opened 2019-12-24
Feature: deprecate an option/command


laggingreflex:
Allow setting `deprecated: true` for an option or a command, like:

```
yargs.options({
  oldOption: {type: 'boolean', deprecated: true }
  newOption: {type: 'boolean' }
})
```

which will show up in help as:

```
Options:
  --old-option [deprecated]
  --new-option
```

bcoe:
@laggingreflex this is a neat idea 👍

laggingreflex:
Cool. I'll try to implement myself this when I get some time. 

Any preferences for the following: 

* Key - "deprecated" 

* Type - I'm thinking boolean "true/false", but also a string like "Please use --newOption instead"

* Placing of the text - 

  Boolean

  1. `[deprecated] --option description [type] [default]`

  1. `--option [deprecated] description [type] [default]`

  1. `--option description [deprecated] [type] [default]`

  String

  1. `... [deprecated: text] ...`

  1. `... [deprecated - text] ...`

* Should `[type]` and `[default]` be hidden when `deprecated` is set? I was thinking they should be hidden as they're not gonna be needed anyway, but maybe that decision should be left to the user (they can always remove those fields themselves).

* Should all the deprecated options be pushed below the normal options? I'd prefer this. 

Anything else?

mleguen:
@laggingreflex Thanks for digging into this.

`deprecated` sounds good to me.

A boolean would be easier to implement, but a string would be more practical (nowing how not to use a deprecated option is good, and IMHO the first reason for having such a deprecation notice).

Perhaps you should have a look at existing keys, to try to mimic as much as possible the behavior and naming of an existing key, for homogeneity (I do not have the time right now to do it myself)?

Concerning text placement, I would suggest looking first as how "reference" CLI (GNU tools, git, docker, npm, etc.) do it (if they do it), not to reinvent the wheel.

I would also suggest keeping type and default, as a deprecated option is still valid and usable.

And if we add this, I would definitely want it also for commands!

laggingreflex:
Was able to finally get this into motion. Please checkout #1559

It implements deprecating just the option for now. Will try to work on command later. 

Also, I couldn't find much references in common tools (git, docker, npm) that do it. If you come across it please let me know. 

I did manage to find a deprecated option in Docker, but it actually didn't mention that option at all in its --help. 

[...]

PR #1585, opened 2020-03-09
feat: deprecate command

laggingreflex:
More work towards #1518

Still a work in progress. 

This is turning out to be a bigger change. Thought I'd ask for opinions before continuing. 

The big parts are the two main refactors: https://github.com/yargs/yargs/commit/9717c2031ae834159f7ff23d8eb943be6f856f9f & https://github.com/yargs/yargs/commit/ffd71e168bfdf830bb2bce0503664e1fdad3fedf, latter one still a work in progress (~50 failing tests ). 

If those are done properly, the feature itself seems pretty simple: https://github.com/yargs/yargs/commit/c71531976d2f30b2f1f566e163db94bb138d5ffe

Should I carry on with the refactor? Or would it be considered too large a change for a small feature?

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -68,7 +68,7 @@ module.exports = function command (yargs, usage, validation, globalMiddleware) {
     if (isDefault) {
       parsedCommand.cmd = parsedAliases[0]
       handler.aliases = parsedAliases.slice(1)
-      handler.command = handler.command.replace(DEFAULT_MARKER, parsedCommand.cmd)
+      handler.cmd = handler.cmd.replace(DEFAULT_MARKER, parsedCommand.cmd)
     }
 
     // populate aliasMap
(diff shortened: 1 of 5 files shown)


bcoe:
> Should I carry on with the refactor? Or would it be considered too large a change for a small feature?

I'm wondering if we could figure out a way to add this feature without a significant overall to commands. Commands being a fairly finicky part of the codebase.

How were you picturing that you would specify a deprecated command? It feels like if we simply made this an additional property on the command itself, like the `handler`, or `builder`, it shouldn't be too hard to output a deprecation warning without a significant refactor.

laggingreflex:
My basic assumption is that this has to be done [here][1], which is what https://github.com/yargs/yargs/commit/c71531976d2f30b2f1f566e163db94bb138d5ffe does (but that uses the refactor). 

Without the refactor the `command` there is an array which traces back to

* [`Usage.command`][2] 
* [`Command.addHandler`][3]

The `cmd` there (or any other object) doesn't seem to be available there to attach a property (`deprecated`). 

It's all arrays and params list. 

I could add the `deprecated` as another param?

```js
  -self.command = function command (cmd, description, isDefault, aliases) {
  +self.command = function command (cmd, description, isDefault, aliases, deprecated) {
```

Initially I did go this route and found a lot of tests failing this way too (hence may have required a big change too). Should I try it this way?



[1]: https://github.com/yargs/yargs/blob/master/lib/usage.js#L214
[2]: https://github.com/yargs/yargs/blob/master/lib/usage.js#L109
[3]: https://github.com/yargs/yargs/blob/master/lib/command.js#L74

bcoe:
@laggingreflex I think adding it as an additional parameter would be reasonable, even though it's a bit ugly. Maybe it's good that your `.command` code is a bit ugly if the command is deprecated, reminds you to actually drop the function eventually.

bcoe:
@laggingreflex want to rebase, and we can see this over the finish line? I think this is a great feature.

laggingreflex:
Please checkout #1624. It makes fewer changes.

Issue #1586, opened 2020-03-11
On the road to typescript


mleguen:
Here is a proposal about how to migrate progressively to typescript, to:
- avoid big-bang merge requests, impossible to review and difficult to merge without freezing other developpments for a while
- limit - and avoid if possible - impacts on projects depending on yargs

Proposed steps:
1. **(done: see #1587)** ~~a first merge request to:~~
   - ~~convert to typescript a first module, not depending on any other yargs' module (`is-promise` could be a good choice), renaming it from `lib/module.js` to `lib/module.ts` (compiled version: `build/lib/module.js` and `build/lib/module.d.ts` for its types)~~
   - ~~adapt other modules depending on this one to look for it in `build`~~
   - ~~and initialize typescript's configuration, dependencies and packaging scripts for this module to have only its compiled version in the final npm package (transparent for users)~~
1. **(done: see #1596, #1601, #1606, #1610, #1614, #1632, #1636 and #1654 )** ~~several other merge requests to convert to typescript all other modules in lib~~
1. **(done: see #1670)** ~~move most of `yargs.js` to `lib/yargs.ts` to keep it as a nearly empty shell (see below)~~
1. **(done: see #1670)** ~~keep `index.js` and `yargs.js` at the root of the project, mostly as empty shells importing typescript-built modules, to keep yargs require calls untouched in calling js code: `const yargs = require('yargs')` or `const yargs = require('yargs/yargs')` (changing to `const { yargs } = require('yargs')` would be necessary if we exposed directly typescript-built modules)~~
1. **(doing)** add `index.d.ts` and `yargs.d.ts` at the root of the project, for calling ts code to be able to access our types (mostly empty shells too)
   - ~start with the type definitions in `@types/yargs`~
   - replace them one by one by type definitions imported from ts files
   - when the new type names diverge from the ones in `@types/yargs`, keep old names as aliases to avoid a breaking change for ts users
1. submit a PR to @types/yargs to replace it by a warning that it is no longer necessary to use it as yargs is now providing its own types
1. update typescript examples
1. convert to typescript other yargs modules we are depending on (set-blocking, y18n and yargs-parser), or merge them into yargs code (require-main-filename, see yargs/require-main-filename#48)
1. contribute to `@types` to add module definitions for other dependencies 

EDITS: fixed typo, updated progress, review type naming strategy

mleguen:
I'm currently working on the 1st step

bcoe:
[...]

PR #1587, opened 2020-03-11
chore(ts): init typescript conf + tsify is-promise

mleguen:
First step of #1586.

It should not have any impact on user projects (neither js nor ts).

Includes linting of ts code.

Tested with `npm pack` too: the only change in the package file list is the replacement of `lib/is-promise.js` by the compiled `build/lib/is-promise.js`.

EDIT: updated path now that ts files are also in the lib dir


bcoe:
@mleguen convince you to try out https://www.npmjs.com/package/gts for linting? this would help keep yargs consistent with the TypeScript configuration I use at work, which would be awesome.

mleguen:
@bcoe In fact, I used standard to keep yargs consistent between js and ts linting, but no problem for me: always ready to try some new tool :-)

bcoe:
@mleguen depending on our timeline, I'd love for us to test `2.x.x`, which requires Node 8, but has a variety of improvements -- it could be used for both our JS and TypeScript linting.

mleguen:
@bcoe What I finally did not to block this PR is to use gts 2.x only for the ts configuration (and script names).

We could later use it for linting too as soon as we drop support for node 8 (but let's not fight 2 battles at the same time).

mleguen:
@bcoe ready for review

mleguen:
@bcoe I tsified test/is-promise too in this PR to stay coherent

PR #1589, opened 2020-03-13
chore(linting): switch from standard to gts 2.x

mleguen:
Follow-up of @bcoe's suggestion in #1587

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,3 @@
+build/
+example/
+test/fixtures/
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,3 @@
+{
+  "extends": "./node_modules/gts/"
+}
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,3 +1,8 @@
 {
-  "extends": "./node_modules/gts/"
+  "extends": "./node_modules/gts/",
+  "rules": {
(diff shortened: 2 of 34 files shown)


mleguen:
@bcoe Looks like gts is not working with node 8. This is confirmed by its package.json:
```json
  "engines": {
    "node": ">=10"
  },
```

So this switch will have to wait...

bcoe:
@mleguen `gts` is officially at `2.x.x` 🎉, also I've been thinking a lot on the topic ([I wrote this blog post on the subject for Node.js](https://medium.com/the-node-js-collection/maintainers-should-consider-following-node-js-release-schedule-ab08ed4de71a)) and would like to drop Node 8 soon.

How do you feel about releasing the last version of `15.x.x` soon (with your initial support for TypeScript), and then moving towards the next `major`, with `gts`, Node 8 being dropped, and any breaking TypeScript changes we need to take on, .e.g., rewriting index.js and yargs.js.

mleguen:
When we drop nodejs 8 support, we should not forget to remove explicit es2017 target in tsconfig.json, to keep the es2018 one from gts default tsconfig.json.

bcoe:
this has gotten a little stale, and I'm not convinced it's worth the trouble if  `standardx` is working for us, let's stick with `standardx` for now.

PR #1595, opened 2020-03-19
Add TOC to API documentation

piotros:


bcoe:
@piotros closing this for now, but I would be very interesting in collaborating on yargs.js.org, and figuring out how we can have a nicer layout that works both on the web and in the docs on GitHub.

PR #1601, opened 2020-03-25
chore(ts): tsify lib/completion

mleguen:
Follow up of the 2nd step of #1586: this PR tsifies the completion module.


mleguen:
@bcoe let's take the same method for async handling once we are done with tyepscript :-)

PR #1606, opened 2020-04-01
chore(ts): tsify usage

mleguen:
Follow up of the 2nd step of #1586: this PR tsifies the usage module.

A check was added too to ensure we never unfreeze usage more than we freezed it (thanks typescript for pointing out that `freezes.pop()` could return `undefined`!)


bcoe:
@mleguen [couldn't find your twitter](https://twitter.com/BenjaminCoe/status/1245936055183785990?s=20).

mleguen:
Thanks @bcoe 

> I appreciate that you're updating all the code first, and not the tests; so that we have a good safety harness as the conversion happens.

Well... It is not exactly what I intended to do at first :-) But as most tests are not depending on the module they test, but directly on yargs, there is no easy way for me to update the tests until yargs is tsified.

This is something I already noticed before, and something I think we should tackle in a near future as *we do not really have unit tests*, only integration ones. This will be a real problem for async developments, for example.

mleguen:
@bcoe 

> couldn't find your twitter

You won't find it as I never felt the need for a twitter account :-)

Thanks anyway for your kind word there.

PR #1610, opened 2020-04-03
chore(ts): tsify lib/argsert

mleguen:
Follow up of the 2nd step of #1586: this PR tsifies the `lib/argsert` module.

To do so, it also:
- tsifies `test/helpers/utils`
- extracts `parseCommand()` from `lib/command`, into an independent `lib/parse-command` module, to break the dependency cycle between `lib/argsert`, `lib/middleware` and `lib/command`
- add a test to ensure `parseCommand()` is called with at least a one word command string (thanks typescript again)
- removes some unused or no longer used function parameters
- replaces uses of `arguments` in `test/argsert` by an `...args` named argument, easier to read and type, and not needing an explicit conversion to array


lib/argsert.ts
```diff
@@ -1,24 +1,22 @@
...
+import { YError } from './yerror'
+import { parseCommand } from './parse-command'
+import { ParsedCommand } from './types'
 
 const positionName = ['first', 'second', 'third', 'fourth', 'fifth', 'sixth']
-function argsert (expected, callerArguments, length) {
+export function argsert (callerArguments: any[], length?: number): void
+export function argsert (expected: string, callerArguments: any[], length?: number): void
+export function argsert (arg1: string | any[], arg2?: any[] | number, arg3?: number): void {
+  function parseArgs (): [Pick<ParsedCommand, 'demanded' | 'optional'>, any[], number?] {
```
bcoe, line 9: I'd never seen `Pick` before, this is neat.

lib/parse-command.ts
```diff
@@ -0,0 +1,33 @@
+import { ParsedCommand } from './types'
+
+export function parseCommand (cmd: string) {
```
bcoe, line 3: I appreciate that this refactor is promoting a bit more decomposition.

test/types/check-output-result.ts
```diff
@@ -0,0 +1,7 @@
+export interface CheckOutputResult<T> {
+  errors: any[]
+  logs: any[]
```
bcoe, line 3: could we be more specific with this type, I think these are `string` arrays?

mleguen: Well, we use them as `string[]`, but their true type for now is `any[]`, as we get their contents directly from `console.{error,log,warn}` first parameter `message`, which is `any`.

So if we want to have real `string[]` (and we should, as we always test them again `string[]` afterwards), we should explicitly transform `message` to `string`, using `util.format` for example.

Although this has nothing to do with tsification, I will add a commit to fix that.

PR #1615, opened 2020-04-08
chore(deps): upgrade decamelize and get rid of @types/decamelize

mleguen:
Follow-up of #1609

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   ],
   "dependencies": {
     "cliui": "^6.0.0",
-    "decamelize": "^4.0.0",
+    "decamelize": "^3.2.0",
     "find-up": "^4.1.0",
     "get-caller-file": "^2.0.1",
     "require-directory": "^2.1.1",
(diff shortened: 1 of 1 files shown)


mleguen:
Oops! I was a bit ambitious here, upgrading directly to 4.0 which requires node >= 10.

I will downgrade to 3.2 next week, as it requires only node >= 6.

bcoe:
@mleguen we should turn on `--strict-engines` for npm installs in CI/CD to catch this sort of thing.

Something I've been thinking about.

coreyfarrell:
@bcoe @mleguen Would you consider reverting this in 15.x?  decamelize was held back to 1.x because 2.x and 3.x both depend on `xregexp` which is massive.  `npm i yargs@15.3.1` on an empty project creates a 1.1MB node_modules where `npm i yargs@15.4.0` creates a 12MB node_modules.  decamelize 4.x drops the xregexp dependency as node.js 10 has Unicode RegExp support natively.

PR #1629, opened 2020-04-17
fix(strictCommands): report unknown argument instead of unknown comma…

devoto13:
…nd if no subcommands configured

This gives user a friendlier error message when unexpected positional argument is passed to the command. It does not really make sense to say "Unknown command: foo" - there are no subcommands to this command. This is an unknown argument and should be reported as such.

A more correct solution would probably be to not report an error at all for this case under strictCommands, but report it under strict mode instead. But this will be a breaking change as it is explicitly [stated in the docs](https://yargs.js.org/docs/#api-strictcommandsenabledtrue) that unknown positional arguments are reported as errors under strictCommands, so current solution should do for now.


test/validation.js
```diff
@@ -972,6 +972,16 @@ describe('validation tests', () => {
         .parse()
     })
 
+    it('fails with error if positional argument is unknown', (done) => {
```
devoto13, line 975: So this test case fails. With current semantics of `strictCommands` I think this should pass. But making it pass will be a breaking change... Not sure what is the best way to proceed here?

test/validation.js
```diff
@@ -985,6 +995,41 @@ describe('validation tests', () => {
         .parse()
     })
 
+    it('fails with error if multiple inner commands are unknown', (done) => {
```
devoto13, line 998: Also while at it. Does this behaviour actually meaningful?
Wouldn't it be enough to only report the first unknown command, even if there are multiple?

mleguen:
@devoto13 I am not sure Iunderstand your use case. Could you please start by describing more precisely the issue you intend to fix with a code sample, and expected and current outputs?

devoto13:
@mleguen The primary issue is that `yargs` reports bad error message when you provide extra positional argument under `strictCommans`.

Give a command line like:

```
yargs('foo blarg')
         .strictCommands()
         .command('foo', 'foo command')
```

The current output:

```
Unknown command: blarg
```

The desired output:

```
Unknown argument: blarg
```

While trying to fix this, I've discovered that current `strictCommans` behaviour is inconsistent works differently whether it is a missing positional argument on root or to the inner command (see first comment). 

Second comment is optional and refers to improving the error message further.

Give a command line like:

```
yargs('foo blarg blarg2')
         .strictCommands()
         .command('foo', 'foo command', (yargs) => {
           yargs.command('bar')
         })
```
The current output:

```
Unknown commands: blarg, blarg2
```

The desired output (it is enough and more clear to report only the first invalid command):

```
Unknown command: blarg
```

Hope this makes it more clear.

devoto13:
@bcoe Any feedback on this?
I don't particularly care about the https://github.com/yargs/yargs/pull/1629#discussion_r410366168 use case, so I can just remove the failing test case if everything else looks good.

bcoe:
@devoto13 there was an error with the GitHub API (_potentially a UI error, as my Internet is running slow today._) that merged this before tests were passing. Would you mind rebasing with master and sending the same PR.

My general feedback, is I'm not sure if it's worth the additional logic to indicate that an unexpected `argument` was received, vs., a `command`. If folks have configured commands in their application, I think they might be in the mindset that positional arguments that are unexpected are unexpected commands.

devoto13:
@bcoe Yeah, the failing test case is [this one](https://github.com/yargs/yargs/pull/1629#discussion_r410366168). If one enables `strictCommands` on the root without having any commands configured, there will be no errors for unexpected positional arguments. While if one configures a command all subsequent positional arguments will become an error. I think this is inconsistent. It should either always complain about positional args (which are not commands) or never complain about positional args. Either of these solutions would be a breaking change, that's why I wanted to have some input here.

> A more correct solution would probably be to not report an error at all for this case *(unknown arguments if there are no commands configured)* under strictCommands, but report it under strict mode instead. But this will be a breaking change as it is explicitly stated in the docs that unknown positional arguments are reported as errors under strictCommands, so the current solution should do for now.

So yeah, I can re-open this PR, but it is still unmergable without a decision.

The motivating example is [Karma](https://github.com/karma-runner/karma/blob/master/lib/cli.js#L148). It is a CLI, which accepts exactly 1 command. It also accepts an optional positional argument after a command. The problem is this:

```
# options is a typo, strat is a typo
$ karma start karma.conf.js option
Unknown commands: karma.conf.js, option
$ karma strat karma.conf.js
Unknown commands: strat, karma.conf.js
```

Which for me as a user of the CLI is misleading: 
1. These are arguments to the `start` command, why are these called "unknown commands"?
2. Okay, `strat` is a typo, but why `karma.conf.js` is referred to as a command? 

Ideally, I would expect something like below instead:

```
$ karma start karma.conf.js option
Unknown argument: option # or no error at all if strictCommands only catches commands, not arguments
$ karma strat karma.conf.js
Unknown command: strat
```

For now, I was able to solve it by explicitly disabling `strictCommands` on subcommands, but it would be better to have it work properly out of the box.

It's also possible that I just completely miss the mental model behind this flag...

---

I tried `strict` just now and it results in:

```
# options is a typo
$ karma start karma.conf.js option
Unknown arguments: karma.conf.js, option
```

So this can probably be improved further... Anyways let me know which approach for `strictCommands` should I choose when re-opening a PR or should I just not re-open it at all (fine by me as I have a workaround)?

PR #1633, opened 2020-04-29
chore(ts): inline sources and source maps in the build

mleguen:
Adds .ts sources and source maps to the package, for easier debugging by yargs' users.

Adds no file, but 140kB to the package (+70%).


bcoe:
@mleguen I'm a little torn on this one:

* on one hand, I want debugging to be easy for folks (_I actually helped contribute source map support to Node.js_).
* on the other hand, 70% is a huge increase in package size:
  * yargs is installed 47,000,000 times a week, so this represents 8TB in bandwidth we're inflicting on the world (not insignificant).

I have an idea for you, since publication is automatic any ways, why don't we add a second automated publish that goes to the dist-tag, `debug`, and includes inline source-maps.

Then, if a user ever needs to debug yargs, they can just `npm i yargs@debug` but, for the general case, we drop the source-maps.

mleguen:
> on the other hand, 70% is a huge increase in package size

This is the reason why first removed it not so long ago :-)

bcoe:
@mleguen sounded like we can close this, but might want to add an additional publish option that includes the source-maps?

PR #1636, opened 2020-04-30
chore(ts): tsify lib/middleware

mleguen:
Follow up of the 2nd step of #1586: this PR tsifies the lib/middleware module.


bcoe:
thanks @mleguen 👌

PR #1642, opened 2020-05-03
docs(api): link to other Yargs documentation

roryokane:
The Yargs Docs page at https://yargs.js.org/docs/ is built from `api.md` (as can be seen in `_layouts/docs.html` in https://github.com/yargs/yargs.github.io/). This means that when browsing the Yargs website, it looks like the API docs are the only documentation Yargs has. These links make the other documents discoverable.


bcoe:
@roryokane I like the idea of adding these links 👍 one thought, could we instead look at whether it's possible to link to a whole docs directory on the website?

roryokane:
I agree it’s a good goal for [the website](https://yargs.js.org/docs/) to contain all relevant docs instead of the API docs being on the website and other docs being on GitHub.

I don’t feel like working on that now, though. I would have to learn the setup of the [yargs.github.io project][yargs.github.io-gh], learn the relevant parts of Jekyll and Flatdoc, add pages for each Markdown file in `docs/`, and then write up a root docs page that links to all those pages. I think just having the links in the Markdown file that is the current root of the docs, `docs/api.md`, is good enough to start with.

[yargs.github.io-gh]: https://github.com/yargs/yargs.github.io

bcoe:
@roryokane we built the yargs site several years ago now, and could probably rethink every decision made today 😄

I'm happy with adding this section in the meantime.

PR #1644, opened 2020-05-03
docs(api): clarify process.argv handling and the order of API methods

roryokane:
The final statement “the rest of these methods below come in just before …” now makes sense for both of these argument-passing styles:

- `require('yargs').argv`
- `require('yargs').parse()`

There is one more argument-passing style I’m not sure how to handle:

- `require('yargs')([ '-x', '1', '-y', '2' ])`

I think I once tried putting methods earlier in the chain like `require('yargs').help().([ '-x', '1', '-y', '2' ])` and got an error. If anyone knows the correct place to put the chain methods like `.option` when passing arguments the above way, I can update this PR to describe it.


bcoe:
@roryokane this seems like a reasonable clarification to me, I'll wait for @mleguen to chime in too.

docs/api.md
```diff
@@ -38,7 +38,15 @@ Calling `.parse()` with no arguments is equivalent to calling `yargs.argv`:
 require('yargs').parse()
 ```
 
-The rest of these methods below come in just before the terminating `.argv`.
+When passing in the arguments yourself, note that Yargs expects the passed array
+to contain only the arguments after the program name, while Node’s
```
mleguen, line 42: Why not being a step more general, as we support electron too now, for example:

```suggestion
to contain only the arguments after the program name, while `process.argv` usually
starts with extra elements. For example, Node’s
```

PR #1654, opened 2020-05-06
chore(ts): tsify lib/command

mleguen:
Follow up (and last PR) of the 2nd step of #1586: this PR tsifies the lib/middleware module and fixes some bugs revealed by type errors:
- `true` was silently accepted as a command description, leading to displaying "true" as the description
- some no longer used function parameters had not been removed
- `options.aliases` was wrongly used as `string` instead of `string[]` in `parseCommand`, leading to the last positional alias only to be taken into account


mleguen:
To be rebased before review, once #1636 is reviewed and merged.

lib/common-types.ts
```diff
@@ -1 +1,13 @@
+import { notStrictEqual } from 'assert'
+
 export type Dictionary<T = any> = { [key: string]: T }
+
+export type NotEmptyArray<T = any> = [T, ...T[]]
+
```
bcoe, line 6: is the idea that these types aren't specific to yargs data structures?

lib/yargs-types.ts
```diff
@@ -1,17 +1,30 @@
...
   argv: Arguments
   customScriptName: boolean
   parsed: DetailedArguments
+  _copyDoubleDash (argv: Arguments): void
   _getLoggerInstance (): LoggerInstance
   _getParseContext(): Object
+  _hasOutput (): boolean
   _hasParseCallback (): boolean
+  // TODO: to be precised once yargs is tsified
+  _parseArgs (args: null, shortCircuit: null, _calledFromCommand: boolean, commandIndex: number): Arguments
```
bcoe, line 19: when we make the next breaking change of yargs, should we migrate any of these `_` methods to being `private` in TypeScript?

bcoe:
@mleguen a couple questions, I'm quite happy with this though 👍

lib/common-types.ts
mleguen: Yes. The main reason for this is `split()` signature is not precise enough: it returns `string[]` as if the first argument could be undefined, which is not the case. Its real signature should be `[string, ...string[]]`.

When you are destructuring the results of `split()`, for example:
```ts
const [cmd, ...args] = something.split();
```
then cmd is `string | undefined` when it should be `string`.

lib/yargs-types.ts
mleguen: Sure.

The next breaking change in yargs should be to switch from factories to classes, and use private and public methods to distinguish inner and API methods.

This would ease a lot understanding, IDE tools use, type inference, etc.

lib/common-types.ts
mleguen: To answer more precisely your question: `common-types` was at first the place to put types several modules would use, but at the moment is mostly used by types which I think should be typescript standards, but are not yet implemented.

PR #1671, opened 2020-06-04
chore(ts): ship yargs.d.ts

mleguen:
1st part of 5th step of #1586: this PR ships a yargs.d.ts adapted from the one in `@types/yargs`.

### DIFF
--- a/lib/yargs.ts
+++ b/lib/yargs.ts
@@ -1388,9 +1388,11 @@ export function Yargs (processArgs: string | string[] = [], cwd = process.cwd(),
 
 // rebase an absolute path to a relative one with respect to a base directory
 // exported for tests
-export function rebase (base: string, dir: string) {
-  return path.relative(base, dir)
+export interface RebaseFunction {
+  (base: string, dir: string): string
 }
+
+export const rebase: RebaseFunction = (base, dir) => path.relative(base, dir)
 
 /** Instance of the yargs module. */
 export interface YargsInstance {
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
   "files": [
     "index.js",
     "yargs.js",
+    "yargs.d.ts",
     "build",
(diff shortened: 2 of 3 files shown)


bcoe:
@mleguen I think in the spirit of SemVer, I would call this step breaking:

* pulling in our own `yargs.d.ts` will mean that users can't easily override `@types/yargs`.
* it sounded like there were a few changes to `@types/yargs` required.

Should we just call this our first commit to `yargs@16.x`, and land #1670 and make this the last patch we land in `15.4.0`.

---

I think we should perhaps let `15.4.0` bake for a few days, after we release it, since it's our first significant TypeScript release.

*but*, after that short baking period, let's march towards the 16.x release.

mleguen:
Sorry, I did not progress much on this the past couple of weeks. I only had time to read the incoming notifications from github... I will try to do better in the coming weeks :-)

bcoe:
@mleguen how do you feel about me getting yargs@15.4.0 released, as I think our continued TypeScript work will be breaking ... perhaps we could try to land the `.mjs` import too, based on your experiments, and then plan that we'll probably be taking breaking changes to get this work over the finish line.

mleguen:
I think it is a good idea, not to block the "normal life of yargs" while we
are terminating the switch to ts.

Landing the `.mjs` import too would be great, if a PR is ready.

Le lun. 22 juin 2020 à 18:16, Benjamin E. Coe <notifications@github.com> a
écrit :

bcoe:
@mleguen I did some testing, Node 14 actually seems pretty smart about hits handling of common JS. With yargs, as it is published today, this just works (I believe):

```js
import yargs from 'yargs'
  
console.info(yargs.argv)
```

Without us needing to add any conditional exports. So, I don't think a ton of value in us adding the conditional exports in our case.

PR #1682, opened 2020-06-28
feat(): add examples

rluvaton:
add function to add multiple examples at once

#1683

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -766,11 +766,22 @@ var argv = require('yargs')
 
 .example(cmd, desc)
 -------------------
+.example(exampleArray)
+-------------------
 
 Give some example invocations of your program. Inside `cmd`, the string
 `$0` will get interpolated to the current script name or node command for the
 present script similar to how `$0` works in bash or perl.
 Examples will be printed out as part of the help message.
+
+If you want to add multiple examples at once, just pass an array of examples, e.g
+```js
+require('yargs')
+  .example([
+    ['$0 --config "~/config.json"', 'Use custom config'],
+    ['$0 --safe', 'Start in safe mode']
+  ]);
+```
 
 <a name="exitprocess"></a>.exitProcess(enable)
 ----------------------------------
--- a/docs/api.md
+++ b/docs/api.md
@@ -766,7 +766,7 @@ var argv = require('yargs')
 
 .example(cmd, desc)
 -------------------
-.example(exampleArray)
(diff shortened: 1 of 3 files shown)


bcoe:
I like this idea 👍 I think I would rather we overrode `example` to accept a `string` or `string[]` rather than adding the new `.examples()` though -- there are already several methods that accept both the singular and array form.

rluvaton:
> I like this idea 👍 I think I would rather we overrode `example` to accept a `string` or `string[]` rather than adding the new `.examples()` though -- there are already several methods that accept both the singular and array form.
> 
> 
> 
> Thank you for the contribution.

Fixed it in the latest commit
_(For some reason this PR hasn’t updated with that commit 😬)_

rluvaton:
@bcoe I recommend you would squash this PR (it contain lots _oh I forgot something_ commit 😬)

docs/api.md
```diff
@@ -766,12 +766,23 @@ var argv = require('yargs')
 
 .example(cmd, desc)
 -------------------
+.example(exampleArray)
```
bcoe, line 769:
```suggestion
.example([[cmd1, desc1], [cmd2, desc2]])
```

☝️  what do you think?

bcoe:
> I recommend you would squash this PR (it contain lots oh I forgot something commit 😬)

@rluvaton no worries, we always squash 😄

docs/api.md
rluvaton:
```suggestion

.example([[cmd1, desc1], [cmd2, desc2], ...])
```

Maybe that (added 3 dots to let them know it can be any size)?

bcoe:
@rluvaton thank you for the contribution.

rluvaton:
@bcoe when this feature should be release?

bcoe:
@rluvaton if you have a few moments, mind testing out a pre-release of this feature:

```
npm i yargs@15.4.0-beta.1
```

I'll release soon if things are looking good.

rluvaton:
> @rluvaton if you have a few moments, mind testing out a pre-release of this feature:
> 
> ```
> npm i yargs@15.4.0-beta.1
> ```
> 
> I'll release soon if things are looking good.

I tried it and it looks good so far (except the missing type definition for the new feature [obviously])

bcoe:
@rluvaton okay, `yargs@15.4.0` is actually released.

PR #1703, opened 2020-08-03
chore(yargs): introduce genericity in YargsInstance

mleguen:
To start aligning yargs.ts on yargs.d.ts

### DIFF
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -17,23 +17,23 @@ const DEFAULT_MARKER = /(^\*)|(^\$0)/
 // handles parsing positional arguments,
 // and populating argv with said positional
 // arguments.
-export function command<T> (
-  yargs: YargsInstance<T>,
+export function command<R> (
+  yargs: YargsInstance<R>,
   usage: UsageInstance,
   validation: ValidationInstance,
-  globalMiddleware: Middleware<T>[] = []
+  globalMiddleware: Middleware<R>[] = []
 ) {
-  const self: CommandInstance<T> = {} as CommandInstance<T>
-  let handlers: Dictionary<CommandHandler<T>> = {}
+  const self: CommandInstance<R> = {} as CommandInstance<R>
+  let handlers: Dictionary<CommandHandler<R>> = {}
   let aliasMap: Dictionary<string> = {}
-  let defaultCommand: CommandHandler<T> | undefined
+  let defaultCommand: CommandHandler<R> | undefined
 
   self.addHandler = function addHandler (
-    cmd: string | string[] | CommandHandlerDefinition<T>,
+    cmd: string | string[] | CommandHandlerDefinition<R>,
     description?: string | false,
-    builder?: CommandBuilder<T> | CommandBuilderDefinition<T>,
+    builder?: CommandBuilder<R> | CommandBuilderDefinition<R>,
     handler?: CommandHandlerCallback,
-    commandMiddleware?: Middleware<T>[],
+    commandMiddleware?: Middleware<R>[],
     deprecated?: boolean
   ) {
     let aliases: string[] = []
@@ -43,7 +43,7 @@ export function command<T> (
     if (Array.isArray(cmd)) {
       aliases = cmd.slice(1)
       cmd = cmd[0]
-    } else if (isCommandHandlerDefinition<T>(cmd)) {
+    } else if (isCommandHandlerDefinition<R>(cmd)) {
       let command = (Array.isArray(cmd.command) || typeof cmd.command === 'string') ? cmd.command : moduleName(cmd)
       if (cmd.aliases) command = ([] as string[]).concat(command).concat(cmd.aliases)
       self.addHandler(command, extractDesc(cmd), cmd.builder, cmd.handler, cmd.middlewares, cmd.deprecated)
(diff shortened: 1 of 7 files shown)


lib/command.ts
```diff
@@ -435,7 +435,7 @@ export function command (
   // the state of commands such that
   // we can apply .parse() multiple times
   // with the same yargs instance.
-  const frozens: FrozenCommandInstance[] = []
+  const frozens: isCommandBFrozenCommandInstanceuilderOptionDefinitions<T>[] = []
```
mleguen, line 438: Strange renaming here, to be fixed

lib/yargs.ts
```diff
@@ -351,7 +355,7 @@ export function Yargs (processArgs: string | string[] = [], cwd = process.cwd(),
     K extends keyof Options[T] & string = keyof Options[T] & string,
     V extends ValueOf<Options[T]> = ValueOf<Options[T]>
   > (
-    builder: (key: K, value: V, ...otherArgs: any[]) => YargsInstance,
+    builder: (key: K, value: V, ...otherArgs: any[]) => YargsInstance<T>,
```
mleguen, line 358: The wrong T is used here.

R (for result) should be used in this whole genericity introduction instead of T which is already used in several places

mleguen: Fixed

lib/command.ts
mleguen: Fixed

mleguen:
Ready for review new

bcoe:
@mleguen 👍 one thing I've been doing with `yargs-parser`, is starting to add a suite that specifically exercises types:

https://github.com/yargs/yargs-parser/pull/295/checks?check_run_id=939183645

I think we could copy this approach as we start to add more sticky typing to this library.

In general, I'm really happy where `yargs-parser` is landing, I've managed to target  `deno`, `ESM`, `CJS`, and TypeScript at the same time 😄

yargs.d.ts
```diff
@@ -28,7 +28,7 @@ import { YargsInstance, RebaseFunction } from './build/lib/yargs'
 import { Parser, DetailedArguments, Configuration } from 'yargs-parser/build/lib/yargs-parser-types'
 
 declare namespace yargs {
-  type BuilderCallback<T, R> = ((args: YargsInstance<T>) => PromiseLike<YargsInstance<R>>) | ((args: YargsInstance<T>) => YargsInstance<R>) | ((args: YargsInstance<T>) => void);
+  type BuilderCallback<R1, R2> = ((args: YargsInstance<R1>) => PromiseLike<YargsInstance<R2>>) | ((args: YargsInstance<R1>) => YargsInstance<R2>) | ((args: YargsInstance<R1>) => void);
```
bcoe, line 31: what's the motivation for moving away from the variable names for generics defined in `yargs.d.ts`. I was picturing it might be easy to simply pull these over.

mleguen: 2 reasons here:
1. T is already used as a generic in several places, so the new T was not usable there (hidden)
2. using T, U, V, etc. for generics makes sense when using them in a few lines long function. However, when you use it application-wide as it is the case here, using letters with more meaning is a necessity to be able to know what you are talking about

I even hesitated using full names, but I finally stuck to one letter generics for readability (it would have lengthen many lines, and people are used to one letter generics).

bcoe: I trust you @mleguen 👍 my concern is  just it becomes a bigger job for you  -- this is outside my comfort zone, with regards to generics.

What I'd like to see, and would be a huge help for my learning, is let's add a TypeScript test suite that literally just exercises type behavior, [like this](https://github.com/yargs/yargs-parser/runs/943541075). Then as the implementation matures, we can write assertions like:

```
it('returns a type of string if .string() is used to configure key')
```

mleguen: @bcoe I would also be much more comfortable if I could rely on a test harness for pointing me out type issues.

However, there are many differences between the "ideal" types described by old @types/yargs now in yargs.d.ts and yargs' code reality, so writing these tests (and having them pass) would not be that straightforward (although necessary, reusing old @types/yargs does is really a pain!).

Sadly, I have to admin I no longer have the necessary bandwidth to be a lead on this...

So let's put yargs.d.ts/yargs.ts aligning in standby for now, until one of us has the time to write such tests.

bcoe: I'm questioning whether it's worth the maintenance burden of pulling in types as complex as `yargs.d.ts` to the library.

It might be better to do without generics when working on the codebase itself, and opt to continue shipping `yargs.d.ts`.

I think this could get us to the goal of a full `ts` conversion, while minimizing risk.

mleguen: I agree with this, but in that case we must keep in mind that:
- `yargs.d.ts` still needs fixing as its signatures does not match exactly the codebase
- every change to `lib/*.ts` would need to be duplicated into `yargs.d.ts` and, without type test harness, we would not have anything broken if we forget to do it (although that would not be worse than when users used `@type/yargs`)
- without test harness, any modification to `yargs.d.ts` would be untested

bcoe: I think we're in agreement. I will start putting some work into building out "typescript" test suite that exercises the types specifically.

I was looking at how DefinitelyTyped approaches this --  it's difficult to use their test runner, but I found this eslint plugin inspired by it:

https://www.npmjs.com/package/eslint-plugin-expect-type

Would it be okay if I close out this PR for now? I would like to put some work into:

1. figuring out ESM/CJS interop.
2. moving towards our next major release.
3. getting back in the swing of addressing a few bug fixes.

I agree with your suggestion of putting this on standby for now while we get this harness in place.

Issue #1701, opened 2020-07-28
Change on strict behaivor was a breaking change, a.k.a strict only for unknown options


fcastilloec:
I know issue #1625 is closed and #1626 merged, but this was a breaking change (major version) from what `yargs` used to do and should have been properly labeled. 

I was actually using `strict` old behavior as the opposite of `strictCommand`, in other words, `strict` will only apply to unrecognized options and not commands/arguments.
I had a command that takes various filenames, and I was counting on `strict` to display an error if the user used an unknown option but not on the filenames, which are seen as unknown arguments now.
Now, I have the opposite problem of #1621, I can't use either `.strict()` or `.strictCommand()`, and I can't prevent the user from adding unrecognized options.

A workaround is to using `check` to manually see any unknow options and throw an error, but this is not ideal, especially because `yargs` was able to do all this. This becomes even harder if you have many options in your command and you're using aliases and/or dashes.
Perhaps, adding a `.strictOption()` method might be a good idea.

bcoe:
@fcastilloec we try to avoid it, but sometimes someone is relying on an edge-case of a feature that isn't covered appropriately by tests, and we accidentally break due to an oversight such as this.

I would certainly be open to the idea of adding a new method like `.strictOption()`.

If you have the time, could you submit a failing test case for the behavior  was working and began failing for you?

bcoe:
@fcastilloec please try `npm i yargs@next` it has the `strictOptions` functionality.

fcastilloec:
@bcoe Thanks! I just tried it and it's working in most cases but I found a bug when combined with: `parserConfiguration({'strip-aliased': true})`. It's related to case-sensitive options, and of course aliases. Here's some code to reproduce the bug:

```js
yargs(['foo', '-a'])
  .parserConfiguration({ 'strip-aliased': true })
  .strictOptions()
  .option('and', { alias: 'A' })
  .parse();
```
The previous code should fail but it passes. If you set `'strip-aliased': false` then it fails. As you can see, there's something happening with capitalization and aliases.
I'll make sure to copy this code and report this in a new issue report.

PR #1710, opened 2020-08-06
add test to reproduce #1701

fcastilloec:
I tried to simplify my scenario as much as possible, so I hope that this actually covers it and shows the difference between the previous and current behaviors.

### DIFF
--- a/test/validation.js
+++ b/test/validation.js
@@ -337,6 +337,18 @@ describe('validation tests', () => {
       expect.fail('no parsing failure')
     })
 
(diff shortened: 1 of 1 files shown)


bcoe:
@fcastilloec thank you for the failing test 👌

bcoe:
@fcastilloec I think you're right that it would be worth adding `strirctOptions` as a complement to `strictCommands`.

bcoe:
closing in favor of https://github.com/yargs/yargs/pull/1738

PR #1724, opened 2020-08-25
Pass argv to error handler.

chevtek:
If you pass a context object into `yargs.parse` it eventually gets merged with `argv` and passed to the command handler. If you need something from that context when an async error occurs there is currently no way to access that. Here is a pared down example:

```js
yargs
  .scriptName("discord-bot")
  .fail((msg, err, yargs) => {
    onFail(err, ???); // Unable to reference parse context
  });

function onFail (err, channel) {
  channel.send(err.message);
};

export function parse (cmdInput, context) {
  const { channel } = context.discord.message;
  try {
    yargs.parse(cmdInput, context, (err, argv, output) => {
      if (err) return onFail(err, channel); // Able to reference parse context
      const { help } = argv;
      if (output) {
        channel.send(output);
      }
    });
  } catch (err) {
    onFail(err, channel); // Able to reference parse context
  }
}
```

In this particular example `yargs.parse` is triggered within a custom `parse` function that is invoked upon a message event from a discord server. The `context` object contains metadata specific to a single message event so that yargs commands can reference them. We can't move the `.fail` call into the `parse` function because it will register a new error handler each time the `parse` function is invoked and there is no way to remove registered fail handlers. The ability to remove a fail handler is another possible solution but the most intuitive solution to me would seem to be to just continue plumping that context through to the fail handler. This PR enables the following modification:

```js
yargs
  .scriptName("discord-bot")
  .fail((msg, err, yargs, argv) => {
    const { channel } = argv.discord.message;
    onFail(err, channel); // We are now able to reference the parsed argv and the merged parse context
  });

function onFail (err, channel) {
  channel.send(err.message);
};
```


bcoe:
@chevtek I'm open to this feature request, but mind adding some through tests to `test/yargs.js`. This will help be better understand the functionality you are looking for.

bcoe:
@chevtek is this still something you'd like to implement? a couple things jump out at me:

1. tests are failing  because `argv` is defined as optional in the Usage interface, but not as optional in the `fail` handler.
2. you would also want to pass the `argv` instance to the `fail` handler in  the `yargs-factory.ts` file.

With this feature however, I'm left wondering if it's  really  needed. Does this provide a significant advantage over writing  something like this:

```
const context = {hello: 'world'}
yargs
  .command('some-command', 'desription', () => {}, () => {})
  .fail((message, err) => {
    console.info(context.hello)
  })
  .demandCommand(1)
  .parse()
```

chevtek:
That's essentially what I'm doing now but this is running in a discord bot where messages can come in asynchronously. Meaning the higher scoped context object has to be different for each message. Currently this means I have to instantiate Yargs with every single message so that I can have unique higher scopes context object. It seems to work fine this way. I'm not sure how much overhead comes with redoing the whole Yargs chain every time a discord message is received but so far it seems okay.

This change would allow a single Yargs instance in situations like mine. Do you think that's a valuable change or does it make more sense to do what I'm currently doing? Perhaps it is the intention to not invoke parse multiple times on a single Yargs instance? If you think this feature would be useful still then I'll happily make the changes to pass the tests :)

bcoe:
@chevtek let's get the tests passing and see if it works invoking the same instance multiple times 👍 

I would like for this to work, but in the past have noticed some bugs related to state (_I hope we've addressed them all_).

bcoe:
@chevtek  is this still something you'd like to land?

bcoe:
@chevtek cleaning up some old PRs, please feel free to reopen this if it is a feature you would like to come back to.

PR #1726, opened 2020-08-27
feat!: tweaks to API surface based on user feedback

bcoe:
based on initial feedback regarding the `v16.x.x` API surface, this PR:

1. switches ESM and Deno entry points to export the `yargs` function as a default export.
2. moves deno's types into `./types.ts`.
3. adds a `helpers.mjs` with a `hidBin` helper, which replaces the verbose `getProcessArgvWithoutBin` helper.

TODO:

- [ ] expose yargs parser.


bcoe:
cc: @mikeal

PR #1742, opened 2020-09-09
fix: code was not passed to process.exit

### LINKED ISSUE #1740: (16.0.0 regression) Missing demandOption gives exit code 0, should be non-zero
In <16.0.0, yargs exited with a non-zero exit code if a required option (`demandOption: true`) is omitted by a user. In 16.0.0, this no longer results in a non-zero exit code. Technically, [the API docs for demandOption](https://github.com/yargs/yargs/blob/master/docs/api.md#demandOption) do only guarantee that yargs exits in this scenario, not that it exits with any particular exit code, but I'm guessing/hoping this is unintentional since it isn't documented among the v16 changelog's breaking changes.

Minimal example:

`cli.js`
```js
require('yargs').option('foo', { demandOption: true }).argv;
```

With yargs 15.4.0 (working as expected):

```powershell
> npm install yargs@15.4.0
> node ./cli.js
Options:
  --foo                                                               [required]

Missing required argument: foo
> $LASTEXITCODE
1
```

With yargs 16.0.0 (regression):

```powershell
> npm install yargs@16.0.0
> node ./cli.js
Options:
  --foo                                                               [required]

Missing required argument: foo
> $LASTEXITCODE
0
```

Motivating test failure in our consuming repo in https://github.com/microsoft/axe-sarif-converter/pull/394

bcoe:
Fixes #1740

### DIFF
--- a/lib/platform-shims/cjs.ts
+++ b/lib/platform-shims/cjs.ts
@@ -27,8 +27,8 @@ export default {
     argv: () => process.argv,
     cwd: process.cwd,
     execPath: () => process.execPath,
-    exit: () => {
-      process.exit()
+    exit: (code: number) => {
+      process.exit(code)
     },
     nextTick: process.nextTick,
     stdColumns: typeof process.stdout.columns !== 'undefined' ? process.stdout.columns : null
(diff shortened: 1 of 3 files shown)


dbjorge:
@bcoe Verified that this fixes the issue in my local repro

dbjorge:
(thanks for the quick turnaround!)

PR #1750, opened 2020-09-10
allow custom version handler by consuming application

fluffynuts:
I'd like to submit a PR against `gulp-cli` to update the `yargs` dependency, for security issues (against a very old version of `yargs-parser`) and the one issue standing in my way is that `yargs` completely takes over handling `--version`. For most people, this probably works well out of the box, but for `gulp-cli`, the app returns two version strings: the local version of gulp and the global version of gulp, if found. So a custom handler for `--version` is required.

This PR simply allows a `yargs` consumer to define a `--version` option which they will then have to handle themselves.


bcoe:
Hey @fluffynuts 👋 I  think this feature is potentially redundant, you can already turn off version handling like so:

```
yargs().version(false).option('version', {describe: 'custom version'})
```

At which point in your application you should be able to do this:

```
if (argv.version) {
  console.info('some version').
}
```

Perhaps we could just add a section  to the docs under version that spells this out better?

fluffynuts:
Yes! documenting would really help -- I tried searching for how to bypass the automatic version handling, and didn't have any joy, leading me to spend 1/2 an hour on this PR (:

fluffynuts:
In addition, adding `.verbose(false)` appears to be breaking parsing of other arguments )':

bcoe:
> In addition, adding .verbose(false) appears to be breaking parsing of other arguments

@fluffynuts do you mean adding `version(false)`? (_verbose isn't in the API as far as I know._).

`.version(false).option.etc` <-- this should definitely work, can you provide an example of it failing.

fluffynuts:
@bcoe yes, I believe that was a typo -- I meant `.version(false)`.

Unfortunately (for this thread), I've deleted the repo and local code that I was using. I provided a PR for gulp-cli to upgrade yargs, taking into account syntax changes since version 7. And,like every other interaction I've had with gulp devs, it was unproductive, so I didn't see the point in keeping the fork or the local clone around. Apologies.

After spending a couple of hours trying to be helpful, I'm just over it, so please ignore all of this.

PR #1752, opened 2020-09-11
fix: make positionals in -- count towards validation

### LINKED ISSUE #1732: demand() ignores anything after `--`
Here's an example:
https://runkit.com/cspotcode/5f4fe1f3c5d0c8001a637b54

```
some-cli --arrayOption foo bar -- positional1 positional2
```
With:
```
.demand(2)
.array('arrayOption')
```

Complains: "Not enough non-option arguments: got 0, need at least 2"

An array option accepts multiple values.  So to separate it from the demanded positionals, we use `--`.
`demand()` ignores all values after `--`.  Removing `demand()` shows that the values are `--` are parsed correctly and included in the `_` array.

Does `demand`'s deprecation mean this cannot be fixed?

bcoe:
@cspotcode this should return things to the old behavior.

Fixes #1732

### DIFF
--- a/lib/validation.ts
+++ b/lib/validation.ts
@@ -20,7 +20,8 @@ export function validation (yargs: YargsInstance, usage: UsageInstance, y18n: Y1
   self.nonOptionCount = function nonOptionCount (argv) {
     const demandedCommands = yargs.getDemandedCommands()
     // don't count currently executing commands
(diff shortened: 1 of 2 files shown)


cspotcode (approved):
Looks good, thanks.

bcoe:
@cspotcode just released this to npm, let me know if you're continuing to bump into any issues?

PR #1768, opened 2020-09-25
feat: expose hideBin helper for CJS

### LINKED ISSUE #1648: Promote the non-singleton interface, require('yargs/yargs'), as the default
## Background: problems with the singleton interface

When I set up Yargs in my command-line app, I learned how to use Yargs from two places: the main documentation page at https://yargs.js.org/docs/ and the examples [in `docs/examples.md`](https://github.com/yargs/yargs/blob/master/docs/examples.md), which I found by searching the web for “yargs example”. According to these instructions, I imported Yargs with `import * as yargs from 'yargs'`, the TypeScript equivalent of `require('yargs')`. This `'yargs'` import, I later learned, is called Yargs’s **singleton interface**.

While following the instructions in those places, I experienced **two bugs** in my app that were ultimately caused by this imported `yargs` object having global state. A summary of these two problems, if you’re curious:

- *Problem:* Tests of my command-line parsing with `.parse()` were failing due to state shared from previous test runs.
    - *Solution:* Call `yargs.reset()` before adding flags to `yargs` so it resets before each test.
- *Problem:* Yargs was acting like I never passed it `.options` and other methods, but only when showing usage with `--help`, not when showing usage due to an argument that failed validation.
    - *Solution:* Set `"esModuleInterop": false` in my `tsconfig.json` for TypeScript (this solution took a long time to track down.)

While searching the Yargs issue tracker for another issue, I found a reference to an alternative to `.reset()`: the Yargs **non-singleton interface**, accessed with `require('yargs/yargs')`.

The documentation I found for it (in [Advanced Topics](https://github.com/yargs/yargs/blob/master/docs/advanced.md)) is slim and didn’t answer all of my questions, but I managed to switch my app to use the non-singleton interface instead of plain `require('yargs')`. After undoing the two previously-mentioned solutions for the state-related problems I experienced, my app still seems to work. I can also see that my code using `require('yargs/yargs')` is no more complicated than the code I had using `require('yargs')` – all I had to change was adding `()` to the result of the import.

## Arguments for the non-singleton interface

My success with the non-singleton interface makes me think that if the documentation had explained and promoted the non-singleton interface from the start, in https://yargs.js.org/docs/ and other places, I **would never have encountered those two bugs**.

I don’t see any pattern of use that the singleton interface allows and the non-singleton interface doesn’t allow. The singleton interface does make it slightly easier to spread out specification of Yargs commands across different files, but that isn’t much harder to do with the non-singleton interface. That pattern seems likely to lead to confusing code anyway, and so it’s not a good default to encourage.

Also, in general, I would call global state a code smell – a property of hard-to-understand programs. It should only be used when the convenience makes up for the increased likelihood of bugs. I see the singleton interface’s enabled convenience as being very minor and its enabled bugs as being serious.

## Question

Is there a reason I am missing to promote the singleton interface as the default, or **should we change the documentation to promote the non-singleton interface as the default?** I could contribute some of those documentation changes.

The singleton interface should probably still be mentioned in the API documentation and in Advanced Topics, but it should be clearly labeled as only for the sake of those with programs already written using that interface.

bcoe:
Exposes the `hideBin` helper for CJS, so that we can standardize the documentation for ESM, CJS, and Deno.

Updates documentation to suggest this approach by default.

Fixes #1648

_TODO: update the rest of the examples in the documentation._


README.md
```diff
@@ -21,9 +21,22 @@ Yargs helps you build interactive command line tools, by parsing arguments and g
 It gives you:
 
 * commands and (grouped) options (`my-program.js serve --port=5000`).
-* a dynamically generated help menu based on your arguments.
+* a dynamically generated help menu based on your arguments:
 
-> <img width="400" src="https://raw.githubusercontent.com/yargs/yargs/master/screen.png">
+```bash
```
roryokane, line 26: I think the `bash` syntax highlighting is less readable than no syntax highlighting, due to its random-seeming coloring of `&` and `return` in the help text.

````suggestion
```
````

Also, this was a problem in the previous version too, but I notice this whole code block should be indented with two spaces so it counts as being within the bullet point, rather than outside of the bulleted list. So ultimately this section of the README should be written like this:

````markdown
* a dynamically generated help menu based on your arguments:

  ```
  mocha [spec..]
  
  Run tests with Mocha
  
  Commands
    mocha inspect [spec..]  Run tests with Mocha                         [default]
    mocha init <path>       create a client-side Mocha setup at <path>
  
  Rules & Behavior
    --allow-uncaught           Allow uncaught errors to propagate        [boolean]
    --async-only, -A           Require all tests to use a callback (async) or
                               return a Promise                          [boolean]
  ```

* bash-completion shortcuts for commands and options.
````

File: docs/api.md [hunk 2]
```diff
@@ -47,6 +47,14 @@ starts with two extra elements:`process.execPath` and the path to the JavaScript
 file being executed. So if you’re getting your arguments from `process.argv` in
 Node, pass `process.argv.slice(2)` to Yargs.
 
+_Note: yargs exposes the helper `hideBin`, handles the `process.argv.slice
+logic for you._
```
roryokane, line 51: There are a few formatting and grammar problems in this paragraph. Also, I personally think the text looks more readable if only the “Note” part of the paragraph is italic (as the majority of notes in `api.md` are formatted). Here is my suggested replacement paragraph that makes all those changes:

```markdown
_**Note:**_ Yargs exposes the helper `hideBin`, which handles the `process.argv.slice`
logic for you.
```

> _**Note:**_ Yargs exposes the helper `hideBin`, which handles the `process.argv.slice` logic for you.

For comparison, here’s how it looks with all of my changes except making the paragraph non-italic:

> _Note: Yargs exposes the helper `hideBin`, which handles the `process.argv.slice` logic for you._

docs/api.md
```diff
@@ -18,25 +18,25 @@ You can run Yargs without any configuration, and it will do its
 best to parse `process.argv`:
 
 ```javascript
-require('yargs').argv
+require('yargs/yargs')(process.argv.slice(2)).argv
```
roryokane, line 21: The introduction to this first example as well as the following example should be changed. Currently, the introductions to these code examples say they show calling `yargs` without passing arguments and then with passing arguments.  However, with `yargs/yargs`, arguments must always be passed, so there’s no such distinction.

First drafts of more appropriate introductions for examples 1 and 2:

> You can run Yargs directly on the relevant part of `process.argv`:

> Yargs accepts any array of options. It doesn’t have to be from `process.argv`:

The change in this PR is nonetheless better than leaving the docs as recommending `require('yargs')`, so I would be fine with merging this PR before this problem is fixed and then fixing it later.

docs/api.md
```diff
@@ -18,25 +18,25 @@ You can run Yargs without any configuration, and it will do its
...
 ```javascript
-require('yargs').parse([ '-x', '1', '-y', '2' ])
+require('yargs/yargs')().parse([ '-x', '1', '-y', '2' ])
 ```
 
 Calling `.parse()` with no arguments is equivalent to calling `yargs.argv`:
 
 ```javascript
-require('yargs').parse()
+require('yargs')(process.argv.slice(2)).parse()
```
roryokane, line 39:
```suggestion
require('yargs/yargs')(process.argv.slice(2)).parse()
```

docs/api.md
```diff
@@ -18,25 +18,25 @@ You can run Yargs without any configuration, and it will do its
...
 ```
 
 or use `.parse()` to do the same thing:
 
 ```javascript
-require('yargs').parse([ '-x', '1', '-y', '2' ])
+require('yargs/yargs')().parse([ '-x', '1', '-y', '2' ])
 ```
 
 Calling `.parse()` with no arguments is equivalent to calling `yargs.argv`:
```
roryokane, line 36:
```suggestion
Calling `.parse()` with no arguments is equivalent to calling `.argv`:
```

The current sentence implies `yargs.parse()` is equivalent to calling `yargs.argv`, but neither of those are a thing when `yargs` is `require('yargs/yargs')`.

docs/advanced.md
```diff
@@ -538,8 +538,10 @@ var argv = require('yargs')
 ### Using the non-singleton interface
 
 To use yargs without running as a singleton, do:
+
 ```js
 const argv = require('yargs/yargs')(process.argv.slice(2))
 ```
 
-This is especially useful when using yargs in a library, as library authors should not pollute the global state.
+We recommend using this approach by default, as the singleton interface is
```
roryokane, line 546: Eventually, all of the docs will use the non-singleton interface `require('yargs/yargs')`. Thus, it might be nice to clarify what the singleton interface is here, to make it clearer which part of the above code block is the singleton/non-singleton interface.

```suggestion
We recommend using this approach by default, as the singleton interface (`require('yargs')`) is
```

roryokane (commented):
Here is my feedback on this PR’s documentation changes. I don’t know Yargs internals well enough to comment on the code changes, beyond saying I don’t see anything obviously wrong in the diff.

Issue #1771, opened 2020-09-30
Incompatible with Deno 1.4.[0,1,2] due to deno adopting tsconfig settings importsNotUsedAsValues and  isolatedModules 


luke-john:
In version 1.4 Deno adopted;

- ~the tsconfig setting [importsNotUsedAsValues](https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues) - https://github.com/denoland/deno/pull/7413~
  This seems to have been reverted in deno 1.4.5, and it appears they aren't planning on making it a default anymore https://github.com/denoland/deno/pull/7800
- the tsconfig setting [isolatedModules](https://www.typescriptlang.org/tsconfig#isolatedModules) when using the deno `--unstable` flag - https://github.com/denoland/deno/issues/7326 ([intended to be enabled by default in 1.5](https://github.com/denoland/deno/issues/7326))

These require the use of;
- type only imports for values which are only used as types (`importsNotUsedAsValues`)
- type only exports for types which are re-exported (`isolatedModules`)

From the errors I'm seeing with our tool which makes use of yargs from deno, this will require updates to the types for yargs, y18n and yargs_parser along the following lines.

```diff
# https://deno.land/x/y18n@v5.0.0-deno/deno.ts:2:1
- import { Y18NOpts } from './build/lib/index.d.ts'
+ import type { Y18NOpts } from './build/lib/index.d.ts'
```

```diff
# https://deno.land/x/yargs@v16.0.3-deno/types.ts#L6
// expose types for the benefit of Deno.
- import { YargsInstance as YargsType, Arguments } from './build/lib/yargs-factory.d.ts'
+ import type { YargsInstance as YargsType, Arguments } from './build/lib/yargs-factory.d.ts'

- export {
+export type {
  Arguments,
  YargsType
}
```

While there is a possible workaround of overriding these in a custom tsconfig, from 1.5 deno have indicated they will ignore any custom value for importsNotUsedAsValues from 1.5 onwards https://github.com/denoland/deno/pull/7413#issuecomment-691127947.

I'll open Pull Requests for these shortly, just getting the repos setup locally.

bcoe:
@luke-john thank you for the patches 👌 

Do you think it would be worth updating our testing matrix too, to move to Deno 1.4, so that we'd catch this?

bcoe:
@luke-john unrelated, there are a few shortcomings in Deno I'd love to address over time:

[...]

PR #1772, opened 2020-09-30
fix(deno): update types for deno ^1.4.0

luke-john:
In version 1.4 Deno adopted;

- the tsconfig setting [importsNotUsedAsValues](https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues) - https://github.com/denoland/deno/pull/7413
- the tsconfig setting [isolatedModules](https://www.typescriptlang.org/tsconfig#isolatedModules) when using the deno `--unstable` flag - https://github.com/denoland/deno/issues/7326 ([intended to be enabled by default in 1.5](https://github.com/denoland/deno/issues/7326))

These require the use of;
- type only imports for values which are only used as types (`importsNotUsedAsValues`)
- type only exports for types which are re-exported (`isolatedModules`)

I've added more details at https://github.com/yargs/yargs/issues/1771.

related prs:
- yargs-parser https://github.com/yargs/yargs-parser/pull/329
- y18n https://github.com/yargs/y18n/pull/100

### DIFF
--- a/deno.ts
+++ b/deno.ts
@@ -1,7 +1,7 @@
 // Bootstrap yargs for Deno platform:
 import denoPlatformShim from './lib/platform-shims/deno.ts'
 import { YargsWithShim } from './build/lib/yargs-factory.js'
-import { YargsInstance as YargsType } from './build/lib/yargs-factory.d.ts'
+import type { YargsInstance as YargsType } from './build/lib/yargs-factory.d.ts'
 
 const WrappedYargs = YargsWithShim(denoPlatformShim)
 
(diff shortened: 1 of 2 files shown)


luke-john:
While preparing this pr and the related ones in yargs-parse and y18n, vscode showed most files as full of errors due to the use of file extensions in imports which don't match the actual source files extension

ie

```ts
// https://github.com/yargs/yargs/blob/master/lib/typings/yargs-parser-types.ts#L3
import type { Dictionary, ValueOf } from './common-types.js'
```

In this example common-types is a typescript file with the extension `.ts` `common-types.ts`

While there's some build tooling and to support the builds to the build folder, I was still seeing issues in the files under `/build`.

Are these intentional (possibly for the cjs builds)?

bcoe:
@luke-john I will need to dig into these CI/CD issues, I just fixed a bug related to how we parse commands, not sure how I failed to update this failing test.

luke-john:
No problems, I've just updated this to be against master instead of Deno.

bcoe (approved):

bcoe:
@luke-john hey Luke, just wanted to let you know that the latest version of yargs should have this problem fixed for you.

Let me know if you continue to bump into any issues?

PR #1788, opened 2020-10-22
fix(validation): edge-case conflicting object bug

alisonailea:
just adding a basic <object>.hasOwnProperty because it's breaking my repo and I can't figure out how else to fix it.


alisonailea:
found an alternative fix

PR #1792, opened 2020-11-02
Expose CommandHandlerDefinition in Deno

ffMathy:
Today, it's hard to have command handlers in separate files that are strong typed, if this is not exposed.

### DIFF
--- a/types.ts
+++ b/types.ts
@@ -3,5 +3,7 @@ import type {
   YargsInstance as YargsType,
   Arguments,
 } from './build/lib/yargs-factory.d.ts';
+  
+import type { CommandHandlerDefinition } from './build/lib/command.js';
 
-export type {Arguments, YargsType};
+export type {Arguments, YargsType, CommandHandlerDefinition};


ffMathy:
Any idea of why this is failing? I'd love some help.

bcoe:
@ffMathy I think there will need to be a fix in `cliui` before Deno starts working again, it also looks like there's just a linting issue happening, which `npm run fix` should address.

bcoe:
@ffMathy I moved Deno's types in to `deno-types.ts`, and had to remove some temporarily to get compilation working for `deno@1.5`; Would love some help getting the types working again for Deno, if we can either:

1. figure out how to use the types in `build/`.
2. figure out how to get Deno and TypeScript working together better, so we can use the types in `src` (_there are some open threads on Deno and TypeScript about how this functionality could be made better), or.
3. worst case scenario, we could write some of our own types in `deno-types`, I'd just like to make sure we don't reproduce too much work we've already done in TypeScript.

bcoe:
Hey @ffMathy, going to close this for now due to inactivity. Just let me know if you want to pick the work back up, I'd love to get types working for Deno again.

PR #1795, opened 2020-11-02
chore: upgrade y18n to 5.0.5 to address prototype pollution issue

DarthHater:
Hi there!

This PR is short and sweet, just updates the package.json to bring in the updates to y18n that address prototype pollution (which I believe is addressed in 5.0.5)

Related: 
- https://github.com/yargs/y18n/issues/96
- https://github.com/yargs/y18n/pull/108

Let me know if you need anything from me!

Cheers!


bcoe (approved):

bcoe:
@DarthHater I don't usually land patch updates like this, as they'll be covered by the SemVer `^` version range. Thanks for taking the time to submit the PR, however.

DarthHater:
@bcoe thanks! Never hurts to bump it even if the semver handles it (in case some other dependency is locked to an older version, npm would in theory select a version to satisfy (project a has exact match to `5.0.2`, your semver is satisfied)).

PR #1799, opened 2020-11-09
fix(deno): get yargs working on deno@1.5.x

bcoe:
This gets yargs working again on deno@1.5.x, by dropping the complex `YargsInstance`.

Unfortunately `deno@1.5.x` seems to break the already fragile path handling logic that was being used to target both Node.js and Deno simultaneously. I'm not sure what the best path forward is, we might want to simply maintain our own type definitions for deno?

Refs https://github.com/yargs/yargs/pull/1792
Fixes https://github.com/yargs/yargs/issues/1794

CC: @ffMathy @KhushrajRathod

Issue #1798, opened 2020-11-05
can't import yargs/helpers in esm-shimmed environment


ixis-doug:
`import { hideBin } from 'yargs/helpers'` fails when invoking like `node -r esm index.js`

The esm module is a production-quality ESM loader and works with old versions of node (and current versions) without any funny configuration required (like setting `"type": "module"` in package.json).

The following error is printed:
```
/home/foobar/my-project/node_modules/yargs/build/lib/utils/apply-extends.js:1
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /home/foobar/my-project/node_modules/yargs/build/lib/utils/apply-extends.js
require() of ES modules is not supported.
require() of /home/foobar/my-project/node_modules/yargs/build/lib/utils/apply-extends.js from /home/foobar/my-project/node_modules/yargs/helpers.mjs is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.  Instead rename apply-extends.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /home/foobar/my-project/node_modules/yargs/package.json
```

tested yargs 16.1.0 on node 12.16 and 12.19 with esm 3.2.25

ljharb:
Duplicate of #1789.

bcoe:
@ixis-doug let me know if `yargs@16.1.1` does the trick for you.

PR #1801, opened 2020-11-11
fix: expose helpers for legacy versions of Node.js

### LINKED ISSUE #1789: Readme CJS example fails with versions of Node older than 12.17
Using the CJS example code from the README.md seems to fail for me on all Node versions I tried before 12.17 (12.16.1, 12.15.0, 12.14.1 and 11.15.0):

```js
#!/usr/bin/env node
const { hideBin } = require('yargs/helpers')
const argv = yargs(hideBin(process.argv)).argv

if (argv.ships > 3 && argv.distance < 53.5) {
  console.log('Plunder more riffiwobbles!')
} else {
  console.log('Retreat from the xupptumblers!')
}
```

The error varies by version.

12.16.1:
```
Error: Cannot find module 'yargs/helpers'
```

12.14.1 and 12.15.0:
```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: <snip>/yargs-node-error/node_modules/yargs/helpers.mjs
```

11.15.0:
```
Error: Cannot find module 'yargs/helpers'
```

Works fine in 12.17 and 12.18.

Here's a sample of how I set it up: https://github.com/gpoole/yargs-node-error

bcoe:
Expose helpers in extensionless `helpers` file, which older Node.js versions will interpret as a JavaScript file.

This is the only approach I have found to circumvent exceptions thrown when loading `.js` files when type is set to `module`.

Fixes #1789


bcoe:
CC: @ljharb, @gpoole

helpers
```diff
@@ -0,0 +1,15 @@
...
+  processArgv,
+} = require('./build/index.cjs');
+
+module.exports = {
+  applyExtends: (config, cwd, mergeExtends) => {
+    return applyExtends(config, cwd, mergeExtends, cjsPlatformShim);
+  },
+  hideBin: processArgv.hideBin,
+  Parser,
+};
```
ljharb, line 15: i'm confused, why an extensionless file instead of a `helpers` dir with a `package.json` that has `"type": "commonjs"` and an `index.js` file?

The current approach will be way more likely cause problems with parsers and resolvers (admittedly, ones with bugs, but i can't even confidently say `resolve` would work properly with this)

bcoe: I appreciate that having a subfolder would cleanup the top level a bit too, we could pull the mjs in there

ljharb:
Does this fix #1799, or does it fix #1798?

bcoe:
@ljharb Deno was type issues, which should be fixed on the main branch already... I might have to move away from TypeScript for Deno unless someone wants to submit and maintain types that work.

PR #1810, opened 2020-11-20
test: add test for dot-path option in strict mode

SkReD:


bcoe:
👋 yargs does not support configuring properties on objects, the idea being that being an object is a property of the argument  `obj`.

We've had this feature request a few times, but I've never dug into how much work it would be to make the parser support this.

Feel free to open a feature request [here](https://github.com/yargs/yargs-parser/issues), if you'd like to make a case for this functionality.

PR #1812, opened 2020-11-22
feat: command() now accepts an array of modules

bcoe:
.command() now accepts an array of modules, the intention being that this
makes it easier for people to build a complex hierarchical CLI in
environments that can not use requireDirectory().

### DIFF
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -26,6 +26,7 @@ import {
 import whichModule from './utils/which-module.js';
 
 const DEFAULT_MARKER = /(^\*)|(^\$0)/;
+export type HandlerArrayItem = string | CommandHandlerDefinition;
 
 // handles parsing positional arguments,
 // and populating argv with said positional
@@ -43,7 +44,7 @@ export function command(
   let defaultCommand: CommandHandler | undefined;
 
   self.addHandler = function addHandler(
-    cmd: string | string[] | CommandHandlerDefinition,
+    cmd: string | CommandHandlerDefinition | HandlerArrayItem[],
     description?: string | false,
     builder?: CommandBuilder | CommandBuilderDefinition,
     handler?: CommandHandlerCallback,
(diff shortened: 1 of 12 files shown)


lib/command.ts
```diff
@@ -54,10 +55,19 @@ export function command(
...
+      for (const command of cmd) {
+        self.addHandler(command);
+      }
+    } else if (Array.isArray(cmd)) {
+      // Assume the array is a cmd, followed by aliases, if the array contains
+      // any strings [cmd, 'alias1', 'alias2']:
+      aliases = cmd.slice(1) as string[];
       cmd = cmd[0];
     } else if (isCommandHandlerDefinition(cmd)) {
+      // An object was provided rather than individual parameters:
```
ljharb, line 70: i suspect the types may narrow better with this:
```suggestion
    if (Array.isArray(cmd)) {
      if (cmd.every(c => typeof c === 'string')) {
        ([cmd, aliases] = cmd);
      } else {
        for (const command of cmd) {
          self.addHandler(command);
        }
      }
    } else if (isCommandHandlerDefinition(cmd)) {
```

lib/command.ts
```diff
@@ -75,10 +85,10 @@ export function command(
       return;
     }
 
-    // allow a module to be provided instead of separate builder and handler
+    // Allow a module to be provided as builder, rather than function:
     if (isCommandBuilderDefinition(builder)) {
       self.addHandler(
-        [cmd].concat(aliases),
+        [cmd].concat(aliases) as HandlerArrayItem[],
```
ljharb, line 91: I think this may avoid the `as`:
```suggestion
      const placeholder: HandlerArrayItem[] = [cmd].concat(aliases);
      self.addHandler(
        placeholder,
```

ljharb (commented):
Overall seems fine, but I'm surprised this is the only path to supporting commands in ESM.

lib/command.ts
bcoe: I managed to tighten up the types quite a bit 👍 

This turned out to be a good audit of the command API.

Issue #1817, opened 2020-12-01
Package subpath './package.json' is not defined by "exports" in /.../node_modules/yargs/package.json


toverux:
For an unknown reason, my build fails at some point with the following error:

> Package subpath './package.json' is not defined by "exports" in /.../node_modules/yargs/package.json

There is no more context on why the package.json from yargs is required.

However, adding this key in the `exports` of yargs' package.json works :

```js
"exports": {
  "./package.json": "./package.json"
}
```

Since yargs decided to add explicit exports, it would be a nice to let external tools read yargs' package description.

bcoe:
@toverux could you provide an example of where this is failing?

mingtsay:
I also have this issue when I run `react-native link` in my react native project.

toverux:
In my case, this is with NativeScript. Maybe it's scanning node_modules for plugins? (not familiar enough with it yet).
I will try to provide more context tomorrow if I can, but the issue seems clear though, any tool `require`ing the package.json of yargs would fail, as `exports` acts like a whitelist, only files listed in it are now reachable through the Node resolution algorithm.

> Like in CommonJS, module files within packages can be accessed by appending a path to the package name unless the package’s package.json contains an "exports" field, in which case files within packages can only be accessed via the paths defined in "exports".

https://nodejs.org/api/esm.html

See also : https://github.com/react-native-community/cli/issues/1168#issuecomment-628034263

bcoe:
@toverux if you've confirmed adding the additional export works, mind submitting a patch?

toverux:
Will do!

bcoe:
give `yargs@16.2.0` a shot.

PR #1818, opened 2020-12-03
chore: add package.json to module exports

toverux:
The reason for this change is explained in #1817.

TL;DR: yargs uses the `package.json`::`exports` field to define importable paths from the outside. Any files that are not explicitly listed in it cannot be required anymore. However, some tools may attempt to `require(node_modules/yargs/package.json)` to read package info and fail. This PR exposes the package.json file to the public.


bcoe (approved):

toverux:
Thanks!

Issue #1791, opened 2020-10-30
Default command invoked when calling .showHelp


wesleytodd:
```javascript
const yargs = require('yargs')

yargs.command('$0', 'a default command', (yargs) => yargs, () => {
  console.log('default command!! why?')
})
yargs.showHelp((output) => {
  console.log(output)
})
```

The above application will also execute the default command even though we never called `.argv` or `.parse`.  This seems like unexpected behavior to me and I could not find any docs on this.  Is this intended for some reason I am unaware of?

bcoe:
@wesleytodd this definitely doesn't seem like intended behavior, thanks for the bug report.

bcoe:
@wesleytodd I believe I fixed this as part of my work in #1823

There's also now a helper `await yargs.getHelp()`, which you could use rather than the callback based approach. I'd love feedback, you can install via:

`npm i yargs@next`

PR #1823, opened 2020-12-06
feat!: improve support for async/await

bcoe:
In `node@15` [top level await](https://v8.dev/features/top-level-await) is now supported 🎉. 

This means that we can have yargs return a promise in more situations, with users writing code that looks like this:

```js
import yargs from './index.mjs';
import { setTimeout } from 'timers/promises';

const argv = await yargs(process.argv.slice(2))
  .command(
    'foo',
    'foo commnd',
    () => {},
    async (argv) => {
      console.info('second')
      await setTimeout(1000)
    }
  )
  .middleware(async (argv) => {
    console.info('first')
    await setTimeout(1000)
    argv.value = 99
  }).argv;
console.info('third', argv)
```

This PR begins work to add better first class support of promises to yargs.

---

TODO:

- [x] ~write unit tests for scenarios outlined in #1420~ broke into several steps.
- [x] add tests for async behavior of `.parse()` (Refs: #1420). 
- [x] add tests for async behavior of top level `.argv` (Refs: #1420).
- [x] add tests for, and document `async getHelp() (Refs: #1420).
- [x] add tests for async getCompletion()/completion (Refs: #1420, #1235).
- [x] write unit tests for any TODO comments added during this work.
- [x] ~should command builders accept a promise?~ see #1042 -- we will land this work separately before the next major.
- [x] ~add async support to coerce/check~, see #859 -- we will land this work separately before the next major.
- [x] ~make global middleware work without defining command (_this work should be in separate breaking PR_).~ broken into ticket https://github.com/yargs/yargs/issues/1849
- [x] add documentation for new async behavior.

BREAKING CHANGES:

* now returns a promise if handler is `async`.
* onFinishCommand removed, in favor of being able to await promise.
* getCompletion now invokes callback with `err` and `completions, returns promise of completions.

Fixes #1791, #1420, #1235

PR #1826, opened 2020-12-08
.showHelp() and .getHelp() now return same output for commands as --help

### LINKED ISSUE #883: Show help for no command and handle unknown command
I trouble to implement folliwng case:

1) Show message `unknown command` if some unknnown command supplied
2) Show help if no command supplied.

To handle unknown command I use default command `*`, it works ok

```ts
  .command({
    command: '*',
    handler: (argv) => {      
      if (argv._[0]) {
        console.log('Unknown commmand', argv._[0])        
      }
    }
  })
 .demand(1)
```

But `demand(1)` here doesn't work in this case so If I supply no command nothing happens at all (it goes to `*` handler)

I tried to use `yargs.showHelp()`  somewhere in '*' handler or just checking main argv

```ts
const argv = yargs...

if (!argv._[0]) {
  yargs.showHelp()
}
```

but it doesn't show help. All it shows:

```
Options:
```

### LINKED ISSUE #1820: Help- and version option description not shown
**The code:**
```
const yargs = require('yargs').locale('en')

let output

const options = {
    exec: { alias: 'e', desc: 'Dis supposed to be exec', type: 'string', requiresArg: true },
    h: { alias: 'help', desc: 'Dis supposed to be help' },
    v: { alias: 'version', desc: 'Dis supposed to be version' }
}

yargs.options(options)

yargs.parse(['--help'], (_err, _argv, _output) => { output = _output })

console.log(output)
```

**Expected output:**
```
Options:
  -e, --exec      Dis supposed to be exec                                [string]
  -h, --help      Dis supposed to be help                                [boolean]
  -v, --version  Dis supposed to be version                              [boolean]
```

**Actual output:**
```
Options:
  -e, --exec     Dis supposed to be exec                                [string]
  -v, --version  Show version number                                   [boolean]
```

**Note**
No need for a workaround since I already got one, but I just wanted to mention this so it would get fixed.

OsmanAltun:
**Closes issues:**
 - closes #883

### DIFF
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -238,60 +238,37 @@ export function command(
       yargs.getUsageInstance().freeze();
     }
     const builder = commandHandler.builder;
-    if (isCommandBuilderCallback(builder)) {
-      // a function can be provided, which builds
-      // up a yargs chain and possibly returns it.
-      const builderOutput = builder(yargs.reset(parsed.aliases));
-      const innerYargs = isYargsInstance(builderOutput) ? builderOutput : yargs;
-      if (!command) innerYargs.getUsageInstance().unfreeze();
-      if (shouldUpdateUsage(innerYargs)) {
-        innerYargs
-          .getUsageInstance()
-          .usage(
-            usageFromParentCommandsCommandHandler(
-              parentCommands,
-              commandHandler
-            ),
-            commandHandler.description
-          );
-      }
-      innerArgv = innerYargs._parseArgs(
-        null,
-        undefined,
-        true,
-        commandIndex,
-        helpOnly
-      );
-      aliases = (innerYargs.parsed as DetailedArguments).aliases;
-    } else if (isCommandBuilderOptionDefinitions(builder)) {
-      // as a short hand, an object can instead be provided, specifying
-      // the options that a command takes.
-      const innerYargs = yargs.reset(parsed.aliases);
 
-      if (!command) innerYargs.getUsageInstance().unfreeze();
-      if (shouldUpdateUsage(innerYargs)) {
-        innerYargs
-          .getUsageInstance()
-          .usage(
-            usageFromParentCommandsCommandHandler(
-              parentCommands,
-              commandHandler
-            ),
-            commandHandler.description
-          );
-      }
+    let innerYargs: YargsInstance;
+    if (isCommandBuilderCallback(builder)) {
+      const builderOutput = builder(yargs.reset(parsed.aliases));
+      innerYargs = isYargsInstance(builderOutput) ? builderOutput : yargs;
+    } else {
+      innerYargs = yargs.reset(parsed.aliases);
+    }
+
+    if (!command) innerYargs.getUsageInstance().unfreeze();
+    if (shouldUpdateUsage(innerYargs)) {
+      innerYargs
+        .getUsageInstance()
+        .usage(
+          usageFromParentCommandsCommandHandler(parentCommands, commandHandler),
+          commandHandler.description
+        );
+    }
+    if (isCommandBuilderOptionDefinitions(builder)) {
       Object.keys(commandHandler.builder).forEach(key => {
         innerYargs.option(key, builder[key]);
       });
-      innerArgv = innerYargs._parseArgs(
-        null,
-        undefined,
-        true,
-        commandIndex,
-        helpOnly
-      );
-      aliases = (innerYargs.parsed as DetailedArguments).aliases;
     }
+    innerArgv = innerYargs._parseArgs(
+      null,
+      undefined,
+      true,
+      commandIndex,
+      helpOnly
+    );
+    aliases = (innerYargs.parsed as DetailedArguments).aliases;
 
     if (!yargs._hasOutput()) {
       positionalMap = populatePositionals(
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -219,6 +219,9 @@ export function command(
     let numFiles = currentContext.files.length;
     const parentCommands = currentContext.commands.slice();
 
+    const tmpCommands = yargs.getUsageInstance().getCommands();
+    const tmpUsage = yargs.getUsageInstance().getUsage();
+
     // what does yargs look like after the builder is run?
     let innerArgv: Arguments | Promise<Arguments> = parsed.argv;
     let positionalMap: Dictionary<string[]> = {};
@@ -233,15 +236,20 @@ export function command(
       const builderOutput = builder(yargs.reset(parsed.aliases));
       const innerYargs = isYargsInstance(builderOutput) ? builderOutput : yargs;
       if (shouldUpdateUsage(innerYargs)) {
-        innerYargs
-          .getUsageInstance()
-          .usage(
-            usageFromParentCommandsCommandHandler(
-              parentCommands,
-              commandHandler
-            ),
-            commandHandler.description
-          );
+        if (!command) {
+          tmpCommands.forEach(e => innerYargs.getUsageInstance().command(...e));
+          tmpUsage.forEach(e => innerYargs.getUsageInstance().usage(...e));
+        } else {
+          innerYargs
+            .getUsageInstance()
+            .usage(
+              usageFromParentCommandsCommandHandler(
+                parentCommands,
+                commandHandler
+              ),
+              commandHandler.description
+            );
+        }
       }
       innerArgv = innerYargs._parseArgs(null, null, true, commandIndex);
       aliases = (innerYargs.parsed as DetailedArguments).aliases;
@@ -250,15 +258,20 @@ export function command(
       // the options that a command takes.
       const innerYargs = yargs.reset(parsed.aliases);
       if (shouldUpdateUsage(innerYargs)) {
-        innerYargs
-          .getUsageInstance()
-          .usage(
-            usageFromParentCommandsCommandHandler(
-              parentCommands,
-              commandHandler
-            ),
-            commandHandler.description
-          );
+        if (!command) {
+          tmpCommands.forEach(e => innerYargs.getUsageInstance().command(...e));
+          tmpUsage.forEach(e => innerYargs.getUsageInstance().usage(...e));
+        } else {
+          innerYargs
+            .getUsageInstance()
+            .usage(
+              usageFromParentCommandsCommandHandler(
+                parentCommands,
+                commandHandler
+              ),
+              commandHandler.description
+            );
+        }
       }
       Object.keys(commandHandler.builder).forEach(key => {
         innerYargs.option(key, builder[key]);
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -231,42 +231,67 @@ export function command(
     if (command) {
       currentContext.commands.push(command);
       currentContext.fullCommands.push(commandHandler.original);
-    } else {
+    }
+    const builder = commandHandler.builder;
+    if (isCommandBuilderCallback(builder)) {
+      // a function can be provided, which builds
+      // up a yargs chain and possibly returns it.
+      const builderOutput = builder(yargs.reset(parsed.aliases));
+      const innerYargs = isYargsInstance(builderOutput) ? builderOutput : yargs;
       // A null command indicates we are running the default command,
       // if this is the case, we should show the root usage instructions
       // rather than the usage instructions for the nested default command:
-      yargs.getUsageInstance().freeze();
-    }
-    const builder = commandHandler.builder;
-
-    let innerYargs: YargsInstance;
-    if (isCommandBuilderCallback(builder)) {
-      const builderOutput = builder(yargs.reset(parsed.aliases));
-      innerYargs = isYargsInstance(builderOutput) ? builderOutput : yargs;
-    } else {
-      innerYargs = yargs.reset(parsed.aliases);
+      if (!command) innerYargs.getUsageInstance().unfreeze();
+      if (shouldUpdateUsage(innerYargs)) {
+        innerYargs
+          .getUsageInstance()
+          .usage(
+            usageFromParentCommandsCommandHandler(
+              parentCommands,
+              commandHandler
+            ),
+            commandHandler.description
+          );
+      }
+      innerArgv = innerYargs._parseArgs(
+        null,
+        undefined,
+        true,
+        commandIndex,
+        helpOnly
+      );
+      aliases = (innerYargs.parsed as DetailedArguments).aliases;
+    } else if (isCommandBuilderOptionDefinitions(builder)) {
+      // as a short hand, an object can instead be provided, specifying
+      // the options that a command takes.
+      const innerYargs = yargs.reset(parsed.aliases);
+      // A null command indicates we are running the default command,
+      // if this is the case, we should show the root usage instructions
+      // rather than the usage instructions for the nested default command:
+      if (!command) innerYargs.getUsageInstance().unfreeze();
+      if (shouldUpdateUsage(innerYargs)) {
+        innerYargs
+          .getUsageInstance()
+          .usage(
+            usageFromParentCommandsCommandHandler(
+              parentCommands,
+              commandHandler
+            ),
+            commandHandler.description
+          );
+      }
       Object.keys(commandHandler.builder).forEach(key => {
         innerYargs.option(key, builder[key]);
       });
+      innerArgv = innerYargs._parseArgs(
+        null,
+        undefined,
+        true,
+        commandIndex,
+        helpOnly
+      );
+      aliases = (innerYargs.parsed as DetailedArguments).aliases;
     }
-
-    if (!command) innerYargs.getUsageInstance().unfreeze();
-    if (shouldUpdateUsage(innerYargs)) {
-      innerYargs
-        .getUsageInstance()
-        .usage(
-          usageFromParentCommandsCommandHandler(parentCommands, commandHandler),
-          commandHandler.description
-        );
-    }
-    innerArgv = innerYargs._parseArgs(
-      null,
-      undefined,
-      true,
-      commandIndex,
-      helpOnly
-    );
-    aliases = (innerYargs.parsed as DetailedArguments).aliases;
 
     if (!yargs._hasOutput()) {
       positionalMap = populatePositionals(
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -266,9 +266,16 @@ export function command(
       );
       aliases = (innerYargs.parsed as DetailedArguments).aliases;
     } else if (isCommandBuilderOptionDefinitions(builder)) {
+      // A null command indicates we are running the default command,
+      // if this is the case, we should show the root usage instructions
+      // rather than the usage instructions for the nested default command:
+      if (!command) yargs.getUsageInstance().freeze();
+
       // as a short hand, an object can instead be provided, specifying
       // the options that a command takes.
       const innerYargs = yargs.reset(parsed.aliases);
+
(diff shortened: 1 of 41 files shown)


package.json
```diff
@@ -71,14 +71,14 @@
     "rimraf": "^3.0.2",
     "rollup": "^2.23.0",
     "rollup-plugin-cleanup": "^3.1.1",
-    "standardx": "^5.0.0",
+    "standardx": "^7.0.0",
```
bcoe, line 74: mind making the `standardx` upgrade in a separate PR?

OsmanAltun: sure

bcoe:
@OsmanAltun 👏 thank you, I've been traveling recently, but will give you review soon.

lib/yargs-factory.ts
```diff
@@ -655,7 +655,7 @@ function Yargs(
     return self;
   };
 
-  self.command = function (
+  self.command = self.commands = function (
```
bcoe, line 658: This seems unrelated to the other work? Could this be a separate PR?

bcoe (changes requested):

lib/command.ts
```diff
@@ -233,15 +236,20 @@ export function command(
...
-            usageFromParentCommandsCommandHandler(
-              parentCommands,
-              commandHandler
-            ),
-            commandHandler.description
-          );
+        if (!command) {
+          tmpCommands.forEach(e => innerYargs.getUsageInstance().command(...e));
+          tmpUsage.forEach(e => innerYargs.getUsageInstance().usage(...e));
+        } else {
```
bcoe, line 242: This fix seems reasonable to me, we will need to rebase with #1823; which is close to landing and touches this same code.

lib/yargs-factory.ts
OsmanAltun: Done

lib/command.ts
OsmanAltun: Merged with master.

OsmanAltun:
To fix issue #1791, would setting `shortCircuit:true` not have fixed it? That's what I had done previously, before merge.

bcoe:
> would setting shortCircuit:true not have fixed it? 

@OsmanAltun this would have perhaps been an alternate approach? But I believe the problem was addressed as I started refactoring how help messages were cached, which was necessary to implement `.getHelp()`.

OsmanAltun:
> which was necessary to implement `.getHelp()`.

Ohkay, I get it now.

bcoe:
@OsmanAltun I was looking at this implementation, and I'm a little concerned by the additional complexity that registering commands with `innerArgv` introduces.

I was wondering if instead we could update the helper `usageFromParentCommandsCommandHandler` so that it populates usage information properly for the default command?

```
function usageFromParentCommandsCommandHandler(
    parentCommands: string[],
    commandHandler: CommandHandler
  ) {
    const c = DEFAULT_MARKER.test(commandHandler.original)
      ? commandHandler.original.replace(DEFAULT_MARKER, '').trim()
      : commandHandler.original;
    const pc = parentCommands.filter(c => {
      return !DEFAULT_MARKER.test(c);
    });
    pc.push(c);
    return `$0 ${pc.join(' ')}`;
  }
```

☝️ couldn't we figure out how the display the proper usage behavior here, if a default command is registered?

OsmanAltun:
The helper function is (as far as I know) only responsible for the first line in the usage output, which is the command call order.

a

If u add all commands here, it would display between the description and the command call order. Besides, to add the commands back, u would still have to loop over every command using `yargsInstance.getUsageInstance().command()` and/or `.usage()`

b

___

The best solution might be to make sure the `reset` function does also preserve all commands (just like the options). I am not too familiar with the reset function though, I don't know if preserving commands would break functionality.

test/usage.cjs
```diff
@@ -2602,9 +2602,10 @@ describe('usage tests', () => {
       r.logs[0]
         .split('\n')
         .should.deep.equal([
-          'usage',
+          'usage [command]',
```
bcoe, line 2605: @OsmanAltun reading both the linked bugs, I'm still a little confused about what bug this is actually addressing.

It would help me see this work over the finish line if you could:

1. add a test that explains the bug you're attempting to address.
2. update the title of the PR, to explain the specific bug fix.

I think there's a systemic bug with how default command is handled, which might be the root cause of both the bugs you've linked, but I'm not sure that the correct solution.

Having the tests, and a clear description of the bug fix, would help me better understand.

bcoe (changes requested):

test/usage.cjs
```diff
@@ -2602,9 +2602,10 @@ describe('usage tests', () => {
...
           '',
-          'do something',
+          'Commands:',
```
bcoe, line 2607: It's by design that we don't show the list of commands unless there is at least one command registered that's not the default command, this is because all the information the user needs can be expressed like this:

```
example.mjs

the default command

Options:
```

I don't think this is the bug you were trying to fix though?

bcoe: I'm guessing your fix addresses another but you were having? but I think we need a test for it, and this test shouldn't change.

bcoe:
@OsmanAltun I want to encourage external contributions to yargs, and I feel bad for keeping this issue open for so long.

To get patches landed quickly, it's best if:

1. each PR addresses one specific issue (this will help me understand the fix).
2. a patch without a test doesn't explain to me what's being fixed (I've found this PR confusing).
3. the PR title/body should also clearly describe a specific bug being addressed (otherwise I don't have a clue what's going on).

If you provide a test, and a clear description of what you're fixing, I will help guide you towards the part of the codebase where I think we can best fix things.

Very much appreciate you taking your time to make a contribution.

bcoe:
@OsmanAltun I did some digging this afternoon, and think I see a simpler patch:

```diff
diff --git a/lib/yargs-factory.ts b/lib/yargs-factory.ts
index 57d1295..69f6308 100644
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -1550,10 +1550,11 @@ function Yargs(
 
       const handlerKeys = command.getCommands();
       const requestCompletions = completion!.completionKey in argv;
-      const skipRecommendation = argv[helpOpt!] || requestCompletions;
+      const skipRecommendation =
+        argv[helpOpt!] || requestCompletions || helpOnly;
       const skipDefaultCommand =
         skipRecommendation &&
-        (handlerKeys.length > 1 || handlerKeys[0] !== '$0');
+        (handlerKeys.length > 1 || !command.hasDefaultCommand());
 
       if (argv._.length) {
         if (handlerKeys.length) {
```

The underlying issue seems to be that we were calling the default command when we shouldn't have.

The default command **should not** be called when:

1. it has been requested that help be output, and there is no default command.
2. it has been requested that help be output, there is a default command, but there are other commands.

The default command **should** be called when:

* it has been requested that help be output, and the only command is the default command.

---

Do you want to run with this patch I've shared? The work that needs to be done is adding the following test cases:

```
it('has same output for --help and .showHelp() when default command is only command')
it('has same output for --help and .showHelp() when there are multiple commands, including default command')
it('has same output for --help and .showHelp() when called from within a handler')
```

OsmanAltun:
With your current suggestion the `--help` option works (Hooray!), but not for my use case unfortunately. Can't we just skip the reset of the usage instance when executing default functions. That would be ideal.

My use-case is as follows: 

- Multiple commands with a single default function
- The default command requires at least one argument
- When no argument is given, the help output should be shown (including ALL commands)

The workaround I currently use requires using `yargs.parse()` twice, which is very hacky. 
____
On another note, (to lift up your confusion lol)
 - #1820 fixed by 80edee12e0b493251c271417660fdd7cbc0e93db
 - #883 fixed by 380782dd1a3af7226ff5ee70a888b7cd4a0c46ec

There were originally three bugs (same amount of commits starting with 'fix'). It might have been better if every commit had a reference to the bug, oh well. Hope that clears things up 😄 .

bcoe:
@OsmanAltun 

If you make a separate PR for https://github.com/yargs/yargs/commit/80edee12e0b493251c271417660fdd7cbc0e93db, it seems reasonable -- although, I'd like to dig a little deeper and see if we can figure out why setting the description for `-h` doesn't also set the description for `--help`, since this is how other options work.

I don't like the patch https://github.com/yargs/yargs/commit/380782dd1a3af7226ff5ee70a888b7cd4a0c46ec, it might give you the output you're looking for, but it's through a weird side effect:

```diff
diff --git a/lib/yargs-factory.ts b/lib/yargs-factory.ts
index 57d1295..69f6308 100644
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -1550,10 +1550,11 @@ function Yargs(
 
       const handlerKeys = command.getCommands();
       const requestCompletions = completion!.completionKey in argv;
-      const skipRecommendation = argv[helpOpt!] || requestCompletions;
+      const skipRecommendation =
+        argv[helpOpt!] || requestCompletions || helpOnly;
       const skipDefaultCommand =
         skipRecommendation &&
-        (handlerKeys.length > 1 || handlerKeys[0] !== '$0');
+        (handlerKeys.length > 1 || !command.hasDefaultCommand());
 
       if (argv._.length) {
         if (handlerKeys.length) {
```

☝️ this solves an actual bug, which is that we should have been treating `helpOnly` the same as `argv[helpOpt!]`, which was why `--help` was behaving differently than `.showHelp()`.

If we're also not showing consistent help output when not enough commands are provided, i.e., validation has failed, then we should address this bug.

bcoe:
@OsmanAltun did some digging, here's a patch that I think would solve all of your problems:

```diff
diff --git a/lib/command.ts b/lib/command.ts
index 81a7b50..bd9ae78 100644
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -261,6 +261,10 @@ export function command(
       // as a short hand, an object can instead be provided, specifying
       // the options that a command takes.
       const innerYargs = yargs.reset(parsed.aliases);
+      // A null command indicates we are running the default command,
+      // if this is the case, we should show the root usage instructions
+      // rather than the usage instructions for
+      if (!command) innerYargs.getUsageInstance().unfreeze();
       if (shouldUpdateUsage(innerYargs)) {
         innerYargs
           .getUsageInstance()
diff --git a/lib/usage.ts b/lib/usage.ts
index d436f20..5053316 100644
--- a/lib/usage.ts
+++ b/lib/usage.ts
@@ -231,7 +231,12 @@ export function usage(yargs: YargsInstance, y18n: Y18N, shim: PlatformShim) {
 
     // your application's commands, i.e., non-option
     // arguments populated in '_'.
-    if (commands.length) {
+    //
+    // If there's only a single command, and it's the default command
+    // (represented by commands[0][2]) don't show command stanza:
+    //
+    // TODO(@bcoe): why isnt commands[0][2] an object with a named property?
+    if (commands.length > 1 || (commands.length === 1 && !commands[0][2])) {
       ui.div(__('Commands:'));
 
       const context = yargs.getContext();
@@ -703,7 +708,10 @@ export function usage(yargs: YargsInstance, y18n: Y18N, shim: PlatformShim) {
   };
   self.unfreeze = function unfreeze() {
     const frozen = frozens.pop();
-    assertNotStrictEqual(frozen, undefined, shim);
+    // In the case of running a defaultCommand, we reset
+    // early, to ensure the correct usage instruction, calling unfreeze
+    // again is a noop:
+    if (!frozen) return;
     ({
       failMessage,
       failureOutput,
diff --git a/lib/yargs-factory.ts b/lib/yargs-factory.ts
index 57d1295..b866b75 100644
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -1483,7 +1483,7 @@ function Yargs(
     commandIndex = 0,
     helpOnly = false
   ) {
-    let skipValidation = !!calledFromCommand;
+    let skipValidation = !!calledFromCommand || helpOnly;
     args = args || processArgs;
 
     options.__ = y18n.__;
@@ -1550,10 +1550,8 @@ function Yargs(
 
       const handlerKeys = command.getCommands();
       const requestCompletions = completion!.completionKey in argv;
-      const skipRecommendation = argv[helpOpt!] || requestCompletions;
-      const skipDefaultCommand =
-        skipRecommendation &&
-        (handlerKeys.length > 1 || handlerKeys[0] !== '$0');
+      const skipRecommendation =
+        argv[helpOpt!] || requestCompletions || helpOnly;
 
       if (argv._.length) {
         if (handlerKeys.length) {
@@ -1584,7 +1582,7 @@ function Yargs(
           }
 
           // run the default command, if defined
-          if (command.hasDefaultCommand() && !skipDefaultCommand) {
+          if (command.hasDefaultCommand() && !skipRecommendation) {
             const innerArgv = command.runCommand(
               null,
               self,
@@ -1617,7 +1615,7 @@ function Yargs(
           self.showCompletionScript();
           self.exit(0);
         }
-      } else if (command.hasDefaultCommand() && !skipDefaultCommand) {
+      } else if (command.hasDefaultCommand() && !skipRecommendation) {
         const innerArgv = command.runCommand(null, self, parsed, 0, helpOnly);
         return self._postProcess(
           innerArgv,
```

You've done a great job of drawing attention to nasty edge case with help output 👍

lib/command.ts
```diff
@@ -262,15 +270,20 @@ export function command(
...
-          .getUsageInstance()
-          .usage(
-            usageFromParentCommandsCommandHandler(
-              parentCommands,
-              commandHandler
-            ),
-            commandHandler.description
-          );
+        if (!command) {
+          tmpCommands.forEach(e => innerYargs.getUsageInstance().command(...e));
```
bcoe, line 274: Now that I dug into the problem, this was 95% of the way there. The advantage of calling `.unfreeze()` instead is it will give you _exactly_ the output of usage at the top level.

OsmanAltun:
Great job, everything works as expected for me now. One small question though. Currently it looks like this:
```typescript
if (isCommandBuilderCallback(builder)) {
 // logic
} else if (isCommandBuilderOptionDefinitions(builder)) {
 // freeze()
 // reset()
 // unfreeze()
 // similar logic to first block
}
```
Should freeze/unfreeze not also be applied in the first block?

bcoe:
> Should freeze/unfreeze not also be applied in the first block?

I'm betting we can pull the logic right to the top of the function, and always `unfreeze` the usage if it's a default command?

If we're running the default command:

1. we want the top level help.
2. we know we definitely **won't** be running a subcommand, as the precondition for running the default command is that no other commands matched?

OsmanAltun:
Unfreeze needs to happen after `reset()` (to not overwrite usage) and before `shouldUpdateUsage()` (to still be able to use `yargs.usage()`). However I could move `freeze()` to root of function.

lib/command.ts
```diff
@@ -231,60 +231,43 @@ export function command(
...
-      }
-      innerArgv = innerYargs._parseArgs(
-        null,
-        undefined,
-        true,
-        commandIndex,
-        helpOnly
-      );
-      aliases = (innerYargs.parsed as DetailedArguments).aliases;
-    } else if (isCommandBuilderOptionDefinitions(builder)) {
```
bcoe, line 260: I'm confused by this code being deleted, since it serves a purpose -- it may be we're just missing tests?

bcoe:
> Right now, I am skipping the isCommandBuilderOptionDefinitions() e.g.:

I'm confused as to why this code needs to be removed as part of the change.

lib/command.ts
```diff
@@ -231,60 +231,43 @@ export function command(
     if (command) {
       currentContext.commands.push(command);
       currentContext.fullCommands.push(commandHandler.original);
+    } else {
+      // A null command indicates we are running the default command,
+      // if this is the case, we should show the root usage instructions
+      // rather than the usage instructions for the nested default command:
+      yargs.getUsageInstance().freeze();
```
bcoe, line 238: I don't understand why this would have been switched from `.unfreeze()` to `.freeze()`.

`unfreeze()` makes sense to me, because it pops us up one level to before the command was run, `freeze()` doesn't make sense to me.

lib/command.ts
```diff
@@ -231,60 +231,43 @@ export function command(
...
-        null,
-        undefined,
-        true,
-        commandIndex,
-        helpOnly
-      );
-      aliases = (innerYargs.parsed as DetailedArguments).aliases;
     }
 
+    if (!command) innerYargs.getUsageInstance().unfreeze();
```
bcoe, line 253:
```suggestion
```

lib/command.ts
bcoe, line 238:
```suggestion
      yargs.getUsageInstance().unfreeze();
```

test/usage.cjs
```diff
@@ -4261,4 +4261,101 @@ describe('usage tests', () => {
...
+            {command: '*', desc: 'Default command description'},
+            {command: 'foo', desc: 'Foo command description'},
+          ])
+          .parse();
+        yargs.showHelp('log');
+      });
+      r.logs[0].split('\n').should.deep.equal(expected);
+    });
+
+    it('should contain the expected output for showHelp when called from within handler', () => {
```
bcoe, line 4344: thank you for the tests 👍

lib/command.ts
```diff
@@ -261,6 +265,10 @@ export function command(
       // as a short hand, an object can instead be provided, specifying
       // the options that a command takes.
       const innerYargs = yargs.reset(parsed.aliases);
+      // A null command indicates we are running the default command,
```
bcoe, line 268: @OsmanAltun I think this does the trick, we want to `unfreeze` if we've called `reset` (which will put the usage output in the original state), and if we're running the default command.

bcoe (approved):
@OsmanAltun I pushed a couple updates to tests, and tweaked the implementation a bit.

Could you try it out for your use case and make sure I haven't broken it?

lib/usage.ts
```diff
@@ -535,10 +540,14 @@ export function usage(yargs: YargsInstance, y18n: Y18N, shim: PlatformShim) {
     (Object.keys(options.alias) || []).forEach(key => {
       options.alias[key].forEach(alias => {
         // copy descriptions.
-        if (descriptions[alias]) self.describe(key, descriptions[alias]);
+        if (descriptions[alias])
```
bcoe, line 543: @OsmanAltun we should land this work on the other PR you've created first.

I'd start working from branches by default, so that it's easier to separate out your work 👍

bcoe: I went ahead and reverted it myself, I think this branch should be ready to go, once you sign off.

OsmanAltun:
I just checked and the tests don't fail, however the sample code I use for testing fails. I realized this is caused by using yargs.argv instead of yargs.parse(). The tests don't take this into consideration. Adding freeze() fixes this. I assumed that you forgot to add freeze(), since the code snippet you sent didn't work at first. 

The code snippet I use for testing:
```typescript
const argv = yargs.usage('This is my awesome program')
  .commands([
    { command: '*', description: 'testing 123', handler: () => {}},
    { command: 'foo', description: 'testing 1932', handler: () => {}}
  ]).argv;

yargs.showHelp();
console.log(argv);
```

bcoe:
> I just checked and the tests don't fail, however the sample code I use for testing fails. I realized this is caused by using yargs.argv instead of yargs.parse(). 

Good catch. The problem is that `.parse()` called `freeze()` (as it should) to store the current state of usage, command, etc. Whereas, `.argv` was a different code path which did not call `freeze()`.

To fix things I switched to making `.argv` call `.parse()` simplifying the logic. Mind checking that this fixes things for you?

OsmanAltun:
Alright, I think everything works as expected now. Go ahead and merge it 👍🏼

bcoe (approved):

PR #1831, opened 2020-12-09
feat: add support for `showVersion`, similar to `showHelp`

### LINKED ISSUE #1829: Feature: showVersion function
**Description:**
A function which has the same signature as the showHelp function. 
Ability to print version information or execute `callback(version: string)`.

OsmanAltun:
**Description**
This PR contains an added feature called showVersion. It has the same signature as the showHelp function. Also updated the docs and tests accordingly. 

 - closes #1829


bcoe:
@OsmanAltun #1823 includes a significant overhaul of `showHelp`, to make it work better with `async`/`await` (_the assumption being that users would rather have control over when they print the help message, or just get the help message as a string_).

I would like to hold off on adding this feature until I've landed updates to `showHelp`, so that we can make sure the APIs match.

docs/api.md
```diff
@@ -1504,6 +1504,33 @@ yargs.showHelp(s => myStream.write(s)); //prints to myStream
...
 
+.showVersion([consoleLevel | printCallback])
+---------------------------
+
+Print the version data.
+
+If no argument is provided, version data is printed using `console.error`.
+
+```js
+var yargs = require("yargs")
```
bcoe, line 1515: Let's show the non-sigleton usage:

```
const { hideBin } = require('yargs/helpers')
const argv = yargs(hideBin(process.argv)).argv
```

I'm trying to steer people away from using `yargs.foo` directly, because it's more bug prone.

test/usage.cjs
```diff
@@ -3198,6 +3198,61 @@ describe('usage tests', () => {
     });
   });
 
+  describe('showVersion', () => {
+    // see #143.
+    it('should show version regardless of whether argv has been called', () => {
```
bcoe, line 3203: Thank you for the tests 👍 

I think you just need to run `npm run fix` to address these linting issues.

bcoe (changes requested):
This looks pretty much ready to go to me, just a couple nits:

1. let's make the help show people the `yargs()` usage, vs., using the `yargs` singleton.
2. need to run the linter.

bcoe (approved):

PR #1836, opened 2020-12-16
update y18n

pjsharpe07:
Attempting to update y18n per [this cve](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-7774)

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "set-blocking": "^2.0.0",
     "string-width": "^1.0.2",
     "which-module": "^1.0.0",
-    "y18n": "^4.0.0",
+    "y18n": "^3.2.2",
     "yargs-parser": "^5.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


bcoe:
I will make an attempt to back port to 15.x, but 7.x is a getting a little bit silly -- is there a dependency using this that could be updated instead?

bcoe:
@pjsharpe07 what's the test matrix in yargs@7.x? I think we'd need to release an older version of y18n that supports older versions of Node.js.

Alternatively, @phated, what would it take for us to get gulp on a slightly newer yargs?

phated:
@bcoe Money. I don't have a job and everyone cancelled their Sponsors/OC donations this year.

bcoe:
@phated loud and clear, I might see if it's remotely possible to quite some of these alerts on yargs' end for you -- unless you would like a patch that attempts to upgrade gulp?

phated:
Silencing is probably the best, as the gulp-cli stuff is coupled to the runner too. So even upgrading the CLI would require a major bump and then a major bump in gulp. We have a team roadmap for v5, but no team members have time to tick off the items needed for the v5 and dropping our support matrix for anything less than node 10.

bcoe:
@pjsharpe07 mind updating this PR to `y18n@3.2.2`?

PR #1850, opened 2021-01-10
Added commands function (similar to options function)

OsmanAltun:
Separate PR as requested.


bcoe:
@OsmanAltun is the goal of this just to provide an alias for `command`?

bcoe (approved):

OsmanAltun:
Yeah, I like symmetry in code, I thought options function should be similar to commands function.

Issue #1013, opened 2017-11-23
Allow us to call the default completion function from a custom one


tkarls:
So, I have a app with a few commands and a couple of arguments to each command.

For some of these arguments I want to provide a custom completion function to be able to provide completion for the argument values.

The problem is that I still want to leverage the automatic completion functionality for the commands and argument names.

Is there a way to call the automatically generated function from the custom one if I conclude that the current parameter is not something I want/can complete?

bcoe:
@tkarls I don't believe there's currently a way to pull this off. If you have any thoughts about how to implement this feature, I'd be grateful for your contribution.

tkarls:
Well, perhaps one way would be to add a new argument to the completion function that passes in the generated one. That way, instead of calling done one can call the generated one if one so chooses?

tkarls:
@bcoe So, I tested the waters some...
https://github.com/yargs/yargs/compare/master...tkarls:enable-fallback-to-standard-completion?expand=1

Can't say that I'm 100% satisfied with the solution. To make it backwards compatible I had to add the yargs completion function as the last argument. I.e. after the done callback. I would have preferred to have the done callback last.

I didn't open a pull request as I'm not sure that this is something to build on. But wanted to share.

bcoe:
@tkarls this is looking great 💯 one thought comes to mind -- you're right, it's definitely a bit ugly that `done` comes before the default completion.

Here's an idea, what if you look at the [number of arguments that the function expects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length) before you invoke it; then invoke the closure with accordingly (always passing done in as the last argument).

Seems like you're making good progress, so please feel free to open the pull request whenever you're comfortable.

_note, I'm sometimes somewhat terrible with git issues; as I get 100s a month; feel free to prod me in [this slack channel](http://devtoolscommunity.herokuapp.com/) if I'm ever moving slow._

tkarls:
Thanks,

I'll take another look at it hopefully later this week and see if I can make the interface better by looking at the number of arguments as you suggest!

gardenappl:
Any progress on this? I'm working on a project where this would be extremely helpful.

gturi:
I tested it out and it works fine for my use case, you can even decide to merge your custom completion with yargs default one.

Recently a well explained example has been added to completion documentation https://yargs.js.org/docs/#api-reference-completioncmd-description-fn

[...]

PR #1855, opened 2021-01-24
Allow calling standard completion function from custom one

gardenappl:
See #1013 

Based on work by @tkarls (https://github.com/yargs/yargs/issues/1013#issuecomment-346814408). This is my first time working with TypeScript so my solution might be a bit ugly.

One thing that worries me is that this API requires you to use callbacks. There is no good way to overload this function:

```ts
export type CompletionFunction =
  | SyncCompletionFunction
  | AsyncCompletionFunction
  | FallbackCompletionFunction;

interface SyncCompletionFunction {
  (current: string, argv: Arguments): string[] | Promise<string[]>;
}

interface AsyncCompletionFunction {
  (current: string, argv: Arguments, done: (completions: string[]) => any): any;
}

interface FallbackCompletionFunction {
  (
    current: string,
    argv: Arguments,
    defaultCompletion: () => any,
    done: (completions: string[]) => any
  ): any;
}
```

Each function takes 2, 3 or 4 arguments respectively.

One possible solution: since you're about to release a new major version, it might be OK to just remove the ability to use callbacks, in favor of Promises. So then the new API would look like this:

```ts
export type CompletionFunction =
  | LegacyCompletionFunction
  | FallbackCompletionFunction;

interface LegacyCompletionFunction {
  (current: string, argv: Arguments): string[] | Promise<string[]>;
}

interface FallbackCompletionFunction {
  (current: string, argv: Arguments, defaultCompletion: () => any): string[] | Promise<string[]> | void;
}
```

### DIFF
--- a/lib/completion.ts
+++ b/lib/completion.ts
@@ -34,6 +34,78 @@ export function completion(
     const current = args.length ? args[args.length - 1] : '';
     const argv = yargs.parse(args, true);
     const parentCommands = yargs.getContext().commands;
+
+    function defaultCompletion(): Arguments | void {
+      const handlers = command.getCommandHandlers();
+      for (let i = 0, ii = args.length; i < ii; ++i) {
+        if (handlers[args[i]] && handlers[args[i]].builder) {
+          const builder = handlers[args[i]].builder;
+          if (isCommandBuilderCallback(builder)) {
+            const y = yargs.reset();
+            builder(y);
+            return y.argv;
+          }
+        }
+      }
+      if (
+        !current.match(/^-/) &&
+        parentCommands[parentCommands.length - 1] !== current
+      ) {
+        usage.getCommands().forEach(usageCommand => {
+          const commandName = parseCommand(usageCommand[0]).cmd;
+          if (args.indexOf(commandName) === -1) {
+            if (!zshShell) {
+              completions.push(commandName);
+            } else {
+              const desc = usageCommand[1] || '';
+              completions.push(commandName.replace(/:/g, '\\:') + ':' + desc);
+            }
+          }
+        });
+      }
+      if (current.match(/^-/) || (current === '' && completions.length === 0)) {
+        const descs = usage.getDescriptions();
+        const options = yargs.getOptions();
+        Object.keys(options.key).forEach(key => {
+          const negable =
+            !!options.configuration['boolean-negation'] &&
+            options.boolean.includes(key);
+          // If the key and its aliases aren't in 'args', add the key to 'completions'
+          let keyAndAliases = [key].concat(aliases[key] || []);
+          if (negable)
+            keyAndAliases = keyAndAliases.concat(
+              keyAndAliases.map(key => `no-${key}`)
+            );
+          function completeOptionKey(key: string) {
+            const notInArgs = keyAndAliases.every(
+              val => args.indexOf(`--${val}`) === -1
+            );
+            if (notInArgs) {
+              const startsByTwoDashes = (s: string) => /^--/.test(s);
+              const isShortOption = (s: string) => /^[^0-9]$/.test(s);
+              const dashes =
+                !startsByTwoDashes(current) && isShortOption(key) ? '-' : '--';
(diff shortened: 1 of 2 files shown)


lib/completion.ts
```diff
@@ -34,6 +34,78 @@ export function completion(
     const current = args.length ? args[args.length - 1] : '';
     const argv = yargs.parse(args, true);
     const parentCommands = yargs.getContext().commands;
+
+    function defaultCompletion(): Arguments | void {
+      const handlers = command.getCommandHandlers();
```
bcoe, line 39: i would be tempted to pull each of these major pieces of logic into a helper function within `completion.ts`:

* `commandCompletions`?
* `optionCompletions`?
* `customCompletions`?

lib/completion.ts
```diff
@@ -34,6 +34,78 @@ export function completion(
...
+      if (current.match(/^-/) || (current === '' && completions.length === 0)) {
+        const descs = usage.getDescriptions();
+        const options = yargs.getOptions();
+        Object.keys(options.key).forEach(key => {
+          const negable =
+            !!options.configuration['boolean-negation'] &&
+            options.boolean.includes(key);
+          // If the key and its aliases aren't in 'args', add the key to 'completions'
+          let keyAndAliases = [key].concat(aliases[key] || []);
+          if (negable)
```
bcoe, line 75: It's this block especially where I think we'd benefit from a bit more decomposition. Could `completeOptionKey` could be pulled out to the top level of the file, along with us breaking apart `defaultCompletion` into a couple more helpers?

I think if we then add comments to each of the helpers, the completion logic might become easier to consume.

lib/completion.ts
```diff
@@ -190,8 +207,23 @@ interface AsyncCompletionFunction {
...
+  ): any;
+}
+
 function isSyncCompletionFunction(
   completionFunction: CompletionFunction
 ): completionFunction is SyncCompletionFunction {
   return completionFunction.length < 3;
 }
+
+function isFallbackCompletionFunction(
```
bcoe, line 225: I like the use of type detection here 👍

lib/completion.ts
```diff
@@ -190,8 +207,23 @@ interface AsyncCompletionFunction {
...
+  completionFunction: CompletionFunction
+): completionFunction is FallbackCompletionFunction {
+  return completionFunction.length > 3;
```
bcoe, line 228: should this be `completionFunction.arguments.length` I'm surprised this worked.

test/completion.cjs
```diff
@@ -340,6 +340,48 @@ describe('Completion', () => {
...
+          if (err) throw err;
+          r.logs.should.include('apple');
+          r.logs.should.include('foo');
+          return done();
+        }
+      );
+    });
+
+    it('allows calling callback instead of default completion function', done => {
+      checkUsage(
```
bcoe, line 365: tests read quite well 👍

bcoe (changes requested):
@gardenappl this is looking really solid to me. What jumps out at me the most is that we've collected enough logic in `completion.ts`, that I think we'd benefit from a tiny bit of refactoring.

gardenappl:
I tried splitting the code into multiple functions, but they required a lot of arguments and it didn't look good at all. So instead I spent some time refactoring the whole thing into a class.
I also fixed handling of short options in aliases (see changes in unit tests).

gardenappl:
I am still kinda bothered by the fact that the API itself is not very clean. Right now the behaviour of `completion(...)` depends on how many arguments the function accepts:
* 2 args = synchronous function or Promise function, can't call default
* 3 args = callback function, can't call default
* 4 args = callback function, can call default

But this means that if you want to call the default function then you have to use callbacks. I already mentioned this when I created the PR and would like to hear some feedback on that.

bcoe (approved):
I think pulling the logic into a class was a very good compromise... and approach we could perhaps use elsewhere in the codebase.

LGTM. I will merge and release to a `next` branch so that you can test.

PR #1878, opened 2021-02-26
feat: allow to access output of default completion in custom completion

jacobator:
### Problem:

When using default completion both `positional` and `option` arguments are treated the same. This results in completion suggesting `positional` arguments in an `option` form, which is incorrect:

```ts
yargs
  .command(
    'exec <pos-a> <pos-b>',
    'Execute command.',
    (yargs) => {
      return yargs
        .positional('pos-a', { type: 'string', desc: 'Pos A' })
        .positional('pos-b', { type: 'string', desc: 'Pos B' })
        .option({
          'opt-a': { type: 'boolean', alias: 'oa', desc: 'Option A' },
          'opt-b': { type: 'string', requiresArg: true, desc: 'Option B' },
        });
    },
  )
  .completion('completion', 'Completion.').argv;
```

```
$ ./index.js exec --
--help     -- Show help
--opt-a    -- Option A
--opt-b    -- Option B
--pos-a    -- Pos A
--pos-b    -- Pos B
--version  -- Show version number
```

```
$ ./index.js exec --pos-a --
--help     -- Show help
--opt-a    -- Option A
--opt-b    -- Option B
--pos-b    -- Pos B
--version  -- Show version number
```

Currently, the solution is to write your on competion funtion from the ground up. There is no way to apply custom completion function that would be able to utilize default completion function output. But this would help significantly in the case above.

### Solution:

This PR adds ability to pass `onCompleted` callback to `defaultCompletion` function to replace the `done` callback, allowing us to do this:

```ts
yargs
  .command(
    'exec <pos-a> <pos-b>',
    'Execute command.',
    (yargs) => {
      return yargs
        .positional('pos-a', { type: 'string', desc: 'Pos A' })
        .positional('pos-b', { type: 'string', desc: 'Pos B' })
        .option({
          'opt-a': { type: 'boolean', alias: 'oa', desc: 'Option A' },
          'opt-b': { type: 'string', requiresArg: true, desc: 'Option B' },
        });
    },
  )
  .completion(
    'completion',
    'Completion.',
    (current, argv, defaultCompletion, done) => {
      // Execute `defaultCompletion` to get default `completions`
      defaultCompletion((_err, completions) => {
        const filteredCompletions = completions.filter(
          (c) => !c.includes('pos'),
        );
        // Execute `done` with `filteredCompletions`
        done(filteredCompletions);
      });
    },
  ).argv;
```
The example above is naive and just filtering based on the `pos` text, but illustrates the point.

```
$ ./index.js exec --
--help     -- Show help
--opt-a    -- Option A
--opt-b    -- Option B
--version  -- Show version number
```

Ideally, the default completion should ignore `positional` arguments out of the box, but this is a simple solution to fix completion right now.

### DIFF
--- a/lib/completion.ts
+++ b/lib/completion.ts
@@ -279,7 +279,7 @@ interface FallbackCompletionFunction {
   (
     current: string,
     argv: Arguments,
-    defaultCompletion: (onCompleted?: CompletionCallback) => any,
+    completionFilter: (onCompleted?: CompletionCallback) => any,
     done: (completions: string[]) => any
   ): any;
 }
--- a/lib/completion.ts
+++ b/lib/completion.ts
@@ -193,7 +193,8 @@ export class Completion implements CompletionInstance {
       return (this.customCompletionFunction as FallbackCompletionFunction)(
         current,
         argv,
-        () => this.defaultCompletion(args, argv, current, done),
+        (onCompleted = done) =>
+          this.defaultCompletion(args, argv, current, onCompleted),
         completions => {
           done(null, completions);
         }
@@ -278,7 +279,7 @@ interface FallbackCompletionFunction {
   (
     current: string,
     argv: Arguments,
-    defaultCompletion: () => any,
+    defaultCompletion: (onCompleted?: CompletionCallback) => any,
     done: (completions: string[]) => any
   ): any;
 }
--- a/test/completion.cjs
+++ b/test/completion.cjs
@@ -341,7 +341,7 @@ describe('Completion', () => {
       r.logs.should.include('success!');
     });
 
-    it('allows the custom completion function to use the standard one', done => {
+    it('allows the custom completion function to use the default completion w/o filter', done => {
       checkUsage(
         () => {
           yargs(['./completion', '--get-yargs-completions'])
@@ -349,8 +349,8 @@ describe('Completion', () => {
             .command('apple', 'banana')
             .completion(
               'completion',
-              (current, argv, defaultCompletion, done) => {
-                defaultCompletion();
+              (current, argv, completionFilter, done) => {
+                completionFilter();
               }
             )
             .parse();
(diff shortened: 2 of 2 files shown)


bcoe:
> Ideally, the default completion should ignore positional arguments out of the box, but this is a simple solution to fix completion right now.

@jacobator I'd rather try to fix the specific bug, rather than growing yargs' API. The problem is we'd then have a new method to support _forever_.

It seems like we should be able to take the disjunction between positionals and options when showing completion, and figure out how to not suggest them?

jacobator:
@bcoe I agree that the disjunction of positionals/options should be done. I will try to look into that to see if I can assist with it.

But the PR here is not only about that. Even when the positionals are excluded from the suggestion list, you would still want to modify the default suggestions. E.g. remove descriptions for all options, reorder options, etc.

The way the default completion function access was added here https://github.com/yargs/yargs/pull/1855 does not make a lot of sense. You can either use default or fully custom completion for every call. This means you have to implement a full-fledged completion for some cases and would probably not require the default completion function for the others.

Having access to default completion suggestions output serves as a great starting point, helping with the custom completion implementation w/o having to do it from the ground-up.

lib/completion.ts
```diff
@@ -278,7 +279,7 @@ interface FallbackCompletionFunction {
   (
     current: string,
     argv: Arguments,
-    defaultCompletion: () => any,
+    defaultCompletion: (onCompleted?: CompletionCallback) => any,
```
bcoe, line 282: I would be tempted to call this `completionFilter`.

lib/completion.ts
```diff
@@ -193,7 +193,8 @@ export class Completion implements CompletionInstance {
       return (this.customCompletionFunction as FallbackCompletionFunction)(
         current,
         argv,
-        () => this.defaultCompletion(args, argv, current, done),
+        (onCompleted = done) =>
```
bcoe, line 196: We should add a test:

`it('allows custom completion to be combined with default completion, using filter')`.

bcoe (changes requested):

bcoe (approved):
This LGTM, as a follow up PR, it would be good to add documentation. I'm fine with that being in a separate PR.

jacobator:
I will update the docs tomorrow

PR #1880, opened 2021-03-01
docs: add completion filter docs

jacobator:


docs/api.md
```diff
@@ -397,6 +397,27 @@ var argv = require('yargs/yargs')(process.argv.slice(2))
   .argv;
 ```
 
+Using default completions in custom implementation. When invoked with no arguments, `completionFilter` will fallback to default completion function. There is no need to call `done` in this case. When provided with a callback function, you can get access to `defaultCompletions` and call `done` with processed version of them.
```
adamaltman, line 400:
```suggestion
Using default completions in a custom implementation. When invoked with no arguments, `completionFilter` will fallback to the default completion function. There is no need to call `done` in this case. When provided with a callback function, you can get access to `defaultCompletions` and call `done` with your processed version of them.
```

bcoe (approved):

bcoe:
@jacobator thank you for the contribution.

PR #1881, opened 2021-03-01
fix: exclude positionals from default completion

jacobator:


jacobator:
@bcoe this is a fix for positional args being present in completion. Please review

bcoe (approved):

Issue #1450, opened 2019-10-15
Support types for TS with `command` fourth argument


garyking:
In TypeScript, with the following:

```typescript
import yargs from 'yargs';

yargs.command(
  'test',
  'Do something.',
  (y) => {
    y.positional('name', {
      type: 'string',
      desc: 'The name of something.',
    });
  },
  ({ name }) => {
    // `name` is of type `unknown`.
  },
);
```

`name` is of type `unknown` when it should instead be of type `string`. Please make it possible for it to inherit its type from the `type` field as specified in any `positional` or `option` methods.

garyking:
Nevermind, just had to `return y.positional`.

PR #1889, opened 2021-03-15
docs(usage): add return to fix TS inheritance

escapedcat:
Should reflect solution of this issue: https://github.com/yargs/yargs/issues/1450  

This seems to be correct in [the deno example](https://github.com/yargs/yargs#deno):  
```ts
...

yargs(Deno.args)
  .command('download <files...>', 'download a list of files', (yargs: any) => {
    return yargs.positional('files', {
      describe: 'a list of files to do something with'
    })

...
```

/cc @garyking

Please Let me know if I can improve this PR somehow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ const { hideBin } = require('yargs/helpers')
 
 yargs(hideBin(process.argv))
   .command('serve [port]', 'start the server', (yargs) => {
-    yargs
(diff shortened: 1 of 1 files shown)


bcoe (approved):

bcoe:
@escapedcat thanks for the contribution 👍

Issue #1841, opened 2020-12-23
Deno bundle yargsParser initialization failure


nchudleigh:
Trying to write a cli with Deno, went to go and try the example from the readme and the bundled code fails at runtime.

Works fine if you just deno run main.ts, but the hope was to use the bundled code (compiled to binary) for distribution.

Any ideas on why this  may be happening? Wasnt sure if this was more of a Deno issue or a yargs issue, so if I am barking up the wrong tree let me know.  

```ts
// main.ts
import yargs from 'https://deno.land/x/yargs/deno.ts'
import { Arguments } from 'https://deno.land/x/yargs/deno-types.ts'

yargs(Deno.args)
  .command('download <files...>', 'download a list of files', (yargs: any) => {
    return yargs.positional('files', {
      describe: 'a list of files to do something with'
    })
  }, (argv: Arguments) => {
    console.info(argv)
  })
  .strictCommands()
  .demandCommand(1)
  .argv
```

`❯  deno bundle ./main.ts bundle.js`
```
Bundle main.ts
Check main.ts
Emit "bundle.js" (227.35KB)
```

`❯  deno run bundle.js`
```
error: Uncaught ReferenceError: Cannot access 'yargsParser' before initialization
yargsParser.camelCase = camelCase;
^
    at bundle.js:517:1
```

bcoe:
@nchudleigh I haven't experimented with Deno much, beyond writing the unit tests to get yargs running on it.

Do you need to explicitly add `yargs-parser` to the bundle?

nchudleigh:
Tried that and didnt have any effect, the issue now seems to be cliui access issue.

```
error: Uncaught ReferenceError: cliui is not defined
    cliui: cliui,
```

I have tried explicitly bundling both of them and it gives the same issue.

I went to the yargs-parser repo and tried the same process of bundling the example in the readme and it had an issue with finding which OS it is on.

```
error: Uncaught ReferenceError: Cannot access 'isWindows' before initialization
const path = isWindows ? _win32 : _posix;
```

Perhaps this is a deeper issue with the dependency, in any case I need to dig deeper, going to close until I find something actionable.

nchudleigh:
Just tried it again with Deno version 1.6.3 and it appears to have been fixed.

seivan:
@bcoe @nchudleigh This is still broken with `deno compile` executables. 
It works when doing `deno run` but running a binary compiled with `deno compile` will fail. `shim.Parser` is undefined. 

[...]

PR #1891, opened 2021-03-23
Use actual names for keys instead of inferring.

seivan:
Fix: https://github.com/yargs/yargs/issues/1841#issuecomment-804770453

Not sure if this is a bug in deno itself, but for now this doesn't work when building distributable binaries. 

There is a distinction between the names of an exported value between  `deno run`  and `deno compile` when importing an object and using it as a key in a new object

In `'https://deno.land/x/yargs_parser@v20.2.4-deno/deno.ts'`
```ts
export default yargsParser
``` 

In `'https://deno.land/x/yargs/deno.ts'`
```ts
import Parser from 'https://deno.land/x/yargs_parser@v20.2.4-deno/deno.ts';

const newObject = {
 Parser
}
``` 
This will create `{yargParser: Parser}` in `deno compile` while work as intended with `deno run`  where you get `{Parser: Parser}`

The codebase relies on `shim.Parser` and that will be `undefined` when running `deno compile` to build binaries as it would be called `shim.yargParser`

### DIFF
--- a/lib/platform-shims/deno.ts
+++ b/lib/platform-shims/deno.ts
@@ -66,7 +66,7 @@ export default {
     notStrictEqual: assertNotEquals,
     strictEqual: assertStrictEquals,
   },
-  cliui,
+  cliui: cliui,
   findUp: escalade,
   getEnv: (key: string) => {
     return env[key];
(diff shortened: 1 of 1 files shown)


bcoe (approved):

bcoe:
@seivan thank you for the fix 👌 I believe it will be a few more weeks before we get to the `v17.x.x` release of yargs -- but I'll do my best to get this out soon.

Issue #1883, opened 2021-03-02
bug: .check() should not have thrown when fn returned a string


zypA13510:
According the [API documentation](https://github.com/yargs/yargs/blame/b368f95a33f610789fff947b5c3e90494ac3639f/docs/api.md#L120):
> If `fn` throws or returns a non-truthy value, Yargs will show the thrown error and usage information.

However, if `fn` returns a non-empty string (which would be a truthy value), yargs will still throw and exit.

Example:
```javascript
const argv = require('yargs')
  .option('a', {type: 'string'})
  .option('b', {type: 'string'})
  .option('c', {type: 'string'})
  .option('d', {type: 'string'})
  .check(argv => (argv.a && argv.b) || (argv.c && argv.d))
  .argv;
```

The intention is to require both a and b, or alternatively both c and d. But since they are string type, yargs will throw at [lib/validation.ts:300](https://github.com/yargs/yargs/blob/b368f95a33f610789fff947b5c3e90494ac3639f/lib/validation.ts#L300-L301):
```javascript
      } else if (typeof result === 'string' || result instanceof Error) {
        usage.fail(result.toString(), result);
```

zypA13510:
I know adding a `Boolean()` call surrounding the return value will workaround this. But still this is a bug. We should either fix the code or fix the docs.

bcoe:
@zypA13510 this behavior is intentional, we decided at some point that we'd allow users to return a string representing an Error, as well as an Error:

```
            } else if (typeof result === 'string' || result instanceof Error) {
              usage.fail(result.toString(), result);
            }
```

Given that the behavior has been like this for years, I think it would be better to update the documentation, vs., changing the behavior. Any interest in submitting the documentation fix?

PR #1907, opened 2021-04-07
docs: clarify .check() behavior

zypA13510:
When `fn` returns a string or error, it will cause .check() to fail. This was not documented. Adding it to docs for clarity.

Close: #1883


bcoe (approved):

bcoe:
@zypA13510 thank you for the contribution.

PR #1908, opened 2021-04-08
`wrap()` should default to `terminalWidth`

seivan:
Since `terminalWidth()` depends on local state (`shim`) it can't be a static variable that you could set
In other words this ***doesn't*** work since it requires the instance. 
```ts
import * as Yargs from "yargs"
Yargs
.default()
.wrap(Yargs.terminalWidth())
```

What does currently work is this 

```ts
import * as Yargs from "yargs"
const yargInstance = Yargs
.default()
.wrap(yargInstance.terminalWidth())
```

But that's a bit cumbersome as it would  require you to unchain, assign to a variable and then use said variable for terminalWidth (and that alone) 

So my suggestion is to make .wrap() default to `terminalWidth` of nothing is set.


bcoe:
@seivan if you run yargs without `wrap()` it should already default to terminal width. Curious what the use case is for this tweak?

seivan:
> @seivan if you run yargs without `wrap()` it should already default to terminal width. Curious what the use case is for this tweak?

And will it listen when the size changes?

bcoe:
@seivan I don't believe it will listen for size changes, so I can understand the use case of calling `getHelp()` multiple times as the terminal is resized?

Mind adding a test that verifies that this approach works?

bcoe (changes requested):
Perhaps try adding a test where we configure a terminal width that's much larger than we'd see in CI/CD, e.g., 2048, and then confirm that terminal width gets set to something less than this after call ing `.wrap()`?

bcoe:
@seivan following up on this one, mind attempting to add a test/

seivan:
> Perhaps try adding a test where we configure a terminal width that's much larger than we'd see in CI/CD, e.g., 2048, and then confirm that terminal width gets set to something less than this after call ing `.wrap()`?


Are we testing the notifications that indicate that the terminal has resized or are we testing that the value of an empty wrap is the same size as the terminal? The former is pretty redundant and I’d assume the latter already had tests and it would just just suffice with a mock to indicate terminalWith() has been called, no? 

I understand if there’s is any hesitation with mocks, in which case I guess just testing it’s the same size as terminal width after calling wrap would do it?

lib/yargs-factory.ts
```diff
@@ -1398,7 +1398,7 @@ export class YargsInstance {
     this.describe(this.#versionOpt, msg);
     return this;
   }
-  wrap(cols: number | null | undefined): YargsInstance {
+  wrap(cols: number | null | undefined = this.terminalWidth()): YargsInstance {
```
bcoe, line 1401: @seivan will `this.terminalWidth()` only be invoked if an `undefined` value is provided?

wondering if we should instead write as:

```js
wrap(cols: number | null | undefined): YargsInstance {
  if (cols === undefined) cols = this.terminalWidth()
}
```

seivan: @bcoe That depends. 
Do you think "null" is a valid option, in which case the current implementation is better. 
```ts
wrap(cols: number | null | undefined = this.terminalWidth())
```

Otherwise, if null is never a valid option, I'd suggest to switch to 
```ts
wrap(cols = this.terminalWidth())
```

bcoe:
@seivan please don't hesitate to dust off this PR if you are able to get back to adding a test, and I'd be happy to land in the future 👍

PR #1915, opened 2021-04-16
Add a list of functions to the start of the API docs for better discoverability

TikiTDO:
Exactly what it says on the tin.


bcoe:
@TikiTDO we have a table of contents automatically generated [here](http://yargs.js.org/docs/#api-reference) perhaps rather than adding this listing, we could add a link to the `yargs.js.org` site, that draws attention to the fact that this table of contents exists/

TikiTDO:
@bcoe That's a way cleaner approach. I've updated the PR to point to the link you provided.

bcoe (approved):

bcoe:
@TikiTDO thank you for the contribution 👏  it's great to have extra sets of eyes on the docs.

PR #1933, opened 2021-05-02
refactor: small simplifications

jly36963:
While looking through the repo, I saw a few lines/sections that could be simplified. 

This is my first OSS PR, so let me know if there are things I can do better.

I'll be looking through more of the repo, and eventually the issues tab, but if there is anything you want me to focus on, let me know!

### DIFF
--- a/lib/cjs.ts
+++ b/lib/cjs.ts
@@ -13,10 +13,9 @@ import cjsPlatformShim from './platform-shims/cjs.js';
 
 // See https://github.com/yargs/yargs#supported-nodejs-versions for our
 // version support policy. The YARGS_MIN_NODE_VERSION is used for testing only.
-const minNodeVersion =
-  process && process.env && process.env.YARGS_MIN_NODE_VERSION
-    ? Number(process.env.YARGS_MIN_NODE_VERSION)
-    : 10;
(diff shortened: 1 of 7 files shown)


lib/cjs.ts
```diff
@@ -13,10 +13,9 @@ import cjsPlatformShim from './platform-shims/cjs.js';
 
 // See https://github.com/yargs/yargs#supported-nodejs-versions for our
 // version support policy. The YARGS_MIN_NODE_VERSION is used for testing only.
-const minNodeVersion =
-  process && process.env && process.env.YARGS_MIN_NODE_VERSION
-    ? Number(process.env.YARGS_MIN_NODE_VERSION)
-    : 10;
+const minNodeVersion = process?.env?.YARGS_MIN_NODE_VERSION
+  ? Number(process.env.YARGS_MIN_NODE_VERSION)
+  : 10;
```
bcoe, line 18: perhaps in a second PR, we should actually make this `: 12` now, since we just dropped Node 10 from the engines field

bcoe (approved):
Looks good to me, with one recommendation for a follow up PR.

In practice I'd tend to make PRs for each distinct thing you're working on, e.g.,

"fixes bug with X", or "refactor if/else to use some/every", but for this small set of changes I'm fine with the PR title.

PR #1944, opened 2021-05-11
refactor: small simplifications 2

jly36963:
## Changes

- Remove shallow copy logic in `lib/yargs-factory.js`.
- Use ternary and || operator where practical.
- Update comments.
- Lint.

I'm still new at this, so let me know if there is anything I can do better.

### DIFF
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -336,19 +336,16 @@ export class CommandInstance {
         commandIndex,
         helpOnly
       );
-    if (isPromise(innerArgv)) {
-      return innerArgv.then(argv => {
-        return {
+
+    return isPromise(innerArgv)
+      ? innerArgv.then(argv => ({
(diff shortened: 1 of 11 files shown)


lib/command.ts
```diff
@@ -602,7 +596,7 @@ export class CommandInstance {
       });
 
       Object.keys(parsed.argv).forEach(key => {
-        if (positionalKeys.indexOf(key) !== -1) {
+        if (~positionalKeys.indexOf(key)) {
```
bcoe, line 599: I think here you could just use `.includes` now?

bcoe (commented):
left one small nit.

lib/command.ts
jly36963: Sounds good to me.

jly36963:
Concede it is.
Thank you for always being kind -- I wish you the best :)

Issue #1949, opened 2021-05-20
[FeatureRequest] make `require("yargs").default` as same as `require("yargs")`


loynoir:
### Brief
`strictOptions` can not be trigger under `require("yargs").default`

### Background
[official document](https://github.com/yargs/yargs#esm)

```ts
// x.ts
import yargs from 'yargs'
```

```js
// tsc output x.js is something like
const yargs = require("yargs").default
```

### Expected
```js
require("yargs").default(["bar", "-a", "10"]).strictOptions().argv
```
```txt
// Error
Options:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]

Unknown argument: a
```

### Actual
```js
require("yargs").default(["bar", "-a", "10"]).strictOptions().argv
```
```txt
// Error is not trigger
{
  '10': undefined,
  _: [],
  bar: undefined,
  '-a': undefined,
  a: undefined,
  '$0': ''
}
```

### Workaround
```ts
// y.ts
import * as yargs from 'yargs'
```

```js
// tsc output y.js is something like
var yargs = require("yargs");
```

### Reproduce Gist
[gist](https://gist.github.com/loynoir/580bb181222bf21452412c0a86ae9cbf)

karizma:
The problem is with conversion from esmodules to commonjs, not a bug. 

You either need to use `import * as yargs from "yargs"` or need the `allowSyntheticDefaultImports` for typescript.

Read more here:
https://www.typescriptlang.org/tsconfig#allowSyntheticDefaultImports

Tried to explain it for this specific scenario since it's a bit unique but the first link should give you enough info:
TL;DR: You are actually accessing [yargs.default()](https://yargs.js.org/docs/#api-reference-defaultkey-value-description), not `yargs`.

Basically `require("yargs");`, exports the function and all the methods the api uses like `.command()` or `.string()`. 
When you are importing it with typescript by `import yargs from "yargs"`, if the file is commonjs, it calls 'default' on it.

so, this code:

```js
import yargs from "yargs";
yargs(['bar', '-a', '10']).strictOptions().argv;
```
turns into:
```js
const yargs_1 = require("yargs");
yargs_1['default'](['bar', '-a', '10']).strictOptions().argv;
```

Like I mentioned before `require("yargs")` already exports the main function and all the methods, so we don't need to access 'default' on it.
Of course typescript doesn't know it so it just has to assume. 

[...]

PR #1950, opened 2021-05-21
Update README.md for compactablity

loynoir:
https://github.com/yargs/yargs/issues/1949

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ yargs(Deno.args)
 As of `v16`,`yargs` supports ESM imports:
 
 ```js
-import * as yargs from 'yargs';
+import * as yargs from 'yargs'
 import { hideBin } from 'yargs/helpers'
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -146,7 +146,7 @@ yargs(Deno.args)
 As of `v16`,`yargs` supports ESM imports:
 
 ```js
-import yargs from 'yargs'
+import * as yargs from 'yargs'
```
bcoe, line 149: I believe this is only necessary if you're using TypeScript?

And the syntax can vary depending on how you have the compiler configured, e.g.,

https://github.com/googleapis/repo-automation-bots/blob/master/packages/owl-bot/src/bin/commands/copy-code.ts#L18

I think we document the appropriate approach here:

https://github.com/yargs/yargs/blob/master/docs/typescript.md

Could we draw attention to this better in the README?

PR #1971, opened 2021-06-19
fix: Generate sourcemap for the minified file build/index.cjs

rpl:
This PR includes a single line change to the rollup configuration, to enable the sourcemap file generation.
The file generated is going to the placed by rollup into the `build/` subdirectory and so it should be already become part of the assets released on npm.

Including it helps to make sure that dependents of the yargs package will get more readable errors stacktraces when the error triggered is coming from one of the minified node modules. 

<details>
<summary>Error stacktrace without sourcemap</summary>

```
Error:
    at someInternalFunctionFromMyOwnProject (...)
    at coerce (...)
    at n.<computed> (.../yargs/build/index.cjs:1:28508)
    at j (.../yargs/build/index.cjs:1:4816)
    at .../yargs/build/index.cjs:1:28483
    at .../yargs/build/index.cjs:1:4674
    at Array.reduce (<anonymous>)
    at C (.../yargs/build/index.cjs:1:4521)
    at _.applyMiddlewareAndGetResult (.../yargs/build/index.cjs:1:8609)
    at _.runCommand (.../yargs/build/index.cjs:1:7127)
```
</details>

<details>
<summary>Error stacktrace with sourcemap</summary>

```
Error:
    at someInternalFunctionFromMyOwnProject (...)
    at coerce (...)
    at getResult (.../yargs/lib/yargs-factory.ts:383:20)
    at maybeAsyncResult (.../yargs/lib/utils/maybe-async-result.ts:17:44)
    at middleware (.../yargs/lib/yargs-factory.ts:378:16)
    at reduce (.../yargs/lib/middleware.ts:107:24)
    at Array.reduce (<anonymous>)
    at applyMiddleware (.../yargs/lib/middleware.ts:89:22)
    at _.applyMiddlewareAndGetResult (.../yargs/lib/command.ts:398:17)
    at _.runCommand (.../yargs/lib/command.ts:240:19)
```
</details>

### DIFF
--- a/rollup.config.cjs
+++ b/rollup.config.cjs
@@ -3,6 +3,7 @@ const ts = require('@wessberg/rollup-plugin-ts');
 const {terser} = require('rollup-plugin-terser');
 
 const output = {
+  sourcemap: true,
   format: 'cjs',
   file: './build/index.cjs',
   exports: 'default',


bcoe:
@rpl I was intentionally not supporting source maps, because they close to double the size of the library.

The reason I think it's okay not to ship them for the CJS version of the library, is that my hope is that more and more people will adopt the ESM version of yargs over the next couple years.

Keeping the module size small makes it feasible to offer a dual-mode module while folks transition from CJS to ESM.

bcoe:
@rpl appreciate the contribution 👏 but I would like to stick with the smaller library size for now. I could imagine adding some docs that describe why we went this route.

rpl:
> @rpl appreciate the contribution clap but I would like to stick with the smaller library size for now. I could imagine adding some docs that describe why we went this route.

@bcoe  yeah, a mention in the doc may help at least to don't be surprised by it, I'm not sure how many projects will keep using the commonjs entrypoint in the near future (e.g. for [web-ext](https://github.com/mozilla/web-ext) we haven't looked into moving to 100% ESM yet, we still support nodejs 12 and I was hoping to defer that until nodejs 14 was the only supported version) but personally I wouldn't exclude that other projects may hit this and wonder why the sourcemap isn't included in the npm package.

For web-ext I will be probably just deferring a bit more updating the yargs dependency to version 17 (and I'll cross my finger that there will not be any security issue in a transitive dependency that would be forcing us too do it sooner than when we will have already evaluated how we will be migrating web-ext to 100% ESM modules).

PR #1973, opened 2021-06-22
Upgrade Levenshtein to Damerau-Levenshtein

thrien:
Minor improvement for convenience. The Damerau-Levenshtein distance considers transpositions (which are common in mistyped strings) as one edit.


bcoe:
@thrien thanks for the contribution 👏 , this code certainly hasn't been touched in a long time.

bcoe:
@thrien tests are failing due to a slight drop in coverage, it looks like line 57 of the new algorithm isn't exercised. Could we add a test case?

thrien:
Line 57 is the new case that counts as one edit (transposition). So it makes sense that it doesn't get tested.

I will try to add a test, even though i don't really know how tests work.

bcoe (approved):

bcoe:
@thrien thank you for the contribution 👏

mrlubos:
Thanks @thrien!

Issue #1861, opened 2021-02-03
strict() and strictCommands() do not appear to complain on unknown commands


biggianteye:
The documentation for `strict()` says that

> Unrecognized commands will also be reported as errors.

But this is not the behaviour I am seeing with version 16.2.0

Here is my test code:

```javascript
const yargs = require('yargs');

const argv = yargs.strict().argv;

console.log(argv);
```

In the above code, no commands or options are defined and so everything is unknown. Here is an example of an unknown command (`foo`) being accepted:

```
[burhan@enceladus yargs-test] {0} $ node index.js foo
{ _: [ 'foo' ], '$0': 'index.js' }
```

Unknown options behave as expected, based on the documentation. Here is an example of an unknown option (`--foo`) being treated as an error.

```
[burhan@enceladus yargs-test] {0} $ node index.js --foo
Options:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]

Unknown argument: foo
```

Am I misunderstanding the `strict()` documentation? If so, how do I ensure that any unknown command is treated as an error?

jly36963:
The check [here](https://github.com/yargs/yargs/blob/master/lib/validation.ts#L175) gets skipped because `currentContext.commands` is an empty array, and so is `commandKeys`. A few lines later, `unknown.length` is 0, so `usage.fail` never gets called.

I'll look into this tomorrow, hopefully I can figure out why this happens.

jly36963:
@biggianteye Let me know if there's anything more to investigate/fix on this

biggianteye:
> @biggianteye Let me know if there's anything more to investigate/fix on this

Looks good to me. 👍🏽

PR #1977, opened 2021-07-04
fix: strict should fail unknown arguments

jly36963:
Addresses: https://github.com/yargs/yargs/issues/1861

As always, let me know if I need to change/add/fix anything.

## Issue

This should fail:
```js
const argv = yargs.strict().parse('foo');
console.log(argv) // { _: [ 'foo' ], '$0': 'example.js' }
```

Neither `unknownArguments` nor `unknownCommands` catch extra args when `commandKeys.length` is 0.

## Fix

I added a block to check if there are any elements in `argv` that aren't in `currentContext.commands`. I made sure to tolerate args when `yargs.demand(number)`  is called, when the number of args falls within the min/max range.

This shouldn't fail:
```js
const argv = yargs.strict().demand(1).parse('foo');
console.log(argv) // { _: [ 'foo' ], '$0': 'example.js' }
```

## Note

I don't understand the `demand` logic well (as it's deprecated and not in the docs), so I'm not sure if the slice in [this line](https://github.com/yargs/yargs/pull/1977/files#diff-4dd83f100006aaa41a4fdf091876a4dd2eee068706221a21c9098ba7164b4e21R190) will always be correct:

I saw that a few other [places](https://github.com/yargs/yargs/blob/a87137143bb4ee33bbb76145d33c3322f8dce26f/lib/validation.ts#L175) do something similar.

### DIFF
--- a/lib/validation.ts
+++ b/lib/validation.ts
@@ -182,17 +182,18 @@ export function validation(
     // https://github.com/yargs/yargs/issues/1861
     if (checkPositionals) {
       // Check for non-option args that are not in currentContext.commands
-      // If yargs.demand called, tolerate non-option args within min/max range
+      // Take into account expected args from commands and yargs.demand(number)
       const demandedCommands = yargs.getDemandedCommands();
       const maxNonOptDemanded = demandedCommands._?.max || 0;
-      if (maxNonOptDemanded < argv._.length) {
-        argv._.slice(maxNonOptDemanded).forEach(nonOptArg => {
-          nonOptArg = String(nonOptArg);
+      const expected = currentContext.commands.length + maxNonOptDemanded;
+      if (expected < argv._.length) {
+        argv._.slice(expected).forEach(key => {
+          key = String(key);
           if (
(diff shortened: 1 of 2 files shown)


bcoe:
@jly36963 thanks for digging into this, will make an effort to review this week 👍 It's great to start squashing some of the long standing bugs.

bcoe (approved):

bcoe:
@biggianteye, @brlodi, I've published @jly36963's fix to `yargs@next`, mind trying it out and seeing if this approach would work for your application?

`npm i yargs@next`, or `npm i yargs@17.0.2-candidate`.

biggianteye:
> @biggianteye, @brlodi, I've published @jly36963's fix to `yargs@next`, mind trying it out and seeing if this approach would work for your application?
> 
> `npm i yargs@next`, or `npm i yargs@17.0.2-candidate`.

I've just tested this version with the code I used in the bug report and it is now working as expected. Thanks!

```
$ node index.js foo
Options:

Unknown argument: foo
```

Issue #1966, opened 2021-06-13
Coerce gets called twice for nested command positionals


CupOfTea696:
Coerce gets called multiple times for nested command positional, this could cause issues when the arg type is converted.

**Example Script:**
```js
import yargs from 'yargs'

const argv = yargs('rule add foo bar baz').showHelpOnFail(false).fail(msg => {
  console.error(`[ERR] ${msg}`)
  process.exit(1)
}).command('rule', '', yargs => {
  return yargs.command('add <name> <desc...>', '', yargs => {
    return yargs.positional('desc', {
      type: 'string',
      coerce: arg => {
        console.log(arg)

        return arg.join(' ')
      }
    })
  })
}).argv

// the parsed data is stored in argv.
console.log(argv)
```

**Expected Output:**
```
[ 'bar', 'baz' ]
{ _: [ 'rule', 'add' ], '$0': 't.js', name: 'foo', desc: 'bar baz' }
```

**Actual Output:**
```
[ 'bar', 'baz' ]
bar baz
[ERR] arg.join is not a function
```

shadowspawn:
This was fixed in #1978. Thanks @CupOfTea696 for report, and fix from @jly36963

Issue #1802, opened 2020-11-11
Coerce is called multiple times


lewisdiamond:
Coerce should always be run once and only once.

```javascript
const yargs = require("yargs/yargs");
const { hideBin } = require("yargs/helpers");
const cmd1 = {
    command: "cmd1",
    desc: "cmd1",
    builder: (yargs) => {
        yargs.options({
            opt1: {
                type: "string",
                coerce: (opt1) => {
                    console.log("opt1", opt1);
                    return opt1;
                },
            },
        });
    },
    handler: (argv) => {
        console.log(argv);
    },
};
const argv = yargs(hideBin(process.argv))
    .options({
        opt: {
            coerce: (opt) => {
                console.log("opt", opt);
                return opt;
            },
        },
    })
    .command(cmd1)
    .help().argv;
```
```
$ node index.js --opt test cmd1 --opt1 test1
opt test
opt test
opt1 test1
{ _: [ 'cmd1' ], opt: 'test', opt1: 'test1', '$0': 'index.js' }
```

shadowspawn:
Fixed by https://github.com/yargs/yargs/pull/1978 (thanks @jly36963)

PR #1978, opened 2021-07-05
fix: coerce middleware should be applied once

jly36963:
Addresses: https://github.com/yargs/yargs/issues/1966 and https://github.com/yargs/yargs/issues/1802

## Debugging

I know that the coerce middleware is being applied twice.
I ran the below example. I logged the execution of the functions that use `applyMiddleware`, and I logged the type/value of the arg during coercion.

```js
const argv = yargs("rule add foo bar baz")
  .command("rule", "rule desc", (yargs) =>
    yargs.command("add <name> <abc...>", "add desc", (yargs) =>
      yargs.positional("abc", {
        type: "string",
        coerce: (arg) => console.log({ argType: typeof arg, arg }) || arg.join(" "),
      })
    )
  ).argv;
```

Logs look like this:
```
parse called
kRunYargsParserAndExecuteCommands called
runCommand called
parseAndUpdateUsage called
kRunYargsParserAndExecuteCommands called
runCommand called
parseAndUpdateUsage called
kRunYargsParserAndExecuteCommands called
pre-validation applyMiddleware called
{ argType: 'object', arg: [ 'bar', 'baz' ] }
post-validation applyMiddleware called
pre-validation applyMiddleware called
{ argType: 'string', arg: 'bar baz' }
```

The coerce middleware is only supposed to run before validation, but the pre-validation `applyMiddleware` gets called twice.

### stack trace

```
get > parse > kRunYargsParserAndExecuteCommands > runCommand > applyMiddlewareAndGetResult > applyMiddleware
```
### The problem

`runCommand` gets called recursively, which means that the middleware gets applied for each nested command.

Coerce middleware mutates argv, and repeating that mutation is undesired and can easily break.
ie: `coerce: (arg) => arg.join()`

### The solution

I added two properties to the Middleware interface:

- mutates (bool): if the middleware mutates argv and should only be applied once.
- applied (bool): if the mutating middleware has already been applied once.

`mutates` defaults to false, but is set to true in `addCoerceMiddleware`.

I then check each middleware during `applyMiddleware`:

- If it mutates argv:
  - If it hasn't been applied, apply it and set `applied` to true.
  - If it mutates and has been applied, skip it.
- If it doesn't mutate argv:
  - handle the middleware like it normally would
  
 ### Feedback
 
If you have recommendations for a better pattern, or better names for the interface properties I added, let me know. I feel like they could definitely be named better.

### DIFF
--- a/test/command.cjs
+++ b/test/command.cjs
@@ -1549,7 +1549,7 @@ describe('Command', () => {
       });
 
       // addresses https://github.com/yargs/yargs/issues/1966
-      it('should not be applied multiple times for nested commands', done => {
+      it('should not be applied multiple times for nested commands', () => {
         let coerceExecutionCount = 0;
 
         const argv = yargs('cmd1 cmd2 foo bar baz')
@@ -1575,7 +1575,6 @@ describe('Command', () => {
 
         argv.rest.should.equal('bar baz');
         coerceExecutionCount.should.equal(1);
-        return done();
       });
     });
 
--- a/test/command.cjs
+++ b/test/command.cjs
@@ -1547,6 +1547,30 @@ describe('Command', () => {
             return done();
           });
       });
+
+      // addresses https://github.com/yargs/yargs/issues/1966
+      it('should not be applied multiple times', done => {
+        yargs('rule add foo bar baz')
+          .command('rule', 'rule desc', yargs =>
+            yargs.command(
+              'add <name> <abc...>',
+              'add desc',
+              yargs =>
+                yargs.positional('abc', {
+                  type: 'string',
+                  coerce: arg => arg.join(' '),
+                }),
+              argv => {
+                argv.abc.should.equal('bar baz');
+                return done();
+              }
+            )
+          )
+          .fail(() => {
+            expect.fail();
+          })
+          .parse();
+      });
     });
 
     describe('defaults', () => {
--- a/test/command.cjs
+++ b/test/command.cjs
@@ -1550,6 +1550,8 @@ describe('Command', () => {
 
       // addresses https://github.com/yargs/yargs/issues/1966
       it('should not be applied multiple times', done => {
+        let coerceExecutionCount = 0;
+
         yargs('rule add foo bar baz')
           .command('rule', 'rule desc', yargs =>
             yargs.command(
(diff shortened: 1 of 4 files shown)


jly36963:
@bcoe 
No worries if you don't have time -- I'll keep digging.
Do you have an idea of what causes the second pre-validation applyMiddleware?

bcoe:
@jly36963 thanks for digging into this one, this would be a great bug to fix. middleware was just overhauled drastically.

test/command.cjs
```diff
@@ -1547,6 +1547,30 @@ describe('Command', () => {
...
+      it('should not be applied multiple times', done => {
+        yargs('rule add foo bar baz')
+          .command('rule', 'rule desc', yargs =>
+            yargs.command(
+              'add <name> <abc...>',
+              'add desc',
+              yargs =>
+                yargs.positional('abc', {
+                  type: 'string',
+                  coerce: arg => arg.join(' '),
```
bcoe, line 1561: I think it would be worth extending this function so that it throws if `coerce` is called more than once.

```js
let executionCount = 0
...
coerce: arg => {
  if (executionCount > 0) throw Error('coerce applied multiple times');
  executionCount++;
  return arg.join(' ');
};
```

I would then confirm that the test fails prior to your fix.

bcoe (changes requested):
Looking reasonable to me, except I think we could extend the test to more explicitly call out the failure that's happening.

jly36963:
I made some changes to the unit test, and this is the result:

- True negative: fails if the fix is not implemented (tested on master branch)
- True positive: passes if the fix is implemented (tested on this branch)
- I intentionally did not put the assertions / `done()` inside a command handler. (Details below)

When I put the assertions/done in the handler and tested on master branch, The test passed when it should have failed. This is what the logs looked like:

- coerce middleware called
- handler called
-  ✓ should not be applied multiple times for nested commands
- coerce middleware called
- fail callback called

`done()` had already been called before the second execution of coerce would have caused it to fail.

test/command.cjs
```diff
@@ -1547,6 +1547,36 @@ describe('Command', () => {
...
+                .fail(() => {
+                  expect.fail();
+                })
+            )
+          )
+          .parse();
+
+        argv.rest.should.equal('bar baz');
+        coerceExecutionCount.should.equal(1);
+        return done();
```
bcoe, line 1578:
```suggestion
```

test/command.cjs
```diff
@@ -1547,6 +1547,36 @@ describe('Command', () => {
             return done();
           });
       });
+
+      // addresses https://github.com/yargs/yargs/issues/1966
+      it('should not be applied multiple times for nested commands', done => {
```
bcoe, line 1552:
```suggestion
      it('should not be applied multiple times for nested commands', () => {
```

bcoe (approved):

bcoe:
@andymcintosh @CupOfTea696 I've published @jly36963's fix to `next`, could you give this a try:

`npm i yargs@next`.

And let us know if it fixes the issues you were bumping into?

CupOfTea696:
@bcoe yup, that fixes the issue!

Issue #1203, opened 2018-08-17
implies doesn't work if value is zero


lilleyse:
```js
return yargs
    .options({
        longitude: {
            type: 'number'
        },
        latitude: {
            type: 'number'
        }
    })
    .implies('longitude', 'latitude')
    .implies('latitude', 'longitude')
    .argv;
```

A command like `node command.js --longitude=0` proceeds when it should have produced an implication failure message.

lilleyse:
Possibly related - a command like `node command.js --longitude=0 --latitude=10` fails:

```
Missing dependent arguments:
 latitude -> longitude
```

bcoe:
thanks for the bug report. If you're so inclined, would happily take a PR for a failing test, or a fix 😄

shadowspawn:
Thanks for the report. This should be fixed by #1985.

Feel free to open a new issue if it comes up again, with new information and renewed interest.
 
Thank you  for your contributions.

Issue #1091, opened 2018-03-20
Validation fails when x => y where y is 0


sangeeth96:
Here's the test case that mimics the issue I encountered:

```
it.only('shouldn\'t fail if an option implies on a number option where value is 0', () => {
  let failCalled = false

  const argv = yargs('--bool --num 0')
  .boolean('bool')
  .number('num')
  .implies('bool', 'num')
  .fail((msg) => {
    failCalled = true
  })
  .argv

  failCalled.should.be.false
  expect(argv.bool).to.be.true
  expect(argv.num).equals(0)
})
```

Running `mocha`, this prints the following:
```
  validation tests
    implies
      1) shouldn't fail if an option implies on a number option where value is 0


  0 passing (57ms)
  1 failing

  1) validation tests implies shouldn't fail if an option implies on a number option where value is 0:

      AssertionError: expected true to be false
      + expected - actual

      -true
      +false

      at Context.it.only (test\validation.js:171:28)
```

This should clearly not be failing as 0 is a valid number but on my project, I get implication error when 0 is passed as the value. Note that other values work fine. I believe this is because the code relies on JS coercion to determine if the implied key exists or not. I have already tried and made a quick-fix which currently passes all the tests including this one but JIC, need to know if I'm missing something regarding this issue.

shadowspawn:
This has hopefully been fixed by #1985
 
Thank you  for your contributions.

PR #1985, opened 2021-07-14
fix: implies should not fail when implied key's value is 0, false or empty string

mathieubergeron:
This is an attemps at fixing #1091 and #1203.

## The issue

The documentation for `implies` is:

> Given the key x is set, it is required that the key y is set.

But this check fails when the value of the key `y` is `0`, `false`, or empty string.

When `y` value is `false`, we might want `implies` to continue failing though. I'd like to have your input on that, and I will adjust that PR accordingly.

The issue came from `keyExists` which, instead of always returning a boolean, could returned the value of the key. That value is later on used like so:
```
if (key && !value) {
  implyFail.push(` ${origKey} -> ${origValue}`);
}
```
Thus, that `if` resolved to `true` when `value` was `0`, `false` or `""`.

## The fix

Having `keyExists` always returning a boolean seems more robust. I used `Object.hasOwnProperty` to check for presence of the key instead.

Let me know if you think that `keyExists` should still return `false` when the value of the key is `false`.

Thanks :)

### DIFF
--- a/test/validation.cjs
+++ b/test/validation.cjs
@@ -102,39 +102,33 @@ describe('validation tests', () => {
     });
 
     it("doesn't fail if implied key exists with value 0", () => {
-      let failCalled = false;
       yargs('--foo --bar 0')
         .implies('foo', 'bar')
-        .fail(msg => {
-          failCalled = true;
+        .fail(() => {
+          expect.fail();
         })
         .parse();
-      failCalled.should.equal(false);
     });
 
     it("doesn't fail if implied key exists with value false", () => {
-      let failCalled = false;
       yargs('--foo --bar false')
         .implies('foo', 'bar')
-        .fail(msg => {
-          failCalled = true;
+        .fail(() => {
+          expect.fail();
         })
         .parse();
-      failCalled.should.equal(false);
     });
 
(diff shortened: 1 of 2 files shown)


bcoe:
@mathieubergeron thanks for the contribution 👍 will make an effort to give review soon.

bcoe:
@jly36963 if you happen to beat me to code review, at a glance this contribution looked solid to me 😄

test/validation.cjs
```diff
@@ -101,6 +101,28 @@ describe('validation tests', () => {
       failCalled.should.equal(true);
     });
 
+    it("doesn't fail if implied key exists with value 0", () => {
+      let failCalled = false;
+      yargs('--foo --bar 0')
+        .implies('foo', 'bar')
+        .fail(msg => {
+          failCalled = true;
```
jly36963, line 109: You could probably remove `failCalled` declaration/update/assertion and instead use `chai.expect.fail()` in the fail callback. `expect` should already be imported from chai.

```js
.fail(() => {
  expect.fail()
}
```

jly36963:
@bcoe I tested it locally with and without his changes, his fix works.

jly36963 (approved):

bcoe (approved):

Issue #1979, opened 2021-07-05
Parameter `version` should not be allowed or should be overrided


giovanni-bertoncelli:
Yargs behaves unexpectedly when declaring a parameter named `version`, like this:

```javascript
const { argv } = require('yargs')
  .string('version')
  .alias('version', 'v')
  .describe('version', 'Target version')
```

Usually in the result you get that `argv.version` is always false and the parameter passed gets hyphenated, like this:
```bash
node . --version abc
```

```javascript
{"_": ["abc"], "version": false}
```

I think this parameter does not correctly override the default `version` parameter expected by yargs. It should be returned a warning for the default behaviour to be overridden or it should return an error if that parameter should not be overridden.

bcoe:
@giovanni-bertoncelli, @jly36963 has added a descriptive warning that will help prevent someone using yargs from falling into this trap (_suggesting that folks either turn off the automatic version behavior, or rely on its default behavior_).

xkn1ght:
actually it did have a way to mock version global and mainwhile maintain subcommand --version functional.https://github.com/yargs/yargs/issues/1918

Issue #1864, opened 2021-02-11
Yargs should throw an error when using reserved names for options


alex3683:
I wanted to add an option called `version` with type string but no matter what I passed as value the resulting argument value was always `false`. I wasted nearly an hour trying to find out what I did wrong. I then stumbled over the `version()` method, enabling the magic `version` option, used to display the app version. Only then I knew that I have to rename my option in order to not collide with this reserved option.

It would have been nice if yargs would have thrown an error stating `version is a reserved option and you are not allowed to use it`, because then I would not have wasted that time. This might save other some time as well ;-) I guess this may also apply to other reserved option names.

antmarot:
A quick note to say that just happened to me as well.
For anyone wondering how to disable the default version option, just use `.version(false)`.

bcoe:
@antmarot @alex3683, @jly36963 has made it so a descriptive warning is displayed if you use the reserved option `version`.

I think we should probably extend this behavior to `help` as well, at which point I'd like to close this issue.

I believe throwing is too heavy handed, as it will break many applications in the wild, whereas a warning is just a nuisance (and points users towards the correct behavior).

PR #1986, opened 2021-07-16
fix: prevent version name collision

jly36963:
Addresses: #1979 and #1864

## Problem

"version" can be passed as a key to `yargs.option()`, but this messes up parsing. 
The `yargs.version()` method already occupies that namespace.

This code

```js
yargs.command(
  "cmd1",
  "cmd1 desc",
  (yargs) =>
    yargs.option("version", {
      alias: "v",
      describe: "version desc",
      type: "string",
    }),
  (argv) => {
    console.log({ argv });
  }
).parse('cmd1 --version 0.25.10')
```

produces this result:

```js
{
  argv: {
    _: [ 'cmd1', '0.25.10' ],
    version: false,
    v: false,
    '$0': 'my-script.js'
  }
}
```

## Solution

I added a check in `yargs.option` that will ~throw an error~ emit a warning if an option (or its alias) is "version", unless they disabled version with `yargs.version(false)`.

### DIFF
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -1470,12 +1470,12 @@ export class YargsInstance {
   }
   [kEmitWarning](
     warning: string,
-    name: string | undefined,
+    type: string | undefined,
     deduplicationID: string
   ) {
     // prevent duplicate warning emissions
     if (!this.#emittedWarnings[deduplicationID]) {
-      this.#shim.process.emitWarning(warning, name);
+      this.#shim.process.emitWarning(warning, type);
       this.#emittedWarnings[deduplicationID] = true;
     }
   }
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -904,10 +904,10 @@ export class YargsInstance {
         opt = {};
       }
 
-      // Prevent version name collision
+      // Warn about version name collision
       // Addresses: https://github.com/yargs/yargs/issues/1979
       if (this.#versionOpt && (key === 'version' || opt?.alias === 'version')) {
-        throw new YError(
+        this.#shim.process.emitWarning(
           [
             '"version" is a reserved word.',
             'Please do one of the following:',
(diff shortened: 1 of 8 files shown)


lib/yargs-factory.ts
```diff
@@ -904,6 +904,21 @@ export class YargsInstance {
         opt = {};
       }
 
+      // Prevent version name collision
+      // Addresses: https://github.com/yargs/yargs/issues/1979
+      if (this.#versionOpt && (key === 'version' || opt?.alias === 'version')) {
+        throw new YError(
+          [
+            '"version" is a reserved word.',
+            'Please do one of the following:',
```
bcoe, line 913: I think throwing an error is too heavy handed, as it would break a lot of applications. Here's what I would do:

1. emit a custom warning with a similar message to this, using the approach described [here](https://nodejs.org/api/process.html#process_emitting_custom_warnings). We should make sure that we don't show it more than once, using the approach outlined in the docs.
2. I think we should turn off our own handling of `.version()` if someone configures it themselves, which the warning message can draw attention to.

What I'm not sure of, is whether we should call this a breaking change.

bcoe (changes requested):
see comments.

lib/yargs-factory.ts
jly36963: A couple of questions:
- I want to capture output and prevent it from going to stdout. Is patching `process.emitWarning` in `checkUsage` the way to go? I see that there are already patches for a few process and console methods there.
- Would emitting a warning be sufficient? It would keep this from becoming a breaking change.

bcoe (approved):

bcoe (commented):
This seems like a really solid work to me 👍 , thank you. I'm sure we'll get additional usage out of `emitWarning`.

My only concern is if a major project, e.g., `mocha`, starts emitting on this warning, and we get a ton of users complaining on yargs.

I'm comfortable with shipping this, and keeping an eye on issues opened against yargs.

Issue #1637, opened 2020-05-01
First argument is lost when variadic arg is explicitly specified


Benjin:
```
let argv = require('yargs')
	.command('$0 [foods...]')
	.help().wrap(null).version(false).argv;

console.log(argv);
```

Arg not explicitly specified (good output):
```
PS F:\Source\scripts> node trial.js apples carrots
{ _: [], '$0': 'trial.js', foods: [ 'apples', 'carrots' ] }
```

Arg explicitly specified (bad output, expected to match above):
```
PS F:\Source\scripts> node trial.js --foods apples carrots
{ _: [], foods: [ 'carrots' ], '$0': 'trial.js' }
```

mleguen:
Also note that:
```
PS F:\Source\scripts> node trial.js --foods apples --foods carrots
{ _: [], '$0': 'trial.js', foods: [ 'apples', 'carrots' ] }
```

So:
- either positionals should overwrite options, and `foods` should be empty here
- or they should be concatenated and `foods` should equal `[apple, carrots]` in @Benjin 's 2nd example

bcoe:
thanks for the bug report, definitely seems like something we should address.

Benjin:
> 
> 
> Also note that:
> 
> ```
> PS F:\Source\scripts> node trial.js --foods apples --foods carrots
> { _: [], '$0': 'trial.js', foods: [ 'apples', 'carrots' ] }
> ```
> 
> So:
> 
>     * either positionals should overwrite options, and `foods` should be empty here
> 
>     * or they should be concatenated and `foods` should equal `[apple, carrots]` in @Benjin 's 2nd example

Wouldn't "overwriting options" mean `foods` would just be `'carrots'` in that example?

Should it even be valid for a parameter to be specified twice?  I think I'd expect an error in that case, because otherwise we're guessing at the user's intention, as I think both concatenation and overwriting are equally iffy.

jly36963:
I believe [my fix](https://github.com/yargs/yargs/pull/1992) was released, can you use yargs 17.3.1 and confirm that it solves the issue you reported?

The solution I chose is as follows:
- A user provides options and positionals under the same namespace
- If neither the option or the positional is array/variadic type, overwrite.
- If at least one is array/variadic, combine.

__NOTE__: The fix had a side-effect that concerns using array/variadic options/positionals in combination with `yargs.config()`. If you come across that behavior, I'm aware of it and currently working on a fix for it.

shadowspawn:
[...]

PR #1992, opened 2021-07-21
fix: positionals should not overwrite options

jly36963:
Addresses: https://github.com/yargs/yargs/issues/1637

## Problem

When executing the default command, if both positionals and options are provided under the same namespace, positionals will overwrite the options in argv. This doesn't happen if a command is provided.

#### Command provided

This will behave correctly. `argv.foods` will have both `apples` and `carrots`

```js
yargs
  .command({
    command: "cmd [foods..]",
    desc: "cmd desc",
    builder: (yargs) =>
      yargs.positional("foods", {
        desc: "foods desc",
        type: "string",
      }),
    handler: (argv) => {
      console.log(argv);
    },
  })
  .help()
  .parse("cmd --foods apples carrots");
```

logs

```
// first time through parser (kRunYargsParserAndExecuteCommands)
{ args: 'cmd --foods apples carrots', config: { 'populate--': true } }
{ 'parsed.argv': { _: [ 'cmd', 'carrots' ], foods: 'apples' } }

// second time through parser
{ args: 'cmd --foods apples carrots', config: { 'populate--': true } }
{ 'parsed.argv': { _: [ 'cmd' ], foods: [ 'apples', 'carrots' ] } }

// argv in handler
{
  _: [ 'cmd' ],
  foods: [ 'apples', 'carrots' ],
  '$0': 'test-issue.js'
}
```

#### Default command

This will behave incorrectly. `argv.foods` will only have `carrots`.

```js
yargs
  .command({
    command: "$0 [foods..]",
    desc: "default desc",
    builder: (yargs) =>
      yargs.positional("foods", {
        desc: "foods desc",
        type: "string",
      }),
    handler: (argv) => {
      console.log(argv);
    },
  })
  .help();
  .parse('--foods apples carrots')
```

logs

```
// first time through parser (kRunYargsParserAndExecuteCommands)
{ args: '--foods apples carrots', config: { 'populate--': true } }
{ 'parsed.argv': { _: [ 'carrots' ], foods: 'apples' } }

// second time through parser
{ args: '--foods apples carrots', config: { 'populate--': true } }
{ 'parsed.argv': { _: [ 'carrots' ], foods: 'apples' } }

// argv in handler
{
  _: [],
  foods: [ 'carrots' ],
  '$0': 'test-issue.js'
}
```

The logic in `postProcessPositionals` will cause the positionals (`['carrots']`) to override the option (`'apples'`) in argv.

## Solution

Check during `postProcessPositionals` if a particular key is already set on argv. If both positionals/options are provided and at least one is an array, combine rather than overwrite.

## Questions

When both positional and option are provided, what should be the behavior?
- if variadic, combine?
- if single value, overwrite?

(This is the assumption I made with this fix.)

### DIFF
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -598,13 +598,7 @@ export class CommandInstance {
           // any new aliases need to be placed in positionalMap, which
           // is used for validation.
           if (!positionalMap[key]) positionalMap[key] = parsed.argv[key];
-          // addresses: https://github.com/yargs/yargs/issues/1637
-          // don't overwrite if key already exists in argv
-          if (argv[key]) {
-            argv[key] = ([] as string[]).concat(argv[key], positionalMap[key]);
-          } else {
-            argv[key] = parsed.argv[key];
-          }
+          argv[key] = parsed.argv[key];
         }
       });
     }
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -598,7 +598,18 @@ export class CommandInstance {
           // any new aliases need to be placed in positionalMap, which
           // is used for validation.
           if (!positionalMap[key]) positionalMap[key] = parsed.argv[key];
-          argv[key] = parsed.argv[key];
+          // Addresses: https://github.com/yargs/yargs/issues/1637
+          // If both positionals/options provided,
+          // and if at least one is an array: don't overwrite, combine.
+          if (
+            argv[key] &&
+            parsed.argv[key] &&
+            (Array.isArray(argv[key]) || Array.isArray(parsed.argv[key]))
+          ) {
+            argv[key] = ([] as string[]).concat(argv[key], parsed.argv[key]);
+          } else {
+            argv[key] = parsed.argv[key];
+          }
         }
       });
     }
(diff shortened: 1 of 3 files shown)


bcoe:
> if single value, overwrite?

I think for this edge case, this assumption seems right to me. Someone writing a command line should avoid a collision like this, for arguments they define. So, this is something that will most likely come up in the edge-case of someone passing an unexpected command.

bcoe (approved):
This fix looks great to me 👍 thank you.

bcoe (approved):

foolip:
Hi all!

It looks like this was a breaking change, at least for how we used yargs in MDN's browser-compat-data. I've worked around it here, which touches all the relevant code: https://github.com/mdn/browser-compat-data/pull/11990

Note that I tried and failed to make only the last positional argument variadic, the default values are combined with the given value even then. This seems like a limitation in yargs now, albeit one that I could work around by not using an array type at all.

jly36963:
@foolip I believe I understand the problem -- I'll work on a fix tonight, and I'll keep you posted

Issue #1848, opened 2021-01-07
Array arguments causes 


gempain:
Hi there, and thanks for the great package, been using it for years now.

We've been having some issues lately with array options conflicting with command arguments in [@getmeli/cli](https://www.npmjs.com/package/@getmeli/cli). Basically:

```
npx @getmeli/cli upload \
    --url https://myurl.tld \
    --site mysiteid \
    --token sometoken \
    --branch "master" \
    ./public
```

throws

```
Not enough non-option arguments: got 0, need at least 1
```

but this works (moved `--branch` up one line):

```
npx @getmeli/cli upload \
    --url https://myurl.tld \
    --site mysiteid \
    --branch "master" \
    --token sometoken \
    ./public
```

The `branch` option is of type `array`, but AFAIK to add items to the array, you need to provide the option multiple times, so `./public` shouldn't be parsed in the array.

For background, we've been discussing this here [issue on our repository](https://github.com/getmeli/meli/issues/176).

bcoe:
Ordering of options should have no effect. Don't suppose you can isolate a minimal reproduction, without pulling in the whole `@getmeli/cli`?

gempain:
@bcoe apologies, I should've given more efforts into trying to provide a minimal reproduction example - thanks for bearing with me on this one !

Here it is (`yargs` 15.4.1):

```js
#!/usr/bin/env node

require('yargs')
  .scriptName('meli')
  .command({
    command: 'upload <directory>',
    builder: (args) => {
      args.option('url', {
        type: 'string',
        demandOption: true,
      })
      args.option('branch', {
        type: 'string',
        array: true,
      });
      return args;
    },
    handler: (args) => {
      console.log(args);
    },
  })
  .command({
    command: '*',
    handler: args => {
      // eslint-disable-next-line no-console
      console.error(`Unknown command "${args._ && args._.length > 0 ? args._[0] : ''}", see --help`);
      process.exit(1);
    },
  })
  // .recommendCommands() // not working with '*'
  .help()
  .parse(process.argv.slice(2));
```

This doesn't work:

```sh
$ ./tmp.js upload --url http://localhost:1234 --site 1234 --token token --branch branch ./tmp
meli upload <directory>

Upload a release

Options:
  --version     Show version number                                    [boolean]
  --help        Show help                                              [boolean]
  --url, -u     URL of the Meli API (e.g. https://meli.sh or
                https://meli.company.com)                    [string] [required]
  --branch, -b                                                           [array]

[...]

PR #1997, opened 2021-08-01
fix: array options should not eat positionals

jly36963:
Addresses: https://github.com/yargs/yargs/issues/1848 https://github.com/yargs/yargs/issues/2017

## Problem

I'm looking at the options that get passed with argv into the parser, and it looks like the parser is unaware of positionals.

## Questions

Possible solution:

- Send the number of expected (required) positional arguments to the parser, 
- Have the parser grab the expected args, before array options are handled.

This solution would have the following problems:

- This wouldn't resolve the ambiguity between array args and optional positional arguments.
- This wouldn't work if the positional args are variadic.

Alternative solutions:

- Output a warning when array options and positionals are used.
- Suggest setting 'greedy-arrays' setting to false.

Any ideas? @bcoe

### DIFF
--- a/test/command.cjs
+++ b/test/command.cjs
@@ -209,6 +209,42 @@ describe('Command', () => {
       argv['--'].should.eql(['apple', 'banana']);
       called.should.equal(true);
     });
+
+    // Addresses: https://github.com/yargs/yargs/issues/1848
+    it('is not stolen by array options', () => {
+      yargs
+        .command({
+          command: 'cmd <entree>',
+          desc: 'cmd desc',
+          builder: yargs =>
+            yargs
+              .positional('entree', {
+                desc: 'entree desc',
+                type: 'string',
+              })
+              .option('drink', {
+                desc: 'drink desc',
+                type: 'string',
+              })
+              .option('desserts', {
+                desc: 'desserts desc',
+                type: 'string',
+                array: 'true',
+              })
+              .fail(() => {
+                expect.fail();
+              }),
+          handler: argv => {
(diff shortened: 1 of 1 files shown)


bcoe:
@jly36963 I'm not sure of the best solution for this problem, will need to page the edge case back into memory.

In hindsight, I've never been a big fan of the array type, which seems too magical. Better that you provide arguments like `--arr foo,bar,snuh`, or `--arr one --arr two`.

> Suggest setting 'greedy-arrays' setting to false

I'm tempted to make this the default behavior when we eventually move to `v18` of yargs ... if there's a tweak that an make behavior a bit more intuitive in the interim, I'm all ears.

bcoe:
@jly36963 just coming back to this. I honestly lean towards a warning, with the recommendation that you set greedy-arrays to false, or handle your own array parsing with an argument like `--arr="foo,bar,apple"`.

I'd be open to adding support for "foo,bar,apple" as a built in feature, perhaps this is another resolution? What if there was an option `greedy-arrays=delimited`?

bcoe:
perhaps we can suggest some of these options to the linked bugs?

jly36963:
I anticipate that an extensive amount of work is required to address this bug, as it would require rethinking how yargs-parser handles combinations of options/positionals/defaults/config/etc. 

I don't want to put a lot of time into this, just to find out that I've broken existing functionality or that I've made changes that don't align with the yargs maintainers' vision for the project.

If at some point in the future I find a more straightforward solution, or have someone to work on this with, I might pick it up again

Issue #1952, opened 2021-05-24
Conflicts validations being skipped with parser configuration skip-dashed


libinvarghese:
When conflicts are set for a 2 (or more) worded option, along with parser configuration  `skip-dashed` enabled, parse does not throw error when both the conflicting arguments are provided.

It works perfectly, if `skip-dashed` is not enabled.

Please find the code demonstrating the issue at [repl.it](https://replit.com/@libin_v/bug-yargs-conflicts-skip-dashedSHARED)

libinvarghese:
Updated [repl.it link](https://replit.com/@libin_v/SHARED-bug-yargs-conflicts-skip-dashed)

Also updated in description.

bcoe:
@libinvarghese will be fixed by @jly36963's work going out in next version of yargs.

PR #1998, opened 2021-08-04
fix: conflicts and strip-dashed

jly36963:
Addresses: https://github.com/yargs/yargs/issues/1952

## Problem

When strip-dashed is true, the kebab case arguments are removed from argv.
When looking for conflicts, the result is a false negative. The conflicts are in kebab case while the argv arguments are in camel case, so the conflicts logic will never match the conflicting arguments.

## Solution

When strip dashed is true, compare camelCase versions of conflicting against argv, to address the disparity between argv (camel) and conflicting (kebab).

## Notes

*This is important; please read*

I didn't want to add any extra dependencies, so I made a very lightweight implementation of camelCase. 
I borrowed heavily from lodash's source code. Let me know if I need to add proper credit, change the solution to avoid copying someone else's code, add unit tests, etc.


lib/utils/camel-case.ts
```diff
@@ -0,0 +1,19 @@
+export function camelCase(s: string) {
+  if (!s) return '';
+  // split camel case, then match ascii words
+  const words = asciiWords(s.replace(/([a-z0-9])([A-Z])/g, '$1 $2'));
+  return words.length > 1
```
bcoe, line 5: I think you should have access to camelCase already through the yargs-parser dependency:

https://github.com/yargs/yargs-parser/blob/main/lib/string-utils.ts#L7

I believe I already use one or two methods off of it elsewhere.

bcoe (changes requested):

bcoe (approved):

PR #2013, opened 2021-08-22
fix: avoid legacy accessors

bmeck:
[These are deprecated and might not exist in every environment.](https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-object.prototype.__defineGetter__) This should also mitigate some false positives as well in object traversals from dynamic property lookups and prototype pollution.


bcoe:
@bmeck thank you for the patch 👍

bcoe (changes requested):
Looks to break some tests, any thoughts?

index.cjs
```diff
@@ -28,15 +28,23 @@ function singletonify(inst) {
...
+      Object.defineProperty(Argv, '$0', {
+        configurable: true,
+        enumerable: true,
+        get() {
+          return inst.$0;
+        }
       });
-      Argv.__defineGetter__('parsed', () => {
-        return inst.parsed;
+      Object.defineProperty(Argv, '$0', {
```
bmeck, line 42:
```suggestion
      Object.defineProperty(Argv, 'parsed', {
```

index.cjs
```diff
@@ -15,6 +15,26 @@ function Argv(processArgs, cwd) {
...
+    enumerable: true,
+    get: getter,
+  });
+}
+function lookupGetter(obj, key) {
+  while (obj) {
+    const desc = Object.getOwnPropertyDescriptor(obj, key);
+    if (typeof desc !== 'undefined') {
+      if (desc.get) {
+        return desc.get;
```
bcoe, line 30: @bmeck can we replace this with:

```js
function lookupGetter(obj, key) {
  const desc = Object.getOwnPropertyDescriptor(obj, key);
  if (typeof desc !== 'undefined') {
    return desc.get;
  }
}
```

So that coverage remains at 100%? I'm assuming that `desc.get` will either be the getter function, or `undefined`?

bcoe:
@bmeck friendly ping on this one.

index.cjs
bmeck: I did a pass to see if it was possible to go through a proto and it only occurs on extremely exotic and tailored Proxy stuff so this replacement should be fine due to it being invoked against the keys/own prop keys.

bcoe (approved):

bcoe:
@bmeck thank you for the contribution.

PR #2018, opened 2021-08-31
feat: autocomplete choices for options

### LINKED ISSUE #2016: TypeScript 4.4 breaks building `yargs`
Hi!

TypeScript 4.4 was released 4 days ago and [introduced](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables) the following:
```typescript
try {
    executeSomeThirdPartyCode();
}
catch (err) { // err: unknown

    // Error! Property 'message' does not exist on type 'unknown'.
    console.error(err.message);

    // Works! We can narrow 'err' from 'unknown' to 'Error'.
    if (err instanceof Error) {
        console.error(err.message);
    }
}
```

This breaks the build of `yargs` (there is no `package-lock.json`, which seems to be on purpose):
```
lib/argsert.ts:76:18 - error TS2571: Object is of type 'unknown'.

76     console.warn(err.stack);
                    ~~~

lib/utils/maybe-async-result.ts:22:25 - error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'Error'.

22     return errorHandler(err);
                           ~~~

Found 2 errors.
```

ansgarm:
Hi!

This PR introduces support for auto completing choices if an option has been specified which has choices but no value has been selected yet. For example:
```
> cdktf convert --language <TAB>
csharp      java        python      typescript
```
when the option has been specified like this: ([source](https://github.com/hashicorp/terraform-cdk/blob/168349149773a969d64063a38d3a69a6d75680cb/packages/cdktf-cli/bin/cmds/convert.ts#L42-L45))
```typescript
.option("language", {
  choices: ["typescript", "python", "csharp", "java"],
  default: "typescript",
})
```

#### Background
We use `yargs` for the [CDK for Terraform](https://github.com/hashicorp/terraform-cdk) and wanted to [introduce support for shell completion](https://github.com/hashicorp/terraform-cdk/issues/676). As we wanted to offer support for auto completing choices for options, I figured I'd do a PR instead of building a custom completion handler in multiple places.

If there's anything I can do to get this merged, please tell me!

P.S. It also fixes #2016 because that caused the build to fail for me and there's no `package-lock.json` or alike.

### DIFF
--- a/lib/argsert.ts
+++ b/lib/argsert.ts
@@ -73,7 +73,7 @@ export function argsert(
       position += 1;
     });
   } catch (err) {
-    console.warn(err.stack);
+    console.warn((err as Error).stack);
   }
 }
 
--- a/lib/completion.ts
+++ b/lib/completion.ts
@@ -68,6 +68,7 @@ export class Completion implements CompletionInstance {
 
     this.commandCompletions(completions, args, current);
     this.optionCompletions(completions, args, argv, current);
+    this.choicesCompletions(completions, args, argv, current);
     done(null, completions);
   }
 
@@ -82,7 +83,8 @@ export class Completion implements CompletionInstance {
       .getContext().commands;
     if (
       !current.match(/^-/) &&
-      parentCommands[parentCommands.length - 1] !== current
+      parentCommands[parentCommands.length - 1] !== current &&
+      !this.previousArgHasChoices(args)
     ) {
       this.usage.getCommands().forEach(usageCommand => {
         const commandName = parseCommand(usageCommand[0]).cmd;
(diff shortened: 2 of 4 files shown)


bcoe (approved):
This seems like a great addition to completion functionality 👍 and I appreciate the fix for TypeScript.

lib/argsert.ts
```diff
@@ -73,7 +73,7 @@ export function argsert(
       position += 1;
     });
   } catch (err) {
-    console.warn(err.stack);
+    console.warn((err as Error).stack);
```
rvitaliy, line 76: 👋 hi all, here my 50cent, sorry!

The scope of usage `unknown` instead of `any` is to prevent fatal errors when catch receive something what is not an `Error`.

`"errorString".stack` > fatal error
to avoid this i suggest to add a check if `err` is instance of `Error` like suggested by documentation instead of cast `err as Error`

https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables

PR #2028, opened 2021-09-09
feat(usage): output the environment variable name

mabels:
This is not tested and  or complete-:) 

I had the idea to display the environment name for an option.
To help the "dockerized" people to understand/help that they could pass environment variables.

So I did this spike. 
Missing is that, it should only printed if envPrefix is set, I didn't find if envPrefix is passed to usage.ts.
And some commands like --help should currently not work if HELP=xx is set.
And my replace should make use of the real converting function:
UnknownFunction("outputSchema")  -> OUTPUT_SCHEMA
which  i didn't found.

currently help looks like this:

Options:
  --help, [$HELP]                                                Show help                                                  [boolean]
  --version, [$VERSION]                                     Show version number                              [boolean]
  --outputSchema, [$OUTPUT_SCHEMA]        output Schema                  [string] [default: "tbd"]

### DIFF
--- a/lib/usage.ts
+++ b/lib/usage.ts
@@ -338,7 +338,7 @@ export function usage(yargs: YargsInstance, shim: PlatformShim) {
         // actually generate the switches string --foo, -f, --bar.
         const switches: Dictionary<string | IndentedText> =
           normalizedKeys.reduce((acc, key) => {
-            acc[key] = [key]
+            const my = [key]
               .concat(options.alias[key] || [])
               .map(sw => {
                 // for the special positional group don't
@@ -366,8 +366,10 @@ export function usage(yargs: YargsInstance, shim: PlatformShim) {
                   ? 1
                   : -1
               )
-              .join(', ');
-
+            my.push(...my.filter(i => i.startsWith("--")).map(i => {
+                 return "[$" + i.slice(2).replace(/([A-Z])/g, '_$1').toUpperCase() + ']';
+            }));
+            acc[key] = my.join(', ');
             return acc;
           }, {} as Dictionary<string>);
 


bcoe:
@mabels I'm not immediately opposed to this feature, but:

1. I'd be interested to see a few examples of CLI applications that display this type of information.
2. I think we'd only want to enable this feature if you have enabled the `.env()` setting.
3. It seems weird to show the environment variable option for flags like `--help` and `--version` (which is what's causing tests to fail today).

If we can't find examples of `1.`, I'm hesitant to add the feature to yargs.

mabels:
1. Which examples you refering to? I provided a sample output.
2. that's a very good idea to enable this only if env is set.
3. I'm agree help und version should exclude from this....but how we identify these excluded flags.

bcoe:
@mabels re: `1.`, is there a common CLI tool (`curl`, `git`, `node`), etc., that displays supported environment variables.

I notice Node.js lists them at them at the bottom:

```bash
Environment variables:
FORCE_COLOR                   when set to 'true', 1, 2, 3, or an empty string causes
                              NO_COLOR and NODE_DISABLE_COLORS to be ignored.
NODE_DEBUG                    ','-separated list of core modules that should print
                              debug information
NODE_DEBUG_NATIVE             ','-separated list of C++ core debug categories that
                              should print debug output
```.

I think rather than show an environment variable with each flag, we could make it so that when you call `.env`, we show:

```bash
Environment variables:

MY_PREFIX_FOO    Hello.
MY_PREFIX_FOO    Goodbye.
```

mabels:
That's a perfect idea, to split the env-vars from the options.
I would scatch this changed out in this PR.
If somebody could point me to these information please:
- how to pass the .env() builder state to "help output function"?
- where is the function to translate from the option "--x-y-z" -> $X_Y_Z, is conversion is somewhere happing in the current code but i didn't found it.

thx in advance

bcoe:
@mabels you can check if `envPrefix` has been set on the `options` returned by `yargs.getOptions`, and then add logic [here](https://github.com/yargs/yargs/blob/main/lib/usage.ts#L452) similar to what already exists for `Options:` and `Commands:`.

bcoe:
@mabels closing this for now, due to inactivity, please feel free to revisit in the future, as I feel this is a good feature request :+1: 

If you don't have the cycles to implement yourself, please go ahead and open an issue for the feature request, and someone else might be able to pick it up.

PR #2029, opened 2021-09-13
Upgrade string-width

adam2k:
Related to https://github.com/chalk/ansi-regex/commit/8d1d7cdb586269882c4bdc1b7325d0c58c8f76f9

Additional information CWE-400 https://cwe.mitre.org/data/definitions/400.html

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "escalade": "^3.1.1",
     "get-caller-file": "^2.0.5",
     "require-directory": "^2.1.1",
-    "string-width": "^4.2.0",
+    "string-width": "^5.0.1",
     "y18n": "^5.0.5",
     "yargs-parser": "^20.2.2"
   },


adam2k:
@bcoe or @substack can I get maintainer approval to run workflows?  I haven't contributed to this repo yet, so let me know if there is some other way I should request access.  Thanks!

ljharb:
Noting: string-width v5 requires node 12 while v4 requires node 8, luckily this isn't a breaking change for yargs since it already requires node 12.

However, string-width v5 is ESM-only, which means it can't be `require`d, which means yargs likely can't use it.

I'd suggest finding an alternative dependency that isn't ESM-only and also lacks the CVE.

bcoe:
@ljharb @adam2k the ESM version of yargs already dropped using `string-width`, ironically because at the time `string-width` wasn't ESM.

The alternate implementation is [here](https://github.com/yargs/yargs/blob/master/lib/platform-shims/esm.mjs#L61).

I suggest we just switch to this approach for both ESM and CJS, the downside is that string width will be slightly off for some unicode characters I believe (_am I correct in my understanding that some unicode characters are two ascii characters wide?_).

Either way, I think this is a minor regression, and worth dropping a dep.

ljharb:
@bcoe that alternate implementation is quite wrong for many things - the spread operator only splits to code points, but many emojis are multiple code points - `'🏳️‍🌈'` for example, has a length of 6 and contains 4 code points.

A proper replacement would likely require using `Intl.Segmenter` and splitting by grapheme clusters, and then checking the resulting array length.

woehr:
The path of least resistance here would probably be to see whether strip-ansi will backport a dependency bump to address the CVE and then to get string-width 4.2.2 to update the strip-ansi dependency. I've asked whether strip-ansi can do a hotfix here: https://github.com/chalk/strip-ansi/issues/40.

bcoe:
@Qix- kindly back-ported a fix to `ansi-regex@5.0.1`, so we should be good now 👍 

https://github.com/chalk/ansi-regex/issues/38

PR #2046, opened 2021-10-12
fix: don't fail if "fileURLToPath(import.meta.url)" throws

Congelli501:
For usage like:

 - built final project used in a javascript vm
 - project built on unix via webpack and used on windows (path format are incompatible with drive letters on windows)

https://github.com/webpack/webpack/issues/14445

### DIFF
--- a/lib/platform-shims/esm.mjs
+++ b/lib/platform-shims/esm.mjs
@@ -15,8 +15,13 @@ import y18n from 'y18n'
 const REQUIRE_ERROR = 'require is not supported by ESM'
 const REQUIRE_DIRECTORY_ERROR = 'loading a directory of commands is not supported yet for ESM'
 
-const mainFilename = fileURLToPath(import.meta.url).split('node_modules')[0]
-const __dirname = fileURLToPath(import.meta.url)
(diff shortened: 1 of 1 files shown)


lib/platform-shims/esm.mjs
```diff
@@ -15,8 +15,13 @@ import y18n from 'y18n'
 const REQUIRE_ERROR = 'require is not supported by ESM'
 const REQUIRE_DIRECTORY_ERROR = 'loading a directory of commands is not supported yet for ESM'
 
-const mainFilename = fileURLToPath(import.meta.url).split('node_modules')[0]
-const __dirname = fileURLToPath(import.meta.url)
+let __dirname;
+try {
+  __dirname = fileURLToPath(import.meta.url);
+} catch (e) {
```
bcoe, line 21: Can you think of a way to add a test for this?

bcoe (changes requested):
I'd love to figure out how to add a test.

bcoe (approved):

bcoe:
@Congelli501 thank you for the contribution.

PR #2052, opened 2021-10-25
Fix dependency vulnerability

MelonBall:
Fix vulnerability with string-width > ansi-regex 

https://github.com/advisories/GHSA-93q8-gq69-wqmw

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "escalade": "^3.1.1",
     "get-caller-file": "^2.0.5",
     "require-directory": "^2.1.1",
-    "string-width": "^4.2.3",
+    "string-width": "^5.0.1",
     "y18n": "^5.0.5",
     "yargs-parser": "^20.2.2"
   },


mecies:
Looking forward for this to be updated 🤗

bcoe:
Merging this would break CommonJS, because `string-width@5` is ESM only.

bcoe:
I'm not sure what the best solution is, because pulling in an ESM dependency is asynchronous, and yargs is unfortuantely synchronous 🤔 

Perhaps we could just break layout for unicode characters in yargs' CJS version.

bcoe:
@MelonBall actually, it looks like the patched version of ansi-regex comes in with yargs@4.2.3:

```
└─┬ yargs@17.2.1
  ├─┬ cliui@7.0.4
  │ ├── string-width@4.2.3 deduped
  │ ├─┬ strip-ansi@6.0.1
  │ │ └── ansi-regex@5.0.1
```

@mecies, @MelonBall are you potentially pinned on an old version of the dependency via a `package-lock.json`?

mecies:
> @mecies, @MelonBall are you potentially pinned on an old version of the dependency via a `package-lock.json`?

I'm relying on a tool called Snyk to explore vulnerabilities in packages but after I explored the lockfile it looks like it points to the ansi-regex@5.0.1 so maybe the report is false on my end 🤔 
I will reach out to someone responsible for the security in my company and get back to you 👍

lirantal:
Thanks folks, looking into it on the Snyk side of things and will get back to you 👀

lirantal:
@mecies @bcoe are you still receiving this issue?
we're seeing that in the Snyk vulnerability database 5.0.1 is marked as the fixed version indeed: https://security.snyk.io/vuln/SNYK-JS-ANSIREGEX-1583908

mecies:
> @mecies @bcoe are you still receiving this issue? we're seeing that in the Snyk vulnerability database 5.0.1 is marked as the fixed version indeed: https://security.snyk.io/vuln/SNYK-JS-ANSIREGEX-1583908

Because I was in rush, I replaced the `yargs` package with `minimist` for my use case but I tested today, and when I do a clean install of `yargs` package, it still reports the vulnerability for me

mecies:
> If you can actually provide your lockfile and project manifest, or a link to an open source repo that is having this issue I can take a further look

I will reach out internally for permission to share it but I doubt that I will get it, but I'm pretty sure it's not somewhere else in the tree because only my PR which introduced the package was causing the vulnerability to appear, and removing the package removes the issue.

MelonBall:
What I have in my audit, if it can help :
image

bcoe:
@MelonBall it looks like you are getting an old version of yargs through `node-sass`, which hasn't used yargs for a few years it would appear:

https://github.com/sass/node-sass/pull/474

I think if you upgrade to the newest version of `node-sass` you should be in a healthy place.

@lirantal thank you for the help.

PR #2056, opened 2021-11-03
feat: fallback to default bash completion

Nevon:
If you do not generate a completion entry, use the default bash completion. If that also does not return a completion, fall back to "readline" default filename completion.

This allows you to complete on for example environment variables. `mycommand --option $OPT<tab>` would complete to `mycommand --option $OPTION` if there was an environment variable called `OPTION` set.

### DIFF
--- a/lib/completion-templates.ts
+++ b/lib/completion-templates.ts
@@ -24,7 +24,7 @@ _{{app_name}}_yargs_completions()
 
     return 0
 }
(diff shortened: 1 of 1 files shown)


bcoe (approved):

bcoe:
@Nevon this seems like a good improvement, have you tested out the behavior (I don't think we have great unit tests for completion, it's hard to test).

PR #2071, opened 2021-11-24
build: fix ts build errors

jly36963:
## Problem

I'm running into a few issues when running `build:cjs`:
- It looks like the ts logic for `Promise.all` has changed. I made a change to address this. (I used [this solution](https://github.com/microsoft/TypeScript/issues/46651#issuecomment-958599609))
- `@wessberg/rollup-plugin-ts` is throwing an error

#### error

```
TypeError: Cannot read properties of undefined (reading 'text')
    at getExportedSymbolFromExportSpecifier (/.../yargs/node_modules/@wessberg/rollup-plugin-ts/src/service/transformer/declaration-bundler/util/create-export-specifier-from-name-and-modifiers.ts:49:35)
```

#### reproduction

You can see a reproduction of the issue [here](https://github.com/yargs/yargs/runs/4316963468?check_suite_focus=true#step:5:34)

#### code

The issue is in [this line](https://github.com/wessberg/rollup-plugin-ts/blob/master/src/service/transformer/declaration-bundler/util/create-export-specifier-from-name-and-modifiers.ts#L50)

## Solution

- Use `rollup-plugin-ts` instead. 
- Install later version of npm for CI, because It wants npm >= 7.0.0, and the CI process was using npm 6.14.15.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,6 +16,7 @@ jobs:
       - uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node }}
+      - run: npm install -g npm@latest
       - run: node --version
       - run: npm install --engine-strict
       - run: npm test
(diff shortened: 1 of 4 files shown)


bcoe (approved):

bcoe:
@jly36963 thank you :clap:

Issue #2069, opened 2021-11-23
options parameter of callback function for check method is always undefined


iddofontijn:
According to https://yargs.js.org/docs/#api-reference-checkfn-globaltrue the callback function of `.check()` has an `options` parameter. In my experience, in version 16 it contained the available options for the specified command. But in version 17 it always returns `undefined`.

shadowspawn:
This was fixed in #2079 (thanks @jly36963).

PR #2079, opened 2021-11-26
fix: re-add options to check callback

jly36963:
Addresses: 

https://github.com/yargs/yargs/issues/2069

## Problem

Check callback `options` is undefined. It should be an object.

The [docs](http://yargs.js.org/docs/#api-reference-checkfn-globaltrue) show `options` in the call signature for the callback in`check`. In [yargs 17](https://github.com/yargs/yargs/blob/main/lib/yargs-factory.ts#L296), options is missing from that call signature.

## Questions

- Not sure if the options object I used is the correct one (or if it's in the correct state). 
- Not sure if mutable access to options should be exposed. Should it be read-only or a copy?

Let me know if there are any changes that need to happen, or if I should add more tests.

### DIFF
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -293,7 +293,10 @@ export class YargsInstance {
     this[kTrackManuallySetKeys](keys);
     return this;
   }
-  check(f: (argv: Arguments) => any, global?: boolean): YargsInstance {
+  check(
+    f: (argv: Arguments, options: Options) => any,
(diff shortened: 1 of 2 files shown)


bcoe (approved):
Looks like a reasonable fix to me, not sure why we dropped this behavior when moving to `v17`.

bcoe:
@jly36963 thank you for the contribution 👍

PR #2083, opened 2021-12-02
multiple: improved completion for choices

gmahomarf:
This is an improvement on #2018. Changes:

- Improvement: Completion for choices will now work for all possible aliases of an option and not just the default long option. This is accomplished by iterating over all known aliases of the provided option, and checking for `choices` on each.
- Fix: changed the RegExp used to remove hypens (`-`), because it removed ALL hyphens from the argument, which broke options like `--long-option-with-dashes`. It now only removes dashes at the beginning of the option.
- Fix: changed the check for option arguments to match options that begin with `'-'`, instead of `'--'`, to include short options

### DIFF
--- a/lib/completion.ts
+++ b/lib/completion.ts
@@ -153,21 +153,33 @@ export class Completion implements CompletionInstance {
     let previousArg = args[args.length - 1];
     let filter = '';
     // use second to last argument if the last one is not an option starting with --
-    if (!previousArg.startsWith('--') && args.length > 1) {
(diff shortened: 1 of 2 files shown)


bcoe:
@gmahomarf thank you for the patch, and appologies for slow review, will work on getting a release of yargs out over the holiday break hopefully.

bcoe (approved):

bcoe:
@gmahomarf thank you for the contribution :clap:

Issue #2085, opened 2021-12-06
`showHelpOnFail` does not show the message when set to false


neha-sri:
From the docs: 
```javascript
showHelpOnFail(enable: boolean, message?: string): Argv<T>;
```
``` text
By default, yargs outputs a usage string if any error is detected. Use the .showHelpOnFail() method to customize this behavior.
Params:
enable – If false, the usage string is not output.
message – Message that is output after the error message.
```

We are using it in the following manner:

```javascript
export const builder = function (yargs: yargs.Argv): yargs.Argv {
  return yargs
    .commandDir('feedback-commands', {
      extensions: ['js'],
    })
    .strict()
    .showHelpOnFail(false, `Use ' --help' for usage`)
    .demandCommand();
};
```
However, the string passed in the parameter `message` is never displayed. Only the error message `Not enough non-option arguments: got 0, need at least 1` is being displayed. 

To be clear, we want to not display the help page but print a string after the error message instead.

PS: The string also doesn't get displayed when the parameter value is set true since `--help` output instead.

We are on version `17.0.1` but  I tried with `17.0.3` as well and same result.

jly36963:
@neha-sri 

`showHelpOnFail` doesn't display the message if it is chained onto yargs globally. It works as expected if chained onto the command builder. 

In the example below, only the `showHelpOnFail` that is chained onto the `cmd1` builder has any effect.

```js
yargs('cmd1 --opt1 hello --extraOpt oops')
  .command(
    'cmd1',
    'cmd1 desc',
    yargs => yargs
      .option('opt1', {
        alias: 'o',
        demandOption: true,
      })
      .showHelpOnFail(false, "What have you done????"), // this one works
    argv => console.log(argv)
  )
  .showHelpOnFail(false, "What have you done????") // this one does nothing
  .strict()
  .parse()
```

This is the output from the above example:

```
Unknown argument: extraOpt

What have you done????
```

bcoe:
The discrepancy between the global configuration of `showHelpOnFail` and the command builder feels a bit weird, but since there's a workaround, I advocate calling this a `p2`.

bcoe:
fixed in #2154, which will go out soon.

PR #2087, opened 2021-12-08
fix: preserve failMessage from showHelpOnFail

jly36963:
Addresses: https://github.com/yargs/yargs/issues/2085

## Problem

`usage.reset()` was setting `failMessage` to null, and as a result the message wasn't shown on fail.

## Solution

I removed the line in `usage.reset()` that set the failMessage to null. 

## Questions

Was there a reason for setting failMessage to null that I'm not aware of?

## Todo

Assuming this is an okay fix, I'll add unit tests

### DIFF
--- a/lib/usage.ts
+++ b/lib/usage.ts
@@ -26,10 +26,8 @@ export function usage(yargs: YargsInstance, shim: PlatformShim) {
     arg1: boolean | string = true,
     arg2?: string
   ) {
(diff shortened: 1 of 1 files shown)


jly36963:
I realized this PR might be unnecessary, and I'm nervous of the side effects associated with my solution. 

I proposed a solution to the issue mentioned [here](https://github.com/yargs/yargs/issues/2085#issuecomment-989479907)

Issue #2033, opened 2021-09-21
Argument Parsing with spaces?


dacbd:
I'm not sure if is an edge case that yargs should address.

When a program is executed from a script like `bash -e /script_run_yargs_program.sh` ie GitHub actions `run:` if there are any trailing spaces after a slash on a multi-line script they get interrupted as an argument and you get a very unhelpful error message

```
Unknown argument:  
```
being akin to `printf("Unknown argument: %s" , " ")`

see: https://github.com/iterative/cml/issues/727#issuecomment-924443434 for an example

I realize this could be an unsupported edge case and Ideally people aren't putting trailing spaces after their \'s 😉

0x2b3bfa0:
It would be nice to quote the unknown options so people can notice whitespace:

```console
Unknown argument: ' '
```

bcoe:
@DanielBarnes, I like @0x2b3bfa0's suggestion... I don't thin kit would be ridiculous to `trim` the strings we parse too, I'd have to look at the parser to see how painful this is.

0x2b3bfa0:
Trimming the parsed strings sounds like a really unexpected behavior for some use cases. 🤔  To my mind, it should be at most a non–default [`parserConfiguration`](https://yargs.js.org/docs/#api-reference-parserconfigurationobj) option.

I can't imagine how puzzling would be if somebody tried to implement a whitespace–sensitive tool like [`tr(1)`](https://man7.org/linux/man-pages/man1/tr.1.html) and invocations like `tr --delete " "` were altered by the parser unexplainably.

bcoe:
@DanielBarnes would you be okay with @0x2b3bfa0's suggestion of quoting the error? vs., trimming.

DanielBarnes:
> @DanielBarnes would you be okay with @0x2b3bfa0's suggestion of quoting the error? vs., trimming.

Yes, I agree that quoting would be a better path. Less/no risk of introducing new behaviors to the existing API.

jly36963:
I want to work on a fix for this, but I'm having trouble reproducing this error:
```
Unknown argument:  
```

If I run a command with a space between `\` and a newline, I can reproduce this error:
```
[...] command not found
```

## Reproduction

This is how I tried to reproduce the error

### Code

```js
// multiline.js

const yargs = require('yargs')

let input

yargs(input)
  .command(
    'cmd1 [pos1]',
    'cmd1 description',
    yargs => yargs
      .option('opt1', { type: 'string' })
[...]

PR #2092, opened 2021-12-13
fix: wrap unknown args in quotes

jly36963:
Addresses:
https://github.com/yargs/yargs/issues/2033

## Problem

If an unknown argument is only whitespace, it causes confusing output: `Unknown argument:  `

### Code

```js
// multiline.js

yargs(process.argv.slice(2))
  .command(
    '$0',
    'default description',
    yargs => yargs
      .option('opt1', { type: 'string' })
      .option('opt2', { type: 'string' }),
    argv => {
      console.log(argv)
    }
  )
  .strict()
  .parse()
```

```
# multiline.sh (version 1)

# space after backslash on opt1 line
node multiline.js \
--opt1=hello \   
--opt2=world
```

```
# multiline.sh (version 2)

# no space after last baskslash on opt1 line
node multiline.js \
--opt1=hello \   \
--opt2=world
```

### Explanation

Calling multiline.js from multiline.sh (version 1) will result in an error like `command not found`. This makes sense, as having a space between the backslash and newline will prevent the newline from being escaped, and the newline would cause the next line to be interpreted as a new command.

Calling multiline.js  from multiline.sh (version 2) will result in the `unknown argument: ` error. I added a `.map()` call to the unknown arguments that will wrap them in quotes. This will make the error described in the issue a bit easier to understand: `Unknown argument: " "`

I don't know what might cause the situation in version 2. The issue specifically mentioned the error happening in a multiline `run: |` command in github actions. Maybe there's some process (like something in the github actions runner) that adds backslashes (where missing) at the end of each line within a multiline command. If something like this is happening, then extra backslashes would be added to the end of lines with trailing whitespace. I only have suspicions, and I feel like investigating it is outside the scope of this fix.

## Solution

The solution mentioned in the comments section of the issue was to wrap arguments listed in the usage output in quotes, rather than trimming/mutating the elements of argv.

## Questions

For the usage output:
- Should I wrap all unknown arguments in quotes? Only whitespace ones?
- Should I modify all similar usage output to also use quotes? (`missing required argument: %s`, etc)

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -7,7 +7,7 @@ jobs:
   release-please:
     runs-on: ubuntu-latest
     steps:
-      - uses: google-github-actions/release-please-action@a4578e0d8575662554aa01d055f7f762c9e3f8bc
+      - uses: google-github-actions/release-please-action@bfd488a47b78a5ba7b51dda30bbd261b4059162b
         id: release
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### [17.3.1](https://github.com/yargs/yargs/compare/v17.3.0...v17.3.1) (2021-12-23)
+
+
+### Bug Fixes
+
+* **translations:** correct Korean translation ([#2095](https://github.com/yargs/yargs/issues/2095)) ([c7c2b9e](https://github.com/yargs/yargs/commit/c7c2b9eb340754ddac7bdd1687c7951332c5ebba))
+
 ## [17.3.0](https://github.com/yargs/yargs/compare/v17.2.1...v17.3.0) (2021-11-30)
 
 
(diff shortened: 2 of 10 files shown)


bcoe:
@jly36963 If I'm understanding correclty, the problem is the form:

```
foo \[space]
--bar\
--final
```

And this would work?

```
foo \
--bar\
--final
```

I like your approach of making this jump out better to people.

jly36963:
@bcoe 

This causes `command not found`.
```
# newline is NOT escaped, --bar is interpreted as new command
foo \[space]
--bar\
--final
```

This causes `unknown argument: `
```
# newline is escaped, space is interpreted as an argument
foo \[space]\
--bar\
--final
```

(I think the second scenario would only happen if there is some process (before execution) adding backslashes at the end of lines that don't have them.)

jly36963:
As I keep thinking about this, I'm more in favor of only wrapping with quotes if the argument _only_ contains whitespace. What are your thoughts?

jly36963:
@bcoe 
To be consistent with different kinds of error messages, should I modify all similar usage output to also use quotes? (missing required argument: %s, etc)

It wouldn't take very long

bcoe:
@jly36963 sounds like consistentcy would be worthwhile, did you want to do so in this PR, or merge this PR and have some follow up work?

jly36963:
@bcoe 
I can definitely do it in this PR, it should go pretty quick

jly36963:
@bcoe 
I'm happy with where it's at now. I took the minimalist approach: If an argument is just whitespace, it's wrapped in quotes; otherwise, it's left alone. I only did it for arguments, because I don't think there is a way to make `' '` an unknown command.

Issue #2096, opened 2021-12-14
variadic positional arguments are duplicated if specified through config


aol-nnov:
minimal example

```js
const yargs = require('yargs/yargs');

yargs(process.argv.slice(2))
    .config(
        {
            "sourceIdx": "source",
            "binIdx": [
                "bin1-file",
                "bin2-file"
            ]
        }
    )
    .command('$0 <packageToSeed> [sourceIdx] [binIdx..]',
        'cmd description',
        {},
        (argv) => console.log(argv))
    .argv;
```
invocation: `node ./minimal.js qwe`

actual output:
```js
{
  _: [],
  sourceIdx: 'source',
  binIdx: [ 'bin1-file', 'bin2-file', 'bin1-file', 'bin2-file' ],
  '$0': 'minimal.js',
  packageToSeed: 'qwe',
  'package-to-seed': 'qwe',
  'bin-idx': [ 'bin1-file', 'bin2-file' ]
}
```

expected output:
```js
{
  _: [],
  sourceIdx: 'source',
  binIdx: [ 'bin1-file', 'bin2-file' ],
  '$0': 'minimal.js',
  packageToSeed: 'qwe',
  'package-to-seed': 'qwe',
  'bin-idx': [ 'bin1-file', 'bin2-file' ]
}
```

Can't help but notice that dashed variant of an array contains correct values.

jly36963:
Thanks for the bug report! I'll work on a fix for this

bcoe:
@jly36963 the fix in #2100 seems to solve the issue when `combine-arrays` is `false`, but behaves a bit oddly when it's `true`. It seems that veriadic arguments are combined with the array defined in the config, but not arguments passed as dash arguments.

jly36963:
@bcoe Looking now  👀

jy95:
@jly36963 I got a similar issue reported by a Chinese user on my [CLI tool](https://github.com/jy95/i18n-tools) - can you investigate please ? 

Minimal example to reproduce : 
```js
const yargs = require('yargs/yargs');

yargs(process.argv.slice(2))
    .config(
        {
            "columns":[
                {"locale":"en","label":"en"},
                {"locale":"zh-CN","label":"zh-CN"}
            ],
            "files":{
                "en":"/translations/en/translation.json",
                "zh-CN":"/translations/zh-CN/translation.json"
            },
            "filename":"translation",
            "outputDir":"/Users/hujin/Projects/socialFi/alienswap-app/temp"
[...]

PR #2100, opened 2021-12-19
fix: passed arguments should take precedence over values in config

jly36963:
Relates to: 

issue: https://github.com/yargs/yargs/issues/2096
PR: https://github.com/yargs/yargs/pull/1992
PR: https://github.com/yargs/yargs/pull/2006

## Problem

There were a few problems with pull 1992:

- Positional arguments combined with defaults (solved by 2006)
- Positional arguments combined with config objects (solved by this PR)
- Not accounting for kebab case arguments (solved by this PR)

## Questions

- Are there any other sources of values that I'm not accounting for? (I would rather not have another follow up PR for a third source 🤦 )
- Is there a better way to write the logic? It works, but I can't help but think there's a more elegant solution.

### DIFF
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -593,8 +593,6 @@ export class CommandInstance {
         positionalKeys.push(...parsed.aliases[key]);
       });
 
-      const {configObjects, default: defaults} = yargs.getOptions();
-
       Object.keys(parsed.argv).forEach(key => {
         if (positionalKeys.includes(key)) {
           // any new aliases need to be placed in positionalMap, which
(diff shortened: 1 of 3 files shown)


bcoe (commented):
What's the specific breaking change this will expose ... if it's behavior that people would never want, and no tests caught it, then I could imagine us not taking this as a major change -- worst case perhaps we could put this behavior behind a flag, to help people bumping into issues, and then swap the default value in the next major.

aol-nnov:
@bcoe I'm not sure I understand you correctly. Does your comment states this change will not be included until next major release unless it is hidden behid a flag or the other way around - it is for such an uncommon (rare) case that the author should not worry about any breakage?

bcoe (changes requested):
Thinking on this one a bit, a couple things jump out at me:

1. Rather than implicitly deciding whether to concat values, based on whether a default value has been configured for a key, better that we figure out a way to track (during parsing) whether a default value has been used -- we've talked in the past about adding an [`isDefaulted` method](https://github.com/yargs/yargs/pull/1472), this would be a great feature for users, and would also allow us to simplify this logic.
2. Providing config _feels_ different than default values, and I could imagine 50% of people wishing that values would be combined, and the other 50% wishing for the opposite behavior. This makes me think that rather than taking this as a breaking change, we'd do better to introduce a parser flag, to allow folks to configure which approach is taken.

jly36963:
Questions:
- I added helpers (eg. `isDefaulted`). Will they need to be changed to accommodate aliases, or configurations like `strip-dashed`?
- Should an option for `overwrite-config` or `combine-config` be added, so that the behavior around merging config & user provided arrays can be configured?

bcoe:
@jly36963 thank you for this great work :clap:, I will make an effort to review soon.

bcoe (approved):

Issue #2102, opened 2021-12-23
Default command options not displayed correctly on command failure


ykolomatski:
Options specified in the builder of a default command are not properly displayed on command failure.

**Issues**:

- Description of required options not displayed
- Optional options not displayed

Running the script with the `--help` option properly displays all the options and descriptions. The issues only occur when the command fails and no `--help` option is provided.

**Code:** 

example.ts:
```typescript
import yargs from "yargs";

yargs(process.argv.slice(2))
  .command(
    "$0",
    "Default command",
    (yargs) => {
      return yargs.options({
        option1: {
          describe: "Option 1 description",
          type: "string",
          demandOption: true,
        },
        option2: {
          describe: "Option 2 description",
          type: "string",
        },
      });
    },
    (argv) => {
      console.log(argv.option);
    }
  )
  .parseSync();
````

**Running:** 
```
npx ts-node example.ts
```

**Expected result:**
```
Default command

Options :
  --help     Affiche l'aide                                            [booléen]
  --version  Affiche le numéro de version                              [booléen]
  --option1  Option 1 description                [chaîne de caractères] [requis]
  --option2  Option 2 description                         [chaîne de caractères]

Argument requis manquant : option1
```

**Actual result:**
```
Default command

Options :
  --help     Affiche l'aide                                            [booléen]
  --version  Affiche le numéro de version                              [booléen]
  --option1                                      [chaîne de caractères] [requis]

Argument requis manquant : option1
```

yargs version: 17.3.0

jly36963:
Relates to: https://github.com/yargs/yargs/issues/2030

xmedeko:
It's working in v17.5.1 and hence may be closed.

bcoe:
@xmedeko that's great news. Closing this issue. @ykolomatski please let us know if you continue to have any problems.

Issue #2030, opened 2021-09-14
Looks like ignore describe of default command


yano3nora:
Thanks great tool. Please tell me if the following is a bug or a specification.

## env
- yargs v17.1.1
- node v14.15.4

## code & reproduction
I want to output a description of the arguments to help for each of the default command and another command.

```ts
import yargs from 'yargs'
const { hideBin } = require('yargs/helpers')

yargs(hideBin(process.argv))
  .command('another <another-arg>', 'Another', (argv: yargs.Argv) => {
    return argv
      .positional('another-arg', { describe: 'ANOTHER_ARG', demandOption: true })
  }, () => {})
  .command(
    '$0 <arg>', 'Default command', (argv: yargs.Argv) => {
      return argv
        .positional('arg', { describe: 'ARG', demandOption: true })
    }, () => {})
  .help()
  .argv
```

I execute another, it outputs `ANOTHER_ARG` and `[required]`  as expected.

```sh
$ node main.js another

===

main.js another <another-arg>

Another

Positionals:
  another-arg  ANOTHER_ARG                                            [required]

Options:
  --version  Show version number                                       [boolean]
  --help     Show help                                                 [boolean]

Not enough non-option arguments: got 0, need at least 1
```

But execute default,`ARG` description was not showing. (seem it works well with `--help`)

```sh
$ node main.js

===

main.js <arg>

Default command

Commands:
  main.js another <another-arg>  Another
  main.js <arg>                  Default command                       [default]

Positionals:
  arg

Options:
  --version  Show version number                                       [boolean]
  --help     Show help                                                 [boolean]

Not enough non-option arguments: got 0, need at least 1
```

When I run the `node main.js`, I want to see the description `ARG` in the help like an another command. What should I do?

jly36963:
@yano3nora 
I can reproduce the issue, and I'm currently working on a fix for this

xmedeko:
Seem it's working in v17.5.1 and may be closed.

bcoe:
@yano3nora let us know if you continue to have any issues in `v17.5.1`.

PR #2105, opened 2021-12-27
fix: failed command usage string is missing arg descriptions and optional args

jly36963:
Addresses: 
https://github.com/yargs/yargs/issues/2030
https://github.com/yargs/yargs/issues/2102

## Problem

The usage from a failed default command is missing the following:
- argument descriptions
- optional arguments
- possibly more (?)

### Example

In the code and output below, you will see that the failed default command is missing descriptions and optional arguments. (I omitted the command usage output, as it's not part of the issue.)

```js

yargs('')
  .command({
    command: '$0 <arg1>',
    desc: 'default desc',
    builder: yargs => yargs
      .option('arg1', {
        type: 'string',
        desc: 'arg1 desc (default command)',
        demandOption: true
      })
      .option('arg2', {
        type: 'string',
        desc: 'arg2 desc (default command)'
      }),
    handler: argv => {
      console.log(argv)
    }
  })
  .command({
    command: 'cmd1 <arg1>',
    desc: 'cmd1 desc',
    builder: yargs => yargs
      .option('arg1', {
        type: 'string',
        desc: 'arg1 desc',
        demandOption: true
      }),
    handler: argv => {
      console.log(argv)
    }
  })
  .strict()
  .parse()


/*

# cmd1
# yargs('cmd1')

Options:
  --arg1     arg1 desc                                       [string] [required]

# default (failed command)
# yargs('')

Options:
  --arg1                                                     [string] [required]

# default (help)
# yargs('--help')

Options:
  --arg1     arg1 desc (default command)                     [string] [required]
  --arg2     arg2 desc (default command)                                [string]

*/
```

## Solution

I'm still trying to figure out the fix for this. I wrote a unit test that should demonstrate the incongruity between expected/actual behavior. 

## Notes

This is what I've found so far, concerning failed default commands and the parts that are missing from the usage output:

- validation functions call `usage.fail` if validation fails
- usage.fail calls `yargs.showHelp('error')`
- yargs.showHelp calls `this.#usage.showHelp`
- usage.showHelp calls `emit(self.help())`

I logged `descriptions` at multiple points in the yargs process, and it's strange because arg1/arg2 show up in descriptions, but later get removed. I don't know if that's a result of resetting or using cached values, but something is wonky there. I'll keep trying to figure out why they get removed.

```
usage.describe { descriptions: { help: '__yargsString__:Show help' } }
usage.describe {
  descriptions: {
    help: '__yargsString__:Show help',
    version: '__yargsString__:Show version number'
  }
}
usage.describe { descriptions: { help: '__yargsString__:Show help' } }
usage.describe {
  descriptions: {
    help: '__yargsString__:Show help',
    version: '__yargsString__:Show version number'
  }
}
yargs.option { key: 'arg1', desc: 'arg1 desc (default command)' }
usage.describe {
  descriptions: {
    help: '__yargsString__:Show help',
    version: '__yargsString__:Show version number',
    arg1: 'arg1 desc (default command)'
  }
}
yargs.option { key: 'arg2', desc: 'arg2 desc (default command)' }
usage.describe {
  descriptions: {
    help: '__yargsString__:Show help',
    version: '__yargsString__:Show version number',
    arg1: 'arg1 desc (default command)',
    arg2: 'arg2 desc (default command)'
  }
}
usage.showHelp {
  descriptions: {
    help: '__yargsString__:Show help',
    version: '__yargsString__:Show version number'
  }
}
usage.help {
  descriptions: {
    help: '__yargsString__:Show help',
    version: '__yargsString__:Show version number'
  }
}
```

### DIFF
--- a/lib/usage.ts
+++ b/lib/usage.ts
@@ -707,22 +707,31 @@ export function usage(yargs: YargsInstance, shim: PlatformShim) {
       descriptions,
     });
   };
-  self.unfreeze = function unfreeze() {
+  self.unfreeze = function unfreeze(defaultCommand = false) {
     const frozen = frozens.pop();
     // In the case of running a defaultCommand, we reset
     // usage early to ensure we receive the top level instructions.
     // unfreezing again should just be a noop:
     if (!frozen) return;
-    ({
-      failMessage,
-      failureOutput,
-      usages,
-      usageDisabled,
-      epilogs,
-      examples,
-      commands,
-      descriptions,
-    } = frozen);
+    // Addresses: https://github.com/yargs/yargs/issues/2030
+    if (defaultCommand) {
+      descriptions = {...frozen.descriptions, ...descriptions};
+      commands = [...frozen.commands, ...commands];
+      usages = [...frozen.usages, ...usages];
+      examples = [...frozen.examples, ...examples];
+      epilogs = [...frozen.epilogs, ...epilogs];
+    } else {
+      ({
+        failMessage,
+        failureOutput,
+        usages,
+        usageDisabled,
+        epilogs,
+        examples,
+        commands,
+        descriptions,
+      } = frozen);
+    }
   };
 
   return self;
@@ -759,7 +768,7 @@ export interface UsageInstance {
   showHelpOnFail(enabled?: boolean | string, message?: string): UsageInstance;
   showVersion(level?: 'error' | 'log' | ((message: string) => void)): void;
   stringifiedValues(values?: any[], separator?: string): string;
-  unfreeze(): void;
+  unfreeze(defaultCommand?: boolean): void;
   usage(msg: string | null, description?: string | false): UsageInstance;
   version(ver: any): void;
   wrap(cols: number | null | undefined): void;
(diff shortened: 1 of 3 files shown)


jly36963:
@bcoe 
The `unfreeze` call on [this line](https://github.com/yargs/yargs/blob/59a86fb83cfeb8533c6dd446c73cf4166cc455f2/lib/command.ts#L315) is causing the weird behavior. It is wiping the descriptions and optional arguments for the default command. If I comment out that block, then I get the expected args/descriptions in the usage output. This would fix the issues reported in 2030 and 2102.

However, I don't think I understand the comment above that block.

https://github.com/yargs/yargs/blob/59a86fb83cfeb8533c6dd446c73cf4166cc455f2/lib/command.ts#L312

If I understand it correctly, then the usage shouldn't contain anything from the default command, eg: `arg1`. What should be the correct behavior here?

lib/usage.ts
```diff
@@ -707,22 +707,31 @@ export function usage(yargs: YargsInstance, shim: PlatformShim) {
...
-      failureOutput,
-      usages,
-      usageDisabled,
-      epilogs,
-      examples,
-      commands,
-      descriptions,
-    } = frozen);
+    // Addresses: https://github.com/yargs/yargs/issues/2030
+    if (defaultCommand) {
```
bcoe, line 717: I like that you just made this a boolean setting rather than adding a new method.

bcoe (approved):

bcoe:
@jly36963 as always, thank you for the contribution.

Issue #1476, opened 2019-11-09
document usage of yargs with input tools like inquirer


bcoe:
It would be good to add a section to our docs that demonstrates using yargs in parallel with tools like [inquirer](https://www.npmjs.com/package/inquirer).

see: https://github.com/yargs/yargs/issues/286

akashroy1:
Can I work on this issue?

bcoe:
@akashroy1 I would love help with this. I would add it as an [Advanced topic](https://github.com/yargs/yargs/blob/main/docs/advanced.md).

shadowspawn:
Yes, I think so. Thanks @nephh. Benjamin is busy elsewhere, but I can give you feedback and review.

nephh:
Thanks @shadowspawn, I just went ahead and made a PR: #2385 This is my first time contributing to an open source project, so please let me know if there's anything I can do differently or any changes that should be made!

shadowspawn:
#2385 has been merged.

Thank you for your contributions.

PR #2114, opened 2022-01-04
added an example using inquirer prompting

chrisdugne:
I've been looking for how to use [inquirer](https://github.com/SBoudrias/Inquirer.js/) with yargs,
and found this issue: https://github.com/yargs/yargs/issues/1476 

So here is how I use an async command to prompt things with inquirer

cheers

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -302,7 +302,7 @@ s.on('end', function () {
 ```
 
 ***
-    $ node line_count.js 
+    $ node line_count.js
     Usage: line_count.js <command> [options]
 
     Commands:
@@ -319,7 +319,7 @@ s.on('end', function () {
     copyright 2019
 
     Missing required argument: f
-    
+
     $ node line_count.js count
     line_count.js count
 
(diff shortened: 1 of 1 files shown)


docs/examples.md
```diff
@@ -337,3 +337,43 @@ s.on('end', function () {
...
+      name: 'name',
+      type: 'string'
+    }
+  ]);
+
+  console.log(`Hello, ${answers.name}!`);
+};
+
+const argv = yargs(process.argv.splice(2))
+  .command('ask', 'use inquirer to prompt for your name')
```
bcoe, line 361: rather than using a switch statement, you could do this:

```js
.command('ask', 'use inquirer to prompt for your name', () => {}, async () => {
   await askName();
}
```

docs/examples.md
```diff
@@ -337,3 +337,43 @@ s.on('end', function () {
...
+
+switch (command) {
+  case 'ask':
+    askName();
+    break;
+  case 'sing':
+    console.log('🎵 Oy oy oy');
+    break;
+  default:
+    console.error(`❌ unknown command ${command}\n`);
```
bcoe, line 376: rather than doing this, you can just use yargs' setting `stricCommand()` and it will enforce that the command is known.

bcoe (changes requested):
Thanks for the contribution @chrisdugne, left a few recommendations.

docs/examples.md
chrisdugne: nice, I've used `.strict()`

docs/examples.md
chrisdugne: ok, it's working without the `async/await` though

```js
.command('ask', 'use inquirer to prompt for your name', () => {}, askName)
```

bcoe: 👍

bcoe: the advantage of async await is that you could `await` the top level `yargs.argv` and it shouln't return until you've filled in the prompted messages ... neat right?

bcoe (approved):

bcoe:
@chrisdugne thank you for the contribution.

docs/examples.md
chrisdugne: actually I've first tested with your suggestion, and it was not working as expected, not sure why;

chrisdugne:
you're welcome, and I thank you for yargs : )

PR #2146, opened 2022-03-16
Add more complex example for typescript

natlibfi-jonollil:


docs/typescript.md
```diff
@@ -133,3 +133,48 @@ const argv = yargs.option('difficulty', {
...
+    .epilog('Copyright (C) 2022 CREATOR')
+    .usage('$0 <file> [options]')
+    .showHelpOnFail(true)
+    .example([
+      ['$ node $0/dist/cli.js file.json -fv true -d output/'],
+      ['$ node $0/dist/cli.js file.xml -v true -f false -d output/']
+      ['$ node $0/dist/cli.js -d output/ file.xml']
+    ])
+    .env('EXAMPLE_CLI')
+    .positional('file', {type: 'string', describe: 'File to be handled'})
```
bcoe, line 154: Should we add a command other than the default command `$0`, to demonstrate how to configure positionals for a sub command?

I think this is a common use case.

bcoe (changes requested):

bcoe:
@natlibfi-jonollil as an advanced example, I might try something more like this:

https://github.com/googleapis/repo-automation-bots/blob/main/packages/owl-bot/src/bin/commands/open-pr.ts

Where we demonstrate how to define types, etc.

bcoe:
Looks like this has become stale, please don't hesitate to submit a PR when you have some time to dust this work off.

PR #2148, opened 2022-03-21
fix(completion): don't show positional args choices with option choices

alan-agius4:
Previously, when both positional arguments and options has choices, during autocompletion we displayed both at the same time.


lib/completion.ts
```diff
@@ -157,20 +157,27 @@ export class Completion implements CompletionInstance {
...
+    }
+
     const positionalKeys =
       this.yargs.getGroups()[this.usage.getPositionalGroupName()] || [];
     const offset = Math.max(
       this.indexAfterLastReset,
       this.yargs.getInternalMethods().getContext().commands.length +
         /* name of the script is first param */ 1
     );
-    const positionalValues = argv._.slice(offset);
```
alan-agius4, line 167: Removed also a redundant slice.

bcoe (approved):

bcoe:
@alan-agius4 thank you for the patch.

PR #2150, opened 2022-03-23
chore: add developing.md

hongaar:
Adds a simple devcontainer configuration to quickly get started with this project as a developer.


bcoe:
@hongaar this is a new technology to me, could you link me some docs?

Thanks for the PR.

hongaar:
It's a convenient way to use a Docker container as an isolated development environment. It makes it really easy to get started with a project: repeatable dev environment without the need to juggle runtime/compiler versions etc.

VSCode has very clever support for it (see https://code.visualstudio.com/docs/remote/containers) and GitHub Codespaces are also built on this (https://github.com/features/codespaces)

Having the configuration included in the repo is merely a convenience, not a requirement to use it. It is not IDE agnostic (yet), unfortunately. There are requests to add support to e.g. JetBrains IDEs, not sure about timelines.

bcoe (approved):

bcoe (changes requested):
I'm on the fence with this PR, over the years we've proliferated a lot of configuration for yargs, and I wonder about adding additional configuration for one specific development environment.

I myself, as an example, use `nvm` and not `vscode` for managing environments, which makes it easy to test in many environments.

What if instead of adding the `.devcontainer` folder, you instead add a `docs/developing.md`, which can have a section that explains how to setup the `.devcontainer` for `vscode`, this could have the exact files you provide here.

hongaar:
@bcoe implemented your suggested change.

I share your hesitancy to include these tool-specific config files in the repo, they're mainly subjective and cater only to some contributors. One reason to include them could be since it's used by GitHub Codespaces - similar to how some repos include a `pull_request_template.md` file.

PR #2154, opened 2022-03-26
fix: Show message when showHelpOnFail is chained globally

jly36963:
Addresses: https://github.com/yargs/yargs/issues/2085

## Problem

When `showHelpOnFail` is used globally, it doesn't work. 
This is because `usage.reset()` sets failMessage to null.

## Solution

I added a `globalFailMessage` variable to capture the global mesage.
On fail, `failMessage || globalFailMessage` is printed. I figured if both are provided, the more specific message should be shown.

## Other

I saw that `null | undefined` was used a lot, so I made type `nil`. (Idea borrowed from lodash isNil)

### DIFF
--- a/lib/usage.ts
+++ b/lib/usage.ts
@@ -2,7 +2,7 @@
 // failures, etc. keeps logging in one place.
 import {Dictionary, PlatformShim} from './typings/common-types.js';
 import {objFilter} from './utils/obj-filter.js';
-import {YargsInstance} from './yargs-factory.js';
+import {YargsInstance, nil} from './yargs-factory.js';
 import {YError} from './yerror.js';
 import {DetailedArguments} from './typings/yargs-parser-types.js';
 import setBlocking from './utils/set-blocking.js';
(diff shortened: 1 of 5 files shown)


lib/usage.ts
```diff
@@ -20,16 +20,22 @@ export function usage(yargs: YargsInstance, shim: PlatformShim) {
...
     arg2?: string
   ) {
-    function parseFunctionArgs(): [boolean, string?] {
-      return typeof arg1 === 'string' ? [true, arg1] : [arg1, arg2];
+    const [enabled, message] =
+      typeof arg1 === 'string' ? [true, arg1] : [arg1, arg2];
+
+    // If global context, set globalFailMessage
+    // Addresses: https://github.com/yargs/yargs/issues/2085
+    if (yargs.getInternalMethods().getContext().commands.length === 0) {
```
jly36963, line 35: Is there an easier way to figure out that this is in the global context? (I can move this logic to a helper function and call it here)

bcoe: I like the idea of pulling this into a private helper method, then if we can think of a better way to detect we're in the global context we can swap out the logic.

bcoe (approved):
One suggested refactor.

bcoe:
LGTM, with a suggested refactor. I would be surprised if we're not already detecting the global context, vs., a command handler context somewhere else in the codebase. Perhaps look in `command.ts` (nothing is immediately coming to my mind).

jly36963:
I realized that checking `commands.length === 0` yields a false positive when running the default command. 
Instead, I set a new property `this.#isGlobalContext` to true, and when the `runCommand` logic begins I set it to false.

Issue #2153, opened 2022-03-25
Can I prevent yargs .describe feature from showing auto-generated info about type, default, etc?


BadIdeaException:
When using yargs [.describe](https://yargs.js.org/docs/#api-reference-describekey-desc) method to generate usage information, is there some way to prevent yargs from auto-adding information about type, default value, choices etc.?

I.e. the following
```
        .option('foo', {
            type: 'string',
            choices: ['bar','baz'],
            required: true,
            default: 'bar',
            description: 'The most important option of all'
        })
```
results in this output when doing --help:
```
--foo      The most important option of all      [string] [required] [choices: "bar", "baz"] [default: "bar"]
```
I have to confess I find that to be a bit ugly, and it is out of line with how help texts are generally formatted in linux. Also, on longer description texts that include line breaks, it breaks formatting.

I haven't found anything in the docs. Is there some way to suppress generation of the bracketed `[]` parts, without having to completely give up using the describe feature?

i-siso:
@bcoe Thanks for the awesome awesome lib. As mentioned in the issue, and also the open #2156, any update on having this change introduced?
Apart from the programmers, the end users generally don't care about the type or can get confused by the mention of the type in help!

bcoe:
@i-siso thank you 😄 

Unfortunately no updates, I've had no spare time I can put towards OSS recently. I think it's just a matter of dusting off @jly36963's PR.

If @jly36963 thinks it's ready, I'm happy to see it land.

bcoe:
There is now an option to turn off type information, implemented in #2156 

Additional usage configuration settings can be added as needed. This will be released in the next update to yargs.

06000208:
Can this feature be released in yargs@next as a bleeding edge feature, per the readme? It [appears](https://www.npmjs.com/package/yargs?activeTab=versions) that yargs@next is several versions behind latest, and does not include it as of writing

Issue #1145, opened 2018-06-06
[Feature Request] Hide default value and flag type


borela:
Currently, my project shows:

```
Commands:
  index.js build     Build the project.
  index.js scaffold  Generate a project from the available templates.
  index.js lint      Check or fix code style.
  index.js test      Run test suites.

Options:
  --help            Show usage instructions.                           [boolean]
  --version         Show toolbox version.                              [boolean]
  --comment-flow    Convert Flow annotations to comments.              [boolean]
  --env             Used to determine the polyfills and fine tune the linter.
        [array] [choices: "browser", "node-js"] [default: ["browser","node-js"]]
  --flow            Enable Flow annotations.                           [boolean]
  --jsx             Enable JSX.                                        [boolean]
  --react           Enable React transformations.                      [boolean]
  --remove-flow     Remove Flow annotations.                           [boolean]
  --production      Remove debug plugins.                              [boolean]
  --target-browser  Browsers targeted by the project.
      [array] [default: ["chrome >= 49",">= 0.5%","last 2 versions","not dead"]]
  --target-node-js  Minimum NodeJS version targeted by the project.
                                                           [string] [default: 6]
  --type-script     Enable TypeScript.                                 [boolean]
```

But the default values and types pollutes it, is it possible already to hide them? I would like something like this:

```
Commands:
  index.js build     Build the project.
  index.js scaffold  Generate a project from the available templates.
  index.js lint      Check or fix code style.
  index.js test      Run test suites.

Options:
  --help            Show usage instructions.                           
  --version         Show toolbox version.                              
  --comment-flow    Convert Flow annotations to comments.              
  --env             Used to determine the polyfills and fine tune the linter.
  --flow            Enable Flow annotations.                           
  --jsx             Enable JSX.                                      
  --react           Enable React transformations.                    
  --remove-flow     Remove Flow annotations.                         
  --production      Remove debug plugins.                            
  --target-browser  Browsers targeted by the project.
  --target-node-js  Minimum NodeJS version targeted by the project.
  --type-script     Enable TypeScript.                                 
```

Linard11:
Afaik, there is no option to change the formatting of the of the help print out yet, besides usage() and epiloge() which will add a header/footer to it

borela:
@bcoe Would you a pull request of a function like "hideExtras()" to hide the types and default values?

borela:
Another solution, suggestion would be `hideTypes()` and `hideDefaultValues()`.

[...]

PR #2156, opened 2022-03-27
feature: add method to hide option extras

jly36963:
Addresses: https://github.com/yargs/yargs/issues/2153

## Description

Given the following:

```js
yargs('cmd1 --help')
  .command(
    'cmd1',
    'cmd1 desc',
    yargs => yargs
      .option('opt1', {
        type: 'string',
        choices: ['foo', 'bar', 'baz'],
        alias: 'o',
        required: true,
        description: "A long description that might break formatting"
      }),
    argv => console.log(argv)
  )
  .hideOptionExtras()
  .strict()
  .parse()
```

The `hideOptionExtras` would cause the following output

```
cmd1 desc

Options:
      --help     Show help                                             [boolean]
      --version  Show version number                                   [boolean]
  -o, --opt1     A long description that might break formatting
                              [string] [required] [choices: "foo", "bar", "baz"]
```

to look like this instead

```
cmd1 desc

Options:
      --help     Show help
      --version  Show version number
  -o, --opt1     A long description that might break formatting
```

## Questions

If this works, I'll add unit tests.
What other things would need to change? (docs, examples, etc)

### DIFF
--- a/lib/usage.ts
+++ b/lib/usage.ts
@@ -768,7 +768,6 @@ export interface UsageInstance {
   getUsage(): [string, string][];
   getUsageDisabled(): boolean;
   help(): string;
-  hideOptionExtras(): void;
   reset(localLookup: Dictionary<boolean>): UsageInstance;
   showHelp(level?: 'error' | 'log' | ((message: string) => void)): void;
   showHelpOnFail(enabled?: boolean | string, message?: string): UsageInstance;
--- a/lib/usage.ts
+++ b/lib/usage.ts
@@ -443,7 +443,8 @@ export function usage(yargs: YargsInstance, shim: PlatformShim) {
           desc
         );
 
-        if (extra) ui.div({text: extra, padding: [0, 0, 0, 2], align: 'right'});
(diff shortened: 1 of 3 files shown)


bcoe (changes requested):
@jly36963 this is great work, thank you :clap: 

One ask, would it be possible to add this setting in the list here: https://github.com/yargs/yargs/blob/main/docs/api.md#parserconfigurationobj

bcoe (approved):

bcoe (approved):
Love this. I think `usageConfiguration` will help us provide nuanced configuration without adding 1,000,000 additional methods.

ljharb:
I’m a bit confused; how can i show types in general, but hide types for an a specific option?

jly36963:
@ljharb I don't think there is a way currently. It would likely be a new property on option/positional objects like `hideUsageTypes`, or potentially a refactor on the usage logic to allow for customizations.

I won't be involved with this package much in the future, so I won't be much help in implementing this change

ljharb:
In that case, this does not solve #1145, thanks for confirming.

Issue #2159, opened 2022-04-05
Parser configuration strip-aliased does not work with coerce


xmedeko:
Even when parser configuration `strip-aliased` is `true`, then it does not work if an option has `coerce`, e.g. see:
```js
const args = require("yargs")
    .parserConfiguration({ "strip-aliased": true })
    .option("log", { alias: ["l", "ll"], coerce: s => s })
    .parse("-l 1");
delete args["$0"]; delete args["_"];
// args is { log: 1, l: 1, ll: 1 } 
```

xmedeko:
@jly36963 Thanks, fixed in 17.4.1. 👍

Issue #2158, opened 2022-04-03
"Coerce" calls always on option, does it correct behaviour?


lis355:
Run this simple script WITHOUT any cli arguments:

```javascript
const path = require("path");

const yargs = require("yargs/yargs");

const argv = yargs(process.argv.slice(2))
	.option("someArg", {
		type: "string",
		coerce: path.resolve
	})
	.parse();

```

You will get error message in output:

```
[C:\Program]() Files\nodejs\node.exe [.\main.js]()
Options:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]
  --someArg                                                             [string]
node_modules/yargs/build/index.cjs:1
The "path" argument must be of type string. Received undefined
node_modules/yargs/build/index.cjs:1
Process exited with code 1
```

Does it correct? If my someArg is not a demand, why he "gets" default value undefined and then coerce is called? And, of cource, path.resolve drop this error 

I think, it's not correct, to call coerce, if user not give an argument

xmedeko:
Seems like duplicate of #2130

lis355:
@xmedeko Yeah, thanks

xmedeko:
@lis355 If you agree it's a duplicate, then close the issue...

xmedeko:
Should be fixed in 17.4.1.

lis355:
Thanks!

Close because duplicate of https://github.com/yargs/yargs/issues/2130

Issue #2130, opened 2022-02-05
Yargs v17 calls coerce with undefined when option is not passed


dsamerotte:
Per #1219, yargs has not historically called `coerce` when the associated option is not passed. However, when porting my code to v17, I see that this is no longer the case, and I'm wondering whether this is an intentional breaking change or a regression. The only potentially relevant breaking change that I see in the release notes is that "coerce is now applied before validation," but I wouldn't expect this change in behavior from that description alone.

I understand that yarg's middleware underwent substantial changes in v17. Based on my two-second investigation, it looks like v16 maps over the supplied options and then calls coerce as needed, such that coerce can never receive `undefined`. However, v17 instead appears to reduce all middleware without checking the existence of the related option.

The obvious result is that code like `.coerce('dir', path.resolve)`, which still appears in the docs, no longer works, and client code now needs to explicitly check for `undefined`.

If this is an intentional change, should this also affect subcommand processing? I'm even seeing coerce called like this in parent commands when calling only a subcommand and when the parent option is not passed and is marked as `global: false`. If I understand the `global` option correctly, this seems especially strange.

I thought I'd keep this short and start with the high-level question first. Based on the answer, I can, of course, provide additional thoughts and information. Thanks!

elitree:
Just to illustrate with an example of one difference in v17 which doesn't seem right (or at least, I didn't anticipate it based on the changelog):

```javascript
const yargs = require('yargs');

let args = yargs
.strict()
.option('abc', {
    boolean: true
})
.option('xyz',{
    implies: ['abc'],
    // removing this coerce method will avoid errors when no params are passed
    coerce: (val) => val && val.toLowerCase()
});

console.log('////yargs x.y.z////');
console.log(JSON.stringify(args));
```
Based on the above, since option 'abc' isn't a required option, I would expect to be able to pass no arguments. This is the case in v16:

```bash
$ node index.js
////yargs 16.2.0////
{"$0":"index.js","parsed":false,"argv":{"_":[],"$0":"index.js"}}
```
In v17, however, the combination of the requires relationship and the coerce causes a "Missing dependent arguments" error:
```bash
$ node index.js
////yargs 17.3.0////
Options:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]
  --abc                                                                [boolean]
  --xyz

Missing dependent arguments:
 xyz -> abc
```
[...]

PR #2161, opened 2022-04-06
fix: coerce pollutes argv

jly36963:
## Related

- https://github.com/yargs/yargs/issues/2130
- https://github.com/yargs/yargs/issues/2158
- https://github.com/yargs/yargs/issues/2159
- https://github.com/yargs/yargs/issues/2162

## TLDR

- coerce always runs, even if arg isn't provided
- coerce sets the value on argv, resulting in keys existing on argv that weren't passed.
- any validation around certain keys (not) existing on argv manifest unintended behavior

## Description

I'm fixing two coerce bugs:
- bug one: coerce and related argument not passed
- bug two: coerce and strip-aliased

### Details

Coerce doesn't behave correctly when the related arg is not passed. 
There are two unintentional side-effects:
- a new arg is set on argv
- coerce function is run

`coerce > maybeAsyncResult > resultHandler` sets argv, which shouldn't happen.

### Reproduction
    
```js
const input = 'cmd1'

yargs(input)
  .command(
    'cmd1',
    'cmd1 desc',
    yargs => yargs
      .option('foo', { alias: 'f', type: 'string' })
      .option('bar', { 
        type: 'string', 
        alias: 'b',
        coerce: val => console.log('Why was I called?') || val,
      }),
    argv => {
      console.log({ argv })
    })
  .strict()
  .parse()
```

```
Why was I called?
{
  argv: {
    _: [ 'cmd1' ],
    '$0': 'example.js',
    bar: undefined,
    b: undefined
  }
}
```

'b' was not provided, so:
- its coerce function should not run
- it should not be defined on `argv`

### Details

Coerce adds aliases to argv (even when strip-aliased is true)

### Reproduction

```js
const input = 'cmd1 -f hello -b world'

yargs(input)
  .parserConfiguration({ "strip-aliased": true })
  .command(
    'cmd1',
    'cmd1 desc',
    yargs => yargs
      .option('foo', { alias: 'f', type: 'string' })
      .option('bar', { 
        type: 'string', 
        alias: 'b',
        coerce: val => val,
      }),
    argv => {
      console.log({ argv })
    })
  .strict()
  .parse()
```

```
{
  argv: {
    _: [ 'cmd1' ],
    foo: 'hello',
    bar: 'world',
    '$0': 'example.js',
    b: 'world'
  }
}
```

`b` should not be defined on `argv`

### DIFF
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -389,7 +389,6 @@ export class YargsInstance {
         let aliases: Dictionary<string[]>;
 
         // Skip coerce logic if related arg was not provided
-        // Addresses: https://github.com/yargs/yargs/issues/2130
         const shouldCoerce = Object.prototype.hasOwnProperty.call(argv, keys);
         if (!shouldCoerce) {
           return argv;
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -404,7 +404,10 @@ export class YargsInstance {
           },
           (result: any): Partial<Arguments> => {
             argv[keys] = result;
-            if (aliases[keys]) {
+            const stripAliased = yargs
+              .getInternalMethods()
+              .getParserConfiguration()['strip-aliased'];
+            if (aliases[keys] && stripAliased !== true) {
               for (const alias of aliases[keys]) {
                 argv[alias] = result;
               }
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -389,7 +389,7 @@ export class YargsInstance {
         let aliases: Dictionary<string[]>;
 
         // Skip coerce logic if related arg was not provided
-        const shouldCoerce = Object.prototype.hasOwnProperty.call(argv, keys);
+        const shouldCoerce = Object.hasOwnProperty.call(argv, keys);
         if (!shouldCoerce) {
           return argv;
         }
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -383,6 +383,14 @@ export class YargsInstance {
         yargs: YargsInstance
       ): Partial<Arguments> | Promise<Partial<Arguments>> => {
         let aliases: Dictionary<string[]>;
+
+        // Skip coerce logic if related arg was not provided
+        // Addresses: https://github.com/yargs/yargs/issues/2130
+        const shouldCoerce = Object.prototype.hasOwnProperty.call(argv, keys);
+        if (!shouldCoerce) {
+          return argv;
+        }
+
         return maybeAsyncResult<
           Partial<Arguments> | Promise<Partial<Arguments>> | any
         >(
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -603,9 +603,9 @@ export class CommandInstance {
           // If both positionals/options provided, no default was set,
           // and if at least one is an array: don't overwrite, combine.
           if (
-            !Object.prototype.hasOwnProperty.call(defaults, key) &&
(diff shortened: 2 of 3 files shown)


lib/yargs-factory.ts
```diff
@@ -387,6 +387,13 @@ export class YargsInstance {
         yargs: YargsInstance
       ): Partial<Arguments> | Promise<Partial<Arguments>> => {
         let aliases: Dictionary<string[]>;
+
+        // Skip coerce logic if related arg was not provided
+        const shouldCoerce = Object.prototype.hasOwnProperty.call(argv, keys);
```
bcoe, line 392: I believe this can just be:
```
Object.hasOwnProperty.call(obj, 'a')
```

bcoe (changes requested):
Small nit, but this bug fix looks good to me.

Great digging @jly36963.

bcoe (approved):

bcoe:
@jly36963 thank you for the great fix, we can likely close the affected PRs?

lib/yargs-factory.ts
dhensby: I'm afraid this is not correct and it should be `Object.prototype.hasOwnProperty`.

The changes relating to this should be reverted.

jly36963: `Object.hasOwnProperty.call({a: 5}, 'a')` and `Object.prototype.hasOwnProperty.call({a: 5}, 'a')` both return `true`. (I thought there might be an issue when using a null object, but both return `false` when `Object.create(null)` is used as the first argument.) I don't know of any scenarios where their behavior differs. Do you have an example where it could cause problems?

jly36963: Also, `Object.hasOwnProperty.call === Object.prototype.hasOwnProperty.call` is true. I'm assuming they are a reference to the same object

dhensby: Notwithstanding the fact that this convention of using the `prototype` instead of the `Object` class directly is widely adopted.

It's the same reason to not use `{a: 1}.hasOwnProperty('a')`. The function could have been overridden via a global polyfill or patch meaning you're not using native behaviour.

I assume that `{}.hasOwnProperty.call === Object.hasOwnProperty.call` as well as the prototype. The point is that they *can* be overridden and are, thus, unreliable to use unless you use the underlying prototype.

bcoe: I mean, `Object.prototype` can just as easily be overwritten:

```
> Object.prototype.hasOwnProperty = () => 'batman'
[Function (anonymous)]
> {}.hasOwnProperty()
'batman'
```

We would have to rewrite all of yargs to use primordials that are frozen at Node.js' startup to really avoid this issue.

bcoe: > We would have to rewrite all of yargs to use primordials that are frozen at Node.js' startup to really avoid this issue.

This is how Node.js itself approaches the problem:

https://github.com/nodejs/node/blob/master/lib/internal/per_context/primordials.js

But it's made easier for them by the fact that they can do so during bootstrapping.

dhensby: Yeah, this is definitely not an issue that is limited to `yargs`, for better or worse it's a language issue.

Whilst I think overriding the built-in methods like `Object.hasOwnProperty()` is pretty crazy, for those crazy enough to do it I think there is enough expectation that the prototype is left alone that it can be used reliably. Hence why it's enforced by standard linting rules in eslint and recommendations elsewhere.

bcoe: @dhensby makes sense to me.

Issue #2124, opened 2022-01-27
`strict()` in combination with `middleware(async, true)` results in `unknown command` error


knobladSICKAG:
### Description
Using the `strict` option in combination with an `async` middleware, which is applied before the validation, results in an `unknown command` error. Removing the `async` from the middleware, removing the `strict` or applying the middleware after the validation produces the expected result (no error).

### Minimal example
The following code fails with the error `Unknown command: test`, although I would expect it to run without error:
```typescript
Yargs
  .strict(true)
  .middleware(async () => {}, true)
  .command("test", false, () => {}, () => {})
  .parseAsync(['test']);
```

PR #2164, opened 2022-04-10
Fix async middleware

jly36963:
Addresses: https://github.com/yargs/yargs/issues/2124

## Description

The result of `applyMiddleware` is `Arguments | Promise<Arguments>`, but the code treats it synchronously. Whenever it returns a promise, that promise isn't handled as such.

This problem is easier to spot when `strict` is used

## Reproduction

```js
const input = 'cmd1 -f Hello -b world'

yargs(input)
  .command(
    'cmd1',
    'cmd1 desc',
    yargs => yargs
      .option('foo', { type: 'string', alias: 'f', required: true })
      .option('bar', { type: 'string', alias: 'b', required: true })
      .middleware(async argv => {
        console.log('sleeping')
        await new Promise(r => setTimeout(r, 2000))
        console.log('waking up')
        return argv
      }, true),
    argv => {
      console.log(argv)
    })
  .strict()
  .parse()
```

I put a console log after `applyMiddleware` is called, and I could see that 'waking up' showed up after `applyMiddleware` finished.

### DIFF
--- a/test/middleware.cjs
+++ b/test/middleware.cjs
@@ -282,21 +282,14 @@ describe('middleware', () => {
     });
 
     // Addresses: https://github.com/yargs/yargs/issues/2124
-    // This test will fail if the result of async middleware is not treated like a promise
-    it('treats result of async middleware as promise', done => {
-      const input = 'cmd1 -f Hello -b world';
+    // This test will fail if result of async middleware is not handled like a promise
+    it('does not cause an unexpected error when async middleware and strict are both used', done => {
+      const input = 'cmd1';
       yargs(input)
         .command(
           'cmd1',
           'cmd1 desc',
-          yargs =>
-            yargs
-              .option('foo', {type: 'string', alias: 'f', required: true})
-              .option('bar', {type: 'string', alias: 'b', required: true})
-              .middleware(async argv => {
-                await new Promise(r => setTimeout(r, 5));
-                return argv;
-              }, true),
+          yargs => yargs.middleware(async argv => argv, true),
           _argv => {
             done();
           }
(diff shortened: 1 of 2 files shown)


test/middleware.cjs
```diff
@@ -281,6 +281,33 @@ describe('middleware', () => {
       }
     });
 
+    // Addresses: https://github.com/yargs/yargs/issues/2124
+    // This test will fail if the result of async middleware is not treated like a promise
+    it('treats result of async middleware as promise', done => {
```
bcoe, line 286: I would be tempted to write a more specific test to #2124 

"Using the strict option in combination with an async middleware, which is applied before the validation, does not result in unknown command"

test/middleware.cjs
```diff
@@ -281,6 +281,33 @@ describe('middleware', () => {
...
+      const input = 'cmd1 -f Hello -b world';
+      yargs(input)
+        .command(
+          'cmd1',
+          'cmd1 desc',
+          yargs =>
+            yargs
```
bcoe, line 293: I had to read this a few times, I'm wondering if it would be worth simplifying to exactly the minimal case outlined in the regression:

```
Yargs
  .strict(true)
  .middleware(async () => {}, true)
  .command("test", false, () => {}, () => {})
  .parseAsync(['test']);
```

I don't have a strong opinion though, if your goal was to exercise parts of the codebase that weren't covered in the above example.

test/middleware.cjs
jly36963: I simplified the test. I kept the `done` logic, because Mocha will pass this test before the unhandled promise would cause it to fail. If there's a regression, this test will fail because `done` will get called twice, and one of the executions will have an error passed as an argument

PR #2165, opened 2022-04-11
refactor: use prototype

jly36963:
Relates to: https://github.com/yargs/yargs/pull/2161#discussion_r847542318

### DIFF
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -603,9 +603,9 @@ export class CommandInstance {
           // If both positionals/options provided, no default was set,
           // and if at least one is an array: don't overwrite, combine.
           if (
-            !Object.hasOwnProperty.call(defaults, key) &&
-            Object.hasOwnProperty.call(argv, key) &&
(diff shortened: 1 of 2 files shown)


bcoe (approved):

bcoe (changes requested):
MDN suggests using `Object.hasOwn`, vs., `Object.hasOwnProperty`, or `Object.prototype.hasOwnProperty`:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty#using_hasownproperty_as_a_property_name

Should we go this route?

jly36963:
@bcoe The second code block in the MDN fragment you linked recommends `Object.prototype.hasOwnProperty.call` as an acceptable alternative, and as @dhensby said, `Object.hasOwn` is a relatively new feature in Node.

bcoe (approved):

bcoe:
👏 works for me, I should have read when `hasOwn` was introduced.

PR #2166, opened 2022-04-13
fix(completion): support for default flags

zuozp8:
following code inside method `argsContainKey` in `completion.ts` have stopped working:
```
    if (this.aliases) {
      // search for aliases in parsed argv
      // can't do the same thing for main option names because argv can contain default values
      for (const alias of this.aliases[key]) {
        if (argv[alias] !== undefined) return true;
```
Currently argv contains default value for aliases too, so I just look at args instead

Test included

### DIFF
--- a/lib/completion.ts
+++ b/lib/completion.ts
@@ -128,7 +128,7 @@ export class Completion implements CompletionInstance {
         if (
           !isPositionalKey &&
           !options.hiddenOptions.includes(key) &&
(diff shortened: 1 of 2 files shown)


zuozp8:
@bcoe @jly36963 rebased

bcoe (approved):

bcoe (approved):

PR #2171, opened 2022-04-26
fix: dont clobber description for multiple option calls

### LINKED ISSUE #2169: subsequent calls to options remove existing description
Hi! 👋 Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `yargs@17.4.1` for the project I'm working on. Here's the issue I was having:

Subsequent calls to `.options()` merge all the properties, except for `description`. Earlier descriptions are always overwritten with `undefined` even if the key is not set in later `.options()` calls. See in the example below how the existing `type` is kept, and `default` is overwritten, but `description` is also removed.

```js
import yargs from 'yargs'

console.log(
  await yargs()
    .options({ arr: { desc: 'my arrrrrg', type: 'string', default: 'old' } })
    .options({ arr: { default: 'new'  } })
    .getHelp()
)

/*

 * Before patch *
Options:
  --arr                                                [string] [default: 'new']

  * After patch *
Options:
  --arr      my arrrrrg                                [string] [default: 'new']

*/
```

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/yargs/build/lib/yargs-factory.js b/node_modules/yargs/build/lib/yargs-factory.js
index f79ad50..2400277 100644
--- a/node_modules/yargs/build/lib/yargs-factory.js
+++ b/node_modules/yargs/build/lib/yargs-factory.js
@@ -667,7 +667,9 @@ export class YargsInstance {
                 this.skipValidation(key);
             }
             const desc = opt.describe || opt.description || opt.desc;
-            this.describe(key, desc);
+            if (typeof desc === 'string') {
+                this.describe(key, desc);
+            }
             if (opt.hidden) {
                 this.hide(key);
             }
```

I couldn't tell if this behavior was intentional or not. All other option properties are guarded in a similar manner, so I thought maybe it was intentional that description was not. But it does seem like a bug to me, or at least I was surprised by this differing behavior.

lukekarrys:
Fixes #2169

My original patch from #2169 broke some existing tests, due to it being necessary to call `this.describe` when initially setting up an option even if no explicit description is set.

So this PR implements this behavior by only calling `this.describe` if there is no existing description OR if the new description is a string.

I added a test that I confirmed fails on main and passes in this PR.

### DIFF
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -1044,7 +1044,11 @@ export class YargsInstance {
       }
 
       const desc = opt.describe || opt.description || opt.desc;
-      this.describe(key, desc);
+      const existingDesc = this.#usage.getDescriptions()[key];
+      if (!existingDesc || typeof desc === 'string') {
+        this.describe(key, desc);
+      }
+
       if (opt.hidden) {
         this.hide(key);
       }
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -1068,8 +1068,11 @@ export class YargsInstance {
       }
 
       const desc = opt.describe || opt.description || opt.desc;
-      const existingDesc = this.#usage.getDescriptions()[key];
-      if (!existingDesc || typeof desc === 'string') {
+      const descriptions = this.#usage.getDescriptions();
+      if (
+        !Object.prototype.hasOwnProperty.call(descriptions, key) ||
+        typeof desc === 'string'
+      ) {
         this.describe(key, desc);
       }
 
--- a/test/usage.cjs
+++ b/test/usage.cjs
@@ -3968,7 +3968,6 @@ describe('usage tests', () => {
           .strict()
           .parse()
       );
-      console.log(r);
       r.errors[0]
         .split('\n')
         .should.deep.equal([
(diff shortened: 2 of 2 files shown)


test/usage.cjs
```diff
@@ -3968,7 +3968,6 @@ describe('usage tests', () => {
           .strict()
           .parse()
       );
-      console.log(r);
```
jly36963, line 3971: [same](https://github.com/yargs/yargs/pull/2157/files)

lib/yargs-factory.ts
```diff
@@ -1044,7 +1044,11 @@ export class YargsInstance {
       }
 
       const desc = opt.describe || opt.description || opt.desc;
-      this.describe(key, desc);
+      const existingDesc = this.#usage.getDescriptions()[key];
+      if (!existingDesc || typeof desc === 'string') {
```
jly36963, line 1048: nit:

We've had bugs pop up in the past because we're checking truthiness of the value rather than if the key is set or not.

Could it be written this way instead?

```js
const descriptions = this.#usage.getDescriptions();
if (
  !Object.prototype.hasOwnProperty.call(descriptions, key) ||
  typeof desc === 'string'
) {
  this.describe(key, desc);
}
```

jly36963:
Left one nit; otherwise, it looks great to me 👍

bcoe:
@lukekarrys looks good to me, mind addressing @jly36963's one nit, and we'll work on merging?

lukekarrys:
@jly36963 @bcoe I pushed a fixup commit addressing the feedback, and I believe this should be good now

jly36963:
Looks good to me 👍

bcoe:
@lukekarrys thank you for the contribution, apologies for the terribly slow review.

Issue #2178, opened 2022-05-06
[Bug]: Infinite loop on bad $LANG


nmattia:
This was originally report to jest, but it appears the issue is related to yargs. I've copied the text and adapted a bit from the [original ticket](https://github.com/facebook/jest/issues/12811).

---

### Version

Jest: 26.0.24, 28.0.3
yargs: 16.0.4

### Steps to reproduce

```
$ LANG=.UTF-8 npx jest --help
RangeError: Maximum call stack size exceeded
    at h (/Users/nicolas/proejct/.npm/_npx/b8d86e6551a4f492/node_modules/yargs/build/index.cjs:1:1423)
    at Xt.locale (/Users/nicolas/proejct/.npm/_npx/b8d86e6551a4f492/node_modules/yargs/build/index.cjs:1:36561)
    at Xt.[guessLocale] (/Users/nicolas/proejct/.npm/_npx/b8d86e6551a4f492/node_modules/yargs/build/index.cjs:1:46290)
    at Xt.locale (/Users/nicolas/proejct/.npm/_npx/b8d86e6551a4f492/node_modules/yargs/build/index.cjs:1:36664)
    at Xt.[guessLocale] (/Users/nicolas/proejct/.npm/_npx/b8d86e6551a4f492/node_modules/yargs/build/index.cjs:1:46290)
    at Xt.locale (/Users/nicolas/proejct/.npm/_npx/b8d86e6551a4f492/node_modules/yargs/build/index.cjs:1:36664)
    at Xt.[guessLocale] (/Users/nicolas/proejct/.npm/_npx/b8d86e6551a4f492/node_modules/yargs/build/index.cjs:1:46290)
    at Xt.locale (/Users/nicolas/proejct/.npm/_npx/b8d86e6551a4f492/node_modules/yargs/build/index.cjs:1:36664)
    at Xt.[guessLocale] (/Users/nicolas/proejct/.npm/_npx/b8d86e6551a4f492/node_modules/yargs/build/index.cjs:1:46290)
    at Xt.locale (/Users/nicolas/proejct/.npm/_npx/b8d86e6551a4f492/node_modules/yargs/build/index.cjs:1:36664)
```

### Expected behavior

I expect `jest` to either warn, raise a proper error, or just continue if `$LANG` is bad.

### Actual behavior

It appears to enter an infinite loop that blows the stack

### Environment

```shell
$ npx envinfo --preset jest
Need to install the following packages:
  envinfo
[...]

PR #2179, opened 2022-05-06
fix: prevent infinite loop with locale

jly36963:
Addresses: https://github.com/yargs/yargs/issues/2178

## Description

locale and kGuessLocale can enter an infinite loop

## Details

There's regex logic that turns '.UTF-8' into '', and downstream there is logic checking if that value is falsey (instead of undefined).

As a result, kGuessLocale and locale call each other infinitely.

## Todo

```js
this.locale(locale.replace(/[.:].*/, ''))
```
This line is what converts `'.UTF-8'` into `''`. I don't know if my fix is sufficient or if the real problem is here.

### DIFF
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -871,7 +871,7 @@ export class YargsInstance {
   }
   locale(locale?: string): YargsInstance | string {
     argsert('[string]', [locale], arguments.length);
-    if (!locale) {
+    if (locale === undefined) {
       this[kGuessLocale]();
       return this.#shim.y18n.getLocale();
     }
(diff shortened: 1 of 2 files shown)


lib/yargs-factory.ts
```diff
@@ -871,7 +871,7 @@ export class YargsInstance {
   }
   locale(locale?: string): YargsInstance | string {
     argsert('[string]', [locale], arguments.length);
-    if (!locale) {
+    if (locale === undefined) {
```
jly36963, line 874: `''` is falsey, and it causes the two functions to infinitely call each other

bcoe (approved):

PR #2151, opened 2022-03-23
fix: import yargs/yargs in esm projects

hongaar:
When using `import yargs from "yargs/yargs"` in an ESM project (`"type": "module"` in package.json), Node.js will throw this error:

```
internal/process/esm_loader.js:74
    internalBinding('errors').triggerUncaughtException(
                              ^

TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension "" for /home/runner/work/bandersnatch/bandersnatch/tests/smoke/esm/node_modules/yargs/yargs
    at new NodeError (internal/errors.js:322:7)
    at Loader.defaultGetFormat [as _getFormat] (internal/modules/esm/get_format.js:71:15)
    at Loader.getFormat (internal/modules/esm/loader.js:105:[42](https://github.com/hongaar/bandersnatch/runs/5661229728?check_suite_focus=true#step:8:42))
    at Loader.getModuleJob (internal/modules/esm/loader.js:2[43](https://github.com/hongaar/bandersnatch/runs/5661229728?check_suite_focus=true#step:8:43):31)
    at async ModuleWrap.<anonymous> (internal/modules/esm/module_job.js:78:21)
    at async Promise.all (index 3)
    at async link (internal/modules/esm/module_job.js:83:9) {
  code: 'ERR_UNKNOWN_FILE_EXTENSION'
```

To solve this, add a new `yargs.mjs` file and update the `./yargs` export in the package.json.

Confirmed that this works for a simple ESM project where `yargs` is a transitive dependency: https://github.com/hongaar/bandersnatch/pull/392 (in `tests/smoke/esm`)

Related to #2014

PR #2185, opened 2022-05-12
fix: add missing entries to published files

hongaar:
Due to the `yargs.mjs` file not being published, the fix merged with #2151 is not effective yet. This change should fix that.

```
Cannot find module './node_modules/yargs/yargs.mjs' imported from ./cli.js
```

_Edit:_ Also added missing `browser.d.ts`

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
   ],
   "files": [
     "browser.mjs",
+    "browser.d.ts",
     "index.cjs",
     "helpers/*.js",
     "helpers/*",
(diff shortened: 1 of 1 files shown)


bcoe (approved):

bcoe:
@hongaar thank you for catching this :+1: 

aside, what do you think of making yargs ESM only, when we drop Node 12?

hongaar:
@bcoe reading some articles on the subject, it looks like this will be a breaking change since consuming ESM modules from CJS can't use `require()`. https://twitter.com/bradleymeck/status/1346824835494719490

PR #2189, opened 2022-05-16
fix: config array combine logic

jly36963:
## Description

I added a unit test to make sure that all the different scenarios around combining arrays work:
- `combine-array: true`
- config
- defaults
- positionals and options

I don't think it's possible to cover every scenario. 

More specifically, my problem is in `postProcessPositionals`. 
I need to know which values in `argv[key]` (options, config, and defaults), and `parsed.argv[key]` (positionals, config, and defaults) are passed by the user. 

Within the context of `postProcessPositionals`:
- I can figure out which values in `parsed.argv[key]` are positionals using `positionalMap`
- I can't figure out which values in `argv[key]` came from user-provided options. (At that moment, `argv` has values from the first run through yargs-parser.) I can compare `options.configObjects` and `options.default` to `argv[key]`, and remove them if they match exactly. The problem is, if the user provided the same values as the default or config, I would accidentally remove those values (because I don't think I can tell the difference).
- `combine-array: true` adds more complexity, as I would need to make sure that there is only one copy of the default values (but both `argv[key]` and `parsed.argv[key]` will both have those values appended to the end.)

## Questions

Do you have an idea how to make this work without a lot of changes to the parser?

I saw that you were experimenting with using parseArgs instead of yargs-parser. Would that change help fix any of the pain points around array options / variadic args? 

I feel like many of the bugs around array values come from the first parser run being positional-agnostic, and the second parser run being option-agnostic. Does that make sense?

## Different approach

At first I wanted to be flexible with a user providing both positional and optional args under the same namespace, but it causes unnecessary complexity and buggy behavior to try and support both. Would it be too heavy-handed to give a warning or throw an error if someone does this? Would it be a more reasonable solution?

### DIFF
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -638,17 +638,31 @@ export class CommandInstance {
           // any new aliases need to be placed in positionalMap, which
           // is used for validation.
           if (!positionalMap[key]) positionalMap[key] = parsed.argv[key];
+
           // Addresses: https://github.com/yargs/yargs/issues/1637
-          // If both positionals/options provided,
-          // and no default or config values were set for that key,
-          // and if at least one is an array: don't overwrite, combine.
+          // In most cases, positionals will overwrite existing values on argv.
+          // They will be combined if both positionals/options are provided, and at least one is array type
           if (
-            !this.isInConfigs(yargs, key) &&
-            !this.isDefaulted(yargs, key) &&
             Object.prototype.hasOwnProperty.call(argv, key) &&
             Object.prototype.hasOwnProperty.call(parsed.argv, key) &&
             (Array.isArray(argv[key]) || Array.isArray(parsed.argv[key]))
           ) {
+            if (this.isInConfigs(yargs, key)) {
+              const shouldCombineArrays =
+                yargs.getInternalMethods().getParserConfiguration()[
+                  'combine-arrays'
+                ] === true;
+              if (shouldCombineArrays) {
+                // TODO: remove config values from argv[key] and parsed.argv[key], then re-add one set
+              } else {
+                // TODO: remove config values from argv[key] and parsed.argv[key],
+                // unless those values were specifically passed as options/positionals.
+              }
+            }
+            if (!this.isDefaulted(yargs, key)) {
+              // TODO: remove default values from argv[key] and parsed.argv[key],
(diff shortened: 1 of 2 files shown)


lib/command.ts
```diff
@@ -638,17 +638,31 @@ export class CommandInstance {
...
           if (!positionalMap[key]) positionalMap[key] = parsed.argv[key];
+
           // Addresses: https://github.com/yargs/yargs/issues/1637
-          // If both positionals/options provided,
-          // and no default or config values were set for that key,
-          // and if at least one is an array: don't overwrite, combine.
+          // In most cases, positionals will overwrite existing values on argv.
+          // They will be combined if both positionals/options are provided, and at least one is array type
           if (
-            !this.isInConfigs(yargs, key) &&
```
jly36963, line 646: These two lines are problematic:
if a key has default values, they will be excluded. This means that option values will be excluded when used with default values. The same will happen with config values (which is only okay when `combine-arrays` is false). 

This is the bug you saw the other day

lib/command.ts
```diff
@@ -638,17 +638,31 @@ export class CommandInstance {
...
             Object.prototype.hasOwnProperty.call(parsed.argv, key) &&
             (Array.isArray(argv[key]) || Array.isArray(parsed.argv[key]))
           ) {
+            if (this.isInConfigs(yargs, key)) {
+              const shouldCombineArrays =
+                yargs.getInternalMethods().getParserConfiguration()[
+                  'combine-arrays'
+                ] === true;
+              if (shouldCombineArrays) {
+                // TODO: remove config values from argv[key] and parsed.argv[key], then re-add one set
```
jly36963, line 656: Instead of excluding keys with defaults/config from the combining logic, I would need to clean them up here and then combine them.

This part is tricky, as I can distinguish between user-provided values and default/config values in most cases (but not all).

jly36963:
I anticipate that an extensive amount of work is required to address this bug, as it would require rethinking how yargs-parser handles combinations of options/positionals/defaults/config/etc. 

I don't want to put a lot of time into this, just to find out that I've broken existing functionality or that I've made changes that don't align with the yargs maintainers' vision for the project.

If at some point in the future I find a more straightforward solution, or have someone to work on this with, I might pick it up again

Issue #2200, opened 2022-06-23
Provide an environment variable to disable columns wrapping


jgoux:
Hello,

I'm currently writing a little tool for generating a CLI reference documentation in markdown from a CLI built with yargs. (I recursively call the `--help` of all the commands/subcommands and parse the result).

The default columns wrapping with `.wrap()` is annoying when parsing the output of all the `--help` commands. I was wondering if it could be possible to add a built-in `YARGS_DISABLE_WRAP` that would set `.wrap(null)` for any yargs CLI?

It would help a ton with the kind of tool I'm building and I could apply myself instead of asking the users to provide such option.

bcoe:
@jgoux I would happily take this PR, and am excited to see the tool you're building.

PR #2210, opened 2022-07-10
feat(usage): add YARGS_DISABLE_WRAP env variable to disable wrap

jgoux:
Resolves https://github.com/yargs/yargs/issues/2200

I had to modify the test `should wrap based on window-size if no wrap is provided` because the actual length of the non-wrapped output is `5` and not `4`. 👍

### DIFF
--- a/lib/usage.ts
+++ b/lib/usage.ts
@@ -164,7 +164,7 @@ export function usage(yargs: YargsInstance, shim: PlatformShim) {
   };
 
   function getWrap() {
-    if (process.env.YARGS_DISABLE_WRAP) {
+    if (shim.getEnv('YARGS_DISABLE_WRAP')) {
       return null;
     }
     if (!wrapSet) {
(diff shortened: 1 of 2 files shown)


bcoe (approved):

lib/usage.ts
```diff
@@ -164,6 +164,9 @@ export function usage(yargs: YargsInstance, shim: PlatformShim) {
   };
 
   function getWrap() {
```
bcoe, line 166: So that we can maintain our coverage #s, I would do this:

```js
this.getWrap = () => {
  ...
}
```

And then your test can simply be:

```js
const yargs = yargs().wrap(99);
process.env.YARGS_DISABLE_WRAP = 'true';
expect(yargs.getInternalMethods().getUsageInstance().getWrap()).to.equal(null);
```

bcoe (changes requested):
Thanks for this contribution, left a suggestion as to how we could potentially keep 100% coverage, and avoid skipping your test.

lib/usage.ts
jgoux: I made the changes 👍

bcoe (approved):

PR #2214, opened 2022-07-19
feat: add fig completions

### LINKED ISSUE #2126: Integrate with Fig?
[Fig](https://www.fig.io) is a very useful tool for autocompletion in the terminal on macOs. They [currently have integrations for autocompletion with popular JS CLI libraries](https://fig.io/docs/guides/autocomplete-for-teams), it would be great to add Yargs to that list.

I therefore wanted to start the conversation of whether this is something Yargs would consider supporting and which approach would be appropriate. I spoke with some of the Fig team on Discord and I believe there are 2 possible approaches:
- Modify the current completions (bash/zsh) functionality so it uses public APIs. Meaning Fig could produce a third-party integration that uses these same APIs. (Currently seems to be using private internal APIs)
- Support fig internally as a 1st party API, perhaps `.completion([cmd], [description], [fn], [options])` where options can specific which completions to generate, e.g `{fig: true, bash: true}` or some-such.

cc @fedeci

fedeci:
Closes #2126 
This adds fig's completions.
I'll wait for a review and add more tests if you think this may get accepted so let me know!

TODOs discovered while adding tests: 
- [x] ignore required positionals/options when running the fig completion generation cmd
- [x] respect `options.configuration['boolean-negation']`
- [x] top level options/positionals added through `.option(s)` are not added
- [x] Async builders
- [ ] Only print completion spec from the position of the command (not parent ones)
- [x] support implies and conflicts


bcoe:
@fedeci thank you for the contribution. At a glance, this is looking good to me, but it will take me a bit of time to review since it's a significant PR.

fedeci:
@bcoe don't worry! I'll implement more test cases in the meanwhile, and fix the CI :)

lib/yargs-factory.ts
```diff
@@ -2135,6 +2146,19 @@ export class YargsInstance {
...
+        ) {
+          if (this.#exitProcess) setBlocking(true);
+          this.showFigCompletion();
+          this.exit(0);
+          return this[kPostProcess](
+            argv,
+            populateDoubleDash,
+            !!calledFromCommand,
+            false
+          );
```
fedeci, line 2161: For reviewers: here I am not sure this is the correct way, I just want to bypass all the middleware/validation logic when the generate fig completion command is run and then exit

fedeci: I changed it in [`31a9467` (#2214)](https://github.com/yargs/yargs/pull/2214/commits/31a9467e036f1d4d7a248bef4b87161cc5a747ed)

test/fig-completion.cjs
```diff
@@ -0,0 +1,1375 @@
...
+        '    }',
+        '  ]',
+        '};',
+        '',
+        'export default completionSpec;',
+        '',
+      ]);
+  });
+
+  // TODO(fedeci): not working properly
```
fedeci, line 359: This change is not required by fig right now so we can keep it for a second PR. 
It involves getting the correct yargs "subinstance" and I found that pretty tricky. If anyone can point me in the right direction I can eventually make the change directly in this PR.

fedeci (commented):
It's ready for review now!

bcoe (commented):
Giving this an initial read, one thought on my mind, is it might be better for us to add functionality for adding a completion plugin as a module, vs., adding a complex completion directly like fig (since this adds a lot of code).

What if it looked something like this:

```
import figCompletions from '@yargs/fig-completions';
yargs.completion(figCompletions);
```

bcoe:
@fedeci I love the idea of adding support for fig, but the amount of code required to do so (compared to bash, fish, etc) is significant.

How do you feel about us adding a helper to configure external completion helpers, you could then publish the fig completion as a separate module (I'd even be willing to add it to the yargs org if you like) and we could document how to configure fig completions in yargs docs.

fedeci:
Sounds great. Please let me know if we can do something to help having the "completions API" available!
/cc @mschrage

bcoe:
@fedeci I added a tracking feature here:

https://github.com/yargs/yargs/issues/2298

I could use help figuring out what the interface should be. What hooks would you need to be able to provide fig completions?

Issue #2186, opened 2022-05-12
Deno: Use `globalThis` instead of `window`


rracariu:
Currently using Yarg in a Deno Worker would not run because there is no `window` global in Workers.

To make this work on all cases using `globalThis` would be better, see:

https://github.com/yargs/yargs/blob/fcb4d38016f2877fa95dcfc4cefc4b743b93795e/lib/platform-shims/deno.ts#L94

bcoe:
Thank you for the suggestion :+1: , would happily take a patch.

PR #2215, opened 2022-07-27
fix(deno): use 'globalThis' instead of 'window' (#2186)

rivy:
Minor PR which implements the fix described in #2186.


bcoe (approved):
Thank you for the fix 👍

Issue #2236, opened 2022-09-15
common-types fails typechecking


timiyay:
I've cloned the repo and run `npm install`. As part of this, it runs `rimraf build && tsc` for typechecking.

For a freshly-cloned repo, this fails with:

```
lib/typings/common-types.ts:65:22 - error TS2769: No overload matches this call.
  Overload 1 of 2, '(o: {}): string[]', gave the following error.
    Argument of type 'T' is not assignable to parameter of type '{}'.
  Overload 2 of 2, '(o: object): string[]', gave the following error.
    Argument of type 'T' is not assignable to parameter of type 'object'.

65   return Object.keys(object) as (keyof T)[];
                        ~~~~~~

  lib/typings/common-types.ts:64:28
    64 export function objectKeys<T>(object: T) {
                                  ~
    This type parameter might need an `extends {}` constraint.
  lib/typings/common-types.ts:64:28
    64 export function objectKeys<T>(object: T) {
                                  ~
    This type parameter might need an `extends object` constraint.


Found 1 error in lib/typings/common-types.ts:65
```

The issue was encountered on macOS Monterey 12.6, using NodeJS 16.17.0 and npm 8.19.2

timiyay:
Fix offered in https://github.com/yargs/yargs/pull/2237.

timiyay:
I think the reason this issue has come up is because this project doesn't have a lockfile. As such, the semver range of `typescript` allows newer minor and patch versions to be installed over time. Assumedly, a more-recent version of Typescript has stricter typechecking than its predecessors.

What's the reasoning behind no lockfile in this project?

PR #2237, opened 2022-09-15
fix(typescript): address warning with objectKeys

timiyay:
This PR fixes https://github.com/yargs/yargs/issues/2236, by following the suggestion of the Typescript compiler.

### DIFF
--- a/lib/typings/common-types.ts
+++ b/lib/typings/common-types.ts
@@ -61,7 +61,7 @@ export function assertSingleKey(
 /**
  * Typing wrapper around Object.keys()
  */
-export function objectKeys<T>(object: T) {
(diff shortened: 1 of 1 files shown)


lib/typings/common-types.ts
```diff
@@ -61,7 +61,7 @@ export function assertSingleKey(
 /**
  * Typing wrapper around Object.keys()
  */
-export function objectKeys<T>(object: T) {
+export function objectKeys<T extends {}>(object: T) {
```
timiyay, line 64: I could've extended from `{}` or `object` here. I had no preference, happy to change as needed.

bcoe (approved):

bcoe:
@timiyay thank you!

PR #2261, opened 2022-11-03
:bug: FIX: esm version to build with vite

jycouet:
I'm using vite & pnpm in a project to bundle everything (a lot of libs are involved)
But during the `build` process I get:

```bash
Error: 'default' is not exported by .../y18n@5.0.8/node_modules/y18n/build/lib/index.js, 
imported by .../yargs@17.6.2/node_modules/yargs/lib/platform-shims/esm.mjs
```

I think that this `import * as y18n from 'y18n'` would fix the issue.

✅ Tests are still passing.


bcoe:
@jycouet the ESM integration tests appear to be surfacing that this is a breaking change.

jycouet:
I'll have to update my `yargs` usage and see if with the latest, I still have the need for this.
I'll close the PR for now.

Thank you

Issue #2259, opened 2022-11-03
Missing entries in locales for unknown command


shadowspawn:
I noticed that only the french translation has an entry for `Unknown command: %s`

- https://github.com/yargs/yargs/blob/f727e71a748847cb32aba329022445ec85fef4f9/locales/fr.json#L36

The error message was added in #1540 (see also #1560).

I think the string should appear in `en.json` as the template for the translations?

Issue #1560, opened 2020-02-19
Missing French translation for "Unknown command: %s"


mleguen:
New translation to be added due to #1540:
```json
"Unknown command: %s": {
    "one": "Commande inconnue : %s",
    "other": "Commandes inconnues : %s"
  },
```

_Originally posted by @mleguen in https://github.com/yargs/yargs/pull/1557_

PR #2262, opened 2022-11-04
chore: add en strings for unknown command

shadowspawn:
Adds the unknown command strings to the `en` locale. This has no affect on the output, since they are the same as in the code! The intent is so people updating or creating other locales can see the strings available for localising.

Placing the strings after unknown argument in the same place as in the `fr` locale from #1560, which looks a good place.

Closes: #2259

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -33,6 +33,10 @@
     "one": "Unknown argument: %s",
     "other": "Unknown arguments: %s"
   },
+  "Unknown command: %s": {
+    "one": "Unknown command: %s",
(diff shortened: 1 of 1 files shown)


bcoe (approved):

bcoe:
@shadowspawn thank you 👏 , I will need to fix up the npm installation before we can merge (looks like 9.x doesn't like Node 12).

Perhaps over holidays this year I'll drop Node 12, and simplify yargs' build.

bcoe:
@shadowspawn thanks for the contribution \o/

Issue #2268, opened 2022-11-21
The completion for zsh does not align the option and its alias.


mshrtsr:
Hi there, and thanks for the great package.  
I encountered following unexpected behavior.

## Environment

- yargs Version: 17.6.2
- Operating System: macOS 12.6
- Node Version: v16.17.0
- Link to reproduction: https://github.com/tasshi-playground/repro-yargs-generate-broken-completion-with-alias

## Summary

When we generate the completion script for zsh, with a option with an alias, the completion shown looks incorrect.
That's because the description of key from alias is missing in the generated completion script for zsh.

```shell
 -- values --
--bar          -- Bar option
--foo          -- Foo option
--help         -- Show help
--version      -- Show version number
-B         -F  --
```

## Reproduction

I created the minimum reproduction in https://github.com/tasshi-playground/repro-yargs-generate-broken-completion-with-alias .

```shell
# If you don't enable zsh-completion, you need to enable it first.
# Clone and configure repository
$ git clone git@github.com:tasshi-playground/repro-yargs-generate-broken-completion-with-alias.git
$ cd repro-yargs-generate-broken-completion-with-alias
$ npm ci
$ npm run build
$ ./cli.js completion > /path/to/your/fpath/_cli.js
# Reload the shell
```

In the repository, I specified the options and aliases as follows.

```javascript
yargs
  .option("foo", {
    describe: "Foo option",
    alias: "F",
    type: "string",
  })
  .option("bar", {
    describe: "Bar option",
    alias: "B",
    type: "string",
  })
  .completion()
  .help().argv;
```

### Expected Behavior

The output when I strike the **TAB** should be the following.

```shell
./cli.js - #Strike TAB
 -- values --
--bar      -B  -- Bar option
--foo      -F  -- Foo option
--help         -- Show help
--version      -- Show version number
```

### Actual Behavior

The actual output is the following.
The aliases (`-B` and `-F`) are aligned to same line with no descriptions.

```shell
./cli.js - #Strike TAB
 -- values --
--bar          -- Bar option
--foo          -- Foo option
--help         -- Show help
--version      -- Show version number
-B         -F  --
```

## Cause

When I request a completion with striking the **TAB** key, completion function calls the original file with `--get-yargs-completions` option.
It is specified in completion script as follows.

```shell
_cli.js_yargs_completions()
{
  local reply
  local si=$IFS
  IFS=$'
[...]

PR #2269, opened 2022-11-21
fix: copy the description of the option to its alias in completion

mshrtsr:
This PR fixes https://github.com/yargs/yargs/issues/2268.  
I find the original option for the alias and copy its description to the alias.
As a result, the option and its alias are collected together and aligned in the completion.

### DIFF
--- a/lib/completion.ts
+++ b/lib/completion.ts
@@ -258,7 +258,9 @@ export class Completion implements CompletionInstance {
     if (!this.zshShell) {
       completions.push(dashes + key);
     } else {
-      const desc = descs[key] || '';
+      const aliasKey = this?.aliases?.[key]?.[0];
+      const descFromAlias = aliasKey ? descs[aliasKey] : undefined;
+      const desc = descs[key] ?? descFromAlias ?? '';
       completions.push(
         dashes +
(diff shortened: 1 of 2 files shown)


lib/completion.ts
```diff
@@ -258,7 +258,9 @@ export class Completion implements CompletionInstance {
     if (!this.zshShell) {
       completions.push(dashes + key);
     } else {
-      const desc = descs[key] || '';
+      const aliasKey = this?.aliases?.[key]?.[0];
```
bcoe, line 261: what about cases where there are multiple aliases associated with a key? should this be a loop?

bcoe (changes requested):

lib/completion.ts
mshrtsr: Oh, Thank you. Yes, sometimes the `this?.aliases?.[key]?.[0]` will be another alias without a description.
I will make it a loop.

mshrtsr:
@bcoe 
Thank you.
Can I ask you for the review again?

bcoe (approved):

bcoe:
@mshrtsr thank you for the contribution :ok_hand:

PR #2281, opened 2022-12-22
Add support for completions in `fish` shell + a few dev scripts to ease workflow

jglanz:
## added support for completions in `fish` shell

To use with `fish`, in fish shell, run `<app-name> completion >> ~/.config/fish/config.fish`

## add scripts making `dev` a bit easier


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "yargs",
-  "version": "17.6.2",
+  "version": "17.6.3-0",
```
bcoe, line 3: The version # is managed by automation, if you don't mind rolling this basck.

package.json
```diff
@@ -94,10 +95,13 @@
     "coverage": "c8 report --check-coverage",
     "prepare": "npm run compile",
     "pretest": "npm run compile -- -p tsconfig.test.json && cross-env NODE_ENV=test npm run build:cjs",
-    "compile": "rimraf build && tsc",
+    "compile": "rimraf build && tsc -b tsconfig.json",
+    "compile:watch": "npm run compile -- --watch",
```
bcoe, line 99: Is `--watch` built into newer versions of npm?

bcoe (changes requested):
Left a couple nits, looks mostly good to me though. Thank you for the contribution.

jglanz:
For sure, i'll get it cleaned up in the next day or so.  Of all the arg parsing frameworks, nothing compares.  Happy to contribute moving forward as well.

package.json
jglanz: running tests now, should have the PR updated within an hr or two at the latest

jglanz:
@bcoe hey - sorry when i rebased I messed up some ordering, everything has been addressed and all tests pass.  If you'd like me to scrap the PR and do up a fresh one just say so, but everything is correct & complete now.  Apologies, but it is all correct now, just LMK

package.json
jglanz: All tests pass - as mentioned, happy to scrap the PR and do a fresh one for cleanliness, but either way - its good to go and rebased to the tip to include the other completion fix from yesterday

jglanz:
> Left a couple nits, looks mostly good to me though. Thank you for the contribution.

@bcoe Happy New Year!  I resolved the nits, can you give it a quick review? I'd love to get this landed as I'm using it at work ;)

bcoe:
@jglanz your work looks good to me, but there seems to be a regression happening in Node 16. Any thoughts?

bcoe (changes requested):
Okay, I think I got tests passing again (there was a regression introduced in Node 16).

The one thing blocking this PR is that I believe coverage drops a little bit, mind taking a look? we might just be missing one test.

package.json
shadowspawn: 1) It does involve some copy-paste repetition, but I prefer the pattern you used for `build:cjs:watch` where you copied the command so you could add the watch.
2) the `dev` script does a compile, then runs compile-watch, which does a compile, which includes a rimraf which runs in parallel to `build:cjs:watch`. I'll make a comment there too...

So if the combination of comments makes sense, I suggest

```suggestion
    "compile:watch": "tsc --watch -p tsconfig.json",
```

package.json
```diff
@@ -93,11 +94,14 @@
...
-    "pretest": "npm run compile -- -p tsconfig.test.json && cross-env NODE_ENV=test npm run build:cjs",
-    "compile": "rimraf build && tsc",
+    "pretest": "npm run compile && tsc -p tsconfig.test.json && cross-env NODE_ENV=test npm run build:cjs",
+    "compile": "rimraf build && tsc -p tsconfig.json",
+    "compile:watch": "npm run compile -- --watch",
     "postcompile": "npm run build:cjs",
     "build:cjs": "rollup -c rollup.config.cjs",
+    "build:cjs:watch": "rollup -w -c rollup.config.cjs",
     "postbuild:cjs": "rimraf ./build/index.cjs.d.ts",
+    "dev": "npm run compile && run-p compile:watch build:cjs:watch",
```
shadowspawn, line 104: I think the current scripts mean this will run compile twice, and invokes rimraf twice. Maybe do the rimraf up front and remove it from the watch scripts. (Disclaimer: making these suggestions online without trying it for real!)
```suggestion
    "dev": "rimraf build && run-p compile:watch build:cjs:watch",
```

shadowspawn:
(Added some light comments on the helper scripts. Inspired by this, I should start using them instead of forgetting to recompile after code changes!)

lib/completion-templates.ts
```diff
@@ -47,3 +47,23 @@ _{{app_name}}_yargs_completions()
 compdef _{{app_name}}_yargs_completions {{app_name}}
 ###-end-{{app_name}}-completions-###
 `;
+
+export const completionFishTemplate = `### {{app_name}} completion - begin. generated by omelette.js ###
```
shadowspawn, line 51: Although we may well wish to acknowledge Omelette somewhere, it seems misleading to say "generated by omelette.js" when it was generated by Yargs?

lib/completion.ts
```diff
@@ -45,6 +51,10 @@ export class Completion implements CompletionInstance {
       (this.shim.getEnv('SHELL')?.includes('zsh') ||
         this.shim.getEnv('ZSH_NAME')?.includes('zsh')) ??
       false;
+    this.fishShell = this.shim.getEnv('SHELL')?.includes('fish') ?? false;
+    this.bashShell =
```
shadowspawn, line 55: I get the intent, but the combination of the name `bashShell` and the use in the later login block are confusing. I'll comment more there.

lib/completion.ts
```diff
@@ -92,8 +102,11 @@ export class Completion implements CompletionInstance {
       this.usage.getCommands().forEach(usageCommand => {
         const commandName = parseCommand(usageCommand[0]).cmd;
         if (args.indexOf(commandName) === -1) {
-          if (!this.zshShell) {
+          if (this.bashShell) {
```
shadowspawn, line 105: It is hard to reason about what shell gets handled where here. The original code had basically zsh and not zsh. I suggest keep the same sort of logic, so it is a more obvious diff, and easier to understand what gets handled where.

```
if (this.fishShell) {
} else if (this.zshShell) {
} else {
// bash et al
}

**Comment by shadowspawn** [comment]
(Darn you GitHub, why are you showing me changesets that have landed in main! 😡 )

File: package.json (hunk 2 above)
Comment (shadowspawn) [reply package.json:99]: I realised I am comfortable passing command-line flags into `npm run-script` from the command-line, just don't normally do it between run-scripts, so weakening my suggestion above. Just an alternative approach!

File: lib/completion-templates.ts [hunk 7]
```diff
@@ -47,3 +47,23 @@ _{{app_name}}_yargs_completions()
...
+#
+# yargs command completion script
+#
+# Installation: {{app_name}} {{completion_command}} >> ~/.config/fish/config.fish.
```
shadowspawn, line 55: I noticed a completions folder when I installed fish to try this out. Would it be more appropriate to suggest writing the completion into `~/.config/fish/completions/{{app_name}}.fish`?

https://fishshell.com/docs/current/completions.html#where-to-put-completions

package.json
shadowspawn: What is `run-p`? (Perhaps run-parallel from some package you have installed?)

shadowspawn:
I checked my suspicion that the current `dev` run-script not going to work as intended. I replaced the actions with echo statements so I could see what was going on to more easily see what got called. This exposed the `postcompile` script getting in on the action too! (The existing script setup is quite subtle!)

```
    "compile": "echo run-script-compile",
    "compile:watch": "npm run compile -- --watch",
    "postcompile": "npm run build:cjs",
    "build:cjs": "echo run-script-build:cjs",
    "build:cjs:watch": "echo run-script-build:cjs:watch",
    "postbuild:cjs": "rimraf ./build/index.cjs.d.ts",
    "dev": "npm run compile && npm run compile:watch && npm run build:cjs:watch",
```

```
% npm run dev

> yargs@17.7.1 dev
> npm run compile && npm run compile:watch && npm run build:cjs:watch


> yargs@17.7.1 compile
> echo run-script-compile

run-script-compile

> yargs@17.7.1 postcompile
> npm run build:cjs


> yargs@17.7.1 build:cjs
> echo run-script-build:cjs

run-script-build:cjs

> yargs@17.7.1 postbuild:cjs
> rimraf ./build/index.cjs.d.ts


> yargs@17.7.1 compile:watch
> npm run compile -- --watch


> yargs@17.7.1 compile
> echo run-script-compile --watch

run-script-compile --watch

> yargs@17.7.1 postcompile
> npm run build:cjs


> yargs@17.7.1 build:cjs
> echo run-script-build:cjs

run-script-build:cjs

> yargs@17.7.1 postbuild:cjs
> rimraf ./build/index.cjs.d.ts


> yargs@17.7.1 build:cjs:watch
> echo run-script-build:cjs:watch

run-script-build:cjs:watch

```

shadowspawn:
In retrospect this PR might have been better as two PR, so apologies that I have feedback about both aspects!

The naming I am familiar with from work and I think is more common in the wild is `npm run watch` rather than `npm run dev` to spin up the watches.

For example:
- https://sebhastian.com/npm-run-dev/
- https://stackoverflow.com/questions/58720485/what-does-scripts-npm-run-dev-and-npm-run-watch-is-used-for

shadowspawn:
Are you still interested in working on this @jglanz ? 

I appreciate it has been slow going for you from initial submission, I have new suggestions, and you have it working locally anyway. If you aren't interested then I'll close this and other people may pick up the pieces.

(I personally do not use fish shell, but do want a watch script! I would list you as co-author if I submit a PR for that.)

zezo2019 (approved):

lib/completion-templates.ts
shadowspawn: I checked what omelette does, which uses the lowest

shadowspawn:
This is a couple of years old, and no longer mergeable. Closing as stale.

Thanks for your contributions.

Issue #2247, opened 2022-10-11
Default subcommand: help on error displays repeated main command


xmedeko:
I have a command with subcommands and one of them is default, see
```js
const options = yargs
  .command(['export'], 'Do export', yargs =>
    yargs
      .command(['csv', '*'], 'Export to CSV',
        {
          'value': { demandOption: true }
        })
      .command('html', 'Export to HTML', { 'value': { demandOption: true } })
      .demandCommand(1)
  )
  .command(['import'], 'Do import', { 'value': { demandOption: true } })
  .help().strict(true).demandCommand(1)
  .parse();
```

### Actual

When a user types wrong subcommand, then the help is wrong:
```
$ node index.js export foo

Export to CSV                                                                     
                                                         
Commands:                                                                         
  index.js export export  Do export
  index.js export import  Do import
...
```

### Expected
When I remove `'*'`, i.e. the default command flag, from the `csv` command, then the help is OK:
```
$ node index.js export foo

Do export                                                                       

Commands:
  index.js export csv   Export to CSV
  index.js export html  Export to HTML
...
```

bcoe:
@xmedeko a fix for this should now be available in `yargs@ 17.7.1`.

PR #2303, opened 2023-02-21
fix: address display bug with default sub-commands

### LINKED ISSUE #2291: Using `$0` (a default command) as an alias breaks the usage output for nested commands
We've run into an issue with yargs when using an alias for the default command.

Rather than outputting the correct usage for the subcommand when there's a typo, it instead is outputting the usage for the top level commands, but like they're nested under the subcommand.

We've create a little reproduction - https://github.com/bglimepoint/yargs-alias-help-text-bug-example.

The expected output would be:

```
$ node index.js test testingtypo

index.js test

parent - test description

Commands:
  index.js test testing   testing description
  index.js test testing2  testing description
  index.js test testing3  testing description

Options:
  --version  Show version number  [boolean]
  --help     Show help  [boolean]

Unknown argument: testingtypo
```

Which is what we get when we remove the `aliases: ['$0']` from one of our subcommands; but when using the default command we get the following bug (afaict) instead:

```
$ node index.js test testingtypo
index.js test testing

testing description

Commands:
  index.js test test        parent - test description
  index.js test command2    parent - command2 description
  index.js test completion  generate completion definition script

Options:
  --version  Show version number  [boolean]
  --help     Show help  [boolean]

Unknown argument: testingtypo
```

Notice that the usage there is for the top level things (`completion`, and then the two fake top level commands), not the usage for the `test` subcommand.

The repro repo has the buggy version on the master branch, and the "working" version on the no-alias branch (due to the removal of the alias).

Hopefully we're not just doing something silly :-D

Thanks for your time and efforts with yargs :-)

bcoe:
freeze() was not being called prior to reset(), resulting in display issues for default sub-commands.

Fixes #2291

CC: @Tobbe, @bglimepoint

### DIFF
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -263,6 +263,7 @@ export class CommandInstance {
     if (isCommandBuilderCallback(builder)) {
       // A function can be provided, which builds
       // up a yargs chain and possibly returns it.
+      yargs.getInternalMethods().getUsageInstance().freeze();
       const builderOutput = builder(
         yargs.getInternalMethods().reset(aliases),
         helpOrVersionSet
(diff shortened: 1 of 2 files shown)


shadowspawn (approved):
LGTM, and manually retesting against the example looks good.

shadowspawn:
I did a quick check on #2247 too. Looks good.

Fixes: #2247

Not sure if the `Fixes:` works in later comments, but still serves as a reminder if it does not. 😄

PR #2308, opened 2023-03-10
fix: coerce should play well with parser configuration

### LINKED ISSUE #2307: parserConfiguration "strip-aliased" causes "coerce" not be applied for expanded-camelcase
```bash
$ node -e 'const p = require("yargs")(["--n-e", "1"]).option("n-e", { coerce: n => n + 1}).parserConfiguration({ "strip-aliased": true }).parse(); console.log(p)'
{ _: [], 'n-e': 2, nE: 1, '$0': '' }
```

without is ok
```bash
$ node -e 'const p = require("yargs")(["--n-e", "1"]).option("n-e", { coerce: n => n + 1}).parse(); console.log(p)'
{ _: [], 'n-e': 2, nE: 2, '$0': '' }
```

### LINKED ISSUE #2311: Dash inside option name with parser option strip-dashed yields undefined value
This is a bug report; minimal example of the bug:

```sh
$ node -e "console.log(require('yargs').parserConfiguration({ 'strip-dashed': true}).option('foo-bar', { coerce: (v) => v }).parse(['executable', '--foo-bar', 'hello']))"
```

outputs:

```js
{
  _: [ 'executable' ],
  fooBar: undefined,
  '$0': '',
  'foo-bar': undefined
}
```

This is incorrect. The value of 'fooBar' should be 'hello', not undefined and the 'foo-bar' property should be entirely removed.

This bug does not appear with options not containing a dash:

```sh
node -e "console.log(require('yargs').parserConfiguration({ 'strip-dashed': true}).option('foobar', { coerce: (v) => v }).parse(['executable', '--foobar', 'hello']))"
```

yields the following correct output:

```js
{ _: [ 'executable' ], foobar: 'hello', '$0': '' }
```

## Workaround No 1

Setting 'strip-dashed' to false yields correct behaviour:

```sh
node -e "console.log(require('yargs').parserConfiguration({ 'strip-dashed': false }).option('foo-bar', { coerce: (v) => v }).parse(['executable', '--foo-bar', 'hello']))"
```

resulting js:

```js
{ _: [ 'executable' ], 'foo-bar': 'hello', fooBar: 'hello', '$0': '' }
```

## Woraround No 2

Using an alias also yields less wrong behavior (still no stripping).

```sh
node -e "console.log(require('yargs').parserConfiguration({ 'strip-dashed': true}).option('foobar', { alias: ['foo-bar'], coerce: (v) => v }).parse(['executable', '--foo-bar', 'hello']))"
```

Output:

```js
{
  _: [ 'executable' ],
  foobar: 'hello',
  fooBar: 'hello',
  '$0': '',
  'foo-bar': 'hello'
}
```

shadowspawn:
Relates to #2307 and #2161. Yargs does the coerce separately after yargs-parser, and should both apply the coerce values to the aliases and camel-case variations, but also not add them back if the parser configuration said not to include them in the results. 

Use the fact that yargs-parser has already decided what variations belong in the results, and update just the naming variations already present in the results.

Fixes #2307, fixes #2311

### DIFF
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -406,12 +406,12 @@ export class YargsInstance {
           },
           (result: any): Partial<Arguments> => {
             argv[keys] = result;
-            const stripAliased = yargs
-              .getInternalMethods()
-              .getParserConfiguration()['strip-aliased'];
-            if (aliases[keys] && stripAliased !== true) {
+            // yargs-parser takes into account the parser options like strip-aliased and strip-dashed and camel-case expansion.
+            // Update any naming variations returned by yargs-parser with the coerced value.
+            if (aliases[keys]) {
               for (const alias of aliases[keys]) {
-                argv[alias] = result;
+                if (Object.prototype.hasOwnProperty.call(argv, alias))
+                  argv[alias] = result;
               }
             }
             return argv;
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -380,20 +380,26 @@ export class YargsInstance {
     if (!value) {
       throw new YError('coerce callback must be provided');
     }
+
+    // Handled multiple above, down to one key.
+    const coerceKey = keys;
     // This noop tells yargs-parser about the existence of the option
-    // represented by "keys", so that it can apply camel case expansion
+    // represented by "key", so that it can apply camel case expansion
     // if needed:
-    this.#options.key[keys] = true;
+    this.#options.key[coerceKey] = true;
     this.#globalMiddleware.addCoerceMiddleware(
       (
         argv: Arguments,
         yargs: YargsInstance
       ): Partial<Arguments> | Promise<Partial<Arguments>> => {
-        let aliases: Dictionary<string[]>;
+        // Narrow down the possible keys to the ones present in argv.
+        const coerceKeyAliases = yargs.getAliases()[coerceKey] ?? [];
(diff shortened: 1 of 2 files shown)


shadowspawn:
I noticed a subtle problem that if `strip-dashed` is true then the argv does not contain the parsed option wi the dashed name so coerce is ignored, although the option was present in the input arguments. May or may not attempt to fix that in this PR too... So many permutations!

shadowspawn:
In #1872 `coerce` handling got promoted out of `yargs-parser` to support async, so the current code is relatively new and has been iteratively improving to fully reimplement the original behaviours and coverage.

shadowspawn:
Looking at the `yargs-parser` [implementation](https://github.com/yargs/yargs-parser/blob/3aba24ceaa1a06ceb982c63a06002526d781e826/lib/yargs-parser.ts#L748), the code works with argv and aliases without checking configuration options. That makes me more confident that heading in the right direction in this PR.

bcoe (approved):

bcoe:
@shadowspawn thanks for the contribution 😄

bcoe (approved):

dleavitt:
@shadowspawn @bcoe anything we could do to help get this released? In addition to the good stuff on main, there are some solid-looking PRs waiting in the wings (e.g. #2332) that a release might clear the way for.

Thanks!

shadowspawn:
Note: here is a change to the minimum node version on main: #2384

Nothing to do to help at the moment, thanks.

Issue #2312, opened 2023-03-14
how can I use default command and help simultaneously


PengBoUESTC:
yargs(hideBin(process.argv))
	.command("$0", "init rspack project", async argv => {
          consosle.log('111')
         })
        .help()
        .parse()

if the `builder` is an async function, the help command will run the builder without the help message

shadowspawn:
I have reproduced the problem (calls `exit()` before displaying help) when the default command has an async builder. Looks like a bug.

4i8:
> yargs(hideBin(process.argv)) .command("$0", "init rspack project", async argv => { consosle.log('111') }) .help() .parse()
> 
> if the `builder` is an async function, the help command will run the builder without the help message

just use [clinei](https://www.npmjs.com/package/clinei) 🤣

shadowspawn:
I got a test working, but uncovered an interesting issue in the process. I think the async triggered during the help processing  is not returned as a promise, so calling await on the parse result does not block and allow the help to display before the test completes. I think this PR still makes sense on its own so not considering that a blocker! I'll open a PR to your branch.

shadowspawn:
This has been fixed by #2313

PR #2313, opened 2023-03-14
fix: exit after async handler done

PengBoUESTC:
#2312

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## [17.7.1](https://github.com/yargs/yargs/compare/v17.7.0...v17.7.1) (2023-02-21)
+
+
+### Bug Fixes
+
+* address display bug with default sub-commands ([#2303](https://github.com/yargs/yargs/issues/2303)) ([9aa2490](https://github.com/yargs/yargs/commit/9aa24908ae4e857161d5084613a402f9dc4895a7))
+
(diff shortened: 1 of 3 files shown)


shadowspawn:
Thanks. My suspicion is adding a callback is a work-around for the async handling not being correct. But it does seem to work! It will take me a while to understand the call paths.

PengBoUESTC:
> Thanks. My suspicion is adding a callback is a work-around for the async handling not being correct. But it does seem to work! It will take me a while to understand the call paths.

callback is not a good idea, i did this because `showHelp ` is not declared  as an async function, I don't want to make a break change

shadowspawn:
`.showHelp()` can take a callback already, so do not need to add another one.

```
          this.showHelp(message => {
            this.#logger.log(message);
            this.exit(0);
          });
```

shadowspawn:
Ideally we would like a unit test for any code changes. Are you up for that?

(I have not looked at how to do this. I assume will need a little setup to prepare for an async operation that calls exit!)

PengBoUESTC:
> Ideally we would like a unit test for any code changes. Are you up for that?
> 
> (I have not looked at how to do this. I assume will need a little setup to prepare for an async operation that calls exit!)

it's not easy to give a test case =_=

shadowspawn:
I'll have a look at doing a test.

shadowspawn:
I got a test working, but uncovered an interesting issue in the process. I think the async triggered during the help processing is not returned as a promise, so calling await on the parse result does not block and allow the help to display before the test completes. I think this PR still makes sense on its own so not considering that a blocker! I'll open a PR to your branch.

shadowspawn (approved):
I think this is a simple change that should work same way for sync and more usefully for async.

PR #2316, opened 2023-03-29
docs: Fix example of coerce

regseb:
The first example of [`.coerce(key, fn)`](https://yargs.js.org/docs/#api-reference-coercekey-fn) is invalid: an `await` is used in a non-`async` function.

```js
  .coerce('file', function (arg) {
    return await require('fs').promises.readFile(arg, 'utf8')
  })
  .argv
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -198,13 +198,14 @@ all other modifications, such as [`.normalize()`](#normalize).
 _Examples:_
 
 ```js
-var fs = require('fs').promises
-var argv = await require('yargs/yargs')(process.argv.slice(2))
+import { readFile } from 'node:fs/promises';
+import yargs from 'yargs';
+const argv = await yargs(process.argv.slice(2))
   .coerce('file', async (arg) => {
-    var content = await fs.readFile(arg, 'utf8')
-    return content
+    const content = await readFile(arg, 'utf8');
+    return JSON.parse(content);
   })
-  .argv
+  .parseAsync();
 ```
 
 Optionally `.coerce()` can take an object that maps several keys to their
--- a/docs/api.md
+++ b/docs/api.md
@@ -198,9 +198,10 @@ all other modifications, such as [`.normalize()`](#normalize).
 _Examples:_
 
 ```js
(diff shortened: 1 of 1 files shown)


shadowspawn:
Making the examples correct is a good thing. The example was updated as part of adding async support to `.coerce()` in https://github.com/yargs/yargs/pull/1872 but was not complete, and I would rather make the example work with async than remove the async. 

Would you like to do that?

regseb:
I modified the example to add an `async` / `await`.

shadowspawn:
I like the direction of getting away from `require` embedded in the code.

It is much easier to run the example in a module where top-level await is available. How about about making this example esm? e.g.

```js
import { readFile } from 'node:fs/promises';
import yargs from 'yargs';
const argv = await yargs(process.argv.slice(2))
   .coerce('file', async (arg) => readFile(arg, 'utf8'))
   .parseAsync();
```

regseb:
In your example, the `async` is useless because the function has no `await` ([_require-await_](https://eslint.org/docs/latest/rules/require-await)). It would be useless to add an `await` before the return: `.coerce('file', async (arg) => await readFile(arg, 'utf8'))` ([_no-return-await_](https://eslint.org/docs/latest/rules/no-return-await)).

I propose to add an intermediate variable (which is useless, but there is no ESLint rule to detect it) to have an excuse to use an `await`.

```js
import { readFile } from 'node:fs/promises';
import yargs from 'yargs';
const argv = await yargs(process.argv.slice(2))
  .coerce('file', async (arg) => {
    const content = await readFile(arg, 'utf8');
    return content;
  })
  .parseAsync();
```

shadowspawn:
Good points. I suggest adding `JSON.parse` on the contents. Then the `await ` is needed and the async function makes sense.

regseb:
I added `JSON.parse()` in the example.

shadowspawn (approved):
LGTM

shadowspawn:
Thanks for the PR and conversation @regseb.

bcoe (approved):

PR #2320, opened 2023-04-22
Add watch run-script

shadowspawn:
Inspired by work in #2281, add an npm run-script for "watch".

Both `tsc` and `rollup` support watch.


bcoe (approved):
LGTM!

PR #2322, opened 2023-04-27
    fix: do not crash completion when having negable options

alan-agius4:
Current completion crashes when using ZSH and having negable options.
  
  ```
  [error] TypeError: Cannot read properties of undefined (reading 'find')
      at D.completeOptionKey (//node_modules/yargs/build/index.cjs:1255:20)
      at //node_modules/yargs/build/index.cjs:1191:37
      at Array.forEach (<anonymous>)
      at D.optionCompletions (//node_modules/yargs/build/index.cjs:1184:26)
      at D.defaultCompletion (//node_modules/yargs/build/index.cjs:1158:12)
      at n (//node_modules/yargs/build/index.cjs:1313:23)
      at D.getCompletion (//node_modules/yargs/build/index.cjs:1314:31)
      at te.[runYargsParserAndExecuteCommands] (//node_modules/yargs/build/index.cjs:3161:27)
      at _.parseAndUpdateUsage (//node_modules/yargs/build/index.cjs:426:8)
      at _.applyBuilderUpdateUsageAndParse (//node_modules/yargs/build/index.cjs:412:17)
  ```

### DIFF
--- a/lib/completion.ts
+++ b/lib/completion.ts
@@ -130,9 +130,12 @@ export class Completion implements CompletionInstance {
           !options.hiddenOptions.includes(key) &&
           !this.argsContainKey(args, key, negable)
         ) {
-          this.completeOptionKey(key, completions, current);
-          if (negable && !!options.default[key])
-            this.completeOptionKey(`no-${key}`, completions, current);
+          this.completeOptionKey(
+            key,
+            completions,
+            current,
+            negable && !!options.default[key]
+          );
         }
       });
     }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -81,6 +81,7 @@
     "rollup-plugin-cleanup": "^3.1.1",
     "rollup-plugin-terser": "^7.0.2",
     "rollup-plugin-ts": "^2.0.4",
+    "tslib": "^2.5.0",
```
bcoe, line 84: wondering why tslib was added?

alan-agius4: As per the commit message I was getting `[!] (plugin Typescript) TS2807: This syntax requires an imported helper named '__classPrivateFieldSet' with 5 parameters, which is not compatible with the one in 'tslib'. Consider upgrading your version of 'tslib'.` when running the tests locally.

bcoe (approved):
LGTM thank you for the fix!

PR #2326, opened 2023-04-29
chore: remove unused coveralls dependency

shadowspawn:
There are multiple warnings during `npm install` in this repo about stale dependencies that turn out to be due to `coveralls`.

For example:
```
$ npm install
...
npm WARN deprecated har-validator@5.1.5: this library is no longer supported


$ npm ls har-validator
yargs@17.7.2 /Users/john/Documents/Sandpits/yargs/my-yargs
└─┬ coveralls@3.1.1
  └─┬ request@2.88.2
    └── har-validator@5.1.5

```

`coveralls` was removed from the coverage workflow in #1578.

So remove `coveralls` as a dependency.


bcoe:
Did this in my cleanup of dependencies to get build passing.

Issue #2113, opened 2022-01-03
zsh completion .positional filename


JacobMGEvans:
The ZSH autocompletion works great for commands and flags. I have been having issues with completion `.positional` filename on a command gets clobbered by the `.completion()` builtin script. 

I noticed that the bashrc has a fallback to filename completion. Not sure if that resolves the clobbering of the filename completion after the command is completed. 

In this instance it will give be the command in tab completes, then immediately gives --name as the next complete option.
```ts
yargs.command(
    "dev <filename>",
    "stuff described here",
    (yargs) => {
      return yargs
        .positional("filename", { describe: "entry point", type: "string" })  
        .option("name", {
          describe: "name of the script",
          type: "string",
        })
```

```sh
{app}_yargs_completions()
{
  local reply
  local si=$IFS
  IFS=$'
' reply=($(COMP_CWORD="$((CURRENT-1))" COMP_LINE="$BUFFER" COMP_POINT="$CURSOR" {script} --get-yargs-completions "${words[@]}"))
  IFS=$si
  _describe 'values' reply
}
```

JacobMGEvans:
Closing, over a year old.

PR #2331, opened 2023-05-09
feat(zsh): Add default completion as fallback

wkunert:
Fallback to default completion if no match was found (see https://zsh.sourceforge.io/Doc/Release/Completion-System.html#Completion-Functions for details on `_default`).

Addresses #2113 and harmonizes the behavior of zsh/bash completion, see `-o default` in the bash completion template: https://github.com/yargs/yargs/blob/e517318cea0087b813f5de414b3cdec7b70efe33/lib/completion-templates.ts#L27


shadowspawn:
This is simple, but I am not sure it is correct. I am researching this to review the PR, and not an expert!

I don't see in that [documentation](https://zsh.sourceforge.io/Doc/Release/Completion-System.html#Completion-Functions) any mention of the result code of calling `_describe`, but I think you are relying on getting a non-zero result if `reply` is empty? Have you seen this pattern used anywhere else? The documentation suggests it might be used in an error condition which I don't think is quite the case here.

There are a huge number of utility functions! I expect there is support for considering multiple or fallback targets somewhere in there... (Styles? Tags? Other functions? So many possibilities!)

For instance, I found howto documentation covering `_describe` (simpler) vs `_alternative` (more complex):
- https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org#writing-simple-completion-functions-using-_describe
- https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org#writing-completion-functions-using-_alternative

wkunert:
Thanks for taking the time @shadowspawn to look into it!

> This is simple, but I am not sure it is correct. I am researching this to review the PR, and not an expert!

I am not an expert either, the missing file completion was just bugging me enough to give it a shot. 😄 

> I don't see in that [documentation](https://zsh.sourceforge.io/Doc/Release/Completion-System.html#Completion-Functions) any mention of the result code of calling `_describe`, but I think you are relying on getting a non-zero result if `reply` is empty? Have you seen this pattern used anywhere else? The documentation suggests it might be used in an error condition which I don't think is quite the case here.

I am sure to have seen it somewhere but I am not able to find it again and I also cannot find any documentation on it. It does work but it might be just a quirk. I went ahead and changed the code to make it more explicit. The docs are mentioning "_completion after an unrecognised subcommand_" as an error condition which I thought does fit here.

> There are a huge number of utility functions! I expect there is support for considering multiple or fallback targets somewhere in there... (Styles? Tags? Other functions? So many possibilities!)

I agree that there might be better solutions but it becomes really complicated very quickly. For now I will stick with my simple approach which is sufficient for my use case.

Thanks again for your time and pointing out this flaw, I hope the change clears your concern.

shadowspawn:
I am happier checking `reply` itself. (Although at some point I'll have to investigate when this triggers and if I agree it is appropriate!)

What does `reply` contain? A string or an array? 

If array, this looks a bit strange, and does not work when I try it locally:
```sh
$ array=(one two three)
$ echo \${#array[@]} 
zsh: no matches found: ${#array[@]}
```

I think the array length check might look like:
```sh
$ array=(one two three)
$ echo ${#array} 

3
```

Using double brackets also gives safer test behaviour, so line might be:
```sh
if [[ ${#reply} -gt 0 ]]; then
```

The difference between `[ condition ]` and `[[ condition ]]` is subtle. I'll leave it to references:
- http://mywiki.wooledge.org/BashFAQ/031
- http://mywiki.wooledge.org/BashGuide/Practices#Bash_Tests

wkunert:
Thanks again for your time and input @shadowspawn!

> What does `reply` contain? A string or an array?

As far as I see it, `reply` should be an array because of the parentheses `reply=(…)`: https://github.com/yargs/yargs/blob/e517318cea0087b813f5de414b3cdec7b70efe33/lib/completion-templates.ts#L43

> If array, this looks a bit strange, and does not work when I try it locally:
> 
> ```shell
> $ array=(one two three)
> $ echo \${#array[@]} 
> zsh: no matches found: ${#array[@]}
> ```

Did you by any chance keep the backslash? This is only in the template because we are within an (JavaScript) template literal and therefore the dollar sign must be escaped. For me (without the backslash) it looks like this in the shell:

```sh
% array=(one two three)
% echo ${#array[@]}    
3
```

> I think the array length check might look like:
> 
> ```shell
> $ array=(one two three)
> $ echo ${#array} 
> 
> 3
> ```

As far as I understand the zsh documentation both versions can be used interchangeably (and there is even a third version: `${#array[*]}`), but I changed it to the more straightforward style you proposed.

> Using double brackets also gives safer test behaviour, …

Agree, changed accordingly. Thanks for the references!

shadowspawn:
I missed being inside a template! Was just looking at the diff and saw all shell code...

shadowspawn (approved):
I gave this a light test comparing zsh and bash completions, and LGTM.

alyssadev:
Hi, please consider pulling this PR, thank you

Goli4thus:
I just stumbled over this PR while searching for a way to make zsh completion work, when I have an option that expect a filepath value.

I simply patched my `~/.zshrc` completion section generated today via `yargs` according to this PR.
But that at first gave me errors (three times the same one):
> bad math expression: illegal character: {

I then removed the leading `backslash` in `\${#reply}`.
EDIT: After actually skimming this PR's discussion it's clear now: was from javascript templating.

Result is: then it works nicely!

All that's needed (after a refresh; e.g. via `exec zsh`) is to force it to switch to the default completion.
That can be done easily via `./<TAB>`, `~/<TAB>`, or (if the desired file/folder-name is known roughly) by writing first letter (or more) and then `<TAB>`.

So this PR certainly seems a nice addition to make zsh completion work regarding filepath completion.

Additional infos:
Tested on manjaro linux with `zsh` version `5.9`.

bcoe:
@wkunert @Goli4thus @alyssadev this is now available running `npm i yargs@next`.

Note: this version of yargs requires Node >=20

PR #2334, opened 2023-05-14
test: use test config for watch compiles

shadowspawn:
I noticed when adding watch to `yargs/cliui` that it would be better if the watch compiles used the test configurations, so source maps available.


shadowspawn:
This is stale after the esm update. It may be less compelling with the simpler setup too, since can manually run
```
npx tsc --watch
```

Issue #2337, opened 2023-06-10
[Feature] Tagline in help text


bojavou:
Hello! I'd like to include a tagline in the help text. Maybe with a description function.

```js
yargs(..)
  .scriptName('gadget')
  .description('Never eat a pirate. He'll spend the rest of his after hunting you.')
```

```
gadget <command>

Never eat a pirate. He'll spend the rest of his afterlife hunting you.

Commands:
  gadget mix   Mix it up well
  gadget stir  Stir a little

Options:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]

Not enough non-option arguments: got 0, need at least 1

```

You can kind of do it with `usage()`, but it shows up in the commands list which I don't want.

```js
yargs(..)
    .scriptName('gadget')
    .usage('$0 <command>', 'Never eat a pirate. The indigestion is terrible.')
```

```
gadget <command>

Never eat a pirate. The indigestion is terrible.

Commands:
  gadget <command>     Never eat a pirate. The indigestion is terrible.[default]
  gadget mix           Mix it up well
  gadget stir          Stir a little

Options:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]

Not enough non-option arguments: got 0, need at least 1
```

The epilog is also nice but it puts it at the end. I'd rather it have right up at the top as a nice header.

```js
yargs(..)
  .scriptName('gadget')
  .epilog('Never eat a pirate. Zombies are much tastier.')
```

```
gadget <command>

Commands:
  gadget mix   Mix it up well
  gadget stir  Stir a little

Options:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]

Never eat a pirate. Zombies are much tastier.

Not enough non-option arguments: got 0, need at least 1
```

bojavou:
Submitted some code in #2338, to give it something real for discussion.

shadowspawn:
My initial impression is that `usage()` is the right method for setting the command usage and description.

But `.usage()` is also used as an alias for ".command()` and setting the default command, as you discovered.

I wonder if the default-command behaviour should only be triggered if there is a builder or handler? (Would this break any reasonable legacy use cases?)

shadowspawn:
Having second thoughts about overloading `.usage()`! It might fit into the existing api but is not a direct match for the goal.

A method called `.desc()` or `.description()` would be more explicit. In practice it will usually only be called on the top-level command since the methods for creating subcommands have a description parameter.

[...]

PR #2338, opened 2023-06-10
feat: add prolog option

bojavou:
Adds a prolog option, for a tagline under the program header. Named prolog for parity with epilog, and because description already has multiple other meanings. This works for me in a local patched version.

Possible solution to #2337.

```js
yargs(..)
  .scriptName('gadget')
  .prolog('Never eat a pirate. He'll spend the rest of his afterlife hunting you.')
```

```
gadget <command>

Never eat a pirate. He'll spend the rest of his afterlife hunting you.

Commands:
  gadget mix   Mix it up well
  gadget stir  Stir a little

Options:

```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1442,6 +1442,19 @@ Valid `opt` keys include:
       - `'number'`: synonymous for `number: true`, see [`number()`](#number)
       - `'string'`: synonymous for `string: true`, see [`string()`](#string)
 
+.prolog(str)
+------------
+.prologue(str)
+--------------
+
+A message to print at the beginning of the usage instructions,
+just after the header, e.g.
+
+```js
+var argv = require('yargs/yargs')(process.argv.slice(2))
+  .prologue('HTTP download tool.')
+```
+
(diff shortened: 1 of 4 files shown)


shadowspawn:
> Named prolog for parity with epilog, and because description already has multiple other meanings.

My impression is that subcommands have a description, and the description is displayed under the usage. So what is missing is a way to set the description for the top-level command. 

What clashing meanings for "description" did you find?

bojavou:
I was thinking of the subcommand description as clashing since it shows up in the commands list, but the way you're thinking of it seems better. There are a whole bunch of descriptions in the docs (config key desc, help option desc, default value desc), but in the new light that seems like the standard term for any freeform text so it should be followed. They all have congruent meanings.

shadowspawn:
Ah, I see what you were thinking. Yes, the subcommand description does get used two ways. In the list of subcommands, and below the usage for the command itself.

bojavou:
Closing this since discussion moved it another way :-)

PR #2343, opened 2023-07-09
Fix/missing command module array type

AlberErre:
This PR fixes a missing type on `yargs` generated bundler. 

Looks like typescript is having issues while concatenating different subtypes. This is happening with `DefinitionOrCommandName` type while compiling 👀

As a result, types such as `cmd: string | CommandHandlerDefinition | DefinitionOrCommandName[]` are missing the `CommandHandlerDefinition[]` type case on the generated bundler. 

I tested this behaviour on `yargs` version `17.7.2`, but looks like this issue is still present.

Under this PR, this issue should be fixed as now the types will be more explicit and won't get lost on typescript compiler 🎉

### DIFF
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -29,6 +29,11 @@ import whichModule from './utils/which-module.js';
 
 const DEFAULT_MARKER = /(^\*)|(^\$0)/;
 export type DefinitionOrCommandName = string | CommandHandlerDefinition;
+export type CommandInput =
+  | string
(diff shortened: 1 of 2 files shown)


AlberErre:
Nevermind, this looks like an issue with `DefinitelyTyped`.

Particularly with `@types/yargs` package.

I new PR on that repository has been created -> https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66012

PR #2353, opened 2023-08-26
Minor edit to documentation for option 'global' example

Cypher1:
The current example shows two different arguments with the same configuration and then explains how they should behave differently due to the default behaviour being used in that case.

This change removes the setting overriding the default behaviour, allowing the example to behave as expected.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -908,8 +908,7 @@ is executed, as an example:
 var argv = require('yargs/yargs')(process.argv.slice(2))
   .option('a', {
     alias: 'all',
-    default: true,
-    global: false
+    default: true
   })
   .option('n', {
     alias: 'none',


shadowspawn:
This example is currently a bit confusing.

I think that line is to change the global behaviour for 'a' to false, so that the later `.global('a')` can change it back to true.

If you are interested in some back-and-forth, we could try and get it simpler and saner?

Cypher1:
Ohhhh! I had misunderstood it!

Yes please.

I think we can do a few things

a) having default behaviour mutated should be one example
b) having non default behaviour should be another
c) the default should be there for contrast (as it already is)

Probably the simplest solution is to keep what is already there and add another small example for case b.

Thoughts?

shadowspawn:
I think the biggest problem is that it is artificial to set global false so that can call `.globals('a')`. 

A smaller issue is mixing the option configuration using options-object with the method. If using the options-object it would be more natural to specify `global: false` in there. Maybe that could be a third example option, so default behaviour, overridden using method, overridden using property. However, in general the examples do not show both everywhere, instead the `.options()` help item lists all the properties that correspond to methods.

So going all-in on configuration using method calls and stripped down: 

```
  .option('all')
  .option('none')
  .global('none', false)
  .command('foo', 'foo command', function (yargs) {
    return yargs.option('bar')
  })
  .strict()
  .argv
```

shadowspawn:
> a) having default behaviour mutated should be one example
> b) having non default behaviour should be another
> c) the default should be there for contrast (as it already is)

So if I understand these correctly, I have left out (b) since it is the property vs method case.

shadowspawn:
> However, in general the examples do not show both everywhere, instead the `.options()` help item lists all the properties that correspond to methods.

That could be considered a historical shortcoming, and perhaps we shouldn't worry about that if we think it would make the example better and more useful.

Cypher1:
I agree.

Perhaps, the API is actually a bit surprising and overriding global-ness
itself is the questionable behaviour.

I appreciate the discussion, but I'm still looking at different libraries
to use and so haven't put in the time to work out if the `.global` function
makes sense.

Based on this experience and the state of the rest of the project though, I
am very keen to recommend `yargs` to my team.

Feel free to continue the discussion, I'll try to chime in where I see an
opportunity.

shadowspawn:
(I had to play with some example code to work out how `global` was working myself.)

shadowspawn:
Thanks @Cypher1 

I'll close this PR. Interesting discussion, thanks for the comments. (Added to my personal list of future possible work...)

PR #2361, opened 2023-09-19
Docs: Update the list of supported locales

s1mpson:
I was on my way to creating _Ukrainian_ translation... but it was just missing in the list of supported languages in docs, along with _Czech_ and _Finnish_ ;)


shadowspawn (approved):
LGTM

PR #2373, opened 2023-11-12
docs: use describe rather than desc

isuvorov:
Probably left over from the previous API version


shadowspawn:
Did you find this did not work, or just that it did not match other documentation?

For historical reasons, I think `desc` or `describe` or `description` are all supported internally.

https://github.com/yargs/yargs/blob/e517318cea0087b813f5de414b3cdec7b70efe33/lib/command.ts#L720

The "Command Module" documentation shows `exports.describe` although `desc` is also used in some examples. I think there is a move away `desc`, so your change is an improvement.

There is one other similar reference to your change in the documentation, so I suggest change this one too:

https://github.com/yargs/yargs/blob/e517318cea0087b813f5de414b3cdec7b70efe33/docs/api.md?plain=1#L653

isuvorov:
> Did you find this did not work, or just that it did not match other documentation?

I just copied the example and catch a TypeScript type error. 
You are right, if I ignore it everything works fine.

> So I suggest change this one too

Agree with you

shadowspawn (approved):

PR #2375, opened 2023-11-25
chore: pin back TypeScript to fix build

shadowspawn:
This is a quick fix to the build for an incompatibility between old TypeScript and old Rollup. Pin back the TypeScript version.

(Same fix as used in `cliui`: https://github.com/yargs/cliui/pull/143#issuecomment-1823917900)


bcoe:
@shadowspawn I believe we can close this now?

shadowspawn:
Yes, a possible interim fix that is no longer required.

PR #2384, opened 2024-01-03
fix(build)!: update typescript / eslint / rollup deps fixing bundling of yargs

bcoe:
build(node)!: require Node >= 16.15.1

@shadowspawn Just dependency updates, but I'm always worried fiddling with build settings happy for additional eyes on review.

_Note: the build tools now require Node 14 or greater, let's just update to Node 14 since 12 has been EOL for quite some time._
_Note 2: had to pin and downgrade a couple build dependencies to make them work on Node 14, since the latest versions require Node 16._


package.json
```diff
@@ -122,6 +123,6 @@
   ],
   "license": "MIT",
   "engines": {
-    "node": ">=12"
+    "node": ">=16.15.1"
```
bcoe, line 126: The rollup deps did not work with Node 14, even with 16 I needed to pin to an older version of rollup.

Given 16 is already EOL, and 14 and 12 have been EOL for quite some time, I suggest we just use 16 as the base node engine.

shadowspawn: (I found trying to get a build going with Node.js 14 in Yargs libraries pretty painful.)

shadowspawn:
For interest, there is a rollup hosted typescript plugin too. I tried switching to it due to targeting Node.js 14 at the time. Mentioning for interest as another option if encounter problems, and not a recommendation as such.

- https://github.com/yargs/yargs-parser/pull/471#issuecomment-1454904018

bcoe:
@shadowspawn if your proposal is still passing, I'm open to a less drastic update where we move to Node 14 rather than 16.

Which option do you prefer?

shadowspawn:
> I'm open to a less drastic update where we move to Node 14 rather than 16.

I vote for Node.js 16, move forward (albeit still not to a version still in support). Multiple Yargs dependencies have dropped Node.js 14 and trying to support 14 is a pain point for maintenance.

shadowspawn:
Looking for version related changes, there is a runtime check of Node.js version in cjs:

https://github.com/yargs/yargs/blob/e517318cea0087b813f5de414b3cdec7b70efe33/lib/cjs.ts#L16-L18

(There is not an equivalent check in esm, perhaps because the esm support in older versions is behind an experimental flag anyway!)

shadowspawn:
Version related changes, there is a second reference to node version number which is still 14 in:

https://github.com/yargs/yargs/blob/e517318cea0087b813f5de414b3cdec7b70efe33/.github/workflows/release-please.yml#L42

PR #2385, opened 2024-01-08
docs: an example of using inquirer to prompt user

nephh:
Added a simple example of using Inquirer prompts to ensure the user inputs a required argument.


docs/advanced.md
```diff
@@ -617,3 +617,46 @@ try {
...
+import yargs from 'yargs/yargs';
+import { hideBin } from 'yargs/helpers';
+import { input } from '@inquirer/prompts';
+
+yargs(hideBin(process.argv))
+  .command({
+    command: 'login [username]',
+    describe: 'Log in with defined username',
+    builder: {
+      username: {
```
shadowspawn, line 637: This builder object is adding an option called `username` rather than supplying a description and type for the positional `username`. Need to use `.positional()` for configuring positional.

shadowspawn: You can confirm this by checking the help, e.g.
```
node index.mjs login --help 
```

shadowspawn:
I wonder whether the inquirer prompts would work well in a middleware callback so "outside" the handler. But not sure makes any practical difference! I'll do a little digging...

shadowspawn:
Putting the code in middleware does work fine but I didn't find an advantage other than the separation, so probably better keeping the code simpler.

docs/advanced.md
nephh: Ah yes I see, didn't realize it was creating an option.

So I guess I have a question, I know there are two ways to create commands, what is the preferred syntax? I'm sure it's just personal preference but just curious.

```js
yargs(hideBin(process.argv))
  .command({
    command: 'login [username]',
    describe: 'Prints out the username',
    builder: (yargs) => {
      yargs.positional('username', {
        describe: 'The username to be printed',
        type: 'string',
      });
    },
    handler: async (argv) => {
      if (!argv.username) {
        argv.username = await input({ message: 'Please enter your username' });
      }

      console.log(`Welcome back, ${argv.username}!`);
    },
  })
  .help()
  .parse();
```
or

```js
yargs(hideBin(process.argv))
  .command(
    'login [username]',
    'Prints out the username',
    (yargs) => {
      yargs.positional('username', {
        describe: 'The username to be printed',
        type: 'string',
      });
    },
    async (argv) => {
      if (!argv.username) {
        argv.username = await input({ message: 'Please enter your username' });
      }

      console.log(`Welcome back, ${argv.username}!`);
    },
  )
  .help()
  .parse();
```

shadowspawn: Yes, personal preference. For this example I prefer the object style you used so that the setings are named (e.g. `command`, `describe`, `handler`) rather than just multiple strings and functions.

shadowspawn:
Good question about async. I don't think need to explicitly refactor code or explain about async since main point of example is inquirer, but I do think should add `await` in example code as good practice when have an async handler.

```
await yargs(hideBin(process.argv))
...
```

nephh:
Sounds good, thanks for all the help. Went ahead and pushed those changes

docs/advanced.md
```diff
@@ -617,3 +617,46 @@ try {
...
+  })
+  .help()
+  .parse();
+```
+
+If username is not given as an argument, inquirer will prompt the user with an input.
+
+```
+$ ./app.js login
+? Please enter your username:
```
shadowspawn, line 658: Shouldn't this show a name being entered after the prompt, and the "Welcome back"? (Looks incomplete at the moment.)

docs/advanced.md
```diff
@@ -617,3 +617,46 @@ try {
 }
 console.info('finish')
 ```
+
+### Using Inquirer to make interactive tools
+
+Using Inquirer together with yargs is a powerful way to make your CLI tools more interactive and responsive to the user.
```
shadowspawn, line 623: I think Inquirer is probably the best known and a good package, but it isn't the only package that could be used. Perhaps we could be explicit somewhere that it is an example?

e.g.
```
Using Inquirer or a similar package with yargs is a powerful way to make your CLI tools more interactive and responsive to the user.
```

docs/advanced.md
nephh: Definitely, looks much more complete now

shadowspawn (approved):
Looks good to me. 

Thanks @nephh

bcoe:
@nephh thank you for the contribution.

Issue #2387, opened 2024-01-12
`scriptName` breaks completion script if it contains whitespace


davidmurdoch:
Reproduction:

One liner:

```javascript
require("yargs").scriptName("yarn whatever").completion().parseSync(["completion"])
```

You can paste this into your terminal to get a quick reproduction:

```console
mkdir scriptNameCompletionTest && cd scriptNameCompletionTest && npm install yargs --no-save >/dev/null && node -e 'require("yargs").scriptName("yarn whatever").completion().parseSync(["completion"])'
```

The output:

```bash
###-begin-yarn whatever-completions-###
#
# yargs command completion script
#
# Installation: yarn whatever completion >> ~/.bashrc
#    or yarn whatever completion >> ~/.bash_profile on OSX.
#
_yarn whatever_yargs_completions()
{
    local cur_word args type_list

    cur_word="${COMP_WORDS[COMP_CWORD]}"
    args=("${COMP_WORDS[@]}")

    # ask yargs to generate completions.
    type_list=$(yarn whatever --get-yargs-completions "${args[@]}")

    COMPREPLY=( $(compgen -W "${type_list}" -- ${cur_word}) )

    # if no match was found, fall back to filename completion
    if [ ${#COMPREPLY[@]} -eq 0 ]; then
      COMPREPLY=()
    fi

    return 0
}
complete -o bashdefault -o default -F _yarn whatever_yargs_completions yarn whatever
###-end-yarn whatever-completions-###
```

Notice the invalid bash function `_yarn whatever_yargs_completions`.

I think if `scriptName` is going to be used in the generated bash completion script it should be sanitized (https://stackoverflow.com/a/28115066/160173). Or completions should allow the use to specify their own function name.

shadowspawn:
I agree it is a short-coming, but under what circumstances would it come up in practice? A script with a space in the name seems unlikely. However, if you actually encountered this bug in real world then definitely of interest!

davidmurdoch:
One use is any file name with a space in it, which might not be common, but is valid everywhere.

But also if your yargs interface is package.json script, e.g., `yarn webpack`, then it makes sense for `--help`, which uses the value of `scriptName` in its Commands, Usage, and Examples outputs, to show the actual command a user should use, i.e. `yarn webpack [command]`, not `some-file.js [command]`.

Example:

Confused users:

[...]

PR #2388, opened 2024-01-12
fix: ensure completion script uses valid function name

davidmurdoch:
Fixes https://github.com/yargs/yargs/issues/2387

### DIFF
--- a/lib/completion.ts
+++ b/lib/completion.ts
@@ -350,13 +350,8 @@ export class Completion implements CompletionInstance {
       ? templates.completionZshTemplate
       : templates.completionShTemplate;
     let name = this.shim.path.basename($0);
-    // Santize `name` to make sure it is a valid bash function name
-    // first, check if the first character is an alphabetic character or underscore
-    if (!str.match(/^[A-Za-z_]/)) {
-      // If not, prepend name with an underscore
-      name = '_' + name;
-    }
-    // Then, replace any characters that are not alphanumeric or underscore with an underscore
+    // Santize `name` to make sure it is a valid bash function name by
+    // replacing any characters that are not alphanumeric or underscore with an underscore
     name = name.replace(/[^A-Za-z0-9_]/g, '_');
 
     // add ./ to applications not yet installed as bin.
--- a/lib/completion.ts
+++ b/lib/completion.ts
@@ -349,7 +349,15 @@ export class Completion implements CompletionInstance {
     let script = this.zshShell
       ? templates.completionZshTemplate
       : templates.completionShTemplate;
-    const name = this.shim.path.basename($0);
+    let name = this.shim.path.basename($0);
+    // Santize `name` to make sure it is a valid bash function name
+    // first, check if the first character is an alphabetic character or underscore
+    if (!str.match(/^[A-Za-z_]/)) {
+      // If not, prepend name with an underscore
+      name = '_' + name;
+    }
+    // Then, replace any characters that are not alphanumeric or underscore with an underscore
+    name = name.replace(/[^A-Za-z0-9_]/g, '_');
 
     // add ./ to applications not yet installed as bin.
     if ($0.match(/\.js$/)) $0 = `./${$0}`;
--- a/test/completion.cjs
+++ b/test/completion.cjs
@@ -608,6 +608,14 @@ describe('Completion', () => {
       r.logs[0].should.match(/ndm flintlock >>/);
     });
 
+    it('santizes scriptName ($0) for use in a bash function', () => {
+      const r = checkUsage(
+        () => yargs([]).scriptName("1 test ing").showCompletionScript()
+      );
+
+      r.logs[0].should.match(/_a_test_ing/);
+    });
+
     it('if $0 has a .js extension, a ./ prefix is added', () => {
       const r = checkUsage(() => yargs([]).showCompletionScript(), ['test.js']);
 
--- a/test/completion.cjs
+++ b/test/completion.cjs
@@ -613,7 +613,7 @@ describe('Completion', () => {
         () => yargs([]).scriptName("1 test ing").showCompletionScript()
       );
 
-      r.logs[0].should.match(/_a_test_ing/);
+      r.logs[0].should.match(/_1_test_ing/);
     });
 
     it('if $0 has a .js extension, a ./ prefix is added', () => {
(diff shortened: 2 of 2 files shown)


davidmurdoch:
Anything else you need for this?

lib/completion.ts
```diff
@@ -349,7 +349,15 @@ export class Completion implements CompletionInstance {
     let script = this.zshShell
       ? templates.completionZshTemplate
       : templates.completionShTemplate;
-    const name = this.shim.path.basename($0);
+    let name = this.shim.path.basename($0);
+    // Santize `name` to make sure it is a valid bash function name
+    // first, check if the first character is an alphabetic character or underscore
```
shadowspawn, line 354: I think the completion template might be putting an underscore on the front already so this step is not needed? (Based on code inspection, not from running code.)

https://github.com/yargs/yargs/blob/0c95f9c79e1810cf9c8964fbf7d139009412f7e7/lib/completion-templates.ts#L8

lib/completion.ts
```diff
@@ -349,7 +349,15 @@ export class Completion implements CompletionInstance {
...
+    if (!str.match(/^[A-Za-z_]/)) {
+      // If not, prepend name with an underscore
+      name = '_' + name;
+    }
+    // Then, replace any characters that are not alphanumeric or underscore with an underscore
+    name = name.replace(/[^A-Za-z0-9_]/g, '_');
```
davidmurdoch, line 360:
```suggestion
    // Santize `name` to make sure it is a valid bash function name by
    // replacing any characters that are not alphanumeric or underscore with an underscore
    name = name.replace(/[^A-Za-z0-9_]/g, '_');
```

lib/completion.ts
davidmurdoch: ah, good catch

test/completion.cjs
```diff
@@ -608,6 +608,14 @@ describe('Completion', () => {
       r.logs[0].should.match(/ndm flintlock >>/);
     });
 
+    it('santizes scriptName ($0) for use in a bash function', () => {
+      const r = checkUsage(
+        () => yargs([]).scriptName("1 test ing").showCompletionScript()
```
davidmurdoch, line 613:
```suggestion
        () => yargs([]).scriptName('1 test ing').showCompletionScript()
```

davidmurdoch:
Is there anything I can do to push this over the line?

shadowspawn:
Ok, my main problem is I am dubious this is useful in practice. Working with command names with spaces in them is pretty clunky and I have never seen it being used. I need convincing that users will benefit from this before I give it a positive review.  (I don't think completion will work through a yarn run script, but haven't tried that scenario mentioned in the issue.)

My secondary concern is whether it even works. I had another look. The current PR completion does not seem to be functional for a script name with a space as the `compdef` line at the end is adding completion for `foo_bar`.

```
% 'foo bar' completion        
#compdef foo_bar
###-begin-foo_bar-completions-###
#
#
# Installation: foo bar completion >> ~/.zshrc
#    or foo bar completion >> ~/.zprofile on OSX.
#
_foo_bar_yargs_completions()
{
  local reply
  local si=$IFS
  IFS=$'
' reply=($(COMP_CWORD="$((CURRENT-1))" COMP_LINE="$BUFFER" COMP_POINT="$CURSOR" foo bar --get-yargs-completions "${words[@]}"))
  IFS=$si
  _describe 'values' reply
}
compdef _foo_bar_yargs_completions foo_bar
###-end-foo_bar-completions-###
```

I wrote out the completion and edited it to have:
```
compdef _foo_bar_yargs_completions 'foo bar'
```

And could then get completion for 

```
% 'foo bar' --<tab>
--boolean  --string  --                                                                                                                                              
--help               -- Show help                                                                                                                                    
--version            -- Show version number                                                                                                                          
```

davidmurdoch:
> I have never seen it being used.

Its in every npm script that ends up calling out out to yargs.

And I'm using it in the MetaMask extension, er, I want to use it in MetaMask but can't without quite a bit of customization because the generated completion script isn't valid.

I've been editing the minified code, so sorry my PR doesn't even work 😢 

Looks like zhs just needs single quotes around commands with spaces. But bash needs a bit more work, as it doesn't work even when quoted. The bash version would be something like:

```
_yarn_webpack_yargs_completions()
{
    local cur prev command_line

    cur="${COMP_WORDS[COMP_CWORD]}"
    prev="${COMP_WORDS[COMP_CWORD-1]}"
    command_line="${COMP_WORDS[*]}"

    # Check if the command line specifically starts with 'yarn webpack'
    if [[ ${command_line} == yarn\ webpack* ]]; then
        # Your completion logic goes here
        # Ensure we are adjusting the argument list appropriately
        local adjusted_args=("${COMP_WORDS[@]:2}") # Skip 'yarn' and 'webpack'
        local type_list

        # Simulate the command as if it started with 'webpack' for yargs
        type_list=$(yarn webpack --get-yargs-completions "${adjusted_args[@]}")
        COMPREPLY=( $(compgen -W "${type_list}" -- ${cur}) )
        return 0
    fi
}

complete -F _yarn_webpack_yargs_completions yarn

shadowspawn:
> It's in every npm script that ends up calling out out to yargs.

This might be where I am missing something. How does this involve a command name with a space in it? I'll explain what I am thinking.

So in my head, I have an npm run-script called `foo`. The run-script uses (say) a command named `bar`. The user types `npm run foo -- --compress` and it invokes `bar --compress`. You suggested the author might specify `.scriptName('npm run foo')` to make the help reflect the main way it is called. I can see that.

Where does completion come into it? I don't think the user will be able to get completion on `npm run foo -- --f<tab>`. What is the command that the user types that gets completion? 

(One step simpler with yarn if can omit the `run`, but I think otherwise similar.)

shadowspawn:
> And I'm using it in the MetaMask extension

This sound like you do have a concrete plan! What does the help show, and what is user typing on command-line when completion works? Or is the completion used in a non-interactive way by some higher level tooling?

shadowspawn:
(I am not expert in completion handlers, and my mental model of one handler per command might be preventing me seeing the goal.)

shadowspawn:
(Oh, multi-word completion handlers, that rings a vague bell...)

davidmurdoch:
Closing, superseded by PR #2422

Issue #2414, opened 2024-05-26
`parserConfiguration` doesn't behave well if called multiple times


dleavitt:
Some surprising behavior as the result of having a global singleton parser instance.

1. Calling it more than once overwrites the whole thing instead of merging:

```typescript
const argv1 = yargs
  .parserConfiguration({ "parse-numbers": false })
  .parserConfiguration({ "boolean-negation": true })
  .option("a", {});

const opts = argv1.parse(["-a", "1"]);
// Expected
opts[a] == "1"
// Actual
opts[a] == 1
```

2. Calling it within a command overwrites the config for the entire yargs instance:

```typescript
const argv2 = yargs
  .parserConfiguration({ "parse-numbers": false })
  .option("a", {})
  .command("cmd1", "", (y) => y.parserConfiguration({}).option("b", {}))
  .command("cmd2", "", (y) => y.option("c", {}));

console.log(argv2.parse(["cmd1", "-a", "1", "-b", "2"]));
console.log(argv2.parse(["cmd2", "-a", "1", "-c", "2"]));

// Expected
{ _: [ 'cmd1' ], a: "1", b: "2" }
{ _: [ 'cmd2' ], a: "1", c: "2" }

// Actual
{ _: [ 'cmd1' ], a: 1, b: 2 }
{ _: [ 'cmd2' ], a: 1, c: 2 }
```

These issues sort of combine to negatively affect the composability of commands - a command in one module changing a single config setting will reset _all_ config settings in _all_ modules.

Ideally:
1. Calling `parserConfiguration` should merge the newly passed options with any previously passed options.
2. Each `CommandBuilder` should probably get its own `#parserConfig`
3.  Also, each yargs instance should probably get its own parser config? Calling it anywhere appears to overwrite values for all yargs instances: **Because I'm using yargs wrong, see responses below**.

```typescript
const cli = yargs.parserConfiguration({ "parse-numbers": false }).option("a", {});

// in some other file perhaps
const otherCli = yargs.parserConfiguration({}).option("a", {});

const opts = cli.parse(["-a", "1"]);
// Expected
opts["a"] == "1"
// Actual
opts["a"] == 1
```

shadowspawn:
> Some surprising behavior as the result of having a global singleton parser instance.

I will response to this part first. The answer here is don't use a global singleton parser instance! 😄 

The singleton was an old pattern which has (hopefully) been removed from the documentation. It has been dropped from the newer exports. The new pattern is the default export is a factory function.

[...]

PR #2415, opened 2024-05-29
Merge rather than replace parserConfiguration

dleavitt:
Additional calls to `parserConfiguration` now merge into the existing configuration rather than replacing all unspecified keys with their default values. 

Addresses one of the items in #2414.

Seems modestly better than the old behavior, though it's technically a breaking change so maybe not worth it?

### DIFF
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -1164,7 +1164,7 @@ export class YargsInstance {
   }
   parserConfiguration(config: Configuration) {
     argsert('<object>', [config], arguments.length);
-    this.#parserConfig = config;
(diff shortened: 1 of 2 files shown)


shadowspawn:
> Seems modestly better than the old behavior, though it's technically a breaking change so maybe not worth it?

Thanks. Unless #2414 gets some likes or comments showing this is a common problem, I think probably not enough of an improvement to change the existing behaviour.

The non-breaking change would be to note the behaviour in the documentation?

dleavitt:
Happy to add a note on this to the documentation. Otherwise feel free to close.

bcoe:
My preference would be to replace not merge. I would happily accept an update to documentation like @shadowspawn suggests 👏 

PR #2393, opened 2024-03-18
fix: propagate Dictionary including undefined in value type

shadowspawn:
The value type we use with Dictionary type declarations often includes undefined. I am guessing this is more to ensure the value type is checked for a possibly unknown key than for setting a value of undefined, but paranoia is ok for either reason.

TypeScript 5.4.0 picks up a case where the `undefined` was not being used consistently,

```
lib/yargs-factory.ts:667:7 - error TS2345: Argument of type 'string | string[] | Dictionary<string | undefined>' is not assignable to parameter of type 'string | string[] | { [x: string]: string; }'.
  Type 'Dictionary<string | undefined>' is not assignable to type 'string | string[] | { [x: string]: string; }'.
    Type 'Dictionary<string | undefined>' is missing the following properties from type 'string[]': length, pop, push, concat, and 29 more.
```

PR #2422, opened 2024-06-28
fix: ensure completion script uses valid function name

### LINKED ISSUE #2387: `scriptName` breaks completion script if it contains whitespace

davidmurdoch:
I finally got around to working on the multi-word completion script. It was a bit more complicated than I thought!

Fixes #2387, supersedes PR #2388

This PR has some other fixes, all in separate commits you can cherry pick or drop/revert as needed:

  * pins TypeScript (typescript minor versions contain breaking changes)
  * renames an eslint rule to match its new name
  * fixes types by using the fix in #2393
  * fixes `--get-yargs-completion` when the parser options strip-dashed setting is `true`
  * and finally, adds support for multi-word completion
   
Please do let me know if you think you'll be able to merge and release this. I can always just patch `yargs` in our codebase (MetaMask) if needed.

### DIFF
--- a/lib/completion-templates.ts
+++ b/lib/completion-templates.ts
@@ -5,30 +5,32 @@ export const completionShTemplate = `###-begin-{{app_name}}-completions-###
 # Installation: {{app_path}} {{completion_command}} >> ~/.bashrc
 #    or {{app_path}} {{completion_command}} >> ~/.bash_profile on OSX.
 #
-_{{app_name}}_yargs_completions()
+_{{safe_app_name}}_yargs_completions()
 {
-    local cur_word args type_list
+    if [[ {{app_name_word_check}} ]]; then
+      local cur_word args type_list
 
-    cur_word="\${COMP_WORDS[COMP_CWORD]}"
-    args=("\${COMP_WORDS[@]}")
+      cur_word="\${COMP_WORDS[COMP_CWORD]}"
+      args=("\${COMP_WORDS[@]}")
 
-    # ask yargs to generate completions.
-    type_list=$({{app_path}} --get-yargs-completions "\${args[@]}")
+      # ask yargs to generate completions.
+      type_list=$({{app_path}} --get-yargs-completions "\${args[@]}")
 
-    COMPREPLY=( $(compgen -W "\${type_list}" -- \${cur_word}) )
+      COMPREPLY=( $(compgen -W "\${type_list}" -- \${cur_word}) )
 
-    # if no match was found, fall back to filename completion
-    if [ \${#COMPREPLY[@]} -eq 0 ]; then
-      COMPREPLY=()
+      # if no match was found, fall back to filename completion
+      if [ \${#COMPREPLY[@]} -eq 0 ]; then
+        COMPREPLY=()
+      fi
     fi
 
     return 0
 }
-complete -o bashdefault -o default -F _{{app_name}}_yargs_completions {{app_name}}
+complete -o bashdefault -o default -F _{{safe_app_name}}_yargs_completions {{app_name_first_word}}
 ###-end-{{app_name}}-completions-###
 `;
 
-export const completionZshTemplate = `#compdef {{app_name}}
+export const completionZshTemplate = `#compdef "{{safe_app_name}}"
 ###-begin-{{app_name}}-completions-###
 #
 # yargs command completion script
@@ -36,7 +38,7 @@ export const completionZshTemplate = `#compdef {{app_name}}
 # Installation: {{app_path}} {{completion_command}} >> ~/.zshrc
 #    or {{app_path}} {{completion_command}} >> ~/.zprofile on OSX.
 #
-_{{app_name}}_yargs_completions()
+_{{safe_app_name}}_yargs_completions()
 {
   local reply
   local si=$IFS
@@ -44,6 +46,6 @@ _{{app_name}}_yargs_completions()
   IFS=$si
   _describe 'values' reply
 }
-compdef _{{app_name}}_yargs_completions {{app_name}}
+compdef _{{safe_app_name}}_yargs_completions "{{zsh_app_name}}"
 ###-end-{{app_name}}-completions-###
 `;
--- a/lib/completion.ts
+++ b/lib/completion.ts
@@ -356,9 +356,32 @@ export class Completion implements CompletionInstance {
       : templates.completionShTemplate;
     const name = this.shim.path.basename($0);
 
+    // Sanitize `name` to make sure it is a valid bash function name by
+    // replacing any characters that are not alphanumeric or underscore with an
+    // underscore
+    const safeName = name.replace(/[^A-Za-z0-9_]/g, '_');
+
     // add ./ to applications not yet installed as bin.
     if ($0.match(/\.js$/)) $0 = `./${$0}`;
 
+    // zsh supports multi-word completions by default, but bash needs custom logic
+    const appNameWords = name.split(' ');
+    if (!this.zshShell) {
+      // to support multi-word completion check each word in the app name
+      const appNameFirstWord = name[0];
+      const wordsToCheck = appNameWords
+        .map((word, index) => {
+          return `"\${COMP_WORDS[${index}]}" == "${word}"`;
+        })
+        .join(' && ');
+      script = script.replace(/{{app_name_first_word}}/g, appNameFirstWord);
+      script = script.replace(/{{app_name_word_check}}/g, wordsToCheck);
+    } else {
+      // zsh multi-word completion works by replacing the space with a '='
+      script = script.replace(/{{zsh_app_name}}/g, appNameWords.join('='));
+    }
+
+    script = script.replace(/{{safe_app_name}}/g, safeName);
     script = script.replace(/{{app_name}}/g, name);
     script = script.replace(/{{completion_command}}/g, cmd);
     return script.replace(/{{app_path}}/g, $0);
--- a/lib/completion.ts
+++ b/lib/completion.ts
@@ -18,6 +18,7 @@ type CompletionCallback = (
 /** Instance of the completion module. */
 export interface CompletionInstance {
   completionKey: string;
+  completionKeys: string[];
   generateCompletionScript($0: string, cmd: string): string;
   getCompletion(
     args: string[],
@@ -29,6 +30,10 @@ export interface CompletionInstance {
 
 export class Completion implements CompletionInstance {
   completionKey = 'get-yargs-completions';
+  // we need to handle the completion key in multiple formats because the
+  // `strip-dashed` parser option might remove the hyphenated version from the
+  // parsed arguments
+  completionKeys = [this.completionKey, 'getYargsCompletions'];
 
   private aliases: DetailedArguments['aliases'] | null = null;
   private customCompletionFunction: CompletionFunction | null = null;
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -1660,7 +1660,7 @@ export class YargsInstance {
   >(
     builder: (key: K, value: V, ...otherArgs: any[]) => YargsInstance,
     type: T,
-    key: K | K[] | {[key in K]: V},
+    key: K | K[] | {[key in K]: V | undefined},
     value?: V
   ) {
     this[kPopulateParserHintDictionary]<T, K, V>(
@@ -1704,7 +1704,7 @@ export class YargsInstance {
   >(
     builder: (key: K, value: V, ...otherArgs: any[]) => YargsInstance,
     type: T,
-    key: K | K[] | {[key in K]: V},
+    key: K | K[] | {[key in K]: V | undefined},
     value: V | undefined,
     singleKeyHandler: (type: T, key: K, value?: V) => void
   ) {
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -2001,11 +2001,17 @@ export class YargsInstance {
 
     let helpOptSet = false;
     let versionOptSet = false;
+    let requestCompletions = false;
     Object.keys(argv).forEach(key => {
       if (key === this.#helpOpt && argv[key]) {
         helpOptSet = true;
       } else if (key === this.#versionOpt && argv[key]) {
         versionOptSet = true;
+      } else if (
+        key in argv &&
+        this.#completion!.completionKeys.includes(key)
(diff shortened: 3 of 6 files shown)


lib/completion.ts
```diff
@@ -351,9 +356,32 @@ export class Completion implements CompletionInstance {
...
+        })
+        .join(' && ');
+      script = script.replace(/{{app_name_first_word}}/g, appNameFirstWord);
+      script = script.replace(/{{app_name_word_check}}/g, wordsToCheck);
+    } else {
+      // zsh multi-word completion works by replacing the space with a '='
+      script = script.replace(/{{zsh_app_name}}/g, appNameWords.join('='));
+    }
+
+    script = script.replace(/{{safe_app_name}}/g, safeName);
```
davidmurdoch, line 384: To save you from having to figure it out. For a script name of "yarn webpack" it'd create this:

```bash
_yarn_webpack_yargs_completions()
{
    if [[ "${COMP_WORDS[0]}" == "yarn" && "${COMP_WORDS[1]}" == "webpack" ]]; then
        local cur_word args type_list

        args=("${COMP_WORDS[@]}")

        type_list=$(yarn webpack --get-yargs-completions "${args[@]}")

          COMPREPLY=()
        fi
    fi

    return 0
}
complete -o bashdefault -o default -F _yarn_webpack_yargs_completions yarn
```

and for zsh it'd create:

```bash
_yarn_webpack_yargs_completions()
{
  local reply
  local si=$IFS
  IFS=$'
' reply=($(COMP_CWORD="$((CURRENT-1))" COMP_LINE="$BUFFER" COMP_POINT="$CURSOR" yarn webpack --get-yargs-completions "${words[@]}"))
  IFS=$si
  _describe 'values' reply
}
compdef _yarn_webpack_yargs_completions "yarn=webpack"
```

lib/yargs-factory.ts
```diff
@@ -1704,7 +1704,7 @@ export class YargsInstance {
   >(
     builder: (key: K, value: V, ...otherArgs: any[]) => YargsInstance,
     type: T,
-    key: K | K[] | {[key in K]: V},
+    key: K | K[] | {[key in K]: V | undefined},
```
davidmurdoch, line 1707: These type changes are from https://github.com/yargs/yargs/pull/2393 and were necessary to get it to compile.

lib/yargs-factory.ts
```diff
@@ -2054,7 +2060,6 @@ export class YargsInstance {
       this.#isGlobalContext = false;
 
       const handlerKeys = this.#command.getCommands();
-      const requestCompletions = this.#completion!.completionKey in argv;
       const skipRecommendation = helpOptSet || requestCompletions || helpOnly;
```
davidmurdoch, line 2063: Since `--get-yargs-completions` is _removed_ from `argv` when `strip-dashed` is set to true I changed this check in two ways:

1. do it the same loop the built-in commands `--help` and `--version` are checked
2. check both possible keys (`--get-yargs-completions` and `--getYargsCompletions`)

I did _not_ have to change the other use of `this.#completion.completionKey` elsewhere in this file as it is used to inspect `args` (not cleaned up), not `argv` (which has been cleaned up).

package.json
```diff
@@ -83,7 +83,7 @@
     "rollup": "^3.29.4",
     "rollup-plugin-cleanup": "^3.2.1",
     "rollup-plugin-ts": "^3.4.5",
-    "typescript": "^5.3.3",
+    "typescript": "5.3.3",
```
davidmurdoch, line 86: TypeScript minor versions almost always have breaking changes. They don't follow semver. :-(

test/completion.cjs
```diff
@@ -594,6 +594,22 @@ describe('Completion', () => {
 
           r.logs.should.have.length(0);
         });
+
+        it('works if strip-dashed is set to `true`', () => {
+          const r = checkUsage(
+            () =>
+              yargs([...firstArguments, './completion', '--foo', '--'])
+                .parserConfiguration({'strip-dashed': true})
```
davidmurdoch, line 602: Notice the `strip-dashed': true` here. We are just testing to ensure this option does _not_ do anything to our `completion` command.

test/completion.cjs
```diff
@@ -608,6 +624,37 @@ describe('Completion', () => {
...
+      it('replaces the bash `complete` word with only the first word of the app name', () => {
+        process.env.SHELL = '/bin/bash';
+        const r = checkUsage(() =>
+          yargs([]).scriptName('1 test ing').showCompletionScript()
+        );
+
+        r.logs[0].should.match(/_yargs_completions 1/);
+      });
+    });
+
```
davidmurdoch, line 657: The bash shell has different replacements than the zsh version, so I test those specifically here.

test/completion.cjs
```diff
@@ -1102,6 +1149,24 @@ describe('Completion', () => {
...
+    it('generates multi-word completion handler', () => {
+      process.env.SHELL = '/bin/zsh';
+      const r = checkUsage(() =>
+        yargs([]).scriptName('1 test ing').showCompletionScript()
+      );
+      console.log(r.logs[0]);
+      r.logs[0].should.match(/_yargs_completions "1=test=ing"/);
+    });
+
```
davidmurdoch, line 1169: These 2 are in the pre-existing zsh `describe`. They are just ensuring multi word support happens as expected.

lib/completion-templates.ts
```diff
@@ -5,45 +5,47 @@ export const completionShTemplate = `###-begin-{{app_name}}-completions-###
 # Installation: {{app_path}} {{completion_command}} >> ~/.bashrc
 #    or {{app_path}} {{completion_command}} >> ~/.bash_profile on OSX.
 #
-_{{app_name}}_yargs_completions()
+_{{safe_app_name}}_yargs_completions()
```
davidmurdoch, line 8: `safe_app_name` is limited to characters that are safe for a bash function name.

lib/completion-templates.ts
```diff
@@ -5,45 +5,47 @@ export const completionShTemplate = `###-begin-{{app_name}}-completions-###
...
 {
-    local cur_word args type_list
+    if [[ {{app_name_word_check}} ]]; then
```
davidmurdoch, line 10: This line turns in to something like `if [[ "${COMP_WORDS[0]}" == "yarn" && "${COMP_WORDS[1]}" == "webpack" ]]; then` (for a script name of `yarn webpack`)

lib/completion-templates.ts
```diff
@@ -5,45 +5,47 @@ export const completionShTemplate = `###-begin-{{app_name}}-completions-###
...
-    COMPREPLY=( $(compgen -W "\${type_list}" -- \${cur_word}) )
+      COMPREPLY=( $(compgen -W "\${type_list}" -- \${cur_word}) )
 
-    # if no match was found, fall back to filename completion
-    if [ \${#COMPREPLY[@]} -eq 0 ]; then
-      COMPREPLY=()
+      # if no match was found, fall back to filename completion
+      if [ \${#COMPREPLY[@]} -eq 0 ]; then
+        COMPREPLY=()
+      fi
```
davidmurdoch, line 24: unchanged

lib/completion-templates.ts
```diff
@@ -5,45 +5,47 @@ export const completionShTemplate = `###-begin-{{app_name}}-completions-###
...
     fi
 
     return 0
 }
-complete -o bashdefault -o default -F _{{app_name}}_yargs_completions {{app_name}}
+complete -o bashdefault -o default -F _{{safe_app_name}}_yargs_completions {{app_name_first_word}}
```
davidmurdoch, line 29: `complete` doesn't have a way of handling multiple words at once, so we have to just give it the first word to kick of the process. The function itself checks if it will need to set the "`COMPREPLY`" or not.

lib/completion-templates.ts
```diff
@@ -5,45 +5,47 @@ export const completionShTemplate = `###-begin-{{app_name}}-completions-###
...
+_{{safe_app_name}}_yargs_completions()
 {
   local reply
   local si=$IFS
   IFS=$'\n' reply=($(COMP_CWORD="$((CURRENT-1))" COMP_LINE="$BUFFER" COMP_POINT="$CURSOR" {{app_path}} --get-yargs-completions "\${words[@]}"))
   IFS=$si
   _describe 'values' reply
 }
-compdef _{{app_name}}_yargs_completions {{app_name}}
+compdef _{{safe_app_name}}_yargs_completions "{{zsh_app_name}}"
```
davidmurdoch, line 49: `compdef` makes it easy, as it supports multiple words right our of the box. We just have to replace any spaces with a `=`. I don't think the quotes I added are required, but they also can't hurt.

lib/completion.ts
```diff
@@ -29,6 +30,10 @@ export interface CompletionInstance {
 
 export class Completion implements CompletionInstance {
   completionKey = 'get-yargs-completions';
+  // we need to handle the completion key in multiple formats because the
+  // `strip-dashed` parser option might remove the hyphenated version from the
+  // parsed arguments
+  completionKeys = [this.completionKey, 'getYargsCompletions'];
```
shadowspawn, line 36: The possibility of user-specified parser configurations like `skip-dashed` makes looking for options in argv rather painful! I did a local fix in #2308 for one case, but I am think we need a utility routine to make it convenient and robust to find options in argv. There are multiple places where it is potentially an issue, like here.

lib/yargs-factory.ts
shadowspawn: The possibility of user-specified parser configurations like `skip-dashed` makes looking for options in `argv` rather painful! I did a local fix in #2308 for one case, but I think we would benefit from a utility routine to make it convenient and robust to find options in `argv`. There are multiple places where it is potentially an issue, like here, for the minority of people who are changing the parser configuration.

shadowspawn:
Lots of useful commentary, thanks.

lib/yargs-factory.ts
davidmurdoch: Do you think I should adjust the method for finding the option? I was thinking it was safe as is here for this specific option as the user should never see it since it'll short circuit when set. Maybe if they made their own option with the same name it would cause trouble, but I think it already would anyway? I haven't verified.

davidmurdoch:
> Lots of useful commentary, thanks.

I ran a semi-popular FOSS package for years so I'm hoping to make this as easy as possible for you. Let me know if there is anything else I can do to help!

lib/yargs-factory.ts
shadowspawn: I don't think you need to fix it here, but I would want to apply the (unwritten) high-level fix here too.

shadowspawn:
> and finally, adds support for multi-word completion

I will have a look focusing on this part of the PR, but might be a few weeks before I get to it.

davidmurdoch:
> > and finally, adds support for multi-word completion
> 
> I will have a look focusing on this part of the PR, but might be a few weeks before I get to it.

a friendly bump :-)

lib/completion-templates.ts
shadowspawn: `yarn=foo` does not install a completion handler for the subcommand `foo` of command `yarn`, it installs a completion handler for command `yarn` _and_ command `foo`.

shadowspawn: https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org#copying-completions-from-another-command

shadowspawn: With code in `index.mjs`:
```
  .scriptName('yarn foo')
```

and script in `package.json`:
```
  "scripts": {
    "foo": "node index.mjs"
  },
```

Setup:
```
% yarn foo completion > comp.sh
% source comp.sh

```

Get completion as intended here:
```console
 % yarn foo --
--boolean  --string  --                                                                        
--help               -- Show help                                                              
--version            -- Show version number     
```

But also here:
```console
% yarn --
--boolean  --string  --                                                                        
--help               -- Show help                                                              
--version            -- Show version number     
```

And here:
```console
% foo --
--boolean  --string  --                                                                        
--help               -- Show help                                                              
--version            -- Show version number     
```

shadowspawn:
I realised when looking at this again that #2387 describes two different things. One is support for a space in a command name, and the other is trying to add completion for a command+argument pair. The first one can be fixed by sanitising the script name, which is what the title of this PR describes.

However, the attempt to allow completion for a multi-word "command" (say command+argument or command+subcommand) is a different thing. In practice for the example `yarn webpack` this means installing a completion for `yarn` which does not really make sense.

lib/completion-templates.ts
shadowspawn: (The above comments all assuming zsh.)

davidmurdoch: Ah, I hadn't noticed this. I'll look into adding a check for the whole command in the competition handler in following week.

shadowspawn:
This was an interesting deep dive into completion. Thanks for the discussion.

I am not clear that the intended outcome of effectively adding a completion to the `yarn` command to allow running a utility through the yarn script support is generally sound. There have not been upvotes on #2387.

The PR is out of date with current code.

> Please do let me know if you think you'll be able to merge and release this. I can always just patch yargs in our codebase (MetaMask) if needed.

Not going to merge at this time. Thanks.

Issue #2402, opened 2024-04-26
Generate zsh autoload completion function file


Nevon:
I'll preface this by saying that the completion system is quite the arcane mystery to me, so please forgive my ignorance here.

We can generate completions for bash and zsh from yargs. This simply prints a piece of bash or zsh code that the user can put into their `.zprofile` or equivalent. The output for this is defined [here](https://github.com/yargs/yargs/blob/main/lib/completion-templates.ts#L1). A zsh example looks like this:

```zsh
###-begin-my-command-completions-###
#
# yargs command completion script
#
# Installation: /Users/nevon/.volta/tools/image/packages/my-package/bin/my-command completion >> ~/.zshrc
#    or /Users/nevon/.volta/tools/image/packages/my-package/bin/my-command completion >> ~/.zsh_profile on OSX.
#
_my_command_yargs_completions()
{
  local reply
  local si=$IFS
  IFS=$'
' reply=($(COMP_CWORD="$((CURRENT-1))" COMP_LINE="$BUFFER" COMP_POINT="$CURSOR" /Users/nevon/.volta/tools/image/packages/my-package/bin/my-command --get-yargs-completions "${words[@]}"))
  IFS=$si
  _describe 'values' reply
}
compdef _my_command_yargs_completions my-command
###-end-my-command-completions-###
```

This is all well and good if the user invokes the completion command to install the completions. However, when this command is packaged up for distribution, such as in homebrew, the package manager would like to manage the completions in files that it can add or remove from the shell path when the package is installed or uninstalled.

Zsh has this functionality baked in, in the form of [autoloaded files](https://zsh.sourceforge.io/Doc/Release/Completion-System.html#Autoloaded-files). The gist is that you put a file named `_[command]` (for example `_git`) into a directory that's in the `fpath`, with the first line of the file being `#compdef command` (for example `#compdef git`). Whenever the shell needs completions for the command, it executes that file. The equivalent of the completion script above would be:

```zsh
#compdef my-command

_my_command_yargs_completions()
{
  local reply
  local si=$IFS
  IFS=$'
' reply=($(COMP_CWORD="$((CURRENT-1))" COMP_LINE="$BUFFER" COMP_POINT="$CURSOR" /Users/nevon/.volta/tools/image/packages/my-package/bin/my-command --get-yargs-completions "${words[@]}"))
  IFS=$si
  _describe 'values' reply
}
_my_command_yargs_completions "$@"
```

[...]

PR #1856, opened 2021-01-25
Zsh completion is now installable system-wide

gardenappl:
I manage [an AUR package](https://aur.archlinux.org/packages/readability-cli/#news) for my own node CLI tool, so I want to be able to install Bash/Zsh completions system-wide.

Bash completion works fine when I put it in `/usr/share/bash-completion/completions/<app name>`. But when I try to put the Zsh completion into `/usr/share/zsh/site-functions/_<app name>` it does not work at all.

This is fixed by adding `#compdef _<app name>` at the top of the completion file.

PR #2424, opened 2024-07-07
Fix Zsh autoloaded completion to not require double tab

segevfiner:
Currently when you use the generated completion script in Zsh via autoloading, it doesn't provide completions on the first tab press. This is because when Zsh autoloads a completion script, it expects it to directly generate completions, this little trick allows it to be directly evaluated and used as an autoloaded script.

P.S. I think if the shell function is named the same as the autoloaded script, Zsh will call it directly on subsequent calls, bypassing the rest of the script, e.g. the `if` code below. See https://zsh.sourceforge.io/Doc/Release/Functions.html#Autoloading-Functions

See https://github.com/kislyuk/argcomplete/pull/475


shadowspawn:
(Just letting you know I am planning to review this, but will need to do some testing and research. I use command-line completion often, but seldom need to modify it!)

shadowspawn:
I was not able to reproduce the problem. I do not need to press the tab key twice for auto completion using a yargs generated completion script. (zsh on Mac)

How have you saved the completion script? (Including to which shell startup script.)

segevfiner:
Ad an autoload script on my `fpath`, it will only happen with autoloading.

shadowspawn:
Did you save the completion script generated by yargs directly into the file?

segevfiner:
Yes you just save it to somewhere on the fpath of Zsh, with the `#autoload` line intact, and then restart Zsh.

shadowspawn:
Related: #2402

shadowspawn:
I was a bit slow understanding the context here, I will have another look! (Reading lots of pages about autoloading...)

shadowspawn:
The possible values for `zsh_eval_context` are listed here: https://zsh.sourceforge.io/Doc/Release/Parameters.html#Parameters-Set-By-The-Shell

lib/completion-templates.ts
```diff
@@ -44,6 +44,10 @@ _{{app_name}}_yargs_completions()
   IFS=$si
   _describe 'values' reply
 }
-compdef _{{app_name}}_yargs_completions {{app_name}}
+if [[ $zsh_eval_context == *func ]]; then
```
shadowspawn, line 47: This looks a bit magical. The possible values matching `*func` are `evalautofunc `, `loadautofunc`, and `shfunc`.

Is there a reason to test for anything other than `loadautofunc`? Which would be clearer.

shadowspawn:
The installation instructions could include an explicit example for autoload or at least a mention that the script can be saved into a file to use with autoload.

https://github.com/yargs/yargs/blob/0c95f9c79e1810cf9c8964fbf7d139009412f7e7/lib/completion-templates.ts#L36

lib/completion-templates.ts
segevfiner: It's `loadautofunc` when first autoloaded and `shfunc` on subsequent calls.

shadowspawn: After it gets autoloaded with context `loadautofunc` , I do not think these lines are executed again.

The completion function (not the whole file) gets loaded into shell and called multiple times with context `shfunc`, but I don't think that is relevant to the code outside the completion function.

segevfiner: It'e actually the whole file that becomes the completion function unless the file defines a function with the exact same name as the one it was autoloaded with. Which is not the case here ATM. But we can change the function name if you want to go that route.

shadowspawn: I did carefully save the completion to a file with the name of the autocomplete function, as per advice. I'll try using a different name to check behaviour. I had a small battle to get autoload working but think I have the hang of it now...

shadowspawn: > It'e actually the whole file that becomes the completion function unless the file defines a function with the exact same name as the one it was autoloaded with. 

Yes. 

> Which is not the case here ATM. But we can change the function name if you want to go that route.

I might be missing something (again). This is the case ATM as long as user creates the file with the name of the function supplied by Yargs?

But I see that if user uses a different filename, the `shfunc` context comes into play. The autoload defines a function based on the filename, which calls/includes the Yargs supplied function. I am slightly concerned `shfunc` could trigger incorrectly in some context outside autoload, but not very likely.

I was thinking the user instructions could mention the autoload case with the "correct" filename, something like:

```
# Installation: foo completion >> ~/.zshrc
#    or foo completion >> ~/.zprofile on OSX.
#    or save completion into executable file named _foo_yargs_completions in autoload fpath
```

segevfiner: It would still be safer to have the if like that, it will work either way in that case. It is customary to call the autoload scripts just `_{cmdname}`, e.g. `_git`, having it require a specific name might just cause mistakes for users or package maintainers in the future...

shadowspawn:
For historical interest, the `#compdef` tag and initial autoload support got added in: #1856

shadowspawn:
I have learnt a lot reviewing this PR! Thanks for the help.

9 GitHub hits for `"*func" "zsh_eval_context" "#compdef" NOT is:fork`
81 GitHub hits for `"loadautofunc"  "zsh_eval_context" "#compdef" NOT is:fork`

Potential problems with `*func`:
- `*func` will match `shfunc` which could occur if user wraps completion code in a function. Low chance false positive.
- `*func` will match `evalautofunc`, but I have not worked out whether it makes a difference in practice.

The failure mode if user saves the completion into a file in `fpath` with the wrong name is need to hit tab twice the first time. That isn't so bad.

1) From looking through other implementations, I like `if [[ "${zsh_eval_context[-1]}" == "loadautofunc" ]]; then` as pretty clear and robust.

2) I think the autoload support should be included in the usage tips. I personally would use the autoload rather than pollute my startup file, if I knew it was an alternative.

3) I do see `_{cmdname}` commonly used for the autoload filename, and ideally it should match the internal function name. I suggest we don't change the internal function name in this PR but consider separately for a major version release.

shadowspawn:
A different approach could be to specify the intended context for the completion script. Like say:
```
my-command completion --autoload
```

It would remove the need for a few lines of context detection code, but probably not worthwhile.

segevfiner:
@shadowspawn So, you want me to change the `if`, and add a tip about using that to the help and usage?

shadowspawn:
Feedback welcome, but yes please.

For a minor version release, I am more concerned about breaking any existing usage than about someone saving the completion for autoloading with the wrong filename and needing to double-tab to activate the completion.

I did wonder about putting in a comment about the naming issue into the completion, saying for autoload the function name and filename should match, and that it is ok to change the function name in the two places it appears so can use a different filename (like `_my-command`). Do you think that would help? The naming issue is quite simple but quite subtle! But might be clear enough to cover lightly without explaining the detail behind the issue.

segevfiner (commented):
BTW Bash also has completion auto-loading via bash-completion, which implements it specifically for completion scripts. It uses it's own set of directories, and the names of the files in it instead of `#compdef` lines to map completion scripts to command.

BTW The quotes are not needed when using `[[` instead of `[` as far as I remember.

Another alternative is something like:
```sh
if [ "$funcstack[1]" = "_{{app_name}}_yargs_completions" ]; then
    _{{app_name}}_yargs_completions
fi
```

Which is used by Cobra https://github.com/spf13/cobra/blob/78bfc837fe358c750faa7e7f0a8016b300044d58/zsh_completions.go#L300-L303

docs/api.md
```diff
@@ -343,7 +343,14 @@ discussion of the advanced features exposed in the Command API.
 Enable bash/zsh-completion shortcuts for commands and options.
 
 `cmd`: When present in `argv._`, will result in the `.bashrc` or `.zshrc` completion script
-being outputted. To enable bash/zsh completions, concat the generated script to your
+being outputted.
+
+To enable bash/zsh completions, you can either: 
+1. Write it to a file in `$XDG_DATA_HOME/bash-completion` or
+   `~/.local/share/bash-completion` with the same name as the command (or
+   for Zsh, to a file in your `$fpath` named `_{{app_name}}_yargs_completions()`
```
shadowspawn, line 351:
```suggestion
   for Zsh, to a file in your `$fpath` named `_{{app_name}}_yargs_completions`
```

shadowspawn:
I think the two directories you mention for Bash are used by the [bash-completion](https://github.com/scop/bash-completion/blob/main/README.md) package, and not vanilla Bash behaviour? (i.e. `$XDG_DATA_HOME/bash-completion` and  `~/.local/share/bash-completion`).

(Although the README says "bash-completion" it is just tying together the two words and not referring to the package.)

README.md
```diff
@@ -38,7 +38,7 @@ Rules & Behavior
                              return a Promise                          [boolean]
 ```
 
-* bash-completion shortcuts for commands and options.
+* bash-completion (and Zsh completion) shortcuts for commands and options.
```
shadowspawn, line 41:
```suggestion
* generate completion scripts for Bash and Zsh for your command
```

shadowspawn: (Good tidy-up adding Zsh mention in the README.)

segevfiner:
> I think the two directories you mention for Bash are used by the [bash-completion](https://github.com/scop/bash-completion/blob/main/README.md) package, and not vanilla Bash behaviour? (i.e. `$XDG_DATA_HOME/bash-completion` and `~/.local/share/bash-completion`).
> 
> (Although the README says "bash-completion" it is just typing together the two words and not referring to the package.)

Yes it is referring to that. It is generally installed by default as the main way to get completion in Bash in most distros I know.

docs/api.md
```diff
@@ -343,7 +343,14 @@ discussion of the advanced features exposed in the Command API.
 Enable bash/zsh-completion shortcuts for commands and options.
 
 `cmd`: When present in `argv._`, will result in the `.bashrc` or `.zshrc` completion script
-being outputted. To enable bash/zsh completions, concat the generated script to your
+being outputted.
+
+To enable bash/zsh completions, you can either: 
+1. Write it to a file in `$XDG_DATA_HOME/bash-completion` or
```
shadowspawn, line 349: This bash description will only work if the `bash-completion` package is installed. I checked macOS and Debian and Ubuntu and Fedora and none of them had it installed by default.

Sadly appending to the shell start-up scripts is the lowest common denominator! I suggest have that first, and then the zsh autoload instructions.

Perhaps rather than using filename placeholder here, could point to the completion text where it can be included. Like:

> 2. For Zsh autoloading, write it to a to a file in your `$FPATH`. See completion script for required file name.

docs/api.md
```diff
@@ -343,8 +343,21 @@ discussion of the advanced features exposed in the Command API.
...
-`.bashrc` or `.bash_profile` (or `.zshrc` for zsh).
+being outputted.
+
+To enable bash/zsh completions, you can either: 
+1. Concat the generated script to your
+   `.bashrc` or `.bash_profile` (or `.zshrc` for zsh).
+
+   e.g. `./command completion > ~/.bashrc`
+
+2. If you have `bash-completion` installed, write it to a file in
```
shadowspawn, line 354: That is better with explicit `bash-completion` context, thanks.

segevfiner:
@shadowspawn Updated

shadowspawn (approved):
LGTM

Thanks @segevfiner

shadowspawn:
(I pushed an empty commit to trigger a build, which then failed because #2393 has not landed yet. No action required here.)

bcoe:
@segevfiner `${` is a reserved character in a template string, so is causing the build to fail.

Bother you to test zsh locally with the generated script 👏  (it should be easier to build locally on v18 of yargs).

bcoe:
@segevfiner were you able to test the generated zsh completion script locally?

segevfiner:
Not yet. I still need to test this again.

bcoe:
> Not yet. I still need to test this again.

👍 

@segevfiner just let me know when you've done so, and I'll merge.

bcoe:
@segevfiner this change is available by installing `yargs@18.0.0-candidate.5`.

Issue #2439, opened 2024-10-30
Change from `yerror.js` to `yerror.ts`


victorfernandesraton:
# Issue
When try to compile yargs in deno 2.0 i received this error 
```
error: Module not found "https://deno.land/x/yargs@v17.7.2/build/lib/yerror.js".
    at https://deno.land/x/yargs@v17.7.2/lib/platform-shims/deno.ts:18:22
````
i thin is a import error there, they try import `js` file but only `ts` file exists

https://github.com/yargs/yargs/blob/3566b84b24a7e8cf24badac73ea1d20a0851924e/lib/platform-shims/deno.ts#L18

# Steps to reproduce
- Create a simple deno project with `deno init`
- Put above code in `main.ts` file
- Trying compile using deno compile -A main.ts

```typescript
import yargs from "https://deno.land/x/yargs@v17.7.2/deno.ts";
import type { Arguments } from "https://deno.land/x/yargs@v17.7.2/deno-types.ts";

// Configura o yargs para capturar os argumentos da linha de comando
const argv = yargs(Deno.args)
    .option("portFacade", {
        alias: "pf",
        type: "number",
        description: "Porta da Facade",
        // demandOption: true,
    })
    .demandCommand(0)
    .strictOptions()
    .parseSync();

function main(args: Arguments) {
    console.log(args);
}

main(argv);
```
# System information
OS: Win11 x64
Deno: deno 2.0.3 (stable, release, x86_64-pc-windows-msvc)
v8 12.9.202.13-rusty
typescript 5.6.2

shadowspawn:
Thanks for the clear description.

The problem is the tags being used for the deno versions of yargs have `-deno` on the end of the version number. Try adding that:

```diff
- import yargs from "https://deno.land/x/yargs@v17.7.2/deno.ts";
+ import yargs from "https://deno.land/x/yargs@v17.7.2-deno/deno.ts";
- import type { Arguments } from "https://deno.land/x/yargs@v17.7.2/deno-types.ts";
+ import type { Arguments } from "https://deno.land/x/yargs@v17.7.2-deno/deno-types.ts";
```

PR #2440, opened 2024-11-02
docs: improve readme

victorfernandesraton:
Add information about `-deno` in the end of version tag url for deno.land and avoid mistakes like https://github.com/yargs/yargs/issues/2439

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,24 @@ yargs(Deno.args)
   .parse()
 ```
 
+> Note: If you specify version tags in url , you also have to add `-deno` flag, like this:
+```typescript
+import yargs from "https://deno.land/x/yargs@v17.7.2-deno/deno.ts";
+import type { Arguments } from "https://deno.land/x/yargs@v17.7.2-deno/deno-types.ts";
+
+yargs(Deno.args)
+  .command('download <files...>', 'download a list of files', (yargs: any) => {
+    return yargs.positional('files', {
+      describe: 'a list of files to do something with'
+    })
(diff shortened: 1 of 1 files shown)


shadowspawn:
I think an example using version tags is a good idea. I suggest rather than adding a copy of the whole example, change the example to use the tagged import. This is the Deno best practice, and getting latest version is just removing the version and same as normal. A comment that the version tag for yargs includes `-deno` is appropriate as it is non-standard.

Q&A section covers version:  https://deno.land/x

We should move or dual publish to JSR too, but that is a separate task:
- https://deno.com/blog/http-imports
- https://deno.com/blog/jsr_open_beta

shadowspawn:
I suggest also changing the imports in the example to use version tags. So effectively your example replaces the old example. (You might already be doing this and my comment is too quick!)

shadowspawn (approved):

README.md
```diff
@@ -141,6 +141,8 @@ yargs(Deno.args)
   .parse()
 ```
 
+> Note: If you define version tags in url , you also have to add `-deno` flag in the end , like `@17.7.2-deno`
```
shadowspawn, line 144:
```suggestion
> Note: If you use version tags in url then you also have to add `-deno` flag on the end, like `@17.7.2-deno`
```

shadowspawn (approved):
Thanks @victorfernandesraton

shadowspawn:
(Build failure is not blocking as this PR is docs only. Build fix is in #2393.)

PR #2445, opened 2025-01-02
fix(build): address problems with typescript compilation

bcoe:
This addresses bit rot in the rollup compilation, which is used to expose both a CJS and MJS module.

## Note

Now that you can require ESM modules in CJS, I think there's an argument for making yargs pure ESM, which would remove the need for this build step, and would reduce the size of yargs by 100%, see:

https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require

### DIFF
--- a/.nycrc
+++ b/.nycrc
@@ -8,7 +8,7 @@
     "html",
     "text"
   ],
-  "lines": 100,
+  "lines": 99,
   "branches": "96",
   "statements": "100"
 }
--- a/lib/platform-shims/cjs.ts
+++ b/lib/platform-shims/cjs.ts
@@ -30,7 +30,7 @@ export default {
       process.emitWarning(warning, type),
     execPath: () => process.execPath,
     exit: (code: number) => {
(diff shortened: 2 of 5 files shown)


shadowspawn (approved):

shadowspawn:
> Now that you can require ESM modules in CJS, I think there's an argument for making yargs pure ESM

Not sure how aggressively you are considering this, although I know you have been looking forward to pure esm for a long time!

Would you only ship this after feature available without flag or warning in all LTS versions of node?

bcoe:
> Not sure how aggressively you are considering this, although I know you have been looking forward to pure esm for a long time! 

I haven't been doing too much OSS these days (as I'm sure you've noticed). But, I'm pretty serious since having this complicated toolchain to do dual-mode ESM:

* makes keeping this library building annoying.
* makes it harder for people to contribute.

> Would you only ship this after feature available without flag or warning in all LTS versions of node?

I think if we do it as a major, we could just say _you-only-live-once_ and ship. There's the option of using `import.then`.

PR #2451, opened 2025-02-18
feat!: yargs is now ESM first

bcoe:
See: https://joyeecheung.github.io/blog/2024/03/18/require-esm-in-node-js/ for what's motivating this work.

Yargs build step to be dual-mode ESM / CJS had created a fragile build process that has been hard to upgrade over the years. I'm quite excited to remove it.

Known Breaking Changes:

* command names are now only derived if you're using `commandDir`. If you are providing a module directly, a `command` property must be set.
* Supported Node.js versions are now `^20.19.0 || ^22.12.0 || >=23` (there were fixes to issues with require ESM in patch releases of 20.x and 22.x which is why the complicated version range is necessary).
* yargs no longer works as a singleton, breaking usage like `yargs.parse()`, `yargs.argv`.
* I'm sure I've broken some other things with this major overhaul, we'll keep this in beta for a healthy length of time.

TODO:
  - [x] Look into adding back command name detection for commandDir.
  - [x] rewrite all docs to not use yargs as singleton (this is no longer a thing in CJS compatibility mode).
      - [x] advanced.md
      - [x] readme.md
      - [x] api.md 
      - [x] examples.md 

---

## Notes

* I left the integration tests in CJS so that the behaviour of requiring yargs in a CJS environment can be tested -- this requires Node 23.
  * Note: Node 22 still seems to be having issues on Windows.
* ~cliui doesn't wrap properly because it's using an old version of wrap-ansi. TODO, unravel this problem.~ fixed thanks @shadowspawn.
* removed which-module functionality (TODO: dig into what the heck I was thinking when I wrote this).
  * figured it out, see: inferring module name from path.
* ~1000 lbs 🦍 in the room, should we figure out a replacement for `commandDir`, perhaps using `createRequire`? could this be a follow up PR to reduce the scope of this work? (CC: @shadowspawn thoughts).~ Figured out how to make `commandDir` work, including inferring name of command from directory.
* Coverage is working again and readable \o/ glad to have rollup out of the mix 🥳


bcoe:
@shadowspawn this is a work in progress, but porting over tests is going pretty smoothly so far.

shadowspawn:
Oh ho, exciting!

shadowspawn:
Probably running ahead of your work in progress. I was trying the branch and found that using require from CJS and experimental-require-module did not get the function with a default import. Tested with node v22 and v23.

```js
const yargsFactory = require('yargs');
console.log(yargsFactory);
```

```
[Module: null prototype] {
  __esModule: true,
  default: [Function (anonymous)]
}
```

The documentation has a work-around, with the downside it hides named exports. But that is fine in current setup, since there are none.

```
# Add at bottom of index.mjs
export {Yargs as 'module.exports'};
```

> When a ES Module contains both named exports and a default export, the result returned by `require()` is the module namespace object, which places the default export in the `.default` property, similar to the results returned by `import()`. To customize what should be returned by require(esm) directly, the ES Module can export the desired value using the string name `"module.exports"`.

Reference: https://nodejs.org/docs/latest/api/modules.html#loading-ecmascript-modules-using-require

package.json
```diff
@@ -2,31 +2,16 @@
   "name": "yargs",
   "version": "17.7.2",
   "description": "yargs the modern, pirate-themed, successor to optimist.",
-  "main": "./index.cjs",
+  "main": "./index.mjs",
   "exports": {
     "./package.json": "./package.json",
-    ".": [
```
shadowspawn, line 8: Still need an export of `"."` to get access to `index.mjs`.

package.json
```diff
@@ -2,31 +2,16 @@
...
-      {
-        "import": "./index.mjs",
-        "require": "./index.cjs"
-      },
-      "./index.cjs"
-    ],
     "./helpers": {
-      "import": "./helpers/helpers.mjs",
-      "require": "./helpers/index.js"
+      "import": "./helpers/helpers.mjs"
```
shadowspawn, line 9: Not sure if the lone import was intentional. I was expecting just file path rather than conditional.

```
".helpers": "./helpers/helpers.mjs"
```

shadowspawn:
(I can postpone comments until later in draft or post-draft if you prefer!)

package.json
bcoe: 👍 thanks reading this as I work through things.

bcoe: Added back ".".

package.json
bcoe: Done.

bcoe:
> Probably running ahead of your work in progress. I was trying the branch and found that using require from CJS and experimental-require-module did not get the function with a default import. Tested with node v22 and v23.

@shadowspawn switched to using `export {foo as 'module.exports'};` for both cliui and this PR, it seems to work well.

> Some cliui PRs open to add wrapping in ESM by working around dependency complications, but perhaps instead moving cliui to ESM-only and use latest dependencies would be in line with this PR?

@shadowspawn thanks for the PR on cliui, I think this is mostly addressed (although I expect we'll have some follow up PRs on that repo).

I was able to stop skipping all the tests related to string wrapping 👍

shadowspawn:
> * 1000 lbs 🦍 in the room, should we figure out a replacement for `commandDir`, perhaps using `createRequire`? could this be a follow up PR to reduce the scope of this work? (CC: @shadowspawn thoughts).

`createRequire` might allow backwards compatible, but would not be moving codebase forward towards supporting `commandDir` like functionality for esm. I would rather work on something that works with esm too.

Assuming we use dynamic import, we need to return a promise somewhere. A way to hook this in with existing `.commandDir()` signature might be to delay the actual work until `.parseAsync()` is called. This would change the implementation details but potentially keep code compatibility for existing users.

I think this could be a followup PR. Comparison of usage, take with salt:
- 3.6k uses of `yargs AND commandDir`: [GitHub search]( https://github.com/search?q=yargs+AND+commandDir+AND+NOT+is%3Afork+AND+%28language%3Ajavascript+OR+language%3Atypescript%29&type=repositories)
- 2.5M uses of `yargs` [GitHub search](https://github.com/search?q=yargs+AND+NOT+is%3Afork+AND+%28language%3Ajavascript+OR+language%3Atypescript%29&type=repositories)

shadowspawn:
> removed which-module functionality (TODO: dig into what the heck I was thinking when I wrote this).

I had never looked at this code and did a little digging. Looks like it got copied from an external package when moving to TypeScript:
- https://github.com/yargs/yargs/blob/028b50d5ced3b41a8dccf74107dbfc7065052a5d/lib/command.ts#L12
- https://github.com/nexdrew/which-module#readme

shadowspawn:
@bcoe Not essential, but are you interested in matching esm-only work on `yargs-parser` and `y18n`, which will include fixing up builds and updating dependencies? (`cliui` was easier than I expected! But it is small.)

bcoe:
> Not essential, but are you interested in matching esm-only work on yargs-parser and y18n, which will include fixing up builds and updating dependencies? (cliui was easier than I expected! But it is small.)

Seems like a good call (just lower priority as you say).

> createRequire might allow backwards compatible, but would not be moving codebase forward towards supporting commandDir like functionality for esm. I would rather work on something that works with esm too. (Edit: based on limited research. I had not come across createRequire until looking into this.)

I agree I'd rather move us forward to trying to make ESM work. I might experiment a bit with `createRequire` just to minimize breaking changes if possible (and give up if it turns out to be too painful).

shadowspawn:
I wondered today whether `createRequire` might actually work with esm files too, since `require(esm)` works... It does! (Only tried in non-module context.)

That changes my opinion, looking into `createRequire` to reimplement `.commandDir()` does seem interesting. 😄 

```js
// foo.cjs
const moduleType = 'commonjs';
module.exports.moduleType = moduleType;
```

```js
// bar.mjs
export const moduleType = 'module'; 
```

```js
import { createRequire } from 'module';
const require = createRequire(import.meta.url);
const foo = require('./foo.cjs');
const bar = require('./bar.mjs');
console.log(foo.moduleType);
console.log(bar.moduleType);
```

```console
$ node --version                
v22.14.0
$ node createRequire.mjs 
commonjs
module
```

bcoe:
> I wondered today whether createRequire might actually work with esm files too, since require(esm) works... It does! (Only tried in non-module context.)

I was thinking the exact same thing. I'm still going to add tests for ESM modules after landing this 👍

shadowspawn:
(I missed you had marked the PR as ready for review a couple of weeks ago, I'll try and have a deeper look this week.)

package.json
```diff
@@ -2,31 +2,16 @@
...
     },
-    "./yargs": [
-      {
-        "import": "./yargs.mjs",
-        "require": "./yargs"
-      },
-      "./yargs"
-    ]
+    ".": "./index.mjs",
+    "./yargs": "./index.mjs"
```
shadowspawn, line 14: What is the short or long term intent with require/import `yargs/yargs`? Is this just for backwards compatibility and could be removed from examples, or stay around for a while?

package.json
```diff
@@ -39,12 +24,9 @@
   "files": [
     "browser.mjs",
     "browser.d.ts",
-    "index.cjs",
     "helpers/*.js",
     "helpers/*",
     "index.mjs",
-    "yargs",
```
shadowspawn, line 46: Glad to see this file with no extension going away. 👍  (I suspect it may have caused a couple of the rare weird issues reported with third party tools getting confused about whether cjs or esm.)

package.json
```diff
@@ -53,58 +35,53 @@
...
-    "postcompile": "npm run build:cjs",
-    "build:cjs": "rollup -c rollup.config.cjs",
-    "postbuild:cjs": "rimraf ./build/index.cjs.d.ts",
     "check": "gts lint && npm run check:js",
-    "check:js": "eslint . --ext cjs --ext mjs --ext js",
+    "check:js": "eslint . --ext mjs --ext js",
     "clean": "gts clean",
-    "watch": "rimraf build && tsc && concurrently npm:watch:tsc npm:watch:cjs",
-    "watch:cjs": "rollup -w -c rollup.config.cjs",
+    "watch": "rimraf build && tsc && concurrently npm:watch:tsc",
```
shadowspawn, line 84: I think the plain `tsc` was to prepare a clean start for `watch:cjs` (I had wondered about it in past). With cjs gone, can remove the plain `tsc` and also the `concurrently`.

```suggestion
    "watch": "rimraf build && npm:watch:tsc",
```

package.json
```diff
@@ -53,58 +35,53 @@
...
+    "@babel/preset-typescript": "^7.26.0",
     "@types/chai": "^4.2.11",
     "@types/mocha": "^9.0.0",
     "@types/node": "^20.0.0",
+    "@typescript-eslint/eslint-plugin": "^8.26.1",
     "browserslist-generator": "^2.0.1",
     "c8": "^9.1.0",
     "chai": "^4.2.0",
     "chalk": "^4.0.0",
     "concurrently": "^7.6.0",
```
shadowspawn, line 57: Not currently using `concurrently` so can be removed.

package.json
```diff
@@ -123,6 +100,6 @@
   ],
   "license": "MIT",
   "engines": {
-    "node": ">=16.15.1"
+    "node": "^20.19.0 || ^22.12.0 || >=23"
```
shadowspawn, line 103: I assume these versions for unflagged support of `require(esm)`? I agree with this. 

Could allow older versions since they would work with esm and flagged support for `require(esm)`, but simpler story and support and user experience to have it Just Work if installs.

package.json
```diff
@@ -53,58 +35,53 @@
...
+    "eslint": "^8.57.1",
     "eslint-plugin-prettier": "^5.1.2",
     "gts": "^5.2.0",
     "hashish": "0.0.4",
-    "mocha": "^9.0.0",
+    "mocha": "^10.8.2",
     "rimraf": "^3.0.2",
-    "rollup": "^3.29.4",
-    "rollup-plugin-cleanup": "^3.2.1",
     "typescript": "^5.7.2",
```
shadowspawn, line 67: Asking a question rather than recommending as such. TypeScript does not follow semver, and the yargs and yargs-parser builds have broken in past for minor versions. We might use "~5.7.2" or even "5.7.2", and only apply upgrades after `renovate` tries them?

.github/workflows/ci.yaml
```diff
@@ -29,31 +29,17 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: 20
+          node-version: 23
```
shadowspawn, line 32: I suggest this uses an LTS version? (i.e. 22)

.github/workflows/release-please.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v1
         with:
-          node-version: 20
+          node-version: 23
```
shadowspawn, line 23: I suggest this uses an LTS version? (i.e. 22)

CHANGELOG.md
```diff
@@ -243,88 +243,3 @@ All notable changes to this project will be documented in this file. See [standa
 * deprecated reset() method is now private (call yargs() instead). ([376f892](https://www.github.com/yargs/yargs/commit/376f89242733dcd4ecb8040685c40ae1d622931d))
 * implicitly private methods are now actually private ([376f892](https://www.github.com/yargs/yargs/commit/376f89242733dcd4ecb8040685c40ae1d622931d))
 * **yargs-factory:** refactor yargs-factory to use class ([#1895](https://www.github.com/yargs/yargs/issues/1895)) ([376f892](https://www.github.com/yargs/yargs/commit/376f89242733dcd4ecb8040685c40ae1d622931d))
```
shadowspawn, line 245: For interest, was there any particular reason for archiving the older changelogs? 

(I did this myself in Commander because CHANGELOG was included in the published package, but npm no longer does that so I shifted it back to a single file. I like being able to search for what versions changes shipped in, but it is only a minor convenience.)

docs/advanced.md
```diff
@@ -14,11 +14,11 @@ commands. tldr; default commands allow you to define the entry point to your
 application using a similar API to subcommands.
 
 ```js
-const argv = require('yargs/yargs')(process.argv.slice(2))
+const argv = require('yargs')(process.argv.slice(2))
   .command('$0', 'the default command', () => {}, (argv) => {
     console.log('this command will be run by default')
   })
-  .parse()
+  .parse(process.argv.slice(2))
```
shadowspawn, line 21: I suggest `process.argv.slice(2))` in either factory call, or in parse, but not both.

docs/advanced.md
```diff
@@ -27,11 +27,11 @@ is run with `./my-cli.js --x=22`.
 Default commands can also be used as a command alias, like so:
 
 ```js
-const argv = require('yargs/yargs')(process.argv.slice(2))
+const argv = require('yargs')(process.argv.slice(2))
   .command(['serve', '$0'], 'the serve command', () => {}, (argv) => {
     console.log('this command will be run by default')
   })
-  .parse()
+  .parse(process.argv.slice(2))
```
shadowspawn, line 34: I suggest `process.argv.slice(2))` in either factory call, or in parse, but not both.

test/command.mjs
```diff
@@ -16,7 +18,7 @@ async function wait() {
 
 describe('Command', () => {
   beforeEach(() => {
-    yargs.getInternalMethods().reset();
+    yargs().getInternalMethods().reset();
```
shadowspawn, line 21: Is this needed? It looks like it creates a yargs instance, calls a method on it, and abandons it to the garbage collection. So surplus to requirements now there is not a singleton in play.

test/integration.mjs
```diff
@@ -228,7 +227,13 @@ function testCmd(cmd, args, cb) {
   const cmds = cmd.split(' ');
 
   const bin = spawn(cmds[0], cmds.slice(1).concat(args.map(String)), {
-    cwd: path.join(__dirname, '/fixtures'),
+    cwd: path.resolve('./test/fixtures'),
+    env: {
+      ...process.env,
+      ...{
+        NODE_OPTIONS: '--experimental-require-module',
```
shadowspawn, line 234: Is this ENV still needed, or perhaps left over from earlier work before enforcing node version?

test/command.mjs
shadowspawn: Found it was deleted in another test file, so probably stale.

test/yargs.mjs
```diff
@@ -663,38 +645,36 @@ describe('yargs dsl tests', () => {
...
-          .help('h')
-          .wrap(null)
-          .detectLocale(false)
-          .parse();
+        y.parse(['snuh', '-h']);
       });
 
-      yargs.locale().should.equal('en');
-      yargs.getDetectLocale().should.equal(false);
+      y.locale().should.equal('en_US');
```
shadowspawn, line 676: Very minor. Do you know the locale changed from `en` to `en_US`? I wasn't expecting a functional change.

test/yargs.mjs
```diff
@@ -924,13 +913,13 @@ describe('yargs dsl tests', () => {
       const r = checkOutput(() => {
         yargs(['--help']).command('blerg', 'blerg command').wrap(null).parse();
       });
-
+      console.info(r.logs[0]);
```
shadowspawn, line 916: Left over logging?

.github/workflows/ci.yaml
shadowspawn: Saw this comment, not sure if still applies:

> I left the integration tests in CJS so that the behaviour of requiring yargs in a CJS environment can be tested -- this requires Node 23.

.github/workflows/release-please.yml
shadowspawn: Saw this comment, not sure if still applies:

> I left the integration tests in CJS so that the behaviour of requiring yargs in a CJS environment can be tested -- this requires Node 23.

package.json
shadowspawn: (To be clear, "Not currently using `concurrently`" if removed from npm run script for `watch`.)

shadowspawn:
I have done a pass through whole PR and made some minor comments and questions, nothing blocking. 

I plan to do a little user testing on old and new commandDir next.

shadowspawn:
The documentation for `.commandDir` describes the `include` and `exclude` options for `require-directory` which haven't been implemented. 

https://github.com/yargs/yargs/blob/ef28c98ae14621fd484ec80b128b46e85b6bf858/docs/advanced.md?plain=1#L272-L278

The previous code was passing through user supplied options object so the user could also supply other `require-directory` options like `renamer`, but that behaviour was not documented.

shadowspawn (approved):
> I'm sure I've broken some other things with this major overhaul, we'll keep this in beta for a healthy length of time.

Qualified approval. I think `.commandDir()` still a work in progress with include/exclude to implement or deprecate. But overall and notwithstanding `.commandDir()`, up to at least beta quality with what I saw and tried.

Nice work!

package.json
bcoe: My thought was it's just cheap to add this export for backwards compatibility. Stops us from breaking people who followed old documentation.

package.json
bcoe: Someone using pure ESM should be able to use a looser version range of Node.js, it's relatively rare that people actually install with `--engine-strict` I think?

CHANGELOG.md
bcoe: I was doing it to keep the bundle size down, I didn't realize npm no longer published CHANGELOG.md.

.github/workflows/ci.yaml
bcoe: LTS version is working now. When I started this work the LTS version wasn't yet working for requiring ESM in CJS.

package.json
shadowspawn: Definitely want it for backwards compatibility, and I liked it did not need a file. Nice "trick".

I was mainly wondering if `yargs/yargs` will be disappearing from customer facing docs.

bcoe: I'll remove any references I see to `yargs/yargs` in docs as part of this PR 👍

package.json
bcoe: It's been long while since TypeScript did this to us, and I guess I have a short memory 😆 

I'm tempted to leave it as is for now. We can revisit the decision in the future if our builds break, and I can be blamed.

CHANGELOG.md
bcoe: went ahead and recombined the CHANGELOGs given this feedback.

docs/advanced.md
bcoe: I believe I've addressed this, please feel free to follow up with a docs PR if I've missed your meaning.

test/integration.mjs
bcoe: I believe this is no longer needed. Removed.

test/yargs.mjs
bcoe: Removed 👍

test/yargs.mjs
bcoe: y18n is a singleton, and I found an edge-case in tests during my refactor where the state of the locale from a prior test would bleed over. To fix the problem I started calling `this.#shim.y18n.setLocale(DEFAULT_LOCALE);` which sets the locale to `en_US`. I chose `en_US` because I believe it's more correct than just `en` without a country code.

shadowspawn: 👍 Thanks for explanation.

bcoe:
Regarding:

> The documentation for .commandDir describes the include and exclude options for require-directory which haven't been implemented.

@shadowspawn I've added back support for `include` / `exclude`, based on how I _think_ it should work. The [documentation for `require-directory`](https://www.npmjs.com/package/require-directory#whitelisting) seemed off to me, with it documenting the same behaviour for include / exclude.

> The previous code was passing through user supplied options object so the user could also supply other require-directory options like renamer, but that behaviour was not documented as such.

Given that this behaviour isn't documented in yargs, I think it's not a breaking change to remove it.

lib/command.ts
```diff
@@ -61,27 +62,60 @@ export class CommandInstance {
     opts?: RequireDirectoryOptions
   ): void {
     opts = opts || {};
-
-    // disable recursion to support nested directories of subcommands
-    if (typeof opts.recurse !== 'boolean') opts.recurse = false;
+    this.requireCache.add(callerFile);
+    const fullDirPath = join(dirname(callerFile), dir);
```
vipzhicheng, line 66: Hey, I found that `commandDir` has been changed to be relative from here on. I just want to know the reason. Is it related to security concerns? Because in my case, I use the `commandDir` feature extensively to load commands from various locations (v17.7.2), not just from a single project directory. For example, I combine project commands, home directory commands, and global directory commands. With this change, my approach has become completely unfeasible. So I'd like to know if there's an option that allows the directory to be an absolute directory.

vipzhicheng: Except for the above case, there is another scenario that would also fail. Put `yargs` in the `@org/core` package, and export an API to launch CLI tools. Then, add commands from another package that imports `@org/core` and passes the directory via the API.

shadowspawn: I think the intent is that the directory can be absolute, and it is probably an oversight (bug) if an absolute path passed into `.commandDir()` does not work.

Have you confirmed this problem at runtime, or is this feedback from code inspection?

(I won't have time to try reproducing and look carefully until later.)

vipzhicheng: I prepared a demo repo for this issue. It can be reproduced. When you have time, you can give it a shot.

https://github.com/vipzhicheng/test-yargs-next

Logically, `fullDirPath` is `callerFile` dir + passed in dir, I hope it is just passed in dir, like before.

vipzhicheng: I have made a PR https://github.com/yargs/yargs/pull/2465, Please take a look. 

Just like you suggested, it works in my test repo and also still passes all `npm run test` after the change.

PR #2452, opened 2025-02-19
fix: handle spaces in bash completion

oldium:
The spaces are not handled well in bash completion, the offered result needs to be escaped. Fix parsing the word list and add quotation around suggestions with spaces. See [here][1] for the discussion.

[1]: https://stackoverflow.com/a/40944195/7080036


shadowspawn:
What is an example of this being a problem with yargs completion? (Did you encounter this?)

I understand the theoretical problem, but I am not sure when it would come up in practice.

oldium:
> What is an example of this being a problem with yargs completion? (Did you encounter this?)

Yes, with https://github.com/firecow/gitlab-ci-local. When you want `yargs` to auto-complete the job name, which has a space in the name (which is usual), it simply does not work. It is expected to actually _have_ spaces in the arguments.

Just compare the following. Before patch `gitlab-ci-local build-deb<TAB>` becomes `gitlab-ci-local build-debian: [debian-11]`. With the patch it becomes `gitlab-ci-local "build-debian: [debian-11]"`. First one does not work, second one works.
 
> I understand the theoretical problem, but I am not sure when it would come up in practice.

Well, that is not a theoretical problem, that is an actual problem 😉 After all, that is why I investigated it and prepared a fix.

oldium:
This is how the empty-arguments completion looks like:

```
$ gitlab-ci-local <TAB>
[centos-10]       [debian-11]       [debian-13]       [fedora-40]       [fedora-42]       [ubuntu-2404]     build-debian:     env-vars          generate-patches
[centos-9]        [debian-12]       [fedora-39]       [fedora-41]       [fedora-43]       [ubuntu-2410]     build-fedora:     fetch-source      upload
```

This is how it looks like with my patch:

```
$ gitlab-ci-local <TAB>
"build-debian: [debian-11]"    "build-debian: [ubuntu-2404]"  "build-fedora: [centos-9]"     "build-fedora: [fedora-41]"    env-vars                       upload
"build-debian: [debian-12]"    "build-debian: [ubuntu-2410]"  "build-fedora: [fedora-39]"    "build-fedora: [fedora-42]"    fetch-source
"build-debian: [debian-13]"    "build-fedora: [centos-10]"    "build-fedora: [fedora-40]"    "build-fedora: [fedora-43]"    generate-patches
```

That is completely different result.

oldium:
Funny part is that in both cases the suggestions provided by `gitlab-ci-local` are exactly the same, so the actual problem is really provisioning the completions to bash.

bcoe:
@oldium I find it slightly alarming how much more complicated this makes the bash logic (since it's hard to QA this throughly). Are we sure `awk` and `mapfile` will be broadly available?

What if someone has bash installed in an environment like alpine, that's really bare bones?

oldium:
> @oldium I find it slightly alarming how much more complicated this makes the bash logic (since it's hard to QA this throughly). Are we sure `awk` and `mapfile` will be broadly available?
> 
> What if someone has bash installed in an environment like alpine, that's really bare bones?

`mapfile` is Bash built-in command, so there should be no issue. I checked bash on latest Alpine Docker image and `mapfile` is available.

```console
$ mapfile --help
mapfile: mapfile [-d delim] [-n count] [-O origin] [-s count] [-t] [-u fd] [-C callback] [-c quantum] [array]
    Read lines from the standard input into an indexed array variable.

    Read lines from the standard input into the indexed array variable ARRAY, or
    from file descriptor FD if the -u option is supplied.  The variable MAPFILE
    is the default ARRAY.
...
```


`awk` is a standard tool and Alpine has it as part of `busybox` tool and supports the used syntax. I made a few tests on Docker Alpine image (after installing `apk add --no-cache bash`):

```console
$ echo test me | awk '/ / { print "\""$0"\"" } /^[^ ]+$/ { print $0 }'
"test me"
$ echo testme | awk '/ / { print "\""$0"\"" } /^[^ ]+$/ { print $0 }'
testme
```

The `awk` tool is defined in POSIX standard IEEE Std 1003.1, so it should be part of any compliant operating system.

bcoe (approved):
LGTM. Thanks for the contribution.

bcoe:
@oldium give `yargs@18.0.0-candidate.2` a try (it requires newer versions of Node.js).

oldium:
@bcoe Works as expected. I re-compiled `gitlab-ci-local` from source with updated `yargs@next` (`18.0.0-candidate.3`)

```console
$ gitlab-ci-local --completion
###-begin-gitlab-ci-local-completions-###
#
#
# Installation: /home/oldium/bin/gitlab-ci-local completion >> ~/.bashrc
#    or /home/oldium/bin/gitlab-ci-local completion >> ~/.bash_profile on OSX.
#
_gitlab-ci-local_yargs_completions()
{
    local cur_word args type_list

    args=("${COMP_WORDS[@]}")

    # see https://stackoverflow.com/a/40944195/7080036 for the spaces-handling awk
    mapfile -t type_list < <(/home/oldium/bin/gitlab-ci-local --get-yargs-completions "${args[@]}")
    mapfile -t COMPREPLY < <(compgen -W "$( printf '%q ' "${type_list[@]}" )" -- "${cur_word}" |
        awk '/ / { print "\""$0"\"" } /^[^ ]+$/ { print $0 }')

      COMPREPLY=()
    fi

    return 0
}
complete -o bashdefault -o default -F _gitlab-ci-local_yargs_completions gitlab-ci-local
###-end-gitlab-ci-local-completions-###
```

```console
$ gitlab-ci-local <TAB>
.git/                                   .gitlab-ci-local-variables.yml
... <redacted>
$ source <(gitlab-ci-local --completion)
$ gitlab-ci-local <TAB>
"build: [debian,debian-11]"    "build: [debian,ubuntu-2410]"  "build: [fedora,fedora-40]"    fetch-source                   upload-debian
"build: [debian,debian-12]"    "build: [fedora,centos-10]"    "build: [fedora,fedora-41]"    generate-patches
"build: [debian,debian-13]"    "build: [fedora,centos-9]"     "build: [fedora,fedora-42]"    prepare
"build: [debian,ubuntu-2404]"  "build: [fedora,fedora-39]"    "build: [fedora,fedora-43]"    upload
```

PR #2461, opened 2025-04-19
feat: commandDir now works with ESM files

bcoe:
Turns out this _just worked_, so this PR just adds tests and updates docs.


docs/advanced.md
```diff
@@ -290,11 +290,12 @@ can either move your module to a different directory or use the `exclude` or
...
 - `include`: RegExp or function
 
-    Allow list certain modules. See [`require-directory`](https://www.npmjs.com/package/require-directory) for details.
+
+    Allow list certain modules. Either a regex or callback can be provided. `true` = load the file.
 
 - `exclude`: RegExp or function
 
-    Block list certain modules. See [`require-directory`](https://www.npmjs.com/package/require-directory) for details.
+    Block list certain modules. Either a regex or callback can be provided. `false` = do not load the file.
```
shadowspawn, line 298: I am not sure what ``false` = do not load the file.` means. If it is the callback, it looks inverted. (I have not checked the code or runtime, just reading the docs alone at the moment.)

shadowspawn (approved):
A couple of queries on wording for true/false in docs, but not blocking.

shadowspawn:
I noticed the trailing section in the docs is about the previous work-around. The section can be retained, but the wording could do with an update since now just an example of building command hierarchy by hand.

```text
<a name="esm-hierarchy"></a>
### Example command hierarchy using index.mjs

To support creating a complex nested CLI when using ESM, the method
`.command()` was extended to accept an array of command modules.
```

shadowspawn:
(Separately: I noticed I can't run the examples directly from the examples folder because they are `.js` files with `require`, and now package is type module.)

docs/advanced.md
bcoe: You're correct, it should be return `true`. Fixed.

PR #2463, opened 2025-04-20
chore: update examples to run from examples folder

shadowspawn:
The examples were not runnable from examples folder since they were CJS with `.js` extension in a project which is ESM. This was probably broken since v17, not a recent problem.

- The files were not set to executable although `examples.md` gave examples of running directly, so fixed that too.
- Left a few of the examples as CJS but changed file extension to `.cjs` so runnable. 
- Changed `var` to `const`.
- Deleted the `strict` example as it didn't run and wasn't clear what it was doing!

Messy diff since files renamed and edited.


bcoe (approved):

bcoe:
Awesome, thanks for helping with the docs @shadowspawn. I'm glad we're cleaning up some of this cruft from past releases.

PR #2465, opened 2025-04-23
fix: addDirectory do not support absolute command dir

vipzhicheng:
The issue context is under https://github.com/yargs/yargs/pull/2451/files/9e920df512e869d3d5f7917f9dbd79f99b26b93c#diff-3378419d97c7d3aa5e4c5a36f8744505164ed98aeda47707e070f8df8496b91f

The main problem is that the commandDir API function no longer supports the absolute commands directory, which was supported previously. @shadowspawn suggested changing from path.join to path.resolve. I tested his solution and it worked. Therefore, I've created this pull request.

BTW, I ran the `npm run test` with all the passes. but `npm run test:esm` with failed "throws an error if commndDir() used in ESM mode" I think it will not be a test case in yargs@next

### DIFF
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -61,7 +61,7 @@ export class CommandInstance {
   ): void {
     opts = opts || {};
     this.requireCache.add(callerFile);
-    const fullDirPath = this.shim.path.join(
+    const fullDirPath = this.shim.path.resolve(
       this.shim.path.dirname(callerFile),
       dir
     );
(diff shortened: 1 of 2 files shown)


lib/command.ts
```diff
@@ -61,7 +61,7 @@ export class CommandInstance {
   ): void {
     opts = opts || {};
     this.requireCache.add(callerFile);
-    const fullDirPath = this.shim.path.join(
+    const fullDirPath = this.shim.path.resolve(
```
bcoe, line 64: Is there a way we could update a test in `test/command.mjs` to catch this, perhaps by passing in a resolved `callerFile`?

vipzhicheng: I have added a test that can catch this issue and pass after the fix.

bcoe (approved):

Issue #2467, opened 2025-04-29
Want command module to be extendable


vipzhicheng:
https://github.com/yargs/yargs/blob/888db19ccebcb5065a7aa415445e41cb15411c50/lib/command.ts#L104-L116

For now:

When I add middleware in `opts.visit` (Why? Because I can get command object)

If the command module have no key: middleares, it will throw an error: （BTW, if the command do have a key: middleware, no error happens）

```
command.middlewares = [middleware];
                                        ^
TypeError: Cannot add property middlewares, object is not extensible
```

I think maybe we can change these lines code to be follows to solve this issue

```diff
        const module = req(joined);
+        const extendableModule = { ...module };
-         const visited = visit(module, joined, file);
+        const visited = visit(extendableModule, joined, file);
        if (visited) {
          if (this.requireCache.has(joined)) continue;
          else this.requireCache.add(joined);
          // Infer command from directory structure if none is given:

-            if (!module.command) {
-            module.command = this.shim.path.basename(
+          if (!extendableModule.command) {
+            extendableModule.command = this.shim.path.basename(
              joined,
              this.shim.path.extname(joined)
            );
          }
-          this.addHandler(module);
+         this.addHandler(extendableModule);
```

Do you think this makes sense?

bcoe:
@vipzhicheng can you provide a longer example of code that fails with the existing API? I think I get the gist of what you're trying to achieve 😄 

Is it worth creating an object with a null prototype, since we're using this like a POJO I believe?

`Object.create(null,  Object.getOwnPropertyDescriptors({... module}));`

vipzhicheng:
OK, I've created a simple example repository to showcase this issue.

https://github.com/vipzhicheng/test-yargs-next

Please take a look.

shadowspawn:
I reproduced the problem, thanks for the repository. 

For interest, were you extending the command previously in `visit` with Yargs 17, or is this something new you were trying with Yargs 18?

vipzhicheng:
I made a Yargs wrapper project years ago. The basic idea is to provide a CLI tool to quickly define and execute commands.

It's just a open-source project but not famous.

https://semo.js.org/en/

The typical scenarios are:

1. Put every npm into REPL for a quick test 

```
semo repl -r lodash:_ -r dayjs:day
```

2. Encapsulate CLI app

```
semo run juejin
```

This is a CLI version of Chinese tech community. The default mode is a Twitter-like flow.

3. Convention

[...]

PR #2468, opened 2025-05-02
fix: allows ESM modules commands to be extensible using visit option

vipzhicheng:
refs to #2467

The test can catch the issue, and pass after fixing, and also pass my test repo.

### DIFF
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -102,18 +102,19 @@ export class CommandInstance {
       if (supportedExtension) {
         const joined = this.shim.path.join(fullDirPath, file);
         const module = req(joined);
-        const visited = visit(module, joined, file);
+        const extendableModule = {...module};
+        const visited = visit(extendableModule, joined, file);
         if (visited) {
           if (this.requireCache.has(joined)) continue;
           else this.requireCache.add(joined);
           // Infer command from directory structure if none is given:
-          if (!module.command) {
-            module.command = this.shim.path.basename(
+          if (!extendableModule.command) {
+            extendableModule.command = this.shim.path.basename(
               joined,
               this.shim.path.extname(joined)
             );
           }
-          this.addHandler(module);
+          this.addHandler(extendableModule);
         }
       }
     }
--- a/test/command.mjs
+++ b/test/command.mjs
@@ -2498,5 +2498,36 @@ describe('Command', () => {
           '  --version  Show version number  [boolean]',
         ]);
     });
+
+    it('allows ESM modules commands to be extensible using visit option', () => {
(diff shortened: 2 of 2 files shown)


lib/command.ts
```diff
@@ -102,18 +102,19 @@ export class CommandInstance {
       if (supportedExtension) {
         const joined = this.shim.path.join(fullDirPath, file);
         const module = req(joined);
-        const visited = visit(module, joined, file);
+        const extendableModule = {...module};
```
bcoe, line 105: nit, let's make this have a `null` prototype.

```
const extendableModule = Object.create(null,  Object.getOwnPropertyDescriptors({... module}));
```

vipzhicheng: > const extendableModule = Object.create(null,  Object.getOwnPropertyDescriptors({... module}));

This code will let the extendableModule lose its prototype, it causes another test to fail.

Test: it('supports a "visit" function option')

> commandObject.should.have.property('command')
> TypeError: Cannot read properties of undefined (reading 'have')

Because chai use should() to put should assertions into `Object.prototype`.

So what do you think?

shadowspawn: In the failing `'supports a "visit" function option'` test, you can replace the pattern `foo.should()` with `expect(foo).to`.

https://www.chaijs.com/guide/styles/

e.g.

```diff
+      commandObject.should.have.property('builder');
-      expect(commandObject).to.have.property('builder');

Comment (shadowspawn) [reply lib/command.ts]: Sorry, I got the diff example in the wrong order. Switch from `should` to `expect` for this test, which does not require mucking with prototypes.

```diff
-      commandObject.should.have.property('builder');
+      expect(commandObject).to.have.property('builder');
```

shadowspawn: `.should` and `expect()` support the same chaining language. `should` uses extended Object prototype, while `expect` is a plain function.

(My impression is that extending prototype was popular in the past, but is not the preferred approach in modern applications.)

shadowspawn: When we are only using an object as a collection of properties, we can get extra safety by using a null-prototype. This avoids collecting any extra properties either from JavaScript itself (like `.toString`) or added by other code (like `.should()`. 

In this case it might be safe without the null prototype, but Yargs has had problems in past leading to CVE from casual use of object without a null prototype. So paranoid, and easy to be robust.

vipzhicheng: Got it and done. Please take a look.

shadowspawn (approved):
LGTM. Thanks for the new test, and updating the old test.

bcoe:
@vipzhicheng thank you for the contribution.

PR #2471, opened 2025-06-19
fix(yargs-parser): support bun for getting cli name

### LINKED ISSUE #2377: Add support for Bun JavaScript Runtime
I have been experimenting with the new Bun JavaScript runtime and have found that yargs does not understand this new runtime. When referencing the $0 variable in yargs, it return "bun" instead of the script being run. As a quick hack, I modified index.cjs to include "bun" to the Regex statement:

`/\b(node|iojs|electron)(\.exe)?$/`

which becomes

`/\b(node|iojs|electron|bun)(\.exe)?$/`

This works and $0 now correctly reflects the script being executed. I would propose that a single line in yargs-factory.js be changed to the following so that index.cjs is updated with the new check:

`if (/\b(node|iojs|electron|bun)(\.exe)?$/.test(__classPrivateFieldGet(this, _YargsInstance_shim, "f").process.argv()[0])) {`

RiskyMH:
When using `bunx --bun package-name`, it claims bun is the cli name. However this incorrect, so this PR changes the behavior to work like node!

the --bun isn't strictly necessarily as if you don't have bun installed, then it force attempts to use bun runtime by changing the path.

(note, I have indeed tested this behavior and it works)

fixes #2377

### DIFF
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -1552,7 +1552,9 @@ export class YargsInstance {
     // ignore the node bin, specify this in your
     // bin file with #!/usr/bin/env node
     let default$0: string[];
-    if (/\b(node|iojs|electron|bun)(\.exe)?$/.test(this.#shim.process.argv()[0])) {
+    if (
+      /\b(node|iojs|electron|bun)(\.exe)?$/.test(this.#shim.process.argv()[0])
+    ) {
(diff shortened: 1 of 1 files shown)


shadowspawn:
I was initially a bit confused by the mention of `bunx`, but fix applies for `bun run index.mjs` too.

bcoe (approved):

PR #2476, opened 2025-07-04
Fix 'count' for de locale

timheiko:
Changing de localization for 'count' to 'Anzahl' because 'Zähler' is closer to 'counter'


shadowspawn:
A maintainer needs to do the merge. I have approved the request (so I think it is ok). In general, I give the primary maintainer the opportunity to check the PR and do the merge and decided when to release.

bcoe:

PR #2490, opened 2025-10-03
fix: suppress webpack warning by destructing import.meta before using…

### LINKED ISSUE #2489: import.meta.resolve gives compile-time warning when bundling with webpack
The line
https://github.com/yargs/yargs/blob/971e351705f0fbc5566c6ed1dfd707fa65e11c0d/lib/utils/apply-extends.ts#L21
introduced in https://github.com/yargs/yargs/pull/2451 gives the following warning
```
WARNING in ./node_modules/yargs/build/lib/utils/apply-extends.js 14:32-43
Critical dependency: Accessing import.meta directly is unsupported (only property access or destructuring is supported)
```
when compiled with webpack (latest version 5.102.0).

FYI1: The webpack warning message seems to be changed yesterday to
> 'import.meta' cannot be used as a standalone expression. For static analysis, its properties must be accessed directly (e.g., 'import.meta.url') or through destructuring.
https://github.com/webpack/webpack/pull/19961

FYI2: While
> webpack 5 supports import.meta
https://github.com/webpack/webpack/pull/11075

there is still an open issue on supporting import.meta.resolve https://github.com/webpack/webpack/issues/16693, and I'm not really sure what to expect to webpack.
Still, I think there is a great chance that there is some (ESM-compatible) way or another to avoid this warning by changing the implementation.

ej-toita:
… resolve

resolves: [2489](https://github.com/yargs/yargs/issues/2489)

### DIFF
--- a/lib/utils/apply-extends.ts
+++ b/lib/utils/apply-extends.ts
@@ -18,7 +18,8 @@ export function applyExtends(
     let pathToDefault: string | null = null;
     if (!isPath) {
       try {
-        pathToDefault = import.meta.resolve(config.extends);
+        const {resolve} = import.meta;
+        pathToDefault = resolve(config.extends);
       } catch (_err) {
         // maybe the module uses key for some other reason,
         // err on side of caution.


ej-toita:
Will be draft until https://github.com/yargs/yargs/issues/2489#issuecomment-3367223102 is addressed.

shadowspawn:
Per https://github.com/yargs/yargs/issues/2489#issuecomment-3367223102, I think this PR is suppressing the warning without addressing it.

PR #2493, opened 2025-11-07
Fix .wrap() docs

### LINKED ISSUE #2477: yargs 18: yargs.terminalWidth() is not a function.
[Documentation for `.wrap()`](https://yargs.js.org/docs/#api-reference-wrapcolumns) says to use `yargs.terminalWidth()`, however `TypeError: yargs_1.default.terminalWidth is not a function`.

```
$ node
Welcome to Node.js v22.15.0.
Type ".help" for more information.
> const y = await import('yargs');
undefined
> y.default.terminalWidth()
Uncaught TypeError: y.default.terminalWidth is not a function
> y.terminalWidth()
Uncaught TypeError: y.terminalWidth is not a function
> 
```

Solution: `yargs().terminalWidth()`

This is slightly less ergonomic for the chained API style, since you need to get your instance of yargs first:

```diff
- const argv = yargs(hideBin(process.argv))
+ const yargsInstance = yargs(hideBin(process.argv))
+ const argv = yargsInstance
    .strict()
    .showHelpOnFail(false, "Specify --help for available options")
    .epilogue(`...`)
-   .wrap(yargs.terminalWidth()).argv;
+   .wrap(yargsInstance.terminalWidth()).argv;
```

Not sure if this is just a docs fix, or if the `yargs.terminalWidth()` method should be restored.

(Also: the TypeScript types say that `import yargs from 'yargs'; yargs.terminalWidth()` is valid.)

benjie:
- Fixes #2477

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1783,5 +1783,5 @@ If the boolean argument `false` is provided, it will disable `--version`.
 Format usage output to wrap at `columns` many columns.
 
 By default wrap will be set to `Math.min(80, windowWidth)`. Use `.wrap(null)` to
-specify no column limit (no right-align). Use `.wrap(yargs.terminalWidth())` to
+specify no column limit (no right-align). Use `.wrap(yargs().terminalWidth())` to
 maximize the width of yargs' usage instructions.
--- a/docs/api.md
+++ b/docs/api.md
@@ -1783,5 +1783,6 @@ If the boolean argument `false` is provided, it will disable `--version`.
 Format usage output to wrap at `columns` many columns.
 
 By default wrap will be set to `Math.min(80, windowWidth)`. Use `.wrap(null)` to
-specify no column limit (no right-align). Use `.wrap(yargs().terminalWidth())` to
-maximize the width of yargs' usage instructions.
+specify no column limit (no right-align). Use `.wrap(y.terminalWidth())` (where
+`y` is your instance of `yargs()`) to maximize the width of yargs' usage
+instructions.


shadowspawn:
Creating a new yargs parser object to get the terminal width does not seem the right pattern to me.

(I am guessing you decided against using `yargsInstance` as the variable name. It isn't used elsewhere in this file so not an established pattern.)

`y` is used in this file in a few places where a variable is used for the top level parser:

```
const y = yargs(process.argv.slice(2));
y.version('1.0.0');
```

Following that compact pattern, how about: 

> Use `y.wrap(y.terminalWidth())` to maximize the width of yargs' usage instructions.

docs/api.md
```diff
@@ -1783,5 +1783,5 @@ If the boolean argument `false` is provided, it will disable `--version`.
 Format usage output to wrap at `columns` many columns.
 
 By default wrap will be set to `Math.min(80, windowWidth)`. Use `.wrap(null)` to
-specify no column limit (no right-align). Use `.wrap(yargs.terminalWidth())` to
+specify no column limit (no right-align). Use `.wrap(yargs().terminalWidth())` to
 maximize the width of yargs' usage instructions.
```
benjie:
```suggestion
specify no column limit (no right-align). Use `.wrap(y.terminalWidth())` (where
`y` is your instance of `yargs()`) to maximize the width of yargs' usage
instructions.
```

Are you thinking something like this @shadowspawn?

shadowspawn: Yes indeed.

shadowspawn (approved):
Looks good, thanks.

Issue #2282, opened 2022-12-23
`fish` shell completion support added in pr #2281 - please review whenever possible


jglanz:
Hi, absolutely love `yargs` - I found sometime to add support for `fish` shell & submitted a PR ( #2281 )

Please let me know if there are any issues in review; I am willing to support `fish` shell completions moving forward if that's helpful.

Happy Holidays 
-Jon

jglanz:
@bcoe aside from the dfish PR, I'd love to introduce a completion provider paradigm on the `Option` & `Positional` types for things like specific files or in my case, `targets` for the source tree being managed, etc.

Any thoughts?

shadowspawn:
Fish completion coming in next release: #2495 #2568

Issue #1904, opened 2021-04-06
Support fish shell completions.


ctwise:
Please add command completion support for fish shell as well as zsh and bash.

ionafan2:
That would be awesome!

shadowspawn:
See #2281 for an attempt.

shadowspawn:
Coming in next release: #2495 #2568

PR #2495, opened 2025-11-14
feat: support creating completions for fish shell

threddast:
Adds support for creating completions in fish shell with `<command> completion > ~/.config/fish/completions/<command>.fish` 

Ref #2282 and #1904

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -351,10 +351,10 @@ discussion of the advanced features exposed in the Command API.
 .completion([cmd], [description], [fn])
 ---------------------------------------
 
-Enable bash/zsh-completion shortcuts for commands and options.
+Enable bash/zsh/fish-completion shortcuts for commands and options.
 
-`cmd`: When present in `argv._`, will result in the `.bashrc` or `.zshrc` completion script
-being outputted.
+`cmd`: When present in `argv._`, will result in the `.bashrc`, `.zshrc`, or fish
+completion script being outputted.
 
 To enable bash/zsh completions, you can either: 
 1. Concat the generated script to your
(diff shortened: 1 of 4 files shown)


shadowspawn:
Related, previous PR: #2281

shadowspawn:
Looks tidy, thanks. I'll give it try before reviewing.

gigamonster256:
itd be great to have a method to set the shell to generate completions for - package managers often pre-generate completions at build time results in hacky workarounds to generate completions for shells other than the one currently running

something like `app completion fish` rather than replying on implicit passing of env vars to determine which to generate

shadowspawn:
I am still getting back to this! A note to self: there is a later PR open (#2506), but I will looking at this one first.

GoodbyeNJN:
Is there any progress on this PR? I'm looking forward to this feature.

lib/completion-templates.ts
```diff
@@ -56,3 +56,13 @@ else
 fi
 ###-end-{{app_name}}-completions-###
 `;
+
+export const completionFishTemplate = `###-begin-{{app_name}}-completions-###
+#
+# yargs command completion script
+#
+# Installation: {{app_path}} {{completion_command}} > $XDG_CONFIG_HOME/fish/completions/{{app_name}}.fish
```
shadowspawn, line 64: I do not have this defined, and took me a while to work out why was not working when I just wrote it to home folder. (I did not know is was `~/.config`.) I think simplest would be to leave out `XDG_CONFIG_HOME`. People using those will know about them?

docs/api.md
```diff
@@ -374,6 +374,11 @@ To enable bash/zsh completions, you can either:
 
    e.g. `./command completion > /usr/local/share/zsh/site-functions/_command_yargs_completions`
 
+For Fish, write it to a file in your fish completions directory (`$XDG_CONFIG_HOME/fish/completions`),
```
shadowspawn, line 377: This is presumably somewhat uncommon, and the example does not include it. I suggest using the default location.

shadowspawn:
Working on landing this in #2568

shadowspawn:
Merged in #2568

Will be available in next release.

Thanks for your contribution, sorry about the long delay.

PR #2498, opened 2025-11-23
fix: prototype pollution vulnerability in apply-extends

### LINKED ISSUE #2497: Vulnerability Report: Prototype Pollution in `apply-extends`
## Summary
A prototype pollution vulnerability was identified in the `apply-extends` helper function within `lib/utils/apply-extends.ts`. This function is responsible for merging configuration objects, including those loaded via the `extends` keyword. The vulnerability allowed an attacker to inject properties into `Object.prototype` by including a `__proto__` key in a configuration file.

### Vulnerability Location
The vulnerability existed in the `mergeDeep` function within `lib/utils/apply-extends.ts`. This function recursively merged two objects without checking if the key being merged was `__proto__`.

### Impact
Prototype pollution can lead to:
- **Denial of Service (DoS)**: By overwriting methods like `toString` or `valueOf`.
- **Logic Errors**: By injecting unexpected properties that alter application flow.
- **Remote Code Execution (RCE)**: In some specific scenarios where polluted properties are used in sensitive operations (e.g., template engines, child process execution).

### Reproduction
The following script demonstrated the vulnerability:

```javascript
// repro_proto.js
import { applyExtends } from './build/lib/utils/apply-extends.js';
// ... setup ...

const config = JSON.parse('{"__proto__":{"polluted":true}}');
applyExtends(config, ...);

console.log(config.polluted); // true (before fix)
```

niklabh:
fixes #2497 

The vulnerability existed in the `mergeDeep` function within `lib/utils/apply-extends.ts`. This function recursively merged two objects without checking if the key being merged was `__proto__`.

### Impact
Prototype pollution can lead to:
- **Denial of Service (DoS)**: By overwriting methods like `toString` or `valueOf`.
- **Logic Errors**: By injecting unexpected properties that alter application flow.
- **Remote Code Execution (RCE)**: In some specific scenarios where polluted properties are used in sensitive operations (e.g., template engines, child process execution).

### Reproduction
The following script demonstrated the vulnerability:

```javascript
// repro_proto.js
import { applyExtends } from './build/lib/utils/apply-extends.js';
// ... setup ...

const config = JSON.parse('{"__proto__":{"polluted":true}}');
applyExtends(config, ...);

console.log(config.polluted); // true (before fix)
```

The fix involves explicitly checking for and skipping the `__proto__` key during the iteration of object keys in the `mergeDeep` function.

BEGIN_COMMIT_OVERRIDE
fix: local prototype pollution vulnerability in apply-extends
END_COMMIT_OVERRIDE

### DIFF
--- a/test/helpers.mjs
+++ b/test/helpers.mjs
@@ -15,4 +15,11 @@ describe('applyExtends', () => {
     expect(conf.batman).to.equal('Caped Crusader');
     expect(conf.version).to.equal('1.0.2');
   });
+
+  it('does not allow prototype pollution via __proto__', () => {
+    const polluted = JSON.parse('{"__proto__": {"polluted": true}}');
+    const conf = applyExtends(polluted, process.cwd(), true);
+    expect(conf.polluted).to.equal(undefined);
+    expect({}.polluted).to.equal(undefined);
+  });
 });
--- a/lib/utils/apply-extends.ts
+++ b/lib/utils/apply-extends.ts
@@ -68,6 +68,7 @@ function mergeDeep(config1: Dictionary, config2: Dictionary) {
   }
   Object.assign(target, config1);
   for (const key of Object.keys(config2)) {
+    if (key === '__proto__') continue;
     if (isObject(config2[key]) && isObject(target[key])) {
       target[key] = mergeDeep(config1[key], config2[key]);
     } else {


shadowspawn:
Are you willing to write a test for the issue?

niklabh:
> Are you willing to write a test for the issue?

Done

test/helpers.mjs
```diff
@@ -15,4 +15,11 @@ describe('applyExtends', () => {
     expect(conf.batman).to.equal('Caped Crusader');
     expect(conf.version).to.equal('1.0.2');
   });
+
+  it('does not allow prototype pollution via __proto__', () => {
+    const polluted = JSON.parse('{"__proto__": {"polluted": true}}');
+    const conf = applyExtends(polluted, process.cwd(), true);
+    expect(conf.polluted).to.equal(undefined);
```
shadowspawn, line 22: This expect fails (without the code fix).

test/helpers.mjs
```diff
@@ -15,4 +15,11 @@ describe('applyExtends', () => {
...
+    expect({}.polluted).to.equal(undefined);
```
shadowspawn, line 23: This expect is not reached, but commenting out the above it does _not_ fail without the code fix. I think the way the deep merge is processed means the global object prototype is not polluted.

test/helpers.mjs
shadowspawn: I added a bunch of examples to #2497, I'll resolve this brief comment.

test/helpers.mjs
shadowspawn: I added a bunch of examples to #2497, I'll resolve this brief comment.

shadowspawn:
For comparison, `yargs-parser` sanitises the name of the key:

https://github.com/yargs/yargs-parser/blob/60f2db5dacde1e94f960ce0941a717f3f422f865/lib/yargs-parser.ts#L1110-L1113

niklabh:
Any updates on this? Recent react vulnerability was also due to prototype pollution. Better have this check in the code.

Or something else needs to be done?

shadowspawn:
After the initial rush to determine if the publicly disclosed issue was a global prototype pollution, and finding it was apparently not, this dropped in priority.

I think the patch is simple and effective. 

In general I would rather use the same approach across the code and there is prior art with renaming `__proto__` to `___proto___` as mentioned in https://github.com/yargs/yargs/pull/2498#issuecomment-3579449810 but that would complicate the code for negligible gain in a rare and probably hostile case.

Good enough.

shadowspawn (approved):
LGTM

shadowspawn:
Would you please correct the reproduction description in this PR and in the issue to show the actual pollution that occurs, which is on the returned config object and not on the global prototype.

```
console.log({}.polluted); // true (before fix)
```

shadowspawn:
Thanks for the PR.

shadowspawn:
BEGIN_COMMIT_OVERRIDE
fix: local prototype pollution vulnerability in apply-extends
END_COMMIT_OVERRIDE

PR #2499, opened 2025-11-27
we don't have en_US.json

Stasshe:
changed it into en.json

### DIFF
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -1594,7 +1594,7 @@ export class YargsInstance {
       this.#shim.getEnv('LC_MESSAGES') ||
       this.#shim.getEnv('LANG') ||
       this.#shim.getEnv('LANGUAGE') ||
-      'en_US';
+      'en';
     this.locale(locale.replace(/[.:].*/, ''));
   }
   [kGuessVersion](): string {


shadowspawn:
This is the default locale. If yargs _did_ have an `en_us.json` it would get used in preference to the `en.json` localisation file. Since there is not, I think the locale matching falls back to `en`.

Unless this causes a fault of some sort, I am tempted to leave it as is. Probably working as intended.

Stasshe:
@shadowspawn 
Thanks you for your reply. 

It seemed that en_us.json file is not in locales dir, so final fallback logic will not work well. ( In my env, when it is en_us.json, an error occurred due to this, and when i changed it into en.json, it worked)


Maybe, you might be better to change this, but it is so trivial matter, and if you intend this, there is no problem.

shadowspawn:
Oh, I was not seeing an error in my testing. I will try some more scenarios. Thanks for info.

shadowspawn:
I still can not reproduce an error, but I am interested in how it is broken for you. `en_US` appears in multiple places in the code, so I don't want to change just one usage without understanding more.

What is your environment? What platform are you on? What error are you seeing? 

I tracked down why I think it is working for me. I am running it on command-line with node on macOS. `y18n` has a fallback to remove the region and use just the language. Presumably your setup is both looking for a locale file but also not falling back to checking the language.

https://github.com/yargs/y18n/blob/ad1a31b730a5034218b546e8542bb691b84992b2/lib/index.ts#L204-L212

Stasshe:
Certainly a special environment.

```typescript
process: {
        env: {
          LANG: 'en',

```

When this global object was registered at runtime, it worked fine. However, when the key field LANG: 'en',
was removed, it caused an error.


Presumably your setup is both looking for a locale file but also not falling back to checking the language.

That's right, _resolveLocaleFile is not being executed.

The reason for this is because it is running in js (browser).
It is a nodejs emulator. Currently, it is already set up to supplement enb data, LOCALE, etc., so it is difficult to reproduce the current situation.

Sorry, I should have explained the situation a little better from the start.
I thought it was just a mistake.
If the design relies on the fallback method of _resolveLocaleFile,
 then the point was wrong. Please close the PR.


By the way, this was the error log.
Error: ENOENT: no such file or directory, open 'en_US.json'


Thank you for your sincere response.

shadowspawn:
Ah, thanks for the details. Defining an environment variable for the context seems a good work-around.

For your interest, Yargs has "shims" to wrap some of the node specific behaviour. Your environment is likely to be hitting the browser shim and not getting the y18n support used when running in node context.

- https://github.com/yargs/yargs/tree/main/lib/platform-shims

There is another `en_US` in there:

- https://github.com/yargs/yargs/blob/971e351705f0fbc5566c6ed1dfd707fa65e11c0d/lib/platform-shims/browser.mjs#L85

PR #2507, opened 2026-02-24
Replace EOL node 23 in CI version matrix with node 24

shadowspawn:
CI tests broken on node 23 with engine dependency changes in typescript-eslint related to supporting eslint 10. Node 23 is EOL, and typescript-eslint does not explicitly support odd node versions,  so time to drop it from matrix and replace with 24.

PR with broken tests: #2505

Explanation: https://github.com/typescript-eslint/typescript-eslint/issues/12081#issuecomment-3950141924

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,7 +3,7 @@ on:
     branches:
       - main
   pull_request:
-    types: [ assigned, opened, synchronize, reopened, labeled ]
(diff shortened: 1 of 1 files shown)


shadowspawn:
Ah yes, tests _also_ broken by prettier changes (?). I'll do a PR for that too... Fixed by `npx gts fix`.

shadowspawn:
The test run failed due to #2511, can see tests passing with both changes in #2512.

PR #2511, opened 2026-02-26
Reapply formatting to fix lint checks

shadowspawn:
Lint tests failing on new PRs due to formatting changes. This PR has the changes from simply running:

```console
npm run fix
```

PR #2512, opened 2026-02-26
Combine test fixes

shadowspawn:
The two PRs to fix the tests still fail tests because of the _other_ problem. This dummy PR is just to see the tests pass...

Combines: #2507 #2511

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,7 +3,7 @@ on:
     branches:
       - main
   pull_request:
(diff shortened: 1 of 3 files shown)


shadowspawn:
Perhaps I should have taken a screenshot! 😢 

I closed this PR after the tests passed, and then could not see the tests. I reopened the PR and could see the passing tests. But reopening the reopen triggered a rerun, and now they fail with some git error...

I'll do a dummy commit, but then give up. 🙊

shadowspawn:
PassingTests

Issue #2356, opened 2023-09-06
show-hidden broken if parser has strip-dashed config


YarnSaw:
Minimal Repro:
```js
const yargs = require('yargs')

yargs
  .help('help', 'help')
  .showHidden('show-hidden', 'Show hidden options')
  .options({
    foo: {
      describe: 'bar',
      type: 'string',
      hidden: true,
    }})
  .parserConfiguration({
    'strip-dashed': true,
  })
  .argv;
```
Outputs:
```
$ node test.js --help --show-hidden
Options:
  --version      Show version number                                   [boolean]
  --help         help                                                  [boolean]
  --show-hidden  Show hidden options                                   [boolean]
```

Where the `foo` option isn't shown, despite --help and --show-hidden being specified. If `'strip-dashed': true,` is commented out, the new output is:
```
$ node test.js --help --show-hidden
Options:
  --version      Show version number                                   [boolean]
  --help         help                                                  [boolean]
  --show-hidden  Show hidden options                                   [boolean]
  --foo          bar                                                    [string]
```

Which is properly displaying the `foo` option.

shadowspawn:
Thanks for report. I reproduced, and think it is a bug.

A work-around would be to use an option name without a dash with `showHidden`.  (Which you likely realised, having helpfully tracked down the dependence on `'strip-dashed'`.)

This is one of several bugs introduced by allowing the author to call `parserConfiguration`. Sigh. Flexibility leads to complications! Yargs needs to do more work to find whether a "known" option is present in the parsed args. The (overly) simple check is:

https://github.com/yargs/yargs/blob/e517318cea0087b813f5de414b3cdec7b70efe33/lib/usage.ts#L615

One of the previous fixes for a situation where Yargs wanted to iterate over all the possible permutations was:
https://github.com/yargs/yargs/pull/2308

This case is simpler, just want to lookup option by any name. That could be done by taking into account the parserConfiguration, or could be done by iterating over the aliases. Given this has come up a few times I would like to add a utility routine to wrap the logic.

YarnSaw:
My current solution was to remove the `strip-dashed` since I didn't like the aesthetic of `showHidden` being camel case with the rest of the options being kebab case :sweat_smile:

moitila:
https://github.com/yargs/yargs/pull/2516

This PR I resolve this BUG

PR #2516, opened 2026-03-09
fix: support show-hidden/help lookup when strip-dashed is enabled

moitila:
## Summary

Fix `show-hidden` handling when `parserConfiguration({ 'strip-dashed': true })` is enabled by centralizing option presence lookup and reusing it in the approved call sites.

## Changes

- add a shared option presence helper
- update `lib/usage.ts` to use the shared lookup for hidden option gating
- update `lib/validation.ts` to reuse the same lookup for dashed option conflict handling
- add regression coverage for:
  - `--show-hidden` with `strip-dashed`
  - custom `show-hidden` name with `strip-dashed`
  - alias for custom `show-hidden`
  - dashed conflicts with `strip-dashed`

## Validation

- targeted TypeScript compile passed
- targeted `usage` tests passed
- targeted `validation` conflict tests passed
- manual checks confirmed hidden options and dashed conflicts behave correctly under `strip-dashed`

## Scope

This PR keeps the change limited to:

- `lib/utils/option-presence.ts`
- `lib/usage.ts`
- `lib/validation.ts`
- `test/usage.mjs`

No scope expansion to `lib/yargs-factory.ts`.

ISSUE
https://github.com/yargs/yargs/issues/2356

### DIFF
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -791,13 +791,9 @@ export function command(
   return new CommandInstance(usage, validation, globalMiddleware, shim);
 }
 
-export interface CommandHandlerDefinition
-  extends Partial<
-    Pick<
-      CommandHandler,
-      'deprecated' | 'description' | 'handler' | 'middlewares'
-    >
-  > {
+export interface CommandHandlerDefinition extends Partial<
+  Pick<CommandHandler, 'deprecated' | 'description' | 'handler' | 'middlewares'>
+> {
   aliases?: string[];
   builder?: CommandBuilder | CommandBuilderDefinition;
   command?: string | string[];
--- a/lib/usage.ts
+++ b/lib/usage.ts
@@ -6,6 +6,7 @@ import {YargsInstance} from './yargs-factory.js';
 import {YError} from './yerror.js';
 import {DetailedArguments} from './typings/yargs-parser-types.js';
 import setBlocking from './utils/set-blocking.js';
+import {optionIsPresent} from './utils/option-presence.js';
 
 function isBoolean(fail: FailureFunction | boolean): fail is boolean {
   return typeof fail === 'boolean';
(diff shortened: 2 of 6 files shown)


shadowspawn:

shadowspawn:
As noted in comment https://github.com/yargs/yargs/issues/2356#issuecomment-1709610310

exposing the `parserConfiguration` to pass into yargs-parser has made things harder for yargs, because the key shape in parsed argv now depends on the configuration.

> This case is simpler, just want to lookup option by any name. That could be done by taking into account the `parserConfiguration`, or could be done by iterating over the aliases. Given this has come up a few times I would like to add a utility routine to wrap the logic.

I would like to add a utility routine to handle the key lookup since this issue has come up multiple times (and there are possibly more lurking in the code.)

moitila:
> To be clear: is this issue entirely AI/LLM generated?

Not entirely.

I used an internal engineering workflow tool I’m building to structure the investigation and testing, but the analysis and implementation were done manually after understanding the issue.

The tool basically helps me generate investigation notes, hypotheses and validation steps so I can reason more systematically about bugs.

In this case I reproduced the issue, traced the lookup logic in `usage.ts`, and implemented the fix and tests accordingly.

moitila:
> As noted in comment [#2356 (comment)](https://github.com/yargs/yargs/issues/2356#issuecomment-1709610310)
> 
> exposing the `parserConfiguration` to pass into yargs-parser has made things harder for yargs, because the key shape in parsed argv now depends on the configuration.
> 
> > This case is simpler, just want to lookup option by any name. That could be done by taking into account the `parserConfiguration`, or could be done by iterating over the aliases. Given this has come up a few times I would like to add a utility routine to wrap the logic.
> 
> I would like to add a utility routine to handle the key lookup since this issue has come up multiple times (and there are possibly more lurking in the code.)

Good point.

My PR fixes the immediate issue in filterHiddenOptions, but I agree the lookup logic likely appears in multiple places.

I'll investigate whether this should be implemented as a small utility to resolve option keys (considering parserConfiguration, camelCase and aliases) and update the PR accordingly.

moitila:
I also noticed that `npm run check` still reports pre-existing lint/formatting issues outside the scope of this PR.

If helpful, I can follow up with a separate cleanup PR to address those remaining 19 `prettier` errors without mixing that unrelated formatting work into this fix.

shadowspawn:
Lint fix PR is open: #2511

moitila:
Can i make anothers issue? Recomend?

shadowspawn:
The changes look too heavy-weight just for this fix, but interesting to see you identified a couple of locations where key lookup is a problem with configuration options. 

> I'll investigate whether this should be implemented as a small utility to resolve option keys (considering parserConfiguration, camelCase and aliases) and update the PR accordingly.

I am interested in this. I used a different approach in the past, following how yargs-parser itself looks for keys by considering all aliases without considering the configuration as such.

shadowspawn:
This PR was a bit large for an edge case (`show-hidden` _and_ `strip-dashed`).

This might get saner when a robust lookup is available.

Thank you for your contributions.

PR #2520, opened 2026-03-16
fix(command): show deprecated label in sub-command help

### LINKED ISSUE #2248: Deprecated command flag is not mentioned in the command help
Having a code like:
```js
const options = yargs
  .command({command: 'cmd', description: 'Do command', deprecated: true})
  .help().strict(true).parse();
```

### Actual
Then the command help does not mentioned in the command help:
```
$ node main.js cmd --help

Do command

Options: ...
```

### Expected
I would expect to see something like:
```
$ node main.js cmd --help

Do command [deprecated]

Options: ...
```

terminalchai:
## Summary

This fixes `command --help` output for deprecated commands.

`yargs` already shows `[deprecated]` / `[deprecated: ...]` in the top-level `Commands:` list, but nested command help used `commandHandler.description` directly when populating the usage header, so the deprecated label was dropped there.

With this change, sub-command help now appends the same deprecated suffix to the header description.

Fixes #2248.

## Changes

- add `maybeAddDeprecatedDescription()` in [lib/command.ts](lib/command.ts)
- use it when nested command help calls `.usage(..., description)`
- add regression tests for boolean and string deprecated values in [test/command.mjs](test/command.mjs)

## Testing

Followed the repo guide by adding unit tests and running the relevant test flow:

- `node node_modules/mocha/bin/mocha.js --enable-source-maps ./test/*.mjs --require ./test/before.mjs --timeout=24000 --check-leaks --grep "deprecated command"`

That targeted suite passes locally.

I also ran `npm test -- --grep "deprecated command"`, which runs the repo's normal pretest/test/posttest pipeline. The new tests pass there as well, but the final repo-wide `check` step currently reports unrelated existing Prettier errors in [lib/yargs-factory.ts](lib/yargs-factory.ts) on my local setup, so I did not include any unrelated formatting changes in this PR.


shadowspawn:

shadowspawn:
It did look like a human started work on this, so not closing despite lack of response: https://github.com/yargs/yargs/issues/2248#issuecomment-4070410830

terminalchai:
The implementation and tests were created with substantial AI assistance. I used the agent to investigate the code path, make the changes, and run the tests documented in the PR. I reviewed the resulting diff and test output before submitting it. I should have disclosed that when you first asked, and I understand if this does not meet the project's contribution policy.

PR #2523, opened 2026-03-25
fix: sanitize scriptName spaces in completion script function names

Luc0-0:
## Problem

When `scriptName` contains spaces (e.g. `.scriptName('my tool')`), the generated bash/zsh completion script uses the raw name as a shell function identifier, producing invalid syntax:

```bash
_my tool_yargs_completions()   # invalid — spaces not allowed in function names
```

This breaks `source`ing the completion script entirely.

Reported in #2387.

## Solution

Introduce a separate `{{app_function_name}}` template token that sanitizes the name for use as a shell identifier by replacing any character outside `[a-zA-Z0-9_]` with an underscore.

The existing `{{app_name}}` token is preserved unchanged for places where the literal command string is needed — the `complete` builtin's command-match argument and `compdef`'s command argument — so tab completion still triggers on the actual invocation (e.g. `my tool <TAB>`).

**Before:**
```bash
_my tool_yargs_completions()             # syntax error
complete ... -F _my tool_yargs_completions my tool
```

**After:**
```bash
_my_tool_yargs_completions()             # valid identifier
complete ... -F _my_tool_yargs_completions my tool
```

## Changes

- `lib/completion-templates.ts` — replace `_{{app_name}}_yargs_completions` with `_{{app_function_name}}_yargs_completions` in both bash and zsh templates
- `lib/completion.ts` — compute `functionName` from `name` via `replace(/[^a-zA-Z0-9_]/g, '_')` and substitute `{{app_function_name}}` before `{{app_name}}`
- `test/completion.mjs` — add test cases for bash and zsh completion scripts with a space in `scriptName`

### DIFF
--- a/lib/completion-templates.ts
+++ b/lib/completion-templates.ts
@@ -5,7 +5,7 @@ export const completionShTemplate = `###-begin-{{app_name}}-completions-###
 # Installation: {{app_path}} {{completion_command}} >> ~/.bashrc
 #    or {{app_path}} {{completion_command}} >> ~/.bash_profile on OSX.
 #
-_{{app_name}}_yargs_completions()
+_{{app_function_name}}_yargs_completions()
 {
     local cur_word args type_list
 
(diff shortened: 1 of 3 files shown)


shadowspawn:
As I said in https://github.com/yargs/yargs/issues/2387#issuecomment-1888537399

> I agree it is a short-coming, but under what circumstances would it come up in practice? A script with a space in the name seems unlikely. However, if you actually encountered this bug in real world then definitely of interest!

shadowspawn:
We might merge the fix for theoretical correctness, but much more interested if it makes a practical difference.

PR #2524, opened 2026-03-28
Fix positional arguments not populated when halt-at-non-option is set

### LINKED ISSUE #2423: `halt-at-non-option` is not working as expected.
I'm trying to write a CLI similar to `ssh`. Here's the minimal example:
```typescript
function main() {
  yargs(process.argv.slice(2))
    .usage(
      "$0 <host> [<cmd>...]",
      "Run command via ssh",
      (yargs) =>
        yargs
          .parserConfiguration({
            "halt-at-non-option": true,
          })
          .positional("host", { type: "string", demandOption: true })
          .positional("cmd", { array: true })
          .options({
            verbose: { type: "boolean", alias: "v", count: true },
          }),
      (args) => {
        console.log(args);
      },
    )
    .strict()
    .parseSync();
}
```

which just adds one option.  When parsing `node my-script.js  my-host ls`, I get this error:
```
```

ordinary9843:
When `halt-at-non-option` is configured, `yargs-parser` places all positional arguments into `argv['--']` instead of `argv._`. The `populatePositionals` function in `command.ts` reads only `argv._.length` for its required-argument count check and uses `argv._` exclusively when mapping positionals to their named keys. This means that even when the correct number of positionals is provided, the validation reports "Not enough non-option arguments: got 0, need at least 1" and no positional values are populated.

This fix detects when `halt-at-non-option` is active and `argv['--']` has content, then merges those entries into `argv._` before validation and mapping run. Both the count check and the positional population now see the correct arguments.

Fixes #2423

### DIFF
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -548,6 +548,17 @@ export class CommandInstance {
     yargs: YargsInstance
   ) {
     argv._ = argv._.slice(context.commands.length); // nuke the current commands
+    // When halt-at-non-option is configured, yargs-parser places positional
+    // arguments in argv['--'] instead of argv._. Move them into argv._ so
+    // positional validation and mapping work correctly.
+    if (
+      yargs.getOptions().configuration['halt-at-non-option'] &&
+      argv['--'] &&
+      argv['--'].length
+    ) {
+      argv._ = argv._.concat(argv['--'] as string[]);
+      argv['--'] = [];
+    }
     const demanded = commandHandler.demanded.slice(0);
     const optional = commandHandler.optional.slice(0);
     const positionalMap: Dictionary<string[]> = {};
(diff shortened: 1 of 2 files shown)


lib/command.ts
```diff
@@ -548,6 +548,17 @@ export class CommandInstance {
     yargs: YargsInstance
   ) {
     argv._ = argv._.slice(context.commands.length); // nuke the current commands
+    // When halt-at-non-option is configured, yargs-parser places positional
+    // arguments in argv['--'] instead of argv._. Move them into argv._ so
+    // positional validation and mapping work correctly.
```
shadowspawn, line 553: It turns out the `--` storage is on purpose. 

https://github.com/yargs/yargs/issues/2423#issuecomment-4197483993

I fear simply moving the arguments between properties will interact with other behaviours and combinations of options. In particular if caller has set `"populate--": true`. 

Aside: the ongoing cost we pay for letting yargs users configure yargs-parser!

PR #2532, opened 2026-04-26
Normalize quoted string positionals

### LINKED ISSUE #2416: Quotes are stripped from options but not positional arguments
```javascript
let r;
r = require("yargs")().command("echo [pos]").parse(`echo "a b" --opt "a b"`);
console.log(r.opt); // 'a b'
console.log(r.pos); // '"a b"'

// behavior originates here, where it makes more sense
r = require("yargs-parser")(`echo "a b" --opt "a b"`);
console.log(r); // { _: [ 'echo', '"a b"' ], opt: 'a b' }
```
Related to https://github.com/yargs/yargs-parser/issues/180. 

Specifically seems weird to me that whether an option is positional or not changes the quote stripping behavior.

Seems like there may be a bunch of history here so feel free to just close if this has already been talked to death: https://github.com/yargs/yargs-parser/issues/300

ShipItAndPray:
## Summary
- strip matching shell-style quotes from positional `_` entries when the original parse input is a single string
- leave array input untouched so literal quote characters are preserved there
- add command regression tests for both behaviors

Closes #2416.

## Validation
- reproduced `pos === ""a b""` with `yargs().command("echo [pos]").parse("echo "a b" --opt "a b"")` before the change
- `npm test -- --grep "strips shell-style quotes from string positionals|preserves literal quotes for array positionals"` passes its mocha assertions before hitting the repo's existing posttest/lint debt
- direct repro now yields `pos === "a b"` for string input and preserves `"a b"` for array input

### DIFF
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -62,6 +62,19 @@ import {isPromise} from './utils/is-promise.js';
 import {maybeAsyncResult} from './utils/maybe-async-result.js';
 import setBlocking from './utils/set-blocking.js';
 
+function stripMatchingQuotes(value: string | number): string | number {
+  if (
+    typeof value === 'string' &&
+    value.length >= 2 &&
+    (value[0] === '"' || value[0] === "'") &&
+    value[0] === value[value.length - 1]
+  ) {
+    return value.slice(1, -1);
+  }
+
+  return value;
+}
+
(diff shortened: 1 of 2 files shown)


shadowspawn:

shadowspawn:
No engagement since being opened. May have been AI submitted with no follow-through.

PR #2533, opened 2026-04-27
Ignore defaulted version collisions

### LINKED ISSUE #2199: yargs.option("version") is admitted, but fails silently
Using something like:

```js
yargs.option("version", {
                type: "string",
                default: "0.0.1",
            });
```

ShipItAndPray:
## Summary
- stop treating parser-defaulted `version` values as if the user explicitly requested the built-in version output
- keep the reserved-word warning in place for colliding `version` option/positional names
- add regressions for both a defaulted `option("version")` and a defaulted `install [version]` positional

Closes #2199.

## Validation
- reproduced both `option("version", { default: ... })` and `command("install [version]")` printing the built-in version before the change
- `npm test -- --grep "defaulted version option|defaulted version positional"` passes the mocha assertions before the repo's pre-existing posttest/lint failures in unrelated files
- direct repro now leaves the built-in version silent while still emitting the reserved-word warning

### DIFF
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -2010,7 +2010,11 @@ export class YargsInstance {
     Object.keys(argv).forEach(key => {
       if (key === this.#helpOpt && argv[key]) {
         helpOptSet = true;
-      } else if (key === this.#versionOpt && argv[key]) {
+      } else if (
+        key === this.#versionOpt &&
+        argv[key] &&
+        !parsed.defaulted[key]
+      ) {
         versionOptSet = true;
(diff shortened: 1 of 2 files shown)


shadowspawn:
I thought there was a timing issue involved too. I don't see any changes related to that, was it not needed or not checked?

https://github.com/yargs/yargs/issues/2199#issuecomment-1483520716

shadowspawn:

shadowspawn:
No engagement since being opened. May have been AI submitted with no follow-through.

PR #2539, opened 2026-05-31
Ignore defaulted version values for version output

### LINKED ISSUE #2199: yargs.option("version") is admitted, but fails silently

yzhkali:
## Summary
- avoid treating parser-defaulted `version` values as an explicit `--version` request
- keep explicit version handling unchanged
- add regression coverage for defaulted option and positional `version` values

Closes #2199

## Tests
- `npm test -- --grep version`
- `npm run check:js`
- `git diff --check`

Note: full `npm test` runs the Mocha suite successfully, but the posttest `gts lint` step currently reports existing Prettier issues on upstream main outside this change.

### DIFF
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -939,9 +939,18 @@ export class YargsInstance {
 
       this[kTrackManuallySetKeys](key);
 
+      const aliases = Array.isArray(opt.alias)
+        ? opt.alias
+        : opt.alias
+          ? [opt.alias]
+          : [];
+
       // Warn about version name collision
       // Addresses: https://github.com/yargs/yargs/issues/1979
-      if (this.#versionOpt && (key === 'version' || opt?.alias === 'version')) {
+      if (
+        this.#versionOpt &&
+        (key === 'version' || aliases.includes('version'))
+      ) {
         this[kEmitWarning](
           [
             '"version" is a reserved word.',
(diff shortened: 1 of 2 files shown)


shadowspawn:

yzhkali:
No, it was not entirely AI/LLM-generated. I used Codex to help with parts of the implementation, but there was human involvement in reviewing the code, deciding the scope, and checking the tests before submitting the PR. Happy to make any adjustments needed to fit the project policy.

shadowspawn:
Won't the built-in version handling still be triggered if the clashing option or positional is supplied on the command-line? This looks like it just suppressed the clash for a default version value.

I think the underlying problem is the warning about the clash is lost.

yzhkali:
Thanks, you were right that the previous patch only covered the defaulted `version` path.

I updated the PR to handle the explicit clash paths too. When an app defines `version` as an option, alias, or positional while the default built-in version option is enabled, yargs now still emits the reserved-word warning, but removes the default built-in version parser/usage handling so the user-defined value is parsed normally instead of triggering package version output.

I also added regression coverage for:

- explicit `--version 0.25.10` with `option("version", { type: "string" })`
- explicit `--version 0.25.10` through an alias
- explicit `install 0.25.10` through a positional `version`
- explicit boolean `--version` as a user option
- defaulted option/positional `version` values not displaying the built-in version

Verified locally with:

`npm test -- --grep "when an option or alias \\\"version\\\" is set"`

Result: 7 tests passed, including the posttest lint/check step. `git diff --check` also passes.

yzhkali:
Closing this iteration because the intended reserved-word clash semantics are still unresolved. The latest patch expanded from the default-value case to explicit option, alias, and positional paths, and that scope needs maintainer agreement before further work. The recorded checks pass and issue #2199 remains open; a smaller follow-up can be based on confirmed behavior.

PR #2542, opened 2026-06-04
fix: complete negated boolean options by default

### LINKED ISSUE #2254: bash completion: negative boolean options does not auto complete

liuzemei:
Complete `--no-*` forms for boolean options whenever boolean negation is enabled, matching the parser default. This also removes the previous requirement that the option have a truthy default before its negated form appears in completion results.

Default `--help` and `--version` completions remain unchanged, so completion does not suggest `--no-help` or `--no-version`.

fixes #2254

Tests:
- `PATH=/Users/neo/.nvm/versions/node/v24.8.0/bin:$PATH npm run compile -- -p tsconfig.test.json && PATH=/Users/neo/.nvm/versions/node/v24.8.0/bin:$PATH npx mocha --enable-source-maps ./test/*.mjs --require ./test/before.mjs --timeout=24000 --check-leaks`
- `PATH=/Users/neo/.nvm/versions/node/v24.8.0/bin:$PATH npx gts lint lib/completion.ts`
- `PATH=/Users/neo/.nvm/versions/node/v24.8.0/bin:$PATH npx eslint test/completion.mjs`

### DIFF
--- a/lib/completion.ts
+++ b/lib/completion.ts
@@ -115,13 +115,17 @@ export class Completion implements CompletionInstance {
       !this.previousArgHasChoices(args)
     ) {
       const options = this.yargs.getOptions();
+      const descriptions = this.usage.getDescriptions();
       const positionalKeys =
         this.yargs.getGroups()[this.usage.getPositionalGroupName()] || [];
 
       Object.keys(options.key).forEach(key => {
         const negable =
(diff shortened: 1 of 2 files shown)


shadowspawn:

PR #2543, opened 2026-06-09
fix: handle halted args for command positionals

### LINKED ISSUE #2423: `halt-at-non-option` is not working as expected.

pupuking723:
Fixes #2423.

When `halt-at-non-option` is enabled, yargs-parser stores the halted positional arguments under `--`. Command positional population currently validates and consumes only `argv._`, so default commands such as `$0 <host> [cmd...]` can fail with “got 0” even though the requested positionals were parsed under `--`.

This updates command positional population to use the halted `--` arguments when `halt-at-non-option` left `argv._` empty, then clears `--` after those values are consumed as command positionals. Existing explicit `--` / `populate--` behavior remains covered by the existing command tests.

The small type-definition formatting changes are from the repo's current Prettier version during `npm test` / `gts lint`.

Verification:

- `npx mocha --enable-source-maps ./test/command.mjs --require ./test/before.mjs --grep "halt-at-non-option arguments" --timeout=24000 --check-leaks`
- `npx mocha --enable-source-maps ./test/command.mjs --require ./test/before.mjs --timeout=24000 --check-leaks`
- `npm test`

### DIFF
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -552,6 +552,16 @@ export class CommandInstance {
     const optional = commandHandler.optional.slice(0);
     const positionalMap: Dictionary<string[]> = {};
 
+    if (
+      yargs.getOptions().configuration['halt-at-non-option'] &&
+      argv._.length === 0 &&
+      Array.isArray(argv['--']) &&
+      argv['--'].length > 0
+    ) {
+      argv._ = argv['--'].slice();
+      delete argv['--'];
+    }
+
     this.validation.positionalCount(demanded.length, argv._.length);
 
     while (demanded.length) {
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -559,7 +559,9 @@ export class CommandInstance {
       argv['--'].length > 0
     ) {
       argv._ = argv['--'].slice();
-      delete argv['--'];
+      if (!yargs.getOptions().configuration['populate--']) {
+        delete argv['--'];
+      }
     }
 
     this.validation.positionalCount(demanded.length, argv._.length);
--- a/test/command.mjs
+++ b/test/command.mjs
@@ -208,6 +208,35 @@ describe('Command', () => {
       called.should.equal(true);
     });
 
+    // see: https://github.com/yargs/yargs/issues/2423
+    it('counts halt-at-non-option arguments for default command positionals', () => {
+      let called = false;
+      const argv = yargs('my-host ls')
+        .usage(
+          '$0 <host> [cmd...]',
+          'Run command via ssh',
+          yargs =>
+            yargs
+              .positional('host', {type: 'string', demandOption: true})
+              .positional('cmd', {array: true, type: 'string'})
+              .option('verbose', {type: 'boolean', alias: 'v', count: true}),
+          argv2 => {
+            argv2.host.should.equal('my-host');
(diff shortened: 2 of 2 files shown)


shadowspawn:

pupuking723:
> To be clear: is this Pull Request entirely AI/LLM generated?

Hi @shadowspawn, no. I used an AI coding assistant, but I reviewed the issue and implementation myself, ran the relevant tests, and I’m responsible for the change. I can add an AI-assistance note to the PR description if that helps.

lib/command.ts
```diff
@@ -552,6 +552,16 @@ export class CommandInstance {
...
     const positionalMap: Dictionary<string[]> = {};
 
+    if (
+      yargs.getOptions().configuration['halt-at-non-option'] &&
+      argv._.length === 0 &&
+      Array.isArray(argv['--']) &&
+      argv['--'].length > 0
+    ) {
+      argv._ = argv['--'].slice();
+      delete argv['--'];
```
shadowspawn: I wonder whether deleting `argv['--']` may interact with author using `populate--` configuration? We opened up author changes to configuration, and there are multiple issues around interactions of non-default settings (like the issue prompting this PR).

Related, these two links show yargs turning on `populate--` when calling yargs-parser. There are multiple references to `populateDoubleDash` in code. 

https://github.com/yargs/yargs/blob/437f3a4e0f4166e1f1a3ce023b0331159582746d/lib/yargs-factory.ts#L1990-L1994

https://github.com/yargs/yargs/blob/437f3a4e0f4166e1f1a3ce023b0331159582746d/lib/command.ts#L442-L450

pupuking723:
Thanks for pointing this out. I updated the PR to preserve `argv['--']` when `populate--` is enabled, and added a regression test that combines `halt-at-non-option` with `populate--` for default command positionals.

Verification after the update:

- `npx mocha --enable-source-maps ./test/command.mjs --require ./test/before.mjs --grep "halt-at-non-option" --timeout=24000 --check-leaks`
- `npx mocha --enable-source-maps ./test/command.mjs --require ./test/before.mjs --timeout=24000 --check-leaks`
- `npm test`

pupuking723:
Clarifying my earlier reply: I am the human account owner and contributor. I used AI coding assistance while preparing this change, then reviewed the implementation and tests and take responsibility for what I submitted. I understand if this repository does not accept AI-assisted contributions.

nfriedly:
I would appreciate this fix and possibly a way to say "positional args can only come after the flags and optionally a --" (that didn't involve sending things to the parser), like a new `.strictPositionals()` or something along those lines.

PR #2544, opened 2026-06-10
fix: redact hidden options in implication errors

### LINKED ISSUE #2540: Hidden option revealed for implies
When some option is hidden but other option is dependent on it by "implies", then the first option is revealed when the implied option is used.

For example, we have an electron app and Electron apps consumes some debugging options by default, e.g. `--remote-debugging-port`. We want to allow it on some secret option only:
```js
const yargsHelpers = require('yargs/helpers');

const options = yargs(yargsHelpers.hideBin(process.argv))
    .option("hidden-debug", {
        type: "boolean",
        hidden: true
    })
    .option("remote-debugging-port", { hidden: true })
    .implies("remote-debugging-port", "hidden-debug")
    .help()
    .strict(true)
    .parse();
``` 

However, when someone tries to run the app with `--remote-debugging-port` only, yargs fails and shows:
```
Missing dependent arguments:
 remote-debugging-port -> hidden-debug
```
So, the user knows what option to use, then 😜 

IMO when `hidden-debug` is hidden, it should not be printed in error messages, too.

sjh9714:
## Summary

Fixes #2540.

When an implication failure involves options marked `hidden`, the failure message now redacts those option names instead of printing the raw keys. Validation behavior is unchanged; the error still reports the failed implication without exposing hidden option names.

## Changes

- Redact hidden option names while formatting implication failure details.
- Add regression coverage for an implication between hidden Electron-style options.

## Testing

- `npm run compile -- -p tsconfig.test.json && npx c8 mocha --enable-source-maps ./test/validation.mjs --require ./test/before.mjs --timeout=24000 --check-leaks`
- `npx eslint lib/validation.ts test/validation.mjs`
- `npm test` (Mocha suite passed: 828 passing, 1 pending; posttest `npm run check` failed on pre-existing Prettier findings in untouched `lib/command.ts` and `lib/yargs-factory.ts`.)

Note: I used Codex while preparing this change, reviewed the final diff, and ran the listed checks locally.


shadowspawn:
Looks reasonably tidy. 

I am still wondering if worth fixing: https://github.com/yargs/yargs/issues/2540#issuecomment-4646050526

sjh9714:
Thanks for the pointer. I added coverage for the one-sided cases too: a visible option implying a hidden option now renders as `visible -> [hidden]`, and a hidden option implying a visible option renders as `[hidden] -> visible`. The existing both-hidden regression still covers `[hidden] -> [hidden]`.

Local checks:
- `npx c8 mocha --enable-source-maps ./test/*.mjs --require ./test/before.mjs --timeout=24000 --check-leaks --grep "implication"`
- `npx gts lint lib/validation.ts test/validation.mjs`
- `npm run check:js -- test/validation.mjs`
- `git diff --check`

I also rechecked the current failing CI: the Node 23 job is failing during `npm install --engine-strict` because `eslint-visitor-keys@5.0.1` requires `^20.19.0 || ^22.13.0 || >=24`, and the Prettier errors are in `lib/command.ts` / `lib/yargs-factory.ts`, which are unchanged by this PR and identical to `upstream/main`.

shadowspawn:
Thanks for looking into the test breakage. I have a PR open for the lint change (#2511) and had not looked into the engine issue yet.

PR #2548, opened 2026-06-17
fix(completion): remove spurious single-quote in zsh loadautofunc detection

JSap0914:
## Bug

The zsh completion template added in #2424 contains a stray single-quote `'` before `${zsh_eval_context[-1]}`, making the autoloaded-function detection always fail:

```sh
# Generated script (broken) — the left side is "'loadautofunc" not "loadautofunc"
if [[ "'${zsh_eval_context[-1]}" == "loadautofunc" ]]; then
```

The double-quoted string `"'${zsh_eval_context[-1]}"` evaluates to a string with a literal `'` prepended to the variable value, so the comparison is `'loadautofunc` vs `loadautofunc` — never equal. The entire `if` branch that makes zsh autoloaded completions work on first Tab press is unreachable.

## Fix

Remove the spurious `'` so the generated script reads:

```sh
if [[ "${zsh_eval_context[-1]}" == "loadautofunc" ]]; then
```

## Verification

```
npx mocha --enable-source-maps ./test/completion.mjs --require ./test/before.mjs --timeout=24000 --grep "spurious"

  Completion
    zsh
      ✓ zsh completion script uses correct zsh_eval_context comparison without spurious single-quote

  1 passing (93ms)
```

The test fails (RED) on the original template and passes (GREEN) after the one-character fix. The full suite (828 tests) continues to pass with no regressions.

> Note: This PR was prepared with AI assistance.

### DIFF
--- a/lib/completion-templates.ts
+++ b/lib/completion-templates.ts
@@ -49,7 +49,7 @@ _{{app_name}}_yargs_completions()
     _default
   fi
 }
-if [[ "'\${zsh_eval_context[-1]}" == "loadautofunc" ]]; then
+if [[ "\${zsh_eval_context[-1]}" == "loadautofunc" ]]; then
   _{{app_name}}_yargs_completions "$@"
 else
   compdef _{{app_name}}_yargs_completions {{app_name}}
--- a/test/completion.mjs
+++ b/test/completion.mjs
@@ -1156,18 +1156,6 @@ describe('Completion', () => {
         '--bar:bar flag',
       ]);
     });
-
-    it('zsh completion script uses correct zsh_eval_context comparison without spurious single-quote', () => {
-      process.env.SHELL = '/bin/zsh';
-      const r = checkOutput(() => yargs([]).showCompletionScript(), ['myapp']);
-      const script = r.logs[0];
-      // A stray single-quote before ${zsh_eval_context[-1]} makes the if-comparison always false
-      script.should.not.include(
-        '"\'${zsh_eval_context[-1]}" == "loadautofunc"'
-      );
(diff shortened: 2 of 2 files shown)


shadowspawn:
Over 120 Pull Requests opened in 3 days.

JSap0914:
Fair question, and thanks for asking directly rather than just closing it. Yes — I use an AI coding assistant to find and fix small, self-contained bugs, and I review and test each change locally before opening it (for this one: a RED test reproducing the issue, then the fix, then the full suite). I realize a burst of PRs across many repos reads as noise and adds review load you didn't ask for. If commander/yargs would prefer not to receive AI-assisted contributions, just say the word and I'll close these and stay out of your tracker — no hard feelings either way.

shadowspawn:
That change got introduced right before release: https://github.com/yargs/yargs/pull/2424/changes/cbf7788d4f92121b0a3368f6f23496fe67a83abb

A lot of discussion in #2424 for something that didn't achieve the goal! 😢 

The RED test is not appropriate. It is testing for s specific typo and not testing the functionality. Given the difficulty in testing and a minor change, I think just leave the test out.

I'll check this by hand before releasing in any case.

JSap0914:
Dropped the test — the PR now contains only the one-line fix removing the spurious single-quote in `lib/completion-templates.ts`.

lbesecker195 (approved):
Verified in `lib/completion-templates.ts`: the string is a JS template literal, so the leading `'` is emitted verbatim into the zsh script and the comparison becomes `"'loadautofunc" == "loadautofunc"`, which never matches. Removing it restores the intended check. LGTM.

PR #2550, opened 2026-06-23
fix: apply coerce on every parse of a reused instance

spokodev:
## Problem

A yargs instance is documented as reusable: you can call `.parse()` on the same instance multiple times. But when an option has a `coerce` function, the coercion only runs on the **first** `parse()`. Every parse after that returns the raw, un-coerced value.

```js
const y = yargs().coerce('bar', v => 'C(' + v + ')');

y.parse('--bar one').bar;   // 'C(one)'   correct
y.parse('--bar two').bar;   // 'two'      BUG: coerce skipped
y.parse('--bar three').bar; // 'three'    BUG: coerce skipped
```

## Root cause

Coerce is implemented as a "mutating" middleware. To avoid running it twice within a single parse (the nested-command case fixed in #1966), the middleware sets an `applied` flag the first time it runs and short-circuits on subsequent runs:

```ts
if (middleware.mutates) {
  if (middleware.applied) return acc;
  middleware.applied = true;
  ...
}
```

The flag was never reset between top-level `parse()` calls, so on the second parse the middleware saw `applied === true` and skipped coercion.

## Fix

Add `GlobalMiddleware.resetApplied()`, which clears the `applied` flag on mutating middleware, and call it at the start of a top-level parse only (`commandIndex === 0`) alongside the existing cached-help-message clearing.

Because the reset happens only at the top level, the `applied` guard is still active while a single parse walks into nested commands. The double-coerce fix from #1966 stays intact (verified: that test still passes).

## Test

Added a regression test in `test/middleware.mjs` that parses one instance three times with different `--bar` values and asserts coerce produces the coerced value every time and the coerce function is called exactly 3 times.

Verified red/green: the test fails on the current tree (second parse returns `'two'` instead of `'C(two)'`) and passes with the fix. Full suite stays green (828 passing), including the #1966 nested-command double-coerce test.

### DIFF
--- a/lib/middleware.ts
+++ b/lib/middleware.ts
@@ -70,6 +70,13 @@ export class GlobalMiddleware {
   reset() {
     this.globalMiddleware = this.globalMiddleware.filter(m => m.global);
   }
+  // Reset the "applied" guard on mutating (coerce) middleware so they run
+  // again on the next top-level parse of a reused yargs instance.
+  resetApplied() {
+    for (const m of this.globalMiddleware) {
+      if (m.mutates) m.applied = false;
+    }
+  }
 }
 
 export function commandMiddlewareFactory(
--- a/lib/yargs-factory.ts
+++ b/lib/yargs-factory.ts
@@ -2024,6 +2024,7 @@ export class YargsInstance {
     // need to clear the cached help message from the previous parse:
     if (commandIndex === 0) {
       this.#usage.clearCachedHelpMessage();
+      this.#globalMiddleware.resetApplied();
     }
 
     try {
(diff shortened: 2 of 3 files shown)


shadowspawn:

spokodev:
It is not auto-generated and submitted blindly. I use an AI assistant as a tool in my workflow and I am open about that. The bug is real: `coerce` is not re-applied when a yargs instance is reused across multiple `parse()` calls. I reduced it to a failing test and checked the fix against yargs's existing coerce semantics. The analysis and the decision of what to ship are human-reviewed, not a machine dump. If you would prefer a different approach to the fix, I can rework it.

shadowspawn:
Account created over 100 Pull Requests in same few days as this PR. Too much AI for this human reviewer to spend more time on this PR. Can revisit if and when the issue is prioritised.

PR #2551, opened 2026-06-24
docs: clarify nargs option-like values

### LINKED ISSUE #2442: option narg does not behave as documented
### Problem

nargs options config does not consume values starting with `-` as documented

### Expected Results
The docs describe nargs to work like this:
```
  .nargs('token', 1)
  .parse(['--token', '-my-token']);
```
> parses as:
> { _: [], token: '-my-token', '$0': 'node test' }

### Actual Result
yargs 17.7.2 instead does not consume the `-my-token` and returns an error:

> Options:
>   --help     Show help                                                 [boolean]
>   --version  Show version number                                       [boolean]
> 
> Not enough arguments following: token

nightt5879:
## Summary
- Updates the `.nargs()` API docs to reflect the current default parser behavior for option-like values.
- Shows how to enable `parserConfiguration({'nargs-eats-options': true})` when callers want `.nargs()` to consume values such as `-my-token`.

Fixes #2442

## Verification
- `npm run check`


shadowspawn:
I do not want a broken example (the original example). I think there are at least two main uses of nargs: supporting multiple option-arguments, and in conjunction with `nargs-eats-options` making option POSIX compatible and consume option-like arguments.

I suggest change the original example to take multiple (non-option-like) option-arguments, and your new example covers the case where want to consume option-like arguments.

nightt5879:
Thanks for the suggestion. I updated the `.nargs()` docs so the first example now shows the regular multiple option-arguments use case, and the `nargs-eats-options` example is kept for the option-like value case such as `-my-token`.

Verification: `npm run check`

docs/api.md
```diff
@@ -1228,14 +1228,30 @@ yargs()
...
+  .nargs('files', 2)
+  .parse(['--files', 'src/index.js', 'test/index.js']);
+```
+
+parses as:
+
+`{ _: [], files: ['src/index.js', 'test/index.js'], '$0': 'node test' }`
+
+By default, values that look like options are not consumed by `.nargs()`. If
+the value should consume option-like arguments, enable the yargs-parser
```
shadowspawn:
```suggestion
the option should consume option-like arguments, enable the yargs-parser
```

PR #2553, opened 2026-06-26
fix: resolve show-hidden key via camelCase when strip-dashed is enabled

mahirhir:
## Problem

When `parserConfiguration({ 'strip-dashed': true })` is set, `--show-hidden` silently does nothing. Hidden options do not appear in `--help` output even when `--show-hidden` is explicitly passed.

Minimal repro (from #2356):

```js
yargs(['--help', '--show-hidden'])
  .option('foo', { describe: 'bar', type: 'string', hidden: true })
  .parse();
// --foo never appears
```

## Root cause

`filterHiddenOptions` in `lib/usage.ts` checks `argv[showHiddenOpt]` where `showHiddenOpt` is the raw option name (e.g. `"show-hidden"`). With `strip-dashed: true`, yargs-parser removes dashed keys from parsed argv and only keeps camelCase variants (`"showHidden"`), so the lookup always returns `undefined`.

## Fix

After the direct lookup fails, also check `argv[shim.Parser.camelCase(showHiddenOpt)]`. The `camelCase` helper is already part of the `Parser` interface exposed on the shim. When `strip-dashed` is off the first lookup hits as before; when it is on the camelCase lookup finds the value.

Change is limited to three lines inside `filterHiddenOptions` in `lib/usage.ts`.

## Test

Added a regression test in `test/usage.mjs` inside the existing `hidden options` describe block:

```
✔ --help should display hidden options with --show-hidden when strip-dashed is enabled
```

Full suite: **821 passing, 0 failures**.

Fixes #2356

### DIFF
--- a/lib/usage.ts
+++ b/lib/usage.ts
@@ -610,9 +610,12 @@ export function usage(yargs: YargsInstance, shim: PlatformShim) {
   }
 
   function filterHiddenOptions(key: string) {
+    const showHiddenOpt = yargs.getOptions().showHiddenOpt;
+    const argv = (yargs.parsed as DetailedArguments).argv;
     return (
       yargs.getOptions().hiddenOptions.indexOf(key) < 0 ||
-      (yargs.parsed as DetailedArguments).argv[yargs.getOptions().showHiddenOpt]
+      argv[showHiddenOpt] ||
+      argv[shim.Parser.camelCase(showHiddenOpt)]
     );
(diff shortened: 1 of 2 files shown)


shadowspawn:

mahirhir:
Yes, this was AI-assisted. I'll withdraw it rather than put that on your review — thanks for taking the time to look.

PR #2554, opened 2026-06-28
chore: improve build robustness

shadowspawn:
- update and pin GitHub actions
- add package-lock.json
- add CI test with updated direct dependencies
- lint changes from updates

Over the last few years the builds have often been broken by in-range updates to dependencies, especially development dependencies. Add `package-lock.json` to make the build more reliable. To reduce the risk of missing breakages introduced by updates to production dependencies (which will affect yargs users), added a CI job which updates the direct dependencies then runs the tests.


shadowspawn:
FYI: @wesleytodd 

Adding `package-lock.json` et al.

(Mentioned in https://github.com/nodejs/citgm/pull/1106)

PR #2556, opened 2026-06-29
fix: replace every $1 in demandCommand min/max messages

patrickwehbe:
The custom `minMsg`/`maxMsg` templating in `demandCommand` replaces `$0` with a global regex but `$1` without one, so only the first `$1` in a message is substituted and any later `$1` is left as a literal.

The docs say otherwise (docs/api.md): "every instance of `$1` will be replaced with the expected value". So the `$1` replace should be global to match `$0` and the documented behavior.

Before, with `'got $0 ($0), max $1 ($1)'`:
```
got 3 (3), max 2 ($1)
```
After:
```
got 3 (3), max 2 (2)
```

Changed both the min and max branches in `lib/validation.ts` and added a test covering repeated tokens. Full suite passes locally.

### DIFF
--- a/lib/validation.ts
+++ b/lib/validation.ts
@@ -44,7 +44,7 @@ export function validation(
             demandedCommands._.minMsg
               ? demandedCommands._.minMsg
                   .replace(/\$0/g, _s.toString())
-                  .replace(/\$1/, demandedCommands._.min.toString())
+                  .replace(/\$1/g, demandedCommands._.min.toString())
               : null
           );
         } else {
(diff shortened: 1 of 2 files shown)


shadowspawn:
(I will rerun tests after #2554 lands to stabilise build.)

lbesecker195 (approved):
`$0` was already replaced globally while `$1` was not, so a custom message using `$1` twice left the second occurrence literal. Making both global is consistent, and the new test covers it. LGTM.

PR #2557, opened 2026-07-02
fix: resolve show-hidden option lookup when strip-dashed is enabled

amitmishra11:
## Problem

When `parserConfiguration({ 'strip-dashed': true })` is set, calling `--show-hidden` (or any custom name with dashes) in the CLI has no effect: hidden options are not displayed even though `--show-hidden` is explicitly passed.

Minimal reproduction from #2356:

```js

yargs(['--help', '--show-hidden'])
  .options({ foo: { describe: 'bar', type: 'string', hidden: true } })
  .parse();
// Expected: --foo appears in help output
// Actual:   --foo is hidden despite --show-hidden being passed
```

## Root cause

`filterHiddenOptions` in `lib/usage.ts` checks `argv[showHiddenOpt]` to determine if hidden options should be shown. When `strip-dashed` is enabled, yargs converts dashed option names to camelCase in the parsed `argv` object. So `argv['show-hidden']` is `undefined`, while `argv['showHidden']` is `true`. The lookup always returns undefined, so hidden options are never revealed.

## Fix

Check both the original option name and its camelCase equivalent in argv:

```ts
function filterHiddenOptions(key: string) {
  const showHiddenOpt = yargs.getOptions().showHiddenOpt;
  const argv = (yargs.parsed as DetailedArguments).argv;
  return (
    yargs.getOptions().hiddenOptions.indexOf(key) < 0 ||
    argv[showHiddenOpt] ||
    argv[shim.Parser.camelCase(showHiddenOpt)]
  );
}
```

## Testing

A new test case was added to `test/usage.mjs` covering `--show-hidden` with `strip-dashed` enabled. All 821 existing tests continue to pass.

Fixes #2356

### DIFF
--- a/lib/usage.ts
+++ b/lib/usage.ts
@@ -610,9 +610,12 @@ export function usage(yargs: YargsInstance, shim: PlatformShim) {
   }
 
   function filterHiddenOptions(key: string) {
+    const showHiddenOpt = yargs.getOptions().showHiddenOpt;
+    const argv = (yargs.parsed as DetailedArguments).argv;
     return (
       yargs.getOptions().hiddenOptions.indexOf(key) < 0 ||
-      (yargs.parsed as DetailedArguments).argv[yargs.getOptions().showHiddenOpt]
+      argv[showHiddenOpt] ||
+      argv[shim.Parser.camelCase(showHiddenOpt)]
     );
   }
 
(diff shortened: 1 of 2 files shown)


shadowspawn:

It looks almost identical to #2553.

amitmishra11:
The changes are AI assisted and have been reviewed by me. Regarding #2553 — that PR is closed, so there is no conflict.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #2559, opened 2026-07-11
fix: reveal hidden options with --show-hidden when strip-dashed is set

c-tonneslan:
Fixes #2356.

`filterHiddenOptions` decided whether to reveal hidden options by reading `argv[showHiddenOpt]` — i.e. the dashed key (`show-hidden`). With `parserConfiguration({'strip-dashed': true})`, the dashed key is stripped from `argv` and only the camel-cased `showHidden` remains, so the lookup missed and `--help --show-hidden` no longer revealed hidden options.

This also checks the camel-cased form via `shim.Parser.camelCase(showHiddenOpt)`, so it works whether or not `strip-dashed` (or `camel-case-expansion`) is set.

Added a regression test in `test/usage.mjs`. Full `usage.mjs` suite (214) and `npm run check` (gts + eslint) pass.

### DIFF
--- a/lib/usage.ts
+++ b/lib/usage.ts
@@ -609,9 +609,14 @@ export function usage(yargs: YargsInstance, shim: PlatformShim) {
   }
 
   function filterHiddenOptions(key: string) {
+    const argv = (yargs.parsed as DetailedArguments).argv;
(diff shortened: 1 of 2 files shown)


shadowspawn:

shadowspawn:
No reply in two weeks. Possibly AI generated.

PR #2571, opened 2026-08-01
fix: complete negated boolean options

### LINKED ISSUE #2254: bash completion: negative boolean options does not auto complete

subotac:
## Summary
  - generate positive and negated completions for boolean options
  - respect explicitly disabled boolean negation
  - add Bash, Zsh, and Fish regression coverage

  Fixes #2254

  ## Testing
  - npm test

### DIFF
--- a/lib/completion.ts
+++ b/lib/completion.ts
@@ -124,7 +124,7 @@ export class Completion implements CompletionInstance {
 
       Object.keys(options.key).forEach(key => {
         const negable =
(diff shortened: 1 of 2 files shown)


shadowspawn:
The `--no-help` and `--no-version` are pretty annoying. I'll add a note to the original issue.

PR #2573, opened 2026-08-07
chore: add FUNDING links

shadowspawn:


shadowspawn:
Can do it directly...

PR #2574, opened 2026-08-12
fix: restore parser state when a parse throws from .fail()

### LINKED ISSUE #2481: unknown command giving the wrong error message with .strict() and .strictCommands()?
I have the following yargs setup for one of my projects:
```ts
export const commands = yargs()
  .command(
    "help",
    "Show help menu for commands",
    () => {} // no args
  )

  .command(
    "exit",
    "Disconnect from the world",
    () => {} // no args
  )

  .command(
    "where",
    "Get current location",
    () => {} // no args
  )

  .command("come", "Walk to the sender", () => {})

  .command("build <prompt>", "Build something based on a prompt", yargs => {
    return yargs.positional("prompt", {
      type: "string",
      describe: "Prompt describing the structure",
    });
  });
```

...and I have a function that parses a string (converted to argv with the help of another library) as a command, and throws an error if the parsing fails:

```ts
export function parseCommand(input: string) {
  return commands
    .exitProcess(false)
    .strictCommands()
    .fail(msg => {
      throw new Error(msg);
    })
    .parseSync(stringArgv(input));
}
```

I was testing this function when I noticed that when I gave it the nonexistent command `fly away`, the error was:
```
Error: Missing required argument: prompt
```
This is not the correct, message, right? The message should say that it's an unknown command, but instead it is interpreting it as the "build" command in my yargs instance. I also had .strictCommands() on (tried .strict() as well), so is this an issue? Or am I misunderstanding how the strict modes work?

MsfPablo:
Fixes #2481.

## The bug

Reusing a `yargs` instance across parses leaks state from any parse that threw, so a later unknown command is reported with the *previous* command's error:

```js
const commands = yargs()
  .command('help', 'h', () => {})
  .command('build <prompt>', 'b', y => y.positional('prompt', {type: 'string'}));

const parse = args => commands.strictCommands().fail(msg => { throw new Error(msg); }).parseSync(args);

parse(['build']);      // throws (expected — missing positional)
parse(['fly', 'away']); // ❌ "Missing required argument: prompt"
                        // ✅ "Unknown commands: fly, away"
```

This is what the issue reporter hit, and it's why moving the instance construction inside the function worked around it.

## Cause

Two independent leaks, both on the throwing path:

1. `parse()` pushes state with `kFreeze()` but the synchronous branch only calls `kUnfreeze()` after a successful `kRunYargsParserAndExecuteCommands()`. A throwing `.fail()` handler skips it, so the `build` command builder's demanded positionals stay applied to the instance. (The async branch already handles this via `.finally()`.)
2. `runCommand()` pushes the command onto the shared context and pops it at the end of `handleValidationAndGetResult()`. If validation fails, the pop never runs, so `context.commands` keeps `'build'` — and `validation.unknownCommands()` does `argv._.slice(currentContext.commands.length)`, silently skipping `fly`.

Both are needed: fixing only (1) changes the message to `Unknown command: away` (still dropping `fly`).

## Fix

- `parse()`: unfreeze and rethrow when the sync path throws.
- `kFreeze`/`kUnfreeze`: record the context depth and truncate back to it, rather than assuming the pushes and pops were balanced.

## Verification

```
$ npm test
  840 passing (4s)
  1 pending
```

Lint (`posttest` → `gts lint && eslint`) is clean. The new test in `test/validation.mjs` fails on `main` with exactly the reported message:

```
AssertionError: expected [Function] to throw error including 'Unknown commands: fly, away'
  but got 'Missing required argument: prompt'
```

---

Courtesy note: this patch was prepared with AI assistance (Claude). The repro, the fix, and the test run were verified locally by me — happy to adjust the approach if you'd prefer the context push/pop made exception-safe in `command.ts` instead of truncating in `kUnfreeze`.


shadowspawn:
Thanks for the AI courtesy note, I appreciate that.

Too much AI for this human reviewer to spend time on this PR at this time. Can revisit if and when the issue is prioritised.

PR #2575, opened 2026-08-16
Show hidden options when strip dashed is enabled

Hashim1999164:
## Summary
Help ignored the show hidden flag when parserConfiguration rewrites dashed keys to camelCase. Hidden options then stayed hidden even when that flag was set.

This looks up the camelCase form the same way conflicts already do.

Fixes #2356

## Test plan
Ran usage tests for hidden options, including the new strip dashed cases. Also ran the existing conflicts test for strip dashed.


shadowspawn:
AI slop

PR #2576, opened 2026-08-21
fix: complete negated boolean options after --no- (fix #2254)

### LINKED ISSUE #2254: bash completion: negative boolean options does not auto complete

tunglambk:
## Summary

- Complete negated boolean options when the current completion token starts with `--no-`.
- Preserve the existing opt-in behavior for showing default-true negations with `boolean-negation: true`.
- Exclude the built-in help and version options, and respect `boolean-negation: false`.

Fixes #2254

## Testing

- `npm test`


shadowspawn:
This looks on the surface reasonably tidy, but a robust solution is harder. For example, the help and versions options can have user supplied names.

Account is doing lots of AI contributions currently.

PR #2577, opened 2026-08-22
fix(completion): escape tabs in choice values for fish

chuanghiduoc:
## Problem

fish splits a completion line on the first tab into the completion value and its description. The fish branch of `choicesFromOptionsCompletions()` and `choicesFromPositionalsCompletions()` pushes choice values through unchanged, so a choice containing a literal tab is emitted as e.g. `pro<TAB>duction` — fish then displays it as value "pro" with description "duction", and the actual value can never be completed.

The bash/zsh path escapes the shell-special character (`:` → `\:`), but the fish path added in #2568 has no equivalent escaping for its special character (`\t`).

Reproduction (fails on main, passes with this PR):

```js
process.env.SHELL = '/usr/bin/fish';
yargs()
  .option('env', {choices: ['dev:test', 'pro\tduction']})
  .completion()
  .getCompletion(['--env', ''], (_err, completions) => {
    console.log(completions);
  });
// main:      ['dev:test', 'pro\tduction']
// this PR:   ['dev:test', 'pro\tduction']
```

## Fix

Escape literal tabs (`\t` → `\t`) in both fish choice paths. fish renders the escaped form as a visible tab in the suggestion while keeping the whole string as a single value.

## Testing

- new test in `test/completion.mjs`: choices containing tabs are escaped under `SHELL=fish`
- full suite: 833 tests passing


shadowspawn:
This looks technically correct, but have you got any practical example of a tab appearing in a completion value?

chuanghiduoc:
Practical cases where a tab lands in a completion value:

1. **Choices derived from data**, not literals — e.g. completing from a TSV source (a spreadsheet export, a Makefile recipe list, a database query) where the delimiter is a tab:

```js
const rows = fs.readFileSync('products.tsv', 'utf8').split('\n');
program.option('--product', {choices: rows.map(r => r.split('\t')[1])});
// a row 'P1\tWidget' yields the choice 'Widget' — fine — but a malformed
// row with an embedded tab yields a choice containing '\t'
```

2. **Filesystem names**: tabs are legal in filenames (`my\treport.pdf`). A CLI completing `--file` choices from `fs.readdirSync` passes the raw name through, and fish then renders the completion as value + bogus description.

3. **Descriptions built by joining fields** with a tab (the same separator fish itself uses), e.g. `choices: [name + '\t' + version]` — the second tab corrupts the display for every entry.

The escape is what bash/zsh paths already do for their special character (colon), so this just brings fish to parity. It is cheap (one replace on values that rarely contain tabs) and only affects fish.

shadowspawn:
Low value. Closing as AI contribution not worth reviewing at this time.

PR #2578, opened 2026-08-24
Keep long commands readable in help

### LINKED ISSUE #2204: help output for commands wraps at half of wrap()
yargs 17.5.1

the help output for commands wraps at half of wrap(), even if the command does not have a description. also, for commands *with* descriptions, the output is still silly.

**example**

```
import yargs from 'yargs';
let SECTION = msg => console.log(`\n======${msg?` ${msg} `:''}======\n`);

// help output for commands is wrapped at half of wrap() even without description:
SECTION('command wrap');
yargs(['--help'])
    .command('obliterate <city|country|continent>')
    .wrap(80)
    .exitProcess(false)
    .argv;
        
// also, the placement of description on the right side is bad, anyways:
SECTION();
yargs(['--help'])
    .command('encabulate <problem> <chaos> [maxdepth]', 'recursively encabulate a problem over an arbitrary chaos manifold')
    .describe('turbo', 'enable turbo mode')
    .wrap(80)
    .exitProcess(false)
    .argv;
```

**output**

```
====== command wrap ======

yargteststuff.js [command]

Commands:
  yargteststuff.js obliterate <city|countr
  y|continent>

Options:

============

yargteststuff.js [command]

Commands:
  yargteststuff.js encabulate <problem> <c  recursively encabulate a problem ove
  haos> [maxdepth]                          r an arbitrary chaos manifold

Options:
  --turbo    enable turbo mode
```

---

**proposed solution**

of course it is possible to specify custom usage text with the desired formatting, but the default could still use some improvement. my suggestion is to let commands take the full width, and to place their descriptions on a separate line, e.g.:

```
Commands:
  yargteststuff.js obliterate <city|country|continent>
```

```
Commands:
  yargteststuff.js encabulate <problem> <chaos> [maxdepth]
      recursively encabulate a problem over an arbitrary chaos manifold
```

wantosure:
Fixes #2204.

When help output had a wrap width set, command names were always laid out in the same left column used for command descriptions. That column is capped at half the wrap width, which meant a long command with no description could still wrap even though the rest of the line was empty.

This keeps the existing two-column layout for normal command rows, but uses a full-width command row when there is no description. If a command has a description but the command name would wrap in the half-width column, the description moves onto the next line instead.

Added coverage for both cases from the issue.

Tested with:

```text
npm run compile -- -p tsconfig.test.json
npx mocha --enable-source-maps ./test/usage.mjs --require ./test/before.mjs --timeout=24000 --check-leaks
npm run check
npm test
```

The first full `npm test` run inside the sandbox hit `spawn EPERM` in integration tests; rerunning outside the sandbox passed with `834 passing, 2 pending`.


shadowspawn:

PR #2579, opened 2026-08-24
Handle YError when stack capture is unavailable

### LINKED ISSUE #2101: Using Yargs in browser with .strict(true) throws error
If you use Yargs in the browser with `.strict(true)`, it throws the following error: 

```
[Error] TypeError: Error.captureStackTrace is not a function. (In 'Error.captureStackTrace(this, YError)', 'Error.captureStackTrace' is undefined)
	parseArgs (yargs-factory.js:1021)
	parse (yargs-factory.js:525)
	(anonymous function) (_display:144)
```

This can easily be reproduced by using the example fiddle from the docs and simply adding `.strict(true)` to it. 

I'd like to be able to capture when a command is not selected so I can output a message. Ideally, I'd like Yargs to provide the message, like the help text and such, but there doesn't seem to be a clear way to do this.

wantosure:
Fixes #2101.

`YError` already skips stack capture when the runtime does not provide it. This tightens that guard to the actual callable case, which keeps browser-like runtimes from throwing while building the yargs error.

Added a small regression test that removes `Error.captureStackTrace` before throwing a `YError`.

Tests:
- `npx mocha --enable-source-maps ./test/yargs.mjs --require ./test/before.mjs --timeout=24000 --check-leaks --grep "captureStackTrace"`
- `npm run check`


shadowspawn:

PR #2580, opened 2026-08-24
Do not treat version defaults as the version flag

### LINKED ISSUE #2199: yargs.option("version") is admitted, but fails silently

wantosure:
Fixes #2199.

A `version` option with a default value was being treated the same as passing `--version`, so yargs printed the package version and exited before the user's option could be used.

This checks the original argv for the version flag or one of its aliases instead of relying on the parsed value, which may come from defaults or config.

Tests:
- `npx mocha --enable-source-maps ./test/usage.mjs --require ./test/before.mjs --timeout=24000 --check-leaks`
- `npm run check`


shadowspawn:

PR #2581, opened 2026-08-24
Populate default command positionals after double dash

### LINKED ISSUE #2017: Default command with variadic option and variadic positional
Shouldn't a test like the following pass or I have a wrong schema? (Example call: `$ cli --variadic 1 2 3 -- file1 file2`).

```js
import { strict as assert, deepEqual } from "assert";

yargs()
    .scriptName("cli")
    .command({
      command: `$0 [files..]`,
      description: `Default command description`,
      builder: (yargs) => {
        return yargs
          .positional("files", {
            describe: "Files description",
            type: "string",
          })
          .option("boolean", {
            description: `Single option description.`,
            type: "boolean",
            default: false,
          })
          .option("variadic", {
            description: `Variadic option description`,
            type: "array",
            default: [],
          });
      },
      handler: (argv) => {
        deepEqual(argv.variadic, [1, 2, 3]);
        deepEqual(argv.files, ["file1", "file2"]);
        assert(argv.boolean === true);
      },
    })
    .parse("--variadic 1 2 3 -- file1 file2");
```

If you put a regular option, instead of `--`, it works. For example: `.parse("--variadic 1 2 3 --boolean file1 file2")`.

It's not a big deal if the `files` positional is optional because you can pick the files from `_`. It becomes a problem when it's required, for example: `command: '$0 <files..>'`, because `yargs` exits.

wantosure:
Fixes #2017.

Default commands were filling positional arguments before the temporary `--` bucket was copied back into `_`. That meant `--variadic 1 2 3 -- file1 file2` left `file1` and `file2` in `_` instead of assigning them to `$0 [files..]`.

For default commands only, this folds the `--` values back before positional mapping when `populate--` is not enabled. Regular commands keep the existing behavior covered by the older #1457 test.

Tests:
- `npx mocha --enable-source-maps ./test/command.mjs --require ./test/before.mjs --timeout=24000 --check-leaks`
- `npm run check`


shadowspawn:

PR #2582, opened 2026-08-28
fix: avoid showing version for defaulted option

### LINKED ISSUE #2199: yargs.option("version") is admitted, but fails silently

justadityaraj:
Fixes #2199.

## Problem

A user-defined `version` option with a default was treated as if the built-in version flag had been supplied, so parsing printed the package version and exited.

## Changes

Use the parser's `defaulted` metadata to distinguish injected defaults from explicit version-option input. Add regression coverage for both the defaulted option and explicit `--version` behavior.

## User impact

Applications can use a defaulted `version` option without unexpectedly printing yargs' version or exiting. Explicit `--version` continues to show the built-in version.

## Verification

- `npm test` — passed (including compile, lint, and the full CommonJS test suite).
- `npm test -- --grep "defaulted version option|--version is explicit"` — passed (2 tests).
- `npm run check` — passed.
- `git diff origin/main...HEAD --check` — passed.
- `npm run test:esm` — known clean-main failure under Node 24.12.0: `test/esm/yargs-test.mjs` commandDir test dereferences `undefined.message` (10 passing, 1 failing); reproduced unchanged with the patch stashed.


shadowspawn:
This looks like drive-by AI slop.

PR #2583, opened 2026-08-31
fix: make showHidden(false) permanently hide hidden options

### LINKED ISSUE #2449: The docs suggest .showHidden(false) should hide the hidden options but it does not

gyanu2507:
Fixes #2449.

`.showHidden(false)` currently only skips *registering* the `--show-hidden` flag — it never stops it from working. `addShowHiddenOpt()` returns early, so `options.showHiddenOpt` keeps its default `'show-hidden'` value, and `filterHiddenOptions()` reveals a hidden key whenever `argv[showHiddenOpt]` is set:

```js
yargs.getOptions().hiddenOptions.indexOf(key) < 0 ||
  (yargs.parsed as DetailedArguments).argv[yargs.getOptions().showHiddenOpt]
```

The result is what @shadowspawn described in the issue: `--show-hidden` is absent from the help output, but passing it alongside `--help` still lists every hidden option.

```js
yargs('--help --show-hidden')
  .options({secret: {describe: 'a secret', hidden: true}})
  .showHidden(false)
  .parse()
// --secret is still listed
```

This clears `showHiddenOpt` when the option is disabled, so there is no name left for `filterHiddenOptions()` to look up.

On the "bug or documentation problem?" question from the issue — `docs/api.md` is already unambiguous about the intent, so I treated it as a code bug and left the docs alone:

> If the first argument is a boolean, it enables/disables this option altogether. i.e. hidden keys will be permanently hidden if first argument is `false`.

Unchanged: the default behaviour (`--show-hidden` works without calling `.showHidden()`), `.showHidden()`, and `.showHidden('custom-name')`.

Two tests added, both failing before the change — a help-output test next to the existing `--show-hidden` cases in `test/usage.mjs`, and an option-state test in the `showHidden` block of `test/yargs.mjs`. `npm run check` is clean. `npm test` goes from 821 to 823 passing; the 18 failures I see locally are the `commandDir` ESM and `integration` suites, and they fail identically on an unmodified `main` in my environment.


shadowspawn:
Well that looks simple enough! 

(It is likely to be weeks before I am able to review.)

shadowspawn:
A side-comment. The original post had a comment at the bottom, "Made with Cursor". I guess that was probably promotional and added by Cursor itself, but I do appreciate people acknowledging when there are high levels of AI use in PRs.

gyanu2507:
That footer was injected by the editor. I strip those when I catch them. The change itself is just `showHidden(false)` sticking for later `.option()` calls.

PR #2585, opened 2026-09-01
fix(parse): don't fire the user callback during a short-circuited completion re-parse

### LINKED ISSUE #1857: #parse() calls callback twice for --get-yargs-completions with #completion()
With `yargs@16.2.0`, the following code:

```js
yargs
  .completion()
  .parse(['--get-yargs-completions'], (...args) => {
    console.info('parse callback called with ', args);
  });
```

Prints:

    parse callback called with  [ null, { _: [], '$0': 'foo.js' }, '' ]
    parse callback called with  [
      undefined,
      {
        _: [],
        'get-yargs-completions': true,
        getYargsCompletions: true,
        '$0': 'foo.js'
      },
      'completion'
    ]

Indicating that the parse callback is called twice.  The first time `getYargsCompletions` is not in `argv`, the second it is.  Presumably this is not intentional?

Thanks,
Kevin

yfwmaniish:
Fixes #1857.

## What

\`completion.ts\`'s \`getCompletion()\` re-parses internally to build out completion suggestions:

\`\`\`ts
const argv = this.yargs.parse(args, true);
```

\`shortCircuit=true\` here is a literal boolean (it's the only call site in the codebase that passes one - everywhere else it's an options object, a callback, or omitted), so this nested call skips validation and other side effects, per the existing comment above it: \"completion short-circuits the parsing process, skipping validation, etc.\"

But \`#parseFn\` (the user's \`.parse(args, callback)\` callback) is stored as instance state rather than being scoped to the call that set it. So when a user does:

\`\`\`js
yargs
  .completion()
  .parse(['--get-yargs-completions'], callback);
\`\`\`

the callback fires **twice**: once from this internal short-circuited re-parse (with \`argv\` that doesn't even have \`getYargsCompletions\` set yet), then again from the outer parse that actually owns the callback.

## Fix

\`shortCircuit\` already uniquely identifies this internal re-parse, so \`parse()\` now gates all three places it invokes \`#parseFn\` on \`!isShortCircuited\`, in addition to the existing \`#parseFn\` truthiness check.

## Testing

Added a regression test asserting the callback fires exactly once and receives the fully-populated \`argv\` (with \`getYargsCompletions\` set), reproducing the exact repro from #1857.

- Negative control: reverted just the source fix, kept the test - fails with \`expected 2 to equal 1\`. Restored.
- Full suite: \`mocha test/*.mjs\` - 833 passing (2 pre-existing pending, unrelated).
- \`test/esm/*.mjs\` - one pre-existing failure (\`commndDir\` in ESM mode), confirmed present on unmodified \`main\` too before my change, unrelated to this fix (path-resolution issue, not completion/parse related).
- \`eslint\` clean on both changed files.


shadowspawn:
This looks like a drive-by AI contribution. The account created 8 other Pull Requests on the same day.

PR #2586, opened 2026-09-02
fix: resolve hidden options with --show-hidden under strip-dashed config

hardiksh28:
When strip-dashed parser configuration is enabled, yargs-parser converts option keys from kebab-case to camelCase in the parsed argv object. For example, --show-hidden becomes showHidden in argv.

filterHiddenOptions() in lib/usage.ts was looking up argv[showHiddenOpt] where showHiddenOpt = 'show-hidden'. Under strip-dashed, this key does not exist in argv (it is stored as 'showHidden'), so the lookup always returned undefined/falsy, causing hidden options to never be revealed even when --show-hidden was explicitly passed.

Fix: also check argv[shim.Parser.camelCase(showHiddenOpt)], which correctly resolves to the camelCase form that yargs-parser stores under strip-dashed.

The fix handles both the default 'show-hidden' option name and any custom name passed to showHidden() that contains a dash.

Fixes #2356


shadowspawn:
This is tidy, but not an urgent fix and I have other plans.

My long-term goal is to add a key lookup helper function either to yargs-parser or to yargs, rather than scatter around patches for the various yargs-parser behaviours. In this case it is a default behaviour that is not covered (camel-case), and in multiple other cases it is an optional behaviour is not covered.

PR #2588, opened 2026-09-05
ci: restrict workflow token permissions

sabtain05:
## Summary

- Restrict the CI workflow token to read-only repository contents

## Why

This follows the principle of least privilege for GitHub Actions. The workflow only checks out code and runs tests, so it does not need write access to repository contents. No test behavior or release behavior is changed.


shadowspawn:
AI slop

PR #2590, opened 2026-09-14
fix: dedupe choices on multiple calls

### LINKED ISSUE #2454: Multiple calls to choices should not create duplicate choices for same option

3bbaas:
## What?
Removes duplicate values when .choices() is called multiple times for the same key.

## Why?
Fixes #2454 — repeated calls merged with duplicates, e.g. help showed `["ahmed","shady","amro","ahmed","shady","amro","seif"]`.

## How?
- fix (lib/yargs-factory.ts): Set-based dedupe, gated to choices only.
- test (test/yargs.mjs): regression test with overlapping calls.
- docs (docs/api.md): clarify merge removes duplicates.

---

I hope this PR looks good for you @shadowspawn :>


shadowspawn:
Looks suitably tidy, thanks. Might be a while before I do a formal review.

PR #2592, opened 2026-09-16
fix: reveal hidden options with --show-hidden when strip-dashed is set

Yahiro025:
## Description

Fixes #2356.

When `parserConfiguration({ 'strip-dashed': true })` is enabled, yargs-parser stores camelCased keys on `argv` (`showHidden` instead of `show-hidden`). `filterHiddenOptions` only looked up `argv[showHiddenOpt]`, so `--show-hidden` (and custom show-hidden flag names) never revealed hidden options in help output.

This is a minimal fix in `lib/usage.ts`: also check the camelCase and kebab-case forms of the configured show-hidden option key. Intentionally smaller than #2516 (closed as too heavy for this edge case).

## Motivation and Context

`--help --show-hidden` should list hidden options even when `strip-dashed` is on. Without this, apps that prefer camelCase argv silently lose the ability to surface hidden options via the documented flag.

## How Has This Been Tested?

- Added regression tests in `test/usage.mjs` for:
  - default `--show-hidden` + `strip-dashed`
  - custom `showHidden('custom-show-hidden')` + `strip-dashed`
  - alias of show-hidden + `strip-dashed`
- Ran existing `hidden options` suite and the strip-dashed conflict validation test — all passing.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly. *(no docs change — behavior restored to documented intent)*
- [ ] I have added tests to cover my changes.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.


shadowspawn:
This looks like a drive-by AI contribution.

PR #2594, opened 2026-09-20
fix: show hidden options when strip-dashed is enabled

dyk1454683243-sudo:
Fixes #2356.

`filterHiddenOptions` checked `argv[showHiddenOpt]` using the dashed name. With `strip-dashed`, yargs-parser drops that key and keeps the camelCase/alias forms, so `--help --show-hidden` never revealed hidden options.

This looks up the configured option via parsed aliases, with a camelCase fallback. That matches how yargs-parser finds keys after configuration rewrites, and stays local to `filterHiddenOptions` (a shared helper was tried in #2516 and closed as too large for this edge case).

Regression tests in `test/usage.mjs` cover:
- default `--show-hidden` + `strip-dashed`
- the issue repro (`showHidden('show-hidden')` + `strip-dashed`)
- hidden options still hidden without the flag
- a custom dashed `showHidden()` name
- an alias of `--show-hidden`


shadowspawn:
This looks like a drive-by AI contribution. The account created over 100 other Pull Requests on the same day.

PR #2595, opened 2026-09-20
fix: look up show-hidden via aliases when strip-dashed is enabled

dyk1454683243-sudo:
Fixes #2356

`filterHiddenOptions()` checked `argv[showHiddenOpt]` using the dashed name (`show-hidden`). When `'strip-dashed': true`, yargs-parser deletes dashed keys and leaves the camelCase alias (`showHidden`), so `--show-hidden` was ignored.

This looks the option up through the aliases yargs-parser already recorded — the same approach coerce uses after #2308. Parser configuration does not need to be re-implemented: aliases already include camelCase, and argv already reflects `strip-dashed` / `strip-aliased` / `camel-case-expansion`.

```js
yargs('--help --show-hidden')
  .option('foo', {describe: 'bar', type: 'string', hidden: true})
  .parserConfiguration({'strip-dashed': true})
  .parse()
// --foo now appears in help
```

Added regression tests for the issue repro, the implicit `--show-hidden` flag, and the case where hidden options should stay hidden.


shadowspawn:
This looks like a drive-by AI contribution. The account created over 100 other Pull Requests on the same day.

